PR #10, opened 2023-05-13
Document Strbuf

lhearachel:
Reiteration of https://github.com/pret/pokeplatinum/pull/5 (because I started from a fresh branch).

Some names (e.g. `Strbuf_IsTrainerName`) are from pokediamond.

## Renamed Files

* `include/unk_02023790.h` -> `include/strbuf.h`
* `src/unk_02023790.c` -> `src/strbuf.c`

## New Files

* `include/charcode.h`

## Renamed Structs and Fields

* `UnkStruct_02023790` -> `Strbuf`
* `UnkStruct_02023790::unk_00` -> `Strbuf::maxsize`
* `UnkStruct_02023790::unk_02` -> `Strbuf::size`
* `UnkStruct_02023790::unk_04` -> `Strbuf::magic`
* `UnkStruct_02023790::unk_08` -> `Strbuf::data`

## Renamed Functions

* `sub_02023790` -> `Strbuf_Init`
* `sub_020237BC` -> `Strbuf_Free`
* `sub_020237E8` -> `Strbuf_Clear`
* `sub_02023810` -> `Strbuf_Copy`
* `sub_02023868` -> `Strbuf_Clone`
* `sub_020238A0` -> `Strbuf_FormatInt`
* `sub_020239D4` -> `Strbuf_FormatU64`
* `sub_02023B38` -> `Strbuf_AtoI`
* `sub_02023BE0` -> `Strbuf_Compare`
* `sub_02023C3C` -> `Strbuf_Length`
* `sub_02023C5C` -> `Strbuf_NumLines`
* `sub_02023C9C` -> `Strbuf_CopyLineNum`
* `sub_02023D28` -> `Strbuf_CopyChars`
* `sub_02023D8C` -> `Strbuf_CopyNumChars`
* `sub_02023DF0` -> `Strbuf_ToChars`
* `sub_02023E2C` -> `Strbuf_GetData`
* `sub_02023E4C` -> `Strbuf_Concat`
* `sub_02023EB0` -> `Strbuf_Append`
* `sub_02023EF8` -> `Strbuf_IsTrainerName`
* `sub_02023F10` -> `Strbuf_ConcatTrainerName`
* `sub_02023F8C` -> `Strbuf_UpperChar`

## New Enums and Typedefs

* `enum NumberPaddingMode`
* `typedef u16 Charcode`


src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+    Strbuf_Check(dst);
+
+    if ((dst->size + 1) < dst->maxsize) {
+        dst->data[dst->size++] = c;
+        dst->data[dst->size] = EOS;
+        return;
+    }
+
+    GF_ASSERT(0);
+}
```
luckytyphlosion, line 425: Not sure if we want to follow this style, but if we want to avoid mid-return statements, try the following two if they match. (discuss in Discord)
```c
    if ((dst->size + 1) < dst->maxsize) {
        dst->data[dst->size++] = c;
        dst->data[dst->size] = EOS;
    } else {
        GF_ASSERT(0);
    }
 ```
 
 ```c

    if ((dst->size + 1) >= dst->maxsize) {
        GF_ASSERT(0);
    } else {
        dst->data[dst->size++] = c;
        dst->data[dst->size] = EOS;        
    }
 ```

File: src/strbuf.c [hunk 2]
```diff
@@ -0,0 +1,483 @@
...
+    return strbuf->size > 0 && strbuf->data[0] == COMPRESSED_LEADER;
+}
+
+void Strbuf_ConcatTrainerName(Strbuf *dst, Strbuf *src)
+{
+    if (Strbuf_IsTrainerName(src)) {
+        Charcode *dstC = &dst->data[dst->size];
+        Charcode *srcC = &src->data[1];
+        s32 bit = 0;
+        s32 outsize = 0;
```
luckytyphlosion, line 443: outsize is too nondescript (looking at it I don't know what it means). Don't try to abbreviate variable names in general if it removes information about what the variable is. Also, it should have been camelCased (i.e. outSize)

src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+        Charcode cur = 0;
```
luckytyphlosion, line 444: cur is too nondescript (c as a catch-all for char would be better, curChar would also be fine)

src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+    } else {
+        Strbuf_Concat(dst, src);
+    }
+}
+
+void Strbuf_UpperChar(Strbuf *strbuf, int i)
+{
+    Strbuf_Check(strbuf);
+
+    if (strbuf->size > i) {
```
luckytyphlosion, line 478: Try `i < strbuf->size` if it matches.

src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+        if (strbuf->data[i] >= EN_a && strbuf->data[i] <= EN_z) {
+            strbuf->data[i] -= 26;
```
luckytyphlosion, line 480: `26` -> `(EN_z + 1 - EN_a)`

src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+        }
+
+        *dstC = EOS;
+        dst->size += outsize;
+    } else {
+        Strbuf_Concat(dst, src);
+    }
+}
+
+void Strbuf_UpperChar(Strbuf *strbuf, int i)
```
luckytyphlosion, line 474: `i` probably fine? maybe `index`, not sure.

include/strbuf.h
```diff
@@ -0,0 +1,43 @@
...
+    PADDING_MODE_ZEROES,
+} PaddingMode;
+
+Strbuf* Strbuf_Init(u32 size, u32 heapID);
+void Strbuf_Free(Strbuf *strbuf);
+void Strbuf_Clear(Strbuf *strbuf);
+void Strbuf_Copy(Strbuf *dst, const Strbuf *src);
+Strbuf* Strbuf_Clone(const Strbuf *src, u32 heapID);
+void Strbuf_FormatInt(Strbuf *dst, int num, u32 maxDigits, PaddingMode mode, BOOL whichCharset);
+void Strbuf_FormatU64(Strbuf *dst, u64 num, u32 maxDigits, PaddingMode mode, int whichCharset);
```
luckytyphlosion, line 27: Double check whether whichCharset is BOOL.

luckytyphlosion (changes requested):

src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+    Strbuf_Check(src);
+
+    if (dst->maxsize > src->size) {
+        memcpy(dst->data, src->data, (src->size + 1) * sizeof(Charcode));
+        dst->size = src->size;
+        return;
+    }
+
+    GF_ASSERT(0);
+}
```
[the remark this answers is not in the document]
lhearachel: I prefer avoiding nested statements wherever possible; the fewer `if`s and `else`s in the code, the better.

src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+    Charcode data[1];
+};
+
+#define SIZEOF_STRBUF_HEADER    (sizeof(Strbuf) - sizeof(Charcode))
+#define STRBUF_MAGIC            (0xB6F8D2EC)
+
+static inline void Strbuf_Check(const Strbuf *strbuf)
+{
+    GF_ASSERT(strbuf != NULL);
+    GF_ASSERT(strbuf->magic == STRBUF_MAGIC);
```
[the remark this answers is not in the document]
luckytyphlosion: Don't follow naming scheme from other repos if the name can be better, because the other repos are volatile and subject to change anyway, plus all three repos currently have vastly different struct names/whatever anyway (so there will need to be one big fix later down the line).

include/strbuf.h
```diff
@@ -0,0 +1,43 @@
+#ifndef POKEPLATINUM_STRBUF_H
+#define POKEPLATINUM_STRBUF_H
+
+typedef struct Strbuf_t Strbuf;
+typedef u16 Charcode;
```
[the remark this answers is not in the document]
lhearachel: Consensus is that we typedef as lowercase with a `_t` suffix.

`typedef u16 charcode_t;`

src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+        memcpy(dst->data + dst->size, src->data, (src->size + 1) * sizeof(Charcode));
+        dst->size += src->size;
+        return;
+    }
+
+    GF_ASSERT(0);
+}
+
+// Strbuf_Append
+void Strbuf_Append(Strbuf *dst, Charcode c)
```
luckytyphlosion, line 414: This is actually AppendChar (pokediamond uses AddChar, but Append is also fine and more in line with how string operations work)

src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+        }
+
+        dst->data[dst->size] = EOS;
+        return;
+    }
+
+    GF_ASSERT(0);
+}
+
+void Strbuf_FormatU64(Strbuf *dst, u64 num, u32 maxDigits, PaddingMode paddingMode, int whichCharset)
```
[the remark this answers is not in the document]
lhearachel: Functionally, they are the same; the awkward thing is that the checks are against 0, so it can't just be something like:

```c
const charcode_t *digitSet = (useJPCharset) ? sDigits_JP : sDigits_EN;
```

It's also kinda weird to me to classify this as a `BOOL` in that sense; what does `Strbuf_FormatU64(..., FALSE)` even mean?

Maybe it's better to take this back to being an enum.

src/strbuf.c
lhearachel: Better, I'll just define the uppercase characters and do `(CHAR_EN_A - CHAR_EN_a)`.

src/strbuf.c
lhearachel: Think `i` is fine here. It's a generically-understood index var.

include/strbuf.h
lhearachel: Oh, now I see what you meant. Yeah, they should be the same type on either function.

src/strbuf.c
lhearachel: We elected to allow mid-return statements.

Like before, the fewer nestings, the better.

src/strbuf.c
```diff
@@ -6,32 +6,32 @@
 #include "strbuf.h"
 
 struct Strbuf_t {
-    u16 maxsize;
+    u16 maxSize;
     u16 size;
-    u32 magic;
-    Charcode data[1];
+    u32 checksum;
```
luckytyphlosion, line 11: I think `integrity` is a better name since it's not actually calculated from a checksum function it's more like a check for if the string exists or not.

src/strbuf.c
```diff
@@ -437,35 +434,35 @@ BOOL Strbuf_IsTrainerName(Strbuf *strbuf)
...
         s32 bit = 0;
-        s32 outsize = 0;
-        Charcode cur = 0;
+        s32 charsAdded = 0;
+        charcode_t curChar = 0;
 
         while (TRUE) {
-            cur = (*srcC >> bit) & COMPRESSED_MASK;
+            curChar = (*srcChar >> bit) & COMPRESSED_MASK;
             bit += 9;
```
luckytyphlosion, line 445: not actually sure what bit is, but maybe comments to explain it would be better. (it might be too abstract to explain in the variable name). you can always add a `// TODO` comment if you don't want to comment it now so the PR can be merged.

luckytyphlosion (changes requested):

src/strbuf.c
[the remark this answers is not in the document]
lhearachel: Yeah, I'm flipping this back to an enum. It matches with it, and I think it reads better.

src/strbuf.c
```diff
@@ -0,0 +1,483 @@
...
+#include "charcode.h"
+#include "heap.h"
+#include "strbuf.h"
+
+struct Strbuf_t {
+    u16 maxsize;
+    u16 size;
+    u32 magic;
+    Charcode data[1];
+};
```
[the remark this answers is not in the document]
lhearachel: Consensus here was to make struct definitions transparent.

PR #17, opened 2023-08-01
Include nitrogfx subproject as a wrap

Nomura-RH:


.gitattributes
```diff
@@ -9,6 +9,7 @@
 *.sha1 text eol=lf
 *.json text eol=lf
 *.gmm text eol=lf
+*.wrap text eol=lf
```
mid-kid, line 12: at this point why not just `* text eol=lf`

mid-kid:
lgtm, merge at will

PR #21, opened 2023-08-10
Partially document roaming_pokemon.c

Nomura-RH:


src/roaming_pokemon.c
```diff
@@ -242,66 +244,66 @@ void sub_0206C404 (UnkStruct_0202D7B0 * param0, const int param1)
...
 {
     Pokemon * v0;
     UnkStruct_0206C638 * v1;
     UnkStruct_0202D7B0 * v2;
-    int v3;
+    int previouslyVisitedMapID;
     UnkStruct_02025E6C * v4;
-    int v5;
-    u8 v6;
+    int roamingSpeciesID;
```
mid-kid, line 254: SpeciesID and MapID are both using `ID` redundantly. A species is always an index, not a property of the species.
Overuse of ID elsewhere as well (slotID)...

And then there's saveData which imo is fine as just "save", but that one's more arguable.

mid-kid (changes requested):

src/roaming_pokemon.c
Nomura-RH: fixed

PR #30, opened 2023-08-27
Document BattleContext and related structs

lhearachel:
This is very much a mass-rename PR. It matches. No underlying code was actually touched or cleaned up.

Most of these names are made to match with those in [`pokeheartgold`](https://github.com/pret/pokeheartgold/blob/master/include/battle/battle.h), with some touch-ups for accuracy. These names might not be *exactly* accurate, but they are meant to be a good starting point for us to work with when looking at the rest of the battle code.


include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+    u32 moveNumLockedInto[MAX_BATTLERS];
+    u16 moveNumProtect[MAX_BATTLERS];
+    u16 moveNumHit[MAX_BATTLERS];
+    u16 moveNumHitBattler[MAX_BATTLERS];
+    u16 moveNumHitType[MAX_BATTLERS];
+    u16 moveNumPrevByBattler[MAX_BATTLERS];
+    u16 moveNumCopied[MAX_BATTLERS];
+    u16 moveNumCopiedHit[MAX_BATTLERS][MAX_BATTLERS];
+    u16 moveNumSketched[MAX_BATTLERS];
+    u16 moveNumSelected[MAX_BATTLERS];
```
mid-kid, line 249: moveNum → move

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+
+    u32 moveNumTemp;
+    u32 moveNumCurr;
+    u32 moveNumPrev;
+    u32 moveNumLockedInto[MAX_BATTLERS];
+    u16 moveNumProtect[MAX_BATTLERS];
+    u16 moveNumHit[MAX_BATTLERS];
+    u16 moveNumHitBattler[MAX_BATTLERS];
+    u16 moveNumHitType[MAX_BATTLERS];
+    u16 moveNumPrevByBattler[MAX_BATTLERS];
```
mid-kid, line 245: Isn't everything here ByBattler?

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+    u32 monSpeedValues[MAX_BATTLERS];
+    
+    u8 ioQueue[MAX_LINK_BATTLERS][MAX_BATTLERS][BATTLE_IO_QUEUE_SIZE];
+    u8 ioBuffer[MAX_BATTLERS][BATTLE_IO_BUFFER_SIZE];
+    int battleScript[BATTLE_SCRIPT_SIZE_MAX];
+
+    UnkStruct_ov16_02252060 battleMons[MAX_BATTLERS];
+
+    u32 moveNumTemp;
+    u32 moveNumCurr;
```
mid-kid, line 238: I'm not gonna list this everywhere, but the most used short for current in this repo is `cur`, not `curr`.

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+    u16 conversion2Battler[MAX_BATTLERS];
+    u16 conversion2Type[MAX_BATTLERS];
+    u16 metronomeMoveNum[MAX_BATTLERS];
+
+    int storedDamage[MAX_BATTLERS];
+    int lastHitByBattler[MAX_BATTLERS];
+    int battlerIdSpeedTemp;
+    u8 battlersSwitchingMask;
+    u8 levelUpMons;
+    u16 padding310A;    // not used?
```
mid-kid, line 261: Is this not some struct boundary padding?

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+    BOOL expJinglePlayed;
+
+    int linkBattleTimeout;
+
+    u8 recordedCommandFlags[MAX_BATTLERS];
+
+    int waitingBattlers;
+
+    u32 battleProgressFlag : 1;
+    u32 padding3154_01 : 31;
```
mid-kid, line 290: The compiler pads this automatically.

include/constants/battle.h
```diff
@@ -19,4 +19,22 @@
 #define BATTLE_TYPE_AI_PARTNER      (BATTLE_TYPE_DOUBLES | BATTLE_TYPE_2vs2 | BATTLE_TYPE_AI)
 #define BATTLE_TYPE_SPECIAL_WILD    (BATTLE_TYPE_SAFARI | BATTLE_TYPE_ROAMER | BATTLE_TYPE_PAL_PARK | BATTLE_TYPE_CATCH_TUTORIAL)
 
+#define MAX_LINK_BATTLERS   4
+#define MAX_BATTLERS        4
+#define MAX_BATTLE_SIDES    2
```
mid-kid: I would go for `NUM_` here.

include/battle/ai_context.h
```diff
@@ -0,0 +1,53 @@
...
+    u8 scriptStackSize;
+
+    u8 trainerItemCounts[MAX_BATTLERS_PER_SIDE];
+    u8 battlerIdAttacker;
+    u8 battlerIdDefender;
+    u8 usedItemType[MAX_BATTLERS_PER_SIDE];
+    u8 usedItemCondition[MAX_BATTLERS_PER_SIDE];
+    u8 padding0082;     // originally implicit, declaring here as explicit
+
+    u16 usedItemNums[MAX_BATTLERS_PER_SIDE];
```
mid-kid, line 44: This name makes no sense, each battler has only one entry, how can there be multiple ItemNums for them?
Confused as to how this is used.

include/battle/ai_context.h
```diff
@@ -0,0 +1,53 @@
...
+
+    u32 scriptStackPointer[8];
+    u8 scriptStackSize;
+
+    u8 trainerItemCounts[MAX_BATTLERS_PER_SIDE];
+    u8 battlerIdAttacker;
+    u8 battlerIdDefender;
+    u8 usedItemType[MAX_BATTLERS_PER_SIDE];
+    u8 usedItemCondition[MAX_BATTLERS_PER_SIDE];
+    u8 padding0082;     // originally implicit, declaring here as explicit
```
mid-kid, line 42: Don't do that.

include/battle/battle_mon.h
```diff
@@ -0,0 +1,132 @@
...
+    
+    u32 battlerIdLockOn     :2,
+        mimickedMoveSlot    :4,
+        battlerIdBinding    :2,
+        battlerIdMeanLook   :2,
+        lastResortCount     :3,
+        magnetRiseTurns     :3,
+        healBlockTurns      :3,
+        embargoTurns        :3,
+        hasHeldItem         :1, // this is specifically for tracking Unburden; 0 if no item was ever held, or if item is suppresed with Knock Off
```
mid-kid, line 31: long comment. Either put it above the line and split it up (easier if you don't do the bitfields comma thing), or make the name reflect its use better.

include/battle/battle_mon.h
```diff
@@ -0,0 +1,132 @@
...
+        micleBerryActive    :1,
+        custapBerryActive   :1,
+        quickClawActive     :1,
+        meFirstActive       :1,
+        padding0004_1F      :1;
+
+    int rechargeTurnCount;
+    int fakeOutTurnCount;
+    int slowStartTurnCount;
+    int meFirstTurnCount;
```
mid-kid, line 42: Why not Turns like above?

include/battle/battle_mon.h
```diff
@@ -0,0 +1,132 @@
...
+
+    s8 statBoosts[NUM_BOOSTABLE_STATS];
+
+    int weight;
+
+    u8 type1;
+    u8 type2;
+    u8 formNum          :5,
+       isShiny          :1,
+       padding0026_06   :2;
```
mid-kid, line 83: more implied padding

include/battle/battle_mon.h
```diff
@@ -0,0 +1,132 @@
...
+    u32 statusVolatile;
+    u32 originalTrainerId;
+
+    u16 heldItem;
+    u16 padding007A;    // unused
+
+    u8 timesDamaged;
+    u8 trainerMessageFlags;
+    u8 gender                   :4,
+       originalTrainerGender    :4;
```
mid-kid, line 124: OTGender

include/battle/battle_mon.h
```diff
@@ -0,0 +1,132 @@
...
+
+    u8 level;
+    u8 friendship;
+
+    u16 nickname[MON_NAME_LEN + 1];
+
+    s32 curHP;
+    u32 maxHP;
+
+    u16 originalTrainerName[PERSON_NAME_LEN + 1];
```
mid-kid, line 110: OTName

mid-kid: (This is why I hate camel case)

include/constants/string.h
```diff
@@ -0,0 +1,7 @@
+#ifndef POKEPLATINUM_CONSTANTS_STRING_H
+#define POKEPLATINUM_CONSTANTS_STRING_H
+
+#define MON_NAME_LEN    10
+#define PERSON_NAME_LEN 7
```
mid-kid, line 5: TRAINER_NAME_LEN

include/battle/battle_message.h
```diff
@@ -0,0 +1,13 @@
+#ifndef POKEPLATINUM_BATTLE_MESSAGE_H
+#define POKEPLATINUM_BATTLE_MESSAGE_H
+
+typedef struct BattleMessage {
+    u8 commandCode;
+    u8 tags;
+    u16 id;
+    int params[6];
+    int digits;
+    int battlerId;
```
mid-kid, line 10: battler

mid-kid (changes requested):
I'd also generally prefer comments before the affecting line, over end-of-line comments, where appropriate. Don't align comments on tabstops or over huge structs, one space is fine.

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+    u8 futureSightTurns[MAX_BATTLERS];
+    u8 wishTurns[MAX_BATTLERS];
+    u16 futureSightMoveNum[MAX_BATTLERS];
+    int futureSightAttackerId[MAX_BATTLERS];
+    s32 futureSightDamage[MAX_BATTLERS];
+    u8 wishTarget[MAX_BATTLERS];
+} FieldConditions;
+
+typedef struct SideConditions {
+    u32 reflectBattlerId        :2,
```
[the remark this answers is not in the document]
lhearachel: That seems like a fine compromise to me. I'm mostly worried about the readability of var names in big blocks like this:

```c
u32 reflectBattlerId:2;
u32 reflectTurns:3;
... etc ...
```

It starts to bleed together after enough entries.

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+        reflectTurns            :3,
+        lightScreenBattlerId    :2,
+        lightScreenTurns        :3,
+        mistBattlerId           :2,
+        mistTurns               :3,
+        safeguardBattlerId      :2,
+        safeguardTurns          :3,
+        followMeActive          :1,
+        followMeBattlerId       :2,
```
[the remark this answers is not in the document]
lhearachel: I actually have some ideas on how I would condense a lot of these names down.

`followMeBattlerId` specifically refers to the battler on a given team that has the highest priority Follow Me (i.e., the first that used it). From [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Follow_Me_(move)):

> If another Pokémon on the same team is already the center of attention, the first user takes priority.

I like `followMeUser`.

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+        knockedOffItemBattlers  :6,
+        padding00_1D            :3;
+
+    u32 spikesLayers        :2,
+        toxicSpikesLayers   :2,
+        padding04_04        :28;
```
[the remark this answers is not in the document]
lhearachel: I would rather we make note of padding instances like these in some capacity rather than not at all. At the very least, I want a comment here noting that there are still 28 bits to constitute a full 32-bit value.

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+        protecting          :1,
+        helpingHandActive   :1,
+        magicCoatActive     :1,
+        snatching           :1,
+        roosting            :1,
+        fleeing             :2,
+        enduring            :1,
+        padding00_0A        :22;
+
+    int physicalDamageTakenByAttacker[MAX_BATTLERS];
```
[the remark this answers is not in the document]
lhearachel: This array is meant to represent physical damage that has been taken by a given Pokemon from a particular source. To illustrate, suppose we're looking at the Pokemon in the player's slot 1 (which is battler 0):

```c
physicalDamageTaken[MAX_BATTLERS] = {
    0, // 0 damage dealt to this Pokemon by itself on this turn
    15, // 15 damage dealt to this Pokemon by slot 1 on the enemy team on this turn
    35, // 35 damage dealt to this Pokemon by its partner on this turn (e.g. by Earthquake)
    10, // 10 damage dealt to this Pokemon by slot 2 on the enemy team on this turn
}
```

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+    int specialDamageLastAttacker;
+    int specialDamageAttackerMask;
+
+    int lastDamageTaken;
+    int lastAttacker;
+    int assuranceDamageMask;
+} TurnFlags;
+
+typedef struct SelfTurnFlags {
+    u32 skipPressureCheck       :1,
```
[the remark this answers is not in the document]
lhearachel: It's not one, actually; it's to make sure that a move like Spikes that can target the opponent but fails during execution does not actually get extra PP deducted by Pressure.

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+
+    int battlerIdAttacker;
+    int battlerIdAttackerTemp;
+    int battlerIdDefender;
+    int battlerIdDefenderTemp;
+    int battlerIdFainted;
+    int battlerIdSwitched;
+    int battlerIdSwitchedTemp;
+    int battlerIdAbility;
+    int battlerIdMagicCoat;
```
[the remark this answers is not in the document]
lhearachel: I would vastly prefer that we drop as much of the `battlerId` nomenclature as possible, even if means that we can't fully standardize the names. There is something elegant about being able to write `ctx->attacker`, though it doesn't quite work for something like `ctx->switching` or `ctx->fainting` (Are those flags denoting that someone is switching or fainting? Or are they battler IDs? Not immediately clear at a glance.)

Maybe we could adopt a mix, but that doesn't feel super clean, either:

```c
int attacker;
int attackerTemp;
int defender;
int defenderTemp;
int faintingMon;
int switchingMon;
int switchingMonTemp;
int abilitySource;
int magicCoatSource;
```

include/battle/battle_context.h
```diff
@@ -0,0 +1,294 @@
...
+
+    BOOL multiHitLoop;
+    u32 multiHitCheckFlags;
+    u32 multiHitAccuracyCheck;
+
+    u32 clearVolatileStatus[MAX_BATTLERS];
+
+    u8 selectedMonIndex[MAX_BATTLERS];
+    u8 switchSelectedMonIndex[MAX_BATTLERS];
+    u8 aiSwitchSelectedMonIndex[MAX_BATTLERS];
```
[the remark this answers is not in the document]
lhearachel: I like:

```c
selectedPartySlot
switchPartySlot
```

include/constants/battle.h
lhearachel: I like `NUM_BATTLE_SIDES`, but the rest are actual maximums, since 1v1 battles exist.

include/battle/ai_context.h
lhearachel: It's definitely one item per battler; my bad on keeping this one plural to make that confusing.

include/battle/battle_mon.h
lhearachel: All the instances of `Turns` refer to a number of relative turns that get modified accordingly. All of these values refer to a *specific* turn on the counter. For example, whenever a mon comes into play, its `fakeOutTurnCount` is set to the current turn + 1, and that value is checked against the current turn whenever it tries to use Fake Out.

mid-kid: I see, that makes more sense! Maybe `fakeOutRemainingTurns` or `fakeOutTurnsRemain`?

include/battle/ai_context.h
```diff
@@ -0,0 +1,52 @@
...
+    u32 scriptStackPointer[8];
+    u8 scriptStackSize;
+
+    u8 trainerItemCounts[MAX_BATTLERS_PER_SIDE];
+    u8 attacker;
+    u8 defender;
+    u8 usedItemType[MAX_BATTLERS_PER_SIDE];
+    u8 usedItemCondition[MAX_BATTLERS_PER_SIDE];
+
+    u16 usedItem[MAX_BATTLERS_PER_SIDE];
```
mid-kid, line 43: does this gap still make sense?

include/constants/battle.h
mid-kid: I see, I didn't think of that. In that case, keep it as `MAX_`.

include/battle/battle_context.h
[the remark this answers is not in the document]
mid-kid: No, in that case `ByAttacker` makes sense, though it's a bit confusing. Leaving it out means that the array contains the total damage taken for each battler, not the damage taken from each different source. Maybe `physicalDamageTakenFrom`?

mid-kid:
LGTM now, resolve the remaining conversations and merge at will.

include/battle/battle_context.h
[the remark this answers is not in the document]
lhearachel: > Leaving it out means that the array contains the total damage taken for each battler, not the damage taken from each different source.

I'd argue that there's some context here that this is a substruct used for the `turnFlags` field, and that field is indexed by the individual battlers already. Maybe that's too much of a jump in context, and I don't think it really costs us anything here to be a little more clear. I like `physicalDamageTakenFrom`.

include/constants/battle.h
lhearachel: I'm keeping `NUM_BATTLE_SIDES`; we agreed in the `pokeheartgold` issue that `NUM` is useful for constant cardinalities, and there are *always* 2 sides in a battle. But there might be 2-4 active battlers in a given battle, so that one is a `MAX`.

include/battle/battle_mon.h
lhearachel: `fakeOutTurnNumber`; the comparisons used for this field (and others like it) are always against the total turn counter in the parent `BattleContext` struct.

PR #34, opened 2023-08-29
Move NitroDWC add-in libs to a separate subproject

Nomura-RH:


lhearachel (commented):
LGTM, just the one weird alignment thing

PR #41, opened 2023-09-01
Document item.c

lhearachel:


include/item.h
```diff
@@ -75,27 +75,266 @@ typedef struct ItemData {
...
+/**
+ * @brief Get an ID of the specified type for an item.
+ *
+ * Each type points to a different value for the item's entry in sItemArchiveIDs.
+ *
+ * @param item
+ * @param type      The type of ID to get. 
+ * @return The item's ID of the specified type.
+ */
+u16 Item_ID(u16 item, enum ItemFileType type);
```
Nomura-RH, line 168: I feel like this function should be called something more descriptive, but I'm having a hard time coming up with a better name. Maybe `Item_GetIDInItemFile`?

Nomura-RH (commented):
Aside from one minor comment about a function name, LGTM

include/item.h
lhearachel: How about `Item_FileID`? All the elements accessed here are pointers to a file in a NARC somewhere, with the exception of the GBA IDs.

Nomura-RH: Sounds good to me

PR #46, opened 2023-09-05
Decompile pl_waza_tbl.narc and rebuild using new tool json2bin

lhearachel:
This is a very rudimentary first approach at building a data NARC out of component JSON files. I expect this tool to evolve over time as we make use of it, but there should be enough here to get an idea for how to work with it at a basic level.


tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+    write(output_bin, output_idx, output_dir)
+
+
+def json2bin(target: str,
+             schema: Parser,
+             output_dir: str | None,
+             index_func: FunctionType,
+             batch_mode: bool):
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
```
luckytyphlosion, line 162:
```python
pathlib.Path(output_dir, exist_ok=True, parents=True)
```

tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+    with open(output_fname, 'wb') as output_file:
+        output_file.write(output_bin)
+
+
+def process(fname_in: str,
+            schema: Parser,
+            output_dir: str | None,
+            index_func: FunctionType):
+    if fname_in[-5:] != '.json':
+        return
```
luckytyphlosion, line 148:
```python
pathlib.Path(fname_in).suffix == ".json"
```

tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+    
+    output_bin = parse(fname_in, schema)
+    simple_name = os.path.basename(os.path.normpath(fname_in))
```
luckytyphlosion, line 151:
```python
pathlib.Path(fname_in).name # create variable for pathlib.Path(fname_in)
```

tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+    output_idx = index_func(simple_name.split(".")[0].upper())
```
luckytyphlosion, line 152:
```python
simple_name.split(".")[0].upper()
```
->
```python
pathlib.Path(simple_name).stem.upper() # don't choke on files with multiple dot
```

tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+    prog='json2bin.py',
+    description='Tool for converting JSON documents to binary data via a constructed parsing schema'
+)
+ARGPARSER.add_argument('-b', '--batch-mode', action='store_true')
+ARGPARSER.add_argument('-i', '--input', required=True)
+ARGPARSER.add_argument('-o', '--output')
+
+
+class Parser():
+    def __init__(self):
```
luckytyphlosion, line 20: add in `__slots__`

tools/json2bin/consts/moves.py
```diff
@@ -0,0 +1,474 @@
+from enum import Enum, auto
+
+class Move(Enum):
+    MOVE_0000 = 0
```
luckytyphlosion, line 4: Something something maybe move constants should be defined from the jsons alone? idk

tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+        input_json = json.load(input_file)
+        return schema.parse(input_json)
+
+
+def write(output_bin: bytes, output_idx: int, output_dir: str | None):
+    output_fname = f'{output_idx:04}.bin'
+    if output_dir:
+        output_fname = os.path.join(output_dir, output_fname)
+    
+    with open(output_fname, 'wb') as output_file:
```
luckytyphlosion, line 139: `wb+`? (To clear the file if it exists)

tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+    description='Tool for converting JSON documents to binary data via a constructed parsing schema'
+)
+ARGPARSER.add_argument('-b', '--batch-mode', action='store_true')
+ARGPARSER.add_argument('-i', '--input', required=True)
+ARGPARSER.add_argument('-o', '--output')
+
+
+class Parser():
+    def __init__(self):
+        self.registry = OrderedDict()
```
luckytyphlosion, line 21: Python dictionaries are ordered by default since 3.7

tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+            Parse the given JSON structure according to the currently-defined
+            data schema.
+        '''
+        binary = bytearray([])
+        for key in self.registry:
+            if key == 'name':
+                self.registry[key](data[key])
+            elif key.startswith('padding_'):
+                size, val = self.registry[key]
+                binary = binary + val.to_bytes(size, 'little')
```
luckytyphlosion, line 81:
```python
binary.extend(val.to_bytes(size, 'little'))
```

tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+            else:
+                parse_func, size, const_type = self.registry[key]
+                data_val = self._walk(data, key.split('.'))
+                binary = binary + parse_func(data_val, size, const_type)
```
luckytyphlosion, line 85:
```python
binary.extend(parse_func(data_val, size, const_type))
```

tools/json2bin/json2bin.py
```diff
@@ -0,0 +1,168 @@
...
+def json2bin(target: str,
+             schema: Parser,
+             output_dir: str | None,
+             index_func: FunctionType,
+             batch_mode: bool):
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
+    
+    if batch_mode:
+        for fname_in in os.listdir(target):
```
luckytyphlosion, line 165: Might eventually want to change to a recursive glob in the future.

luckytyphlosion (changes requested):

tools/json2bin/json2bin.py
luckytyphlosion: Also you should use `split(".", maxsplit=1)` as python will use a separate algorithm for splitting there.

tools/json2bin/consts/moves.py
lhearachel: That's what pfero and I have been trying to figure out

luckytyphlosion: Probably not necessary yet.

tools/json2bin/json2bin.py
lhearachel: Yeah, it's on my mind. Out of scope for now.

tools/json2bin/consts/moves.py
lhearachel: Yeah, calling this one out of scope for now; we can revisit it as a problem later if it comes to it.

tools/json2bin/json2bin.py
```diff
@@ -134,22 +137,22 @@ def parse(fname_in: str, schema: Parser) -> bytes:
...
             index_func: FunctionType):
-    if fname_in[-5:] != '.json':
+    if pathlib.Path(fname_in).suffix != '.json':
         return
     
     output_bin = parse(fname_in, schema)
-    simple_name = os.path.basename(os.path.normpath(fname_in))
-    output_idx = index_func(simple_name.split(".")[0].upper())
+    simple_name = pathlib.Path(fname_in).name
+    output_idx = index_func(pathlib.Path(simple_name).stem.upper())
```
luckytyphlosion, line 155: Apparently I misread this statement but rewrite it to be:
```python
    fname_inpath = pathlib.Path(fname_in)
    if fname_inpath .suffix != '.json':
        return

    output_bin = parse(fname_in, schema)
    simple_name = pathlib.Path(fname_in).name
    output_idx = index_func(fname_inpath.stem.upper())
```

luckytyphlosion: Sorry, it should be:
```python
```python
    fname_inpath = pathlib.Path(fname_in)
    if fname_inpath .suffix != '.json':
        return

    output_bin = parse(fname_in, schema)
    output_idx = index_func(fname_inpath.stem.upper())
```

lhearachel: Yeah that's exactly what I ended up with.

res/poketool/waza/meson.build
```diff
@@ -1,5 +1,30 @@
+subdir('pl_waza_tbl')
+
+pl_waza_tbl_files = custom_target('pl_waza_tbl_files',
+    output: 'pl_waza_tbl_files',
```
Nomura-RH, line 4: The value of `output` should match the names of the files that actually get generated by the command, or else meson will always rerun the latter, because it can't find the build artifacts.

Nomura-RH (changes requested):
I was trying to time converting each json separately, instead of using batch mode, but I keep getting this error:

`res/poketool/waza/pl_waza_tbl.narc.p/0468.bin
"C:/msys64/mingw64/bin/python.exe" "C:/Users/fseme/Desktop/pokeplatinum/tools/json2bin/movedata.py" "-i" "../res/poketool/waza/pl_waza_tbl/0468.json" "-o" "res/poketool/waza/pl_waza_tbl.narc.p/0468.bin"
Traceback (most recent call last):
  File "C:\Users\fseme\Desktop\pokeplatinum\tools\json2bin\movedata.py", line 32, in <module>
    j2b.json2bin(args.input, SCHEMA, args.output, indexer, args.batch_mode)
  File "C:/Users\fseme\Desktop\pokeplatinum\tools\json2bin/json2bin.py", line 170, in json2bin
    process(pathlib.Path(target) / fname_in, schema, output_dir, index_func)
UnboundLocalError: local variable 'fname_in' referenced before assignment`

lhearachel:
> I was trying to time converting each json separately, instead of using batch mode, but I keep getting this error

That's my bad, I didn't test that functionality out again before pushing a recent change. Fixed.

res/poketool/waza/meson.build
lhearachel: I refactored the build script to generate the targets sequentially rather than build all the binaries at once. The new code in `res/poketool/waza/pl_waza_tbl/meson.build` is:

```
foreach src_file : pl_waza_tbl_srcs
    if i < 10
        bin_name = '000' + i.to_string() + '.bin'
    elif i < 100
        bin_name = '00' + i.to_string() + '.bin'
    elif i < 1000
        bin_name = '0' + i.to_string() + '.bin'
    else
        bin_name = i.to_string() + '.bin'
    endif

    pl_waza_tbl_bins += custom_target(bin_name,
        output: bin_name,
        input: files(src_file),
        command: [
            movedata_py,
            '-i', '@INPUT@',
            '-o', '@OUTDIR@'
        ]
    )

    i += 1
endforeach
```

I modeled this after how the `pl_msg` narc is built: https://github.com/pret/pokeplatinum/blob/9af28d6d28434d51dad12ead44fd18560a7d0ab7/res/msgdata/pl_msg/meson.build#L730

tools/json2bin/json2bin.py
lhearachel: I wound up doing this anyways.

res/moves/meson.build
```diff
@@ -0,0 +1,19 @@
+pl_waza_tbl_subdirs = run_command('ls', '-d', '*/').stdout().splitlines()
```
Nomura-RH, line 1: As a general rule, source files and directories should be explicitly listed. Globbing/listing them through an external tool is hack-ish, as Meson will not pick up new files when they are added, Refer to: https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard

tools/json2bin/json2bin.py
```diff
@@ -9,11 +9,14 @@
 
 ARGPARSER = ArgumentParser(
     prog='json2bin.py',
-    description='Tool for converting JSON documents to binary data via a constructed parsing schema'
+    description='Tool for converting a collection of JSON documents into\na NARC via a constructed parsing schema'
 )
-ARGPARSER.add_argument('-b', '--batch-mode', action='store_true')
-ARGPARSER.add_argument('-i', '--input', required=True)
-ARGPARSER.add_argument('-o', '--output')
+ARGPARSER.add_argument('--knarc', required=True,
```
Nomura-RH, line 14: Don't remember if I already expresses my opinion on the matter, but I really dislike calling knarc from other tools. If we don't want to have two separate steps (i.e. json -> bins, and bins -> NARC), I'd rather do what the final rom target does (i.e. bakes several steps into one custom target by calling the shell executable and passing it the commands).

.gitignore
```diff
@@ -83,3 +83,5 @@ sdk/
 
 !tools/asmdiff/asmdiff.sh
 
+tools/**/__pycache__
```
Nomura-RH, line 86: nitpicking: couldn't we just have ` __pycache__`  here?

res/moves/meson.build
lhearachel: As I've outlined previously, assuming that we do that, this means that there are two options:

1. We pass all of the input sources through the arg-list. This is, frankly, untenable; there are almost 500 moves here.
2. We create each of the binary files individually and stitch them together later. This isn't unreasonable, but it means that adding a new move is a more cumbersome process for an end-user, because now they have to interface with the build-system to make sure their files are included.

tools/json2bin/json2bin.py
lhearachel: csv2bin basically already does this chain implicitly, just with its own internal narc implementation. What's the difference for you?

res/moves/meson.build
Nomura-RH: > As I've outlined previously, assuming that we do that, this means that there are two options:
> 
> 
> 
> 1. We pass all of the input sources through the arg-list. This is, frankly, untenable; there are almost 500 moves here.
> 

I'm not sure I explained myself correctly. You would just have to list the move subdirectories explicitly instead of relying on `ls`. It's totally fine to then iterate over that list to build source file paths, like the rest of your code is already doing.

> 2. This isn't unreasonable, but it means that adding a new move is a more cumbersome process for an end-user, because now they have to interface with the build-system to make sure their files are included. Indeed, I can foresee many a query pertaining to "I made a new move, why isn't it showing up?" whenever people start trying to use this for hacks because the fail-case in this scenario is silent.

That's gonna happen even as things are now. Meson won't regenerate the build system just because you added a new move. You would have to manually wipe your build tree and reconfigure directly. I personally believe having users add an entry into a list in a meson.build file isn't the end of the world. It's better than having to perform a reconfigure.

tools/json2bin/json2bin.py
Nomura-RH: The difference is knarc is a separate tool, so if we were to make any breaking changes to how it works, or even replace it with something else, we'd have to also update all the other tools that are calling it.

Whereas if you have two separate steps, changing knarc will just require you to modify the build files.

lhearachel: Can I make a proposition, then, that we accept this as a stopgap and update json2bin to have an operable narc mode that compiles the narc, like csv2bin? Would that be amenable? I can file an issue for it.

Nomura-RH: I also want to add that csv2bin having its own NARC implementation is an issue of its own: we should have only one common NARC implementation across all tools.

Nomura-RH: > Can I make a proposition, then, that we accept this as a stopgap and update json2bin to have an operable narc mode that compiles the narc, like csv2bin? Would that be amenable? I can file an issue for it.

Yes, I'll definitely take the current setup as a stopgap. I also really want to see this PR merged. 
Definitely file an issue to discuss about what we should do with NARC implementations.

lhearachel: Done: https://github.com/pret/pokeplatinum/issues/53

PR #49, opened 2023-09-08
Dump the contents of pl_pokegra and rebuild from sprites

lhearachel:
The first commit is just generated by a dump script (BluRosie gave me a zip of all of these). The more important part is the second commit, which contains the actual build logic.

All the sprites are optimized using `optipng`, specifically `v0.7.7`. The following command was used to optimize them:

```
find res/poketool/pokegra/pl_pokegra -name '*.png' -exec optipng -o7 -strip all {} +
```


res/poketool/pokegra/pl_pokegra/meson.build
```diff
@@ -0,0 +1,92 @@
...
+sprites_flags = ['-scanfronttoback', '-handleempty']
+palette_flags = ['-bitdepth', '8', '-nopad', '-comp', '10']
+
+foreach i : range(494)
+    padded_i = run_command('printf', '%04d', i.to_string(), check: false).stdout()
+
+    female_back = files(padded_i / 'female' / 'back.png')
+    male_back = files(padded_i / 'male' / 'back.png')
+    female_front = files(padded_i / 'female' / 'front.png')
+    male_front = files(padded_i / 'male' / 'front.png')
```
mid-kid: Is explicit files() here really necessary? In many cases, these are automatically casted.

res/poketool/pokegra/pl_pokegra/meson.build
```diff
@@ -0,0 +1,92 @@
...
+    )
+
+    if i == 0
+        # nitrogfx doesn't like these two files, so copy over the canonical versions
+        pl_pokegra_files += fs.copyfile('0000/0000-04.NCLR', '0000-04.NCLR')
+        pl_pokegra_files += fs.copyfile('0000/0000-05.NCLR', '0000-05.NCLR')
+        continue
+    endif
+
+    if run_command('test', '-s', male_front, check: false).returncode() == 0
```
mid-kid: Use the filesystem module instead
https://mesonbuild.com/Fs-module.html#is_symlink

res/poketool/pokegra/pl_pokegra/meson.build
```diff
@@ -0,0 +1,92 @@
...
+palette_flags = ['-bitdepth', '8', '-nopad', '-comp', '10']
+
+foreach i : range(494)
+    padded_i = run_command('printf', '%04d', i.to_string(), check: false).stdout()
+
+    female_back = files(padded_i / 'female' / 'back.png')
+    male_back = files(padded_i / 'male' / 'back.png')
+    female_front = files(padded_i / 'female' / 'front.png')
+    male_front = files(padded_i / 'male' / 'front.png')
+    pl_pokegra_srcs += [female_back, male_back, female_front, male_front]
```
mid-kid: What are you using this array for?

res/poketool/pokegra/pl_pokegra/meson.build
```diff
@@ -0,0 +1,92 @@
...
+        command: [
+            nitrogfx_exe,
+            '@INPUT@',
+            '@OUTPUT@',
+            sprites_flags
+        ]
+    )
+
+    if i == 0
+        # nitrogfx doesn't like these two files, so copy over the canonical versions
```
mid-kid: Research pending, of course, but I wonder if the original tools would compress/format these properly.

mid-kid (changes requested):

res/poketool/pokegra/pl_pokegra/meson.build
lhearachel: Does that work? The files *do* exist in these cases, they just have size 0.

mid-kid: Oh, in that case https://mesonbuild.com/Fs-module.html#size
I'm not entirely sure what the check is for, anyhow.

lhearachel: It's just to make sure we're checking a file that actually has data for compiling the palette out of it. Ungendered mons have their sprites stored as `male`, so it's really just to check if the 'mon in question is female-only (e.g., Nidoqueen).

mid-kid: Why does the male sprite exist at all, then?

lhearachel: Because the NARC mandates it. Vanilla has the same structure, having 0-length files for non-existent gender-forms.

mid-kid: In that case I'd create empty files with meson if the source doesn't exist. You can loop through a {male,female,front,back} array to simplify the logic.

mid-kid: Instead of storing empty PNGs I mean

res/poketool/pokegra/pl_pokegra/meson.build
lhearachel: I imagine so. `nitrogfx` gets close, but it replaces the long pairs of `FF FF` at the end of the palette file with `FF 7F`. I'm not quite sure why, and I'm not familiar enough with `nitrogfx` myself to debug it.

res/poketool/pokegra/pl_pokegra/meson.build
lhearachel: > In that case I'd create empty files with meson if the source doesn't exist. You can loop through a {male,female,front,back} array to simplify the logic.

That's fair. I'll implement that tomorrow and clean up the original dump commit.

res/poketool/pokegra/pl_pokegra/meson.build
mid-kid: >I'm not quite sure why
In an RGB555 palette, when stored in 16 bits, the highest bit is unused. I figure nitrogfx used it to signify a missing color maybe?

lhearachel: It's probably a missing color, then, yeah. Is there a way we could modify nitrogfx to support it? Maybe a new input flag?

lhearachel:
This PR now also includes a build for `pl_otherpoke`, and all the sprites have been moved to make use of the new structure under `res/pokemon`.

PR #62, opened 2023-09-18
Simplify building message data

Nomura-RH:
This PR refactors msgenc to read an embedded encryption key from a gmm, and uses this to build the binaries for `pl_msg.narc` with `generator()` instead of a loop that calls `custom_target()`.


lhearachel (approved):
LGTM

PR #67, opened 2023-09-23
Document town map

gensarde:


include/overlay080/struct_town_map_city_holder.h
```diff
@@ -0,0 +1,14 @@
+#ifndef POKEPLATINUM_STRUCT_TOWN_MAP_CITY_HOLDER_H
+#define POKEPLATINUM_STRUCT_TOWN_MAP_CITY_HOLDER_H
+
+#include "overlay080/struct_town_map_city.h"
+
+typedef struct TownMapCityHolder {
```
lhearachel, line 6: Is this just the full map struct with the blocks arranged?

Nomura-RH: > Is this just the full map struct with the blocks arranged?

Nope, it's just a collection of blocks (actually cities, which are blocks + some other data)

lhearachel: Right, but what I'm curious about is what it represents. Is it all the blocks that go on the map? Is it a set of blocks for those multi-block cities (like how Jubilife or Hearthome are 2x2 squares, whereas Sandgem is a single block)?

Nomura-RH: > Right, but what I'm curious about is what it represents. Is it all the blocks that go on the map? Is it a set of blocks for those multi-block cities (like how Jubilife or Hearthome are 2x2 squares, whereas Sandgem is a single block)?

Yes, it holds the data for all the city (plus Pal Park and Victory Road) blocks that go on the map. There are various kind of blocks: 1x1 (like Sandgem), 2x2 (like Jubilife), L-shaped and Pal-Park style (green circle). Each city -> 1 block.

lhearachel:
Closing as abandoned

PR #68, opened 2023-09-23
Add Knarc cmd support for multiple input files and response files

wildfire248:
Title. Command format is changed a bit to line up with the changes.
Also added checksum tests for individual files in res/ ,was useful for catching issues while i edited knarc.
(Note: naming is a bit inconsistent, but it was inconsistent before i touched it anyways)


platinum.us/res.sha1
```diff
@@ -0,0 +1,341 @@
+42e2b71655a74d6e5bffb9ec75a25df930b20c59 res/battle/moves/pl_waza_tbl.narc
+683a7691d685c153e738b136b8266deefb0b4552 res/items/nuts_data.narc
+70d103831556bb58ae8e602adbcce99333a22889 res/items/pl_item_data.narc
+3614557aa04eb1ad7e5f260578de925f73da584d res/text/pl_msg.narc
+
+1f8689cbc763d9efedac9e6f12e940dbd361f7a9 res/prebuilt/poketool/personal/pl_personal.narc
```
lhearachel, line 6: Do we have to include the prebuilt narcs? Does it make sense for us to include the prebuilt narcs' shasums?

tools/knarc/Narc.cpp
```diff
@@ -60,125 +85,184 @@ bool Narc::Cleanup(ofstream& ofs, const NarcError& e)
...
-
-                  for (size_t i = 0; i < bStr.size(); ++i)
-                  {
-                      bStr[i] = tolower(bStr[i]);
-                  }
-
-                  return aStr < bStr;
-              });
+
+    if(lexSort)
```
lhearachel, line 180: [nit] `if (lexSort)`

tools/knarc/Narc.cpp
```diff
@@ -60,125 +85,184 @@ bool Narc::Cleanup(ofstream& ofs, const NarcError& e)
...
-            // read the filenames in the order file and add the corresponding directory entries to the ordered files vector
-            std::string filename;
-            while (std::getline(order_file, filename))
+
+            // if recursive flag is set, search for knarcorder files in subdirectories and process them recursively
+            if (recursive)
             {
-                fs::path file_path = path / filename;
-                if (fs::exists(file_path))
+                for (auto &entry : fs::directory_iterator(path))
```
lhearachel, line 145: [nit] `auto& entry`

tools/knarc/Narc.cpp
```diff
@@ -60,125 +85,184 @@ bool Narc::Cleanup(ofstream& ofs, const NarcError& e)
...
-        });
+                    for (size_t i = 0; i < bStr.size(); ++i)
+                    {
+                        bStr[i] = tolower(bStr[i]);
+                    }
+                    // std::transform(aStr.begin(), aStr.end(), aStr.begin(), ::tolower);
+                    // std::transform(bStr.begin(), bStr.end(), bStr.begin(), ::tolower);
+
+                    return aStr < bStr;
+                }
```
lhearachel, line 251: Make this a separate function since you're using it twice.

platinum.us/res.sha1
lhearachel: I thought about this again and realized that it would only spit out one `OK` at the end, so I'm changing my mind on this bit: I'm fine with it. I was worried about getting a ton of `OK`s and polluting the output.

lhearachel:
Closing in favor of `knarc` subproject

PR #71, opened 2023-09-25
Document mart data

Nomura-RH:


include/data/mart_items.h
```diff
@@ -0,0 +1,348 @@
+#include "constants/items.h"
+#include "constants/seals.h"
+
+typedef struct {
+    u16 itemID;
+    u16 unk_02;
```
lhearachel, line 6: Is this for the badge-based shop levels?

Nomura-RH: Yes, I forgot to update this

PR #74, opened 2023-09-29
Clean up routines for replacing fainted battlers and flagging a battle result.

lhearachel:


src/overlay016/battle_controller.c
```diff
@@ -4035,279 +4036,251 @@ static void ov16_02250798 (BattleSystem * param0, BattleContext * param1)
...
-                        param1->commandNext = v4;
-                        param1->command = 21;
-                        param1->battlerStatusFlags[v1] |= 0x1;
-                    }
+            for (j = 0; j < Party_GetCurrentCount(party); j++) {
+                Pokemon *pokemon = Party_GetPokemonBySlotIndex(party, j);
+
+                if (Pokemon_GetValue(pokemon, MON_DATA_SPECIES_EGG, NULL) != FALSE
+                        && Pokemon_GetValue(pokemon, MON_DATA_SPECIES_EGG, NULL) != SPECIES_EGG) {
+                    curHP += Pokemon_GetValue(pokemon, MON_DATA_CURRENT_HP, NULL);
```
lhearachel, line 4098: I tried to figure out an inline for this general procedure (since it's used in a bunch of places), but wasn't able to quite nail it down. Leaving it as-is.

src/overlay016/battle_controller.c
```diff
@@ -4035,279 +4036,251 @@ static void ov16_02250798 (BattleSystem * param0, BattleContext * param1)
...
             }
+
+            if (curHP == 0) {
+                battleCtx->battlersSwitchingMask |= FlagIndex(i);
+                battleCtx->selectedPartySlot[i] = 6;
+            } else {
+                battleCtx->commandNext = retCommand;
+                battleCtx->command = BATTLE_CONTROL_EXEC_SCRIPT;
+                battleCtx->battlerStatusFlags[i] |= BATTLER_STATUS_SWITCHING;
+            }
```
lhearachel, line 4109: These fields need to be renamed now. I'll address that in my last big sweep through this file once all the functions are cleaned up.

PR #90, opened 2023-10-23
Add support for build on Mac OSX <= 10.14

lhearachel:


mid-kid:
Use `find_program(['sha1sum', 'shasum'])` instead of checking for the OS

PR #93, opened 2023-10-25
Add Dockerfile

PikalaxALT:


INSTALL.md
```diff
@@ -153,3 +153,11 @@ To build the rom, run:
...
 If everything works, then the following ROM should be built:
 - [build/pokeplatinum.us.nds](https://datomatic.no-intro.org/index.php?page=show_record&s=28&n=3541) `sha1: ce81046eda7d232513069519cb2085349896dec7`
+
+# 4. Docker
+
+A Dockerfile is provided for your convenience. To begin, setup docker on your local machine following the instructions at https://docs.docker.com/desktop/. Then, run
+
+    ./clean.sh
+    docker build . -t pret/pokeplatinum
+    docker run -w /rom -v .:/rom pret/pokeplatinum sh -c "./config.sh && ./build.sh"
```
lhearachel, line 163: 1. Does the user need to run `config` and `build` each time? Or is it just `build` after the initial `config` like in the existing instructions for desktop environments?
2. Is the built ROM accessible to the user after it's been built?

PikalaxALT: they only need to run config once, and yes the rom is accessible to the user because of the volume mount (-v .:/rom)

lhearachel (approved):
Pending confirmation that this works as expected. LGTM otherwise

SethBarberee (approved):
Cloned fresh repo and built rom on Arch Linux. Game booted just fine on melonDS.

lhearachel:
I also got an OK on Mac, but I had to install an X server ([XQuartz](https://www.xquartz.org/), specifically).

mid-kid:
It's odd that you needed an X server. This is bad also on linux as more and more people transition to wayland. Try this maybe https://superuser.com/questions/902175/run-wine-totally-headless#948200

Also, why not pip install --user? If the PATH variable isn't picked up properly you can always run /bin/bash -l instead of /bin/sh

PR #101, opened 2023-11-02
Story/archive evolutions

princepangolin:
Working to archive evolution trees. Initially I was messing around with stuff so I also need to clean up some of the extra files that were generated.


tools/json2bin/evo.py
```diff
@@ -0,0 +1,108 @@
...
+import json2b as j2b
+
+
+from consts import (
+    item
+    moves
+)
+
+
+consts.pokemon import (
```
lhearachel, line 14:
```suggestion
from consts.pokemon import (
```

tools/json2bin/evo.py
```diff
@@ -0,0 +1,108 @@
...
+    evo_methods
+    species
+)
+
+
+def get_evo_params(method: EvoMethod, evo: list):
```
lhearachel, line 20:
```suggestion
def get_evo_params(method: evo_methods.EvoMethod, evo: list):
```

tools/json2bin/evo.py
```diff
@@ -0,0 +1,108 @@
...
+    elif method == EvoMethod.KNOW_MOVE:
+        final_param = moves.Move[maybe_param].value
+    elif method == EvoMethod.MON_IN_PARTY:
+        final_param = species.Species[maybe_param].value
+
+    return final_param
+
+
+def table_line(evo_method: int, evo_params: int, species: int) -> bytes:
+    binary = bytearray([])
```
lhearachel, line 66: Generally, you can (and should) use `.extend` instead of `+=` to add to a list.

tools/json2bin/consts/pokemon/evo_methods.py
```diff
@@ -0,0 +1,42 @@
+from enum import Enum, Flag, auto
+
+
+class EvoMethod(Flag):
+    NONE                = 0
```
lhearachel, line 5: I would prefix these with `EVO_` to explicitly define their purpose in the JSON files. We probably need to update the files to account for that, but that should be a simple script.

tools/json2bin/consts/pokemon/evo_methods.py
```diff
@@ -0,0 +1,42 @@
...
+    USE_ITEM_DAY        = auto()
+    USE_ITEM_NIGHT      = auto()
+    KNOW_MOVE           = auto()
+    MON_IN_PARTY        = auto()
+    LEVEL_MALE          = auto()
+    LEVEL_FEMALE        = auto()
+    ELECTRIC_FIELD      = auto()
+    MOSSY_STONE         = auto()
+    ICY_STONE           = auto()
+    LEVEL_DAY           = auto()
```
lhearachel, line 32: Don't include the ones from here and below; they don't exist in vanilla.

lhearachel (commented):
Thanks for taking this on. Here's some light initial feedback. Do you feel stuck on anything in particular for getting this NARC to match?

tools/json2bin/evo.py
princepangolin: Update Made

tools/json2bin/evo.py
princepangolin: Update made

lhearachel:
> I had to look at a few of the different scripts to see how to do it, especially since the evolutions key was a bit different from the most other keys. The big thing is I'm not sure how to figure out the final byte size, alignment, and padding.

You shouldn't need to worry about byte-alignment for this archive in particular, since there is just a single binary file for each species, and each table is uniform in structure. I believe that the entries are all 44 bytes, though. You can extend your binary arrays in `parse_evolutions` like so:

```py
def parse_evolutions(table: list, _size: int, _enum: None):
    out = []
    for j in range(min(len(table), 7)):
        evo = table[j]
        method = evo_methods.EvoMethod[evo[0]]
        params = get_evo_params(method, evo)
        target = species.Species[evo[-1]]
        out.extend(table_line(evo_methods.EvoMethod[method].value, params, species.Species[target].value))
    
    if len(out) < 42:
        out.extend((0).to_bytes(42 - len(out), 'little'))
    out.extend((0).to_bytes(2, 'little'))
    
    return out
```

And then the individual files should match accordingly.

lhearachel:
Tagging this one to issue #103.

princepangolin:
Building the rom with the changes fails the checksum. Something is not quite right.

tools/json2bin/consts/pokemon/evo_methods.py
```diff
@@ -0,0 +1,31 @@
+from enum import Enum, Flag, auto
+
+
+class EvoMethod(Flag):
```
lhearachel, line 4: This needs to derive from `Enum`, not from `Flag`.

```suggestion
class EvoMethod(Enum):
```

tools/json2bin/evo.py
```diff
@@ -0,0 +1,137 @@
...
+    binary.extend((evo_method & 0x7F).to_bytes(2, 'little'))
+    binary.extend((evo_params & 0x7F).to_bytes(2, 'little'))
+    binary.extend((species & 0x7F).to_bytes(2, 'little'))
+    return bytes(binary)
+    
+
+def parse_evolutions(table: list, _size: int, _enum: None):
+    out = []
```
lhearachel, line 74:
```suggestion
    out = bytearray([])
```

tools/json2bin/evo.py
```diff
@@ -0,0 +1,137 @@
...
+
+def table_line(evo_method: int, evo_params: int, species: int) -> bytes:
+    binary = bytearray([])
+    binary.extend((evo_method & 0x7F).to_bytes(2, 'little'))
+    binary.extend((evo_params & 0x7F).to_bytes(2, 'little'))
+    binary.extend((species & 0x7F).to_bytes(2, 'little'))
+    return bytes(binary)
+    
+
+def parse_evolutions(table: list, _size: int, _enum: None):
```
lhearachel, line 73:
```suggestion
def parse_evolutions(table: list, _size: int, _enum: None) -> bytes:
```

tools/json2bin/evo.py
```diff
@@ -0,0 +1,137 @@
...
+        final_param = species.PokemonSpecies[maybe_param].value
+
+    return final_param
+
+
+def table_line(evo_method: int, evo_params: int, species: int) -> bytes:
+    binary = bytearray([])
+    binary.extend((evo_method & 0x7F).to_bytes(2, 'little'))
+    binary.extend((evo_params & 0x7F).to_bytes(2, 'little'))
+    binary.extend((species & 0x7F).to_bytes(2, 'little'))
```
lhearachel, line 69: Don't perform the bitwise-and here; it'll truncate entries unnecessarily down to calc down to 7 bits in length (which is not enough to hold all the possible values, and you're extending to a 16-bit integer anyways).

```suggestion
    binary.extend(evo_method.to_bytes(2, 'little'))
    binary.extend(evo_params.to_bytes(2, 'little'))
    binary.extend(species.to_bytes(2, 'little'))
```

tools/json2bin/evo.py
```diff
@@ -0,0 +1,137 @@
...
+    return bytes(binary)
+    
+
+def parse_evolutions(table: list, _size: int, _enum: None):
+    out = []
+    for j in range(min(len(table), 7)):
+        evo = table[j]
+        method = evo_methods.EvoMethod['EVO_' + evo[0]]
```
lhearachel, line 77: Let's update the data files themselves to have this prefix automatically. I can write up a shell script for you to accomplish that, if you'd like! 😁

tools/json2bin/evo.py
```diff
@@ -0,0 +1,137 @@
...
+        params = get_evo_params(method, evo)
+        target = species.PokemonSpecies[evo[-1]]
+        out.extend(table_line(method.value, params, target.value))
+        
+    if len(out) < 42:
+        out.extend((0).to_bytes(42 - len(out), 'little'))
+    out.extend((0).to_bytes(2, 'little'))
```
lhearachel, line 84:
```suggestion
    out.extend((0).to_bytes(44 - len(out), 'little'))
```

tools/json2bin/evo.py
```diff
@@ -0,0 +1,137 @@
...
+    },
+    'ROTOM': {
+        'HEAT': 503,
+        'WASH': 504,
+        'FROST': 505,
+        'FAN': 506,
+        'MOW': 507,
+    },
+}
+def indexer(file_path: pathlib.Path) -> int:
```
lhearachel, line 117: Whitespace between these.

tools/json2bin/meson.build
```diff
@@ -2,3 +2,4 @@ encounters_py = find_program('encounters.py', native: true)
 movedata_py = find_program('movedata.py', native: true)
 pokemon_personal_data_py = find_program('pokemon_personal_data.py', native: true)
 pokemon_wotbl_data_py = find_program('pokemon_wotbl_data.py', native: true)
+evo_py = find_program('evo.py', native: true)
```
lhearachel, line 5: Make sure this file ends with an empty new-line.

lhearachel (changes requested):

tools/json2bin/evo.py
princepangolin: I should be ablet to that myself. I just wanted to get the script working properly first.

princepangolin: I've updated the 'evolutions' entries to have the evolution methods be 'EVO_' prepended and same thing for the moves if a pokemon evolved by knowing a move (the moves were not prepended with 'MOVE_' before)

tools/json2bin/meson.build
princepangolin: On my end it looks like I have the empty new-line at the end, but it's not showing up on git.

lhearachel: Missing new-lines show up in github diffs as a ⛔ symbol at the end of a file.

princepangolin:
I think I messed up the rebase a little bit since it seems to have added the trainer stuff to this fork as changes. But the ROM builds fine and my tests shows that editing the 'evolutions' field causes changes in the generated rom so everything looks good.

princepangolin:
Closing because of some issues merging with master. PR #109 was created to merge the files into master.

PR #118, opened 2023-11-27
GCC-friendly compile_commands.json maker

PikalaxALT:


mid-kid:
Assuming this is for an IDE, can't a meson plugin be figured out? Either by making compile_commands call meson/ninja or by providing a configuration for such a plugin.

This copies a lot of the behavior from meson.build, and I fear they will get out of sync.

mid-kid:
The meson files might need to support building with clangd but that's something that would need to happen anyway.

lhearachel (approved):
Merging this in for now; I'll investigate the usage of a Meson plugin for the various major editors (primarily looking at VSCode, CLion, [neo]vim, and [Doom ]emacs).

PR #119, opened 2023-12-22
Generate GMM headers for pl_msg at build-time

lhearachel:


res/text/meson.build
```diff
@@ -1,739 +1,749 @@
...
+    'pl_msg_0722.gmm',
+    'pl_msg_0723.gmm',
+)
+
+gmm_header_dir = custom_target('gmm_header_dir',
+    output: 'gmm_header_dir',
+    command: [
+        'sh', '-c', '"$@"',
+        '--',
+        'mkdir', '-p', '../include/constants/text'
```
mid-kid, line 733: mkdir doesn't work without needing a shell?

mid-kid (commented):
This change expects that the build dir is *always* directly one level under the source directory. I think there's ways to get a full path to the source directory in meson, but putting build artifacts in the source directory is really bad practice.

Please explore the possibility for generating these files in the build directory, and adding the include path to the compiler command line instead. See https://mesonbuild.com/Include-directories.html#include-directories

res/text/meson.build
lhearachel: I tried using `run_command`, but that didn't take.

lhearachel:
> Please explore the possibility for generating these files in the build directory, and adding the include path to the compiler include path (through its command line) instead.

I got this working, with the added caveat that it requires `res` to be walked before `src`, but that's not a huge deal, IMO.

res/text/meson.build
```diff
@@ -1,739 +1,752 @@
...
+    'pl_msg_0720.gmm',
+    'pl_msg_0721.gmm',
+    'pl_msg_0722.gmm',
+    'pl_msg_0723.gmm',
+)
+
+gmm_header_target = meson.current_build_dir() / 'gmm'
+gmm_headers_inc = include_directories('gmm')
+
+gmm_header_dir = custom_target('gmm_header_dir',
```
mid-kid, line 731: Does this target not get re-executed each time you run build.sh?
If it does, "touch" a file with the same name or set capture:true.

mid-kid:
Thanks! It looks nice now.

res/text/meson.build
lhearachel: It doesn't, but we can add `capture: true` anyways to keep it clean.

mid-kid: If it's not re-executed every time then it's fine. I just want build.sh to say "nothing to do" if there's nothing to do.

lhearachel: Tested this out just to be sure, and it does indeed say "nothing to do" if the headers have already been generated, so I'd say it's good without the `capture` arg.

lhearachel:
Tested this out a bit locally with some simple modifications to use the generated headers in `battle_script.c`, and it works as expected!

PR #120, opened 2024-01-07
Export debugging symbols

PikalaxALT:
For use with GDB


lhearachel (approved):
Based and debug-pilled

PR #121, opened 2024-01-07
Generate enum-based battle constants at build-time

lhearachel:
This makes use of a subproject that I wrote for generating constants files from a basic manifest schema. The goal here is to support having a single source of truth for such definitions that can be utilized by both the compiler and a direct assembler invocation for building data files (which we will need for battle scripts). The assembler does not support reading `enum` definitions out-of-the-box, so these definitions would need to either be:

1. Duplicated, and thus separately maintained
2. Moved to `#define` syntax, which is less-than-ideal for self-documentation of function signatures
3. Generated from a separate source-of-truth


include/constants/battle/condition.h
```diff
@@ -103,28 +103,26 @@
...
-    OVERWORLD_WEATHER_SPIRITS,
-    OVERWORLD_WEATHER_CLEAR_13,
-    OVERWORLD_WEATHER_FOG,
-    OVERWORLD_WEATHER_DEEP_FOG,
-    OVERWORLD_WEATHER_DARK_FLASH,
-    
-    // these are only for the Battle Frontier
-    OVERWORLD_WEATHER_HARSH_SUN = 1001,
-    OVERWORLD_WEATHER_TRICK_ROOM,
-};
```
lhearachel, line 128: I moved these to `#define` syntax, since there is a break in values outside the usual step-increments.

tools/meson.build
```diff
@@ -18,3 +18,9 @@ if not nitrogfx_exe.found()
     subproject('nitrogfx', default_options: 'native=true')
     nitrogfx_exe = find_program('nitrogfx', native: true)
 endif
+
+constgen_py = find_program('constgen_py', native: true, required: false)
```
mid-kid, line 22: Can't you use the `fallback:` parameter? Granted, that's a new meson feature I think.

mid-kid:
My *only* complaint is that nobody is gonna be able to find the source of truth through neither their IDE nor a simple `grep`. This is because you took away their prefixes in the definitions files, which I think is a questionable feature.

Either:

* Write the constant names out in full without transforming them. Yes it's more verbose, but it's also easier to deal with.
* Output a comment at the top of the header telling the reader that it's a generated file, what program it was generated with, what file the constants originate from, and if relevant the command line arguments passed to the program.

Or both, preferably both.

tools/meson.build
lhearachel: This isn't really a `fallback` candidate, since it's not a different name. We're expecting to find it under this name.

lhearachel:
> Or both, preferably both.

I decided both was worthwhile, too.

mid-kid (approved):
Looks good!

PR #132, opened 2024-01-28
Update INSTALL.md with correct msys2 binutils package

lhearachel:


mid-kid:
Use $MINGW_PACKAGE_PREFIX instead

PR #144, opened 2024-02-10
Determine required pokedex expansion constants

bbepis:
Required const replacements to facilitate dex expansion

I'm 95% sure this matches, it gives me the same SHA1 as the main branch, but the main branch gives me a different than expected hash on MSYS (`22a6a027581f8422940e2e1c6d0a41e3ab945e49`)

I've identified the `UnkStruct_02026324` typedef struct as the PokedexData struct, but I didn't do a full typedef rename of it since it'll double the size of this already complex PR

For the .py files edited here, I've improved them so that only one number has to change for the rest of the constants to also change auomatically. However I'm not sure if the existing numbers in them are correct; in code they're directly after NATIONAL_DEX_COUNT instead of SPECIES_BAD_EGG. But since I'm not able to get the expected SHA1 I don't know if I'm paranoid for no reason


include/struct_decls/struct_02026324_decl.h
```diff
@@ -1,6 +1,6 @@
 #ifndef POKEPLATINUM_STRUCT_02026324_DECL_H
 #define POKEPLATINUM_STRUCT_02026324_DECL_H
 
-typedef struct UnkStruct_02026324_t UnkStruct_02026324;
+typedef struct PokedexData UnkStruct_02026324;
```
lhearachel, line 4: Label the typedef as well:

```suggestion
typedef struct PokedexData PokedexData;
```

src/overlay005/ov5_021F77A8.c
```diff
@@ -45,512 +47,513 @@ typedef struct {
...
-	{ 0x0, 0x22, 0x22, 0x32, 0x28 },
-	{ 0xE, 0xEE, 0xC, 0x5, 0x10 },
-	{ 0x2, 0xA8, 0x50, 0x48, 0x10 },
-	{ 0x2, 0x68, 0x20, 0x0, 0x39 },
-	{ 0x2, 0x68, 0x20, 0x0, 0x39 },
-	{ 0x2, 0x68, 0x20, 0x0, 0x39 },
-	{ 0x2, 0x68, 0x20, 0x0, 0x39 },
-	{ 0x2, 0x68, 0x20, 0x0, 0x39 }
+// I've got next to zero clue what this horseshit does. All I know is that there's one for every species + form, so it's important when increasing the pokedex size
+const UnkStruct_ov5_02200EAC Unk_ov5_02200EAC[FORMS_MAX] = {
```
lhearachel, line 51: Couple things:

1. This comment doesn't really add anything to the code; get rid of it.
2. We discerned, if I recall, that this is related to learnsets, possibly to the move relearner. It could use with a label to reflect that.
3. Get rid of the comments to the right of these table entries; if you intend to index them according to a value, do so with bracket notation, like so:

```c
const UnkStruct_ov5_02200EAC Unk_ov5_02200EAC[FORMS_MAX] = {
    [0] = { 0x6, 0x20, 0x10, 0x8, 0x4 },
    [1] = { 0x6, 0x20, 0x10, 0x8, 0x4 },
    // etc.
};
```

It looks to me like these are dex numbers, just without the leading `SPECIES_NONE` entry. So, the indices should probably be `SPECIES_` constants, offset by 1.

src/overlay021/ov21_021D5600.c
```diff
@@ -117,27 +119,27 @@ Strbuf* ov21_021D56BC (int param0, int param1, int param2, int param3)
...
-    if ((param0 > 493) && (param1 != 6)) {
+    if ((param0 > NATIONAL_DEX_COUNT) && (param1 != 6)) {
         return 0;
     }
 
     return 1;
 }
 
-static Strbuf* ov21_021D5724 (int param0, int param1, int param2)
+static Strbuf* LoadMessage (int bankID, int entryID, int heapID)
```
lhearachel, line 134: This routine could do with some cleanup on variables names, while you're in here.

src/unk_0202631C.c
```diff
@@ -1252,29 +1260,29 @@ u32 sub_0202756C (const UnkStruct_02026324 * param0, int param1, int param2)
     return 0;
 }
 
-u32 sub_020276C8 (const UnkStruct_02026324 * param0, int param1)
+u32 sub_020276C8 (const UnkStruct_02026324 * pokedex, int param1)
```
lhearachel, line 1263: This looks like forms to me.

- 201 is Unown
- 422 is Shellos
- 423 is Gastrodon
- 412 is Burmy
- 413 is Wormadam
- 386 is Deoxys
- 492 is Shaymin
- 487 is Giratina
- 479 is Rotom

tools/json2bin/evo.py
```diff
@@ -82,36 +82,39 @@ def parse_evolutions(table: Sequence, _size: int, _enum: None) -> bytes:
     .register('evolutions', 44, parse_evolutions, optional=j2b.OptionalBehavior.PAD)
 
 
+NATIONAL_DEX_COUNT = 493
+
+# I'm not sure if this is correct? The form indices in-game are offset at NATIONAL_DEX_COUNT, not SPECIES_BAD_EGG
```
lhearachel, line 87: They are correct; all of the references to these target entries in the `pl_personal` NARC, which has entries 494 and 495 as `Egg` and `Bad Egg`, respectively. Deoxys forms start at 496.

lhearachel (changes requested):

include/struct_decls/struct_02026324_decl.h
bbepis: Done, I did this in two separate commits to make it easier to look over the new changes without getting assaulted by 200 file changes

src/overlay005/ov5_021F77A8.c
bbepis: Removed the comments, I decided against using indexing here because I feel like this array should eventually be automatically generated from something like `data.json` for each species

src/overlay021/ov21_021D5600.c
bbepis: Done

src/unk_0202631C.c
bbepis: Done, I added the relevant constants

tools/json2bin/evo.py
bbepis: Removed the comments

src/overlay005/ov5_021F77A8.c
lhearachel: I like the idea of building this table from a data file. Good shout. Can you file an issue for it?

src/overlay005/ov5_021F77A8.c
```diff
@@ -44,526 +44,571 @@
...
+#define MOVESET_MASK_SIZE (NELEMS(TeachableMoves) + 7) / 8
+
+// Each of the bits in this array correspond to a move in TeachableMoves
+// Therefore the size is dependent on how many teachable moves exist
+
+typedef struct {
+    u8 maskData[MOVESET_MASK_SIZE];
+} MovesetMask;
+
+const MovesetMask TeachableMovesets[MOVESET_MAX] = {
```
lhearachel, line 105: `sTeachableMovesets`

src/overlay005/ov5_021F77A8.c
```diff
@@ -44,526 +44,571 @@
...
     u8 unk_02;
     u8 unk_03;
     u8 unk_04;
     u8 unk_05;
-    u32 unk_08;
-} UnkStruct_ov5_02200CE4;
+    u32 shardColor; // the type of underground shard you use to trade to the move tutors
+} TeachableMove;
+
+static const TeachableMove TeachableMoves[] = {
```
lhearachel, line 55: `sTeachableMoves`

lhearachel (changes requested):

lhearachel (approved):

src/overlay005/ov5_021F77A8.c
bbepis: Done: #146

PR #147, opened 2024-02-12
Build experience tables from source

wildfire248:
Builds tables for mapping level to experience (`pl_growtbl.narc`) from source.
Also edits associated functions to use `enum PokemonExpRate` from generated constants instead of `int` for which table is read.


lhearachel (commented):
I think this is an entry where I would argue in favor of CSV, since the user experience of trying to tune level curves on a particular level breakpoint isn't great with JSON. Thoughts?

PR #150, opened 2024-02-15
SDAT Building/Names

froggestspirit:
Added all the naming from the SDAT to sdat.h (preserved the existing naming, and added prefix "SE" to the sfx to fall in line with the internal naming.

SDAT is also built using my gen4 fork of SDATTool


.gitignore
```diff
@@ -52,6 +53,12 @@ symbols.csv
 res/**/*.c
 *.o
 
+# SDAT Files
+res/sound/*.sdat
+res/sound/*/Files/BANK/*.sbnk
+res/sound/*/Files/SEQ/*.sseq
+res/sound/*/Files/WAVARC/*.swar
```
lhearachel, line 60: Are these files generated inside the source-tree? If so, can we make a patch to SDATTool to generate them into a specified folder?

lhearachel: They are not, we gucci

lhearachel (approved):

PR #151, opened 2024-02-17
Use relative paths for directories in battle script build

lhearachel:
Tested this out on WSL1, WSL2, and MSYS2; got an OK on all of them.

I'm not the biggest fan of this solution, but it's functional.


mid-kid:
We need a CI that puts the build directory somewhere in /var/tmp to prevent changas like this...

PR #155, opened 2024-03-08
Improve mwrap to allow out-of-tree builds

mid-kid:
- Use meson subcommands in build.sh and clean.sh
- mwrap: Improve pathsep shenanigans
- mwrap: Separate source and build directory

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
 
 target="test"
 if [ "$#" -ge 1 ]; then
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -10,4 +11,4 @@ fi
 export NINJA_STATUS="[%p %f/%t] "
 
 # Build the project
-ninja -C build "$target" "$@"
+"${MESON:-meson}" compile -C build "$target" "$@"
```
lhearachel, line 14: Shouldn't this be `meson test`?

lhearachel (approved):

PR #163, opened 2024-03-15
Add comments & definitions to map_header.h and map_header.c

Mr-Toodguy:


lhearachel (changes requested):
This is a little hard to review due to the overall diff size of `map_headers.c`. Could we pull the `mapHeaders` table out into a separate file? e.g. `include/data/map_headers.h`, then include the file in `map_headers.c` to bring the total diff size down. While we're at it, let's update the name to be `sMapHeaders`.

Mr-Toodguy:
Ok sounds good, I agree that the giant table in map_headers.c is undesirable.

src/map_header.c
```diff
@@ -2,13059 +2,11 @@
...
-		0xC,
-		0x3,
-		0xA,
-		0x1,
-		0x1,
-		0x1,
-		0x0,
-	},
-};
+#include "data/map_headers.h"
```
lhearachel, line 5: :raised_hands:

lhearachel (approved):
Love to see it.

PR #171, opened 2024-03-30
2nd Chunk of Save Data Documentation

ravepossum:
More documentation for savedata.c.


src/savedata.c
```diff
@@ -579,274 +578,269 @@ static void SaveDataExtra_LoadCheck (SaveData *saveData, int *frontierResult, in
...
 
-    GF_ASSERT(param0 < gSaveTableSize);
-    v0 = v1[param0].sizeFunc();
+    GF_ASSERT(saveTableID < gSaveTableSize);
+    size = saveTable[saveTableID].sizeFunc();
 
-    v0 += 4 - (v0 % 4);
-    v0 += 4;
+    size += 4 - (size % 4);
+    size += 4;
```
lhearachel, line 841: Is there a potential const for this 4?

include/savedata.h
```diff
@@ -33,17 +33,17 @@ typedef struct SaveDataBody {
...
-    u32 unk_1C[2];
-    volatile BOOL unk_24;
-} UnkStruct_020250DC;
+typedef struct SaveDataState {
+    BOOL fullSaveMode;
+    int startBlock;
+    int currentBlock;
+    int endBlock;
+    s32 lockID;
+    int stateSequence;
```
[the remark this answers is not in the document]
ravepossum: mainSequence could be a bit more appropriate there.

src/savedata.c
ravepossum: From what I can tell it looks like padding, so the significance just seems to be that it's half a byte. I'll plan on leaving as-is unless you disagree.

lhearachel: Fine by me.

lhearachel (approved):

PR #173, opened 2024-04-01
Partially document field_comm_manager

adrienntindall:


include/struct_defs/struct_0205964C.h
```diff
@@ -9,22 +9,22 @@
...
 #include <nitro/math.h>
 
-typedef struct {
-    UnkStruct_02072014 * unk_00[4];
+typedef struct FieldCommunicationManager {
+    TrainerCard * trainerCard[4];
     u8 unk_10[4];
     u8 padding_14[4];
-    UnkStruct_0203CDB0 * unk_18;
+    FieldSystem * fieldSys;
```
cbt6, line 16:
```suggestion
    FieldSystem * fieldSystem;
```

include/struct_decls/struct_0203CDB0_decl.h
```diff
@@ -1,6 +1,6 @@
 #ifndef POKEPLATINUM_STRUCT_0203CDB0_DECL_H
 #define POKEPLATINUM_STRUCT_0203CDB0_DECL_H
 
-typedef struct UnkStruct_0203CDB0_t UnkStruct_0203CDB0;
+typedef struct UnkStruct_0203CDB0_t FieldSystem;
```
cbt6, line 4: Remove this file

include/struct_defs/struct_0203CDB0.h
```diff
@@ -95,6 +95,6 @@ typedef struct UnkStruct_0203CDB0_t {
     UnkStruct_0209C370 * unk_BC;
     u32 unk_C0;
     UnkStruct_ov66_0222DCE0_sub1 unk_C4;
-} UnkStruct_0203CDB0;
+} FieldSystem;
```
cbt6, line 98: Rename file to `include/field_system.h`

cbt6 (commented):
Rename the other header files too where the structs have been documented.

include/struct_defs/struct_0205964C.h
lhearachel: I think `fieldSys` is fine here; we already use `battleSys` in battle code.

include/struct_defs/struct_0203CDB0.h
lhearachel: Or `include/field/field_system.h` (which is probably where it will eventually live).

include/struct_decls/struct_0203CDB0_decl.h
adrienntindall: will double check in a second to make sure but if it's the same as hgss there are struct pointers in FieldSystem that also include pointers to FieldSystem, so this just defines FieldSystem to avoid the error caused by the mutual inclusion

cbt6: If it's needed then rename `UnkStruct_0203CDB0_t` and move the forward declaration to wherever the struct is defined and then remove this file

adrienntindall: marking as out of scope for this pr cause it wasn't meant to be field system focused, I just wanted to doc a file that used field system

include/field/field_system_decl.h
```diff
@@ -0,0 +1,6 @@
+#ifndef POKEPLATINUM_FIELD_SYSTEM_DECL_H
+#define POKEPLATINUM_FIELD_SYSTEM_DECL_H
+
+typedef struct FieldSystem_t FieldSystem;
```
lhearachel, line 4: Per conversation on Discord, consolidation, etc. will happen in a later PR.

lhearachel (approved):

PR #176, opened 2024-04-04
Integrate clang-format

lhearachel:
## Usage

For convenience, a `format.sh` script is included; the following command will format the full source tree according to rules specified in `.clang-format`:
```
./format.sh
```

A pre-commit git hook is also included, along with instructions for contributors on how to integrate it into their commit workflow.

## Next Steps

- `clang-format` does not support renaming constants, variables, etc. We will need a separate `clang-tidy` integration if we want to standardize that process.


lhearachel:
This PR is nonessential, and functionally on-hold; I'll revisit it here and there as I have time to do so.

lhearachel:
> I think it's awkward to invoke the build system for this, as it's not building anything, rather modifying the build tree. It's better as a maintenance script, not something to run as part of any build.

I don't disagree, but Meson gives it to us for free out of the box, and it saves us having to write and maintain a script to traverse/walk our source tree. Maybe it should just be a separate `./format.sh` script that serves to invoke the `ninja` target?

> Regarding auto-formatting on PR, I'd rather it give a CI warning/error (preferably warning) if the autoformatter detects formatting changes in the changed lines. This allows for the possibility to evaluate the formatting changes and possibly choose to override the formatter's decision. Kind of like linux' checkpatch.pl

What about a git hook?

PR #182, opened 2024-04-07
Finish Documenting savedata.c/.h

ravepossum:
This finishes out the rest of the documentation for savedata.c and .h.


src/savedata.c
```diff
@@ -845,585 +838,558 @@ int SaveTableEntry_BodySize (int saveTableID)
     return size;
 }
 
-static void sub_020251DC (SavePageInfo * param0)
+static void SavePageInfo_Init (SavePageInfo *pageInfo)
 {
-    const SaveTableEntry * v0 = gSaveTable;
-    int v1;
-    int v2 = 0;
+    const SaveTableEntry *saveTable = gSaveTable;
```
lhearachel, line 843: Odd choice. Do the accesses below not match unless this is a stack-variable?

src/savedata.c
```diff
@@ -845,585 +838,558 @@ int SaveTableEntry_BodySize (int saveTableID)
...
 
-    for (v3 = 0; v3 < 2; v3++) {
-        param0[v3].saveBlockID = v3;
-        param0[v3].size = 0;
-
-        for (; param1[v4].blockID == v3 && v4 < gSaveTableSize; v4++) {
-            param0[v3].size += param1[v4].size;
+        for (; pageInfo[page].blockID == block && page < gSaveTableSize; page++) {
+            blockInfo[block].size += pageInfo[page].size;
         }
```
lhearachel, line 878: I wonder if this is more easily grokked as a `do-while`:

```suggestion
        do {
            blockInfo[block].size += pageInfo[page].size;
            page++;
        } while (pageInfo[page].blockID == block && page < gSaveTableSize);
```

src/savedata.c
```diff
@@ -845,585 +838,558 @@ int SaveTableEntry_BodySize (int saveTableID)
...
-    for (v1 = 0; v1 < gSaveTableSize; v1++) {
-        v4 = param1[v1].location;
-        v3 = &param0->data[v4];
-        v2 = param1[v1].size;
-        MI_CpuClearFast(v3, v2);
-        v0[v1].initFunc(v3);
+    const SaveTableEntry *saveTable = gSaveTable;
+    int i, size;
+    void *page;
+    u32 location;
```
lhearachel, line 898: Do these need to declared C89-style, or can they be declared inline?

Applies to other instances in the file.

src/savedata.c
```diff
@@ -845,585 +838,558 @@ int SaveTableEntry_BodySize (int saveTableID)
...
+    *loadResult = LOAD_RESULT_OK;
+    *isOld = FALSE;
 
-    if ((v3 == 1) && (v4 == 0)) {
-        if (v7 == v5) {
-            if (v9 == 1) {
-                sub_02025A18(param0, param2, v8, v8, 0);
-                *param4 = 1;
+    if (primaryResult == TRUE && backupResult == FALSE) {
+        if (currKey == primaryKey) {
```
lhearachel, line 1165: I think it reads better if this second condition is folded into the first (and it reduces block-indentation).

```suggestion
    if (primaryResult == TRUE
            && backupResult == FALSE
            && currKey == primaryKey) {
```

src/savedata.c
```diff
@@ -845,585 +838,558 @@ int SaveTableEntry_BodySize (int saveTableID)
...
         }
     }
 
-    if ((v3 == 0) && (v4 == 1)) {
-        if (v7 == v6) {
-            if (v9 == 0) {
-                sub_02025A18(param0, param2, v8, v8, 1);
-                *param4 = 1;
+    if (primaryResult == FALSE && backupResult == TRUE) {
+        if (currKey == backupKey) {
```
lhearachel, line 1177: As above.

lhearachel (changes requested):

src/savedata.c
ravepossum: Nope, it only matches as a stack variable.

lhearachel: Lame. Just GameFreak things.

lhearachel (approved):

PR #187, opened 2024-04-13
Rename unk_0C -> saveData in FieldSystem struct

adrienntindall:
Also some more FieldMenu documentation, just figured the common struct element rename alone is worth opening a PR for


include/unk_020508D4.h
```diff
@@ -16,7 +16,7 @@ void sub_020509D4(FieldSystem * param0);
 BOOL sub_020509DC(FieldSystem * param0);
 void sub_02050A38(TaskManager * param0, const OverlayManagerTemplate * param1, void * param2);
 FieldSystem * TaskMan_FieldSystem(TaskManager * param0);
```
lhearachel, line 18: Should this function also be renamed to match the `TaskManager` prefix?

lhearachel (approved):
LGTM, just the one question

include/unk_020508D4.h
adrienntindall: oops yes lemme fix it

PR #188, opened 2024-04-14
Cleanup field scripts and events

cbt6:


lhearachel:
Why the reversion? I'm not clear on why that's necessary.

cbt6:
Those are supposed to be some sort of indexes for scripts to be called (in map scripts), and are not actually movement actions.

PR #190, opened 2024-04-14
Document Game Options

ravepossum:
- Rename options files
- Document options
- Make Options struct transparent
- Create constants file for options

Note that the clunky naming on the new enums (e.g. `OPTIONS_BUTTON_MODE_NORMAL` instead of simply `BUTTON_MODE_NORMAL`) is to distinguish it from the enums not directly related to the options choices for those things. Following the same example, the `BUTTON_MODE_NORMAL` enum value in `core_sys.h`. You can see the (unfortunately necessary) distinction pretty clearly in `Options_SetSystemButtonMode()`.


include/game_options.h
```diff
@@ -0,0 +1,34 @@
...
+    u16 frame : 5;
+    u16 : 1;
+} Options;
+
+Options* Options_AllocMemory(u32 heapID);
+void Options_Copy(const Options *src, Options *dest);
+void Options_Init(Options *options);
+void Options_SetSystemButtonMode(SaveData *saveData, int buttonMode);
+int Options_TextSpeed(const Options *options);
+void Options_SetTextSpeed(Options *options, int textSpeed);
```
lhearachel, line 21: For the `Set` methods, do they match if you use the `enum` types defined in `constants/game_options.h` in the signature over raw `int`?

src/game_options.c
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    switch (buttonMode) {
+    case OPTIONS_BUTTON_MODE_START_IS_X:
+        gCoreSys.buttonMode = BUTTON_MODE_START_IS_X;
+        break;
+    case OPTIONS_BUTTON_MODE_L_IS_A:
+        gCoreSys.buttonMode = BUTTON_MODE_L_IS_A;
+        break;
+    case OPTIONS_BUTTON_MODE_NORMAL:
```
lhearachel, line 52: Does this `case` branch get optimized out? Does the binary match if you remove it?

ravepossum: Nope, it doesn't match without that there.

src/game_options.c
```diff
@@ -0,0 +1,130 @@
...
+{
+    int speed = Options_TextSpeed(options);
+
+    if (speed == OPTIONS_TEXT_SPEED_SLOW) {
+        return 8;
+    } else if (speed == OPTIONS_TEXT_SPEED_NORMAL) {
+        return 4;
+    } else {
+        return 1;
+    }
```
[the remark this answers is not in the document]
ravepossum: Unfortunately not :(

include/game_options.h
ravepossum: That does match, but I thought we weren't typedef'ing enums. Is it an exception if it's a scenario where it increase clarity to typedef it?

lhearachel: What I'm suggesting is signatures like this:

```c
void Options_SetTextSpeed(Options *options, enum OptionsTextSpeed textSpeed);
```

You don't need to `typedef` the `enum` to get that functionality; the compiler will just look for an `enum` definition that matches that name.

ravepossum: Ah, gotcha, sounds good. Will do, and I'll trim the parameter names a bit to make it a bit less verbose.

src/game_options.c
```diff
@@ -0,0 +1,127 @@
...
+u8 Options_TextFrameDelay (const Options *options)
+{
+    int speed = Options_TextSpeed(options);
+
+    if (speed == OPTIONS_TEXT_SPEED_SLOW) {
+        return 8;
+    } else if (speed == OPTIONS_TEXT_SPEED_NORMAL) {
+        return 4;
+    } else {
+        return 1;
```
lhearachel, line 75: This line pains me so much. One character per frame. :skull:

lhearachel (approved):

PR #193, opened 2024-04-17
Document script context

cbt6:


include/script.h
```diff
@@ -1,5 +1,5 @@
-#ifndef POKEPLATINUM_UNK_0203E724_H
-#define POKEPLATINUM_UNK_0203E724_H
+#ifndef POKEPLATINUM_SCRIPT_H
```
lhearachel, line 1: The name here should specify that this is explicitly for field-scripting, to delineate them from other forms of scripting (AI, Battle, Frontier, etc.).

include/inlines.h
```diff
@@ -161,14 +159,14 @@ static inline void inline_02049FA8 (UnkStruct_02049FA8 * param0, int param1, int
     param0->unk_10 = param5;
 }
 
-static inline u16 * inline_0204FCAC (UnkStruct_0203E724 * param0)
+static inline u16 * ScriptContext_GetVarPointer (ScriptContext * ctx)
```
lhearachel, line 162: Should these live instead in `script.h`?

src/script.c
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+    }
+
+    return TRUE;
+}
+
+static BOOL ScriptContext_Push (ScriptContext * ctx, const u8 * ptr)
+{
+    if (ctx->stackPointer + 1 >= (int)NELEMS(ctx->stack)) {
```
lhearachel, line 102: Is this cast necessary to match?

src/script.c
```diff
@@ -0,0 +1,162 @@
...
+
+void ScriptContext_Return (ScriptContext * ctx)
+{
+    ctx->scriptPtr = ScriptContext_Pop(ctx);
+}
+
+u16 ScriptContext_ReadHalfWord (ScriptContext * ctx)
+{
+    u16 value = ScriptContext_ReadByte(ctx);
+    value += ScriptContext_ReadByte(ctx) << 8;
```
lhearachel, line 140: I'm assuming that you tried this, but does this not match as `u16 value = ScriptContext_ReadByte(ctx) + (ScriptContext_ReadByte(ctx) << 8);`?

src/script.c
```diff
@@ -0,0 +1,162 @@
...
+
+    value += byte3;
+    value <<= 8;
+    value += byte2;
+    value <<= 8;
+    value += byte1;
+    value <<= 8;
+    value += byte0;
+
+    return value;
```
lhearachel, line 161: gdi Game Freak

include/inlines.h
cbt6: That would require `#include "unk_0203E880.h"` in `script.h`, which will not compile since `unk_0203E880.h` already does `#include "script.h"`

include/script.h
cbt6: Any suggestions for names?

src/script.c
cbt6: Yes

src/script.c
cbt6: That does not match

include/script.h
lhearachel: `field_script.h/c`?

lhearachel: That or `field_script_context.h/c`

PR #195, opened 2024-05-01
Name some script command functions

cbt6:


src/unk_0203F6C4.c
```diff
@@ -253,15 +253,15 @@ static BOOL ScrCmd_007(ScriptContext * param0);
 static BOOL ScrCmd_008(ScriptContext * param0);
 static BOOL ScrCmd_009(ScriptContext * param0);
 static BOOL ScrCmd_00A(ScriptContext * param0);
-static BOOL sub_0203F808(u16 param0, u16 param1);
+static BOOL Compare(u16 value0, u16 value1);
```
lhearachel: Can we move this declaration out of the block of command handler declarations? Caught me off-guard seeing it here, at first.

src/unk_0203F6C4.c
```diff
@@ -1745,15 +1745,15 @@ static BOOL ScrCmd_00A (ScriptContext * param0)
...
         return 0;
-    } else if (param0 == param1) {
+    } else if (value0 == value1) {
         return 1;
+    } else {
+        return 2;
     }
-
-    return 2;
 }
```
lhearachel: So this *doesn't* return a `BOOL`, then? Should it be `enum ComparisonResult`?

src/unk_0203F6C4.c
```diff
@@ -1817,34 +1817,25 @@ static BOOL ScrCmd_010 (ScriptContext * param0)
...
-
-    v0 = ScriptContext_GetVarPointer(param0);
-    v1 = *v0;
-    v2 = ScriptContext_ReadHalfWord(param0);
-
-    param0->comparisonResult = sub_0203F808(v1, v2);
-    return 0;
+    u16 value0 = *ScriptContext_GetVarPointer(ctx);
+    u16 value1 = ScriptContext_ReadHalfWord(ctx);
+    ctx->comparisonResult = Compare(value0, value1);
```
lhearachel, line 1830: Does this not match as:

```c
u16 *value0 = ScriptContext_GetVarPointer(ctx);
u16 value1 = ScriptContext_ReadHalfWord(ctx);
ctx->comparisonResult = Compare(*value0, value1);
```

src/unk_0203F6C4.c
```diff
@@ -1817,34 +1817,25 @@ static BOOL ScrCmd_010 (ScriptContext * param0)
...
-    u16 * v0;
-    u16 * v1;
-
-    v0 = ScriptContext_GetVarPointer(param0);
-    v1 = ScriptContext_GetVarPointer(param0);
-    param0->comparisonResult = sub_0203F808(*v0, *v1);
-
-    return 0;
+    u16 * pointer0 = ScriptContext_GetVarPointer(ctx);
+    u16 * pointer1 = ScriptContext_GetVarPointer(ctx);
```
lhearachel, line 1837: I think these would be better named as `varN` to align with the function name.

src/unk_0203F6C4.c
cbt6: Does not match.

lhearachel (approved):

PR #198, opened 2024-05-04
Use descriptive names for file entries in sub_seq.narc

### LINKED ISSUE #196: Add generated headers as dependencies for source compilations
https://github.com/mesonbuild/meson/discussions/13107

lhearachel:
This makes use of a new feature in `knarc` 2.0 to specify an arbitrary file which describes the order of files within the generated archive. 

Use of the NAIX header generated by `knarc` is limited in this PR to C sources. Some clean-up will of our constants-generation will be required to incorporate it into the ASM files (as there is a circular dependency).

Fixes #196


lhearachel:
Fixes #196

PR #215, opened 2024-06-05
Simplify comparisons in field scripts

cbt6:


asm/macros/scrcmd.inc
```diff
@@ -4544,3 +4544,55 @@
...
+    CheckFlag \flagID
+    GoToIf 0, \offset
+    .endm
+
+    .macro GoToIfSet flagID, offset
+    CheckFlag \flagID
+    GoToIf 1, \offset
+    .endm
+
+    .macro GoToIfUndefeated trainerID, offset
```
lhearachel, line 4590: Not sure about this name, just because `undefeated` has existing connotation; maybe just `GoToIfNotDefeated`?

lhearachel (approved):

PR #218, opened 2024-06-09
Name field event structs

cbt6:


include/map_header_data.h
```diff
@@ -0,0 +1,91 @@
...
+    u16 z;
+    fx32 y;
+} ObjectEvent;
+
+typedef struct WarpEvent {
+    u16 x;
+    u16 z;
+    u16 destHeaderID;
+    u16 destWarpID;
+    u8 unused[4];
```
lhearachel, line 39: Should this be a `u32`?

src/map_header_data.c
```diff
@@ -0,0 +1,281 @@
...
+    return (index >= fieldSystem->mapHeaderData->numWarpEvents)
+        ? NULL
+        : &fieldSystem->mapHeaderData->warpEvents[index];
+}
+
+int MapHeaderData_GetIndexOfWarpEventAtPos (const FieldSystem * fieldSystem, int x, int z)
+{
+    for (int i = 0; i < fieldSystem->mapHeaderData->numWarpEvents; i++) {
+        if (fieldSystem->mapHeaderData->warpEvents[i].x == x
+            && fieldSystem->mapHeaderData->warpEvents[i].z == z) {
```
lhearachel, line 80: Maybe macro-able? There's at least one other place in this file with a similar construct. I'm thinking, loosely:

```c
#define AT_COORDS(event, xPos, zPos) event.x == xPos && event.z == zPos
```

src/map_header_data.c
```diff
@@ -0,0 +1,281 @@
...
+
+    // bgEvents
+    data->numBgEvents = *(u32 *)events;
+    events += sizeof(u32);
+    if (data->numBgEvents != 0) {
+        data->bgEvents = (const BgEvent *)events;
+    } else {
+        data->bgEvents = NULL;
+    }
+    events += sizeof(BgEvent) * data->numBgEvents;
```
lhearachel, line 205: I'm wondering if there is a way to inline this, since it's kind of smelly to have such similar code repeated 4 times with very slight differences. I feel like a macro would be unwieldy.

src/map_header_data.c
```diff
@@ -0,0 +1,281 @@
...
+
+const u8 * MapHeaderData_GetScripts (const FieldSystem * fieldSystem)
+{
+    GF_ASSERT(fieldSystem->mapHeaderData != NULL);
+    return (const u8 *)&fieldSystem->mapHeaderData->scripts;
+}
+
+BOOL MapHeaderData_IsAnyObjectEventAtPos (const FieldSystem * fieldSystem, u16 x, u16 z)
+{
+    const MapHeaderData * v0 = fieldSystem->mapHeaderData;
```
lhearachel, line 272: Document `v0`

include/map_header_data.h
```diff
@@ -0,0 +1,91 @@
...
+#include "overlay006/struct_ov6_02242634.h"
+
+typedef struct BgEvent {
+    u16 script;
+    u16 type;
+    int x;
+    int z;
+    int y;
+    u16 playerFacingDir;
+    u8 padding[2];
```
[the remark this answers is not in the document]
cbt6: Yes. Should I remove the explicit field here?

include/map_header_data.h
cbt6: What's the difference?

src/map_header_data.c
cbt6: I don't see the point of adding another layer of indirection? The code is imo already pretty readable to me and this pattern occurs only one other time.

include/map_header_data.h
lhearachel: Nothing, really, I guess. It's fine as is.

include/map_header_data.h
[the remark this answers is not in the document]
lhearachel: I misremembered how we handle it in this repo. It's fine to keep. Some of the structs in the battle overlay explicitly mention the byte offset for alignment padding, but that's optional.

lhearachel (approved):

PR #219, opened 2024-06-12
Document Encounter Effects

Fexty12573:
Documents `encounter_effect.c` and associated files. "Encounter effect" as in the visual transitions when starting a battle.


include/overlay005/encounter_effect_core.h
```diff
@@ -0,0 +1,38 @@
+#ifndef POKEPLATINUM_ENCOUNTER_EFFECT_CORE_H
+#define POKEPLATINUM_ENCOUNTER_EFFECT_CORE_H
+
+#include "struct_decls/sys_task.h"
+
+void EncounterEffect_TallGrass_HigherLevel(SysTask *task, void *param);
+void EncounterEffect_TallGrass_LowerLevel(SysTask *task, void *param);
```
lhearachel, line 7: I think this can just be `Grass` to get the idea across well-enough (and as well for the Trainer encounters).

src/overlay005/encounter_effect.c
```diff
@@ -0,0 +1,1673 @@
...
+        *(screenFlash->done) = FALSE;
+    }
+
+    screenFlash->screen = screen;
+    screenFlash->screenFlashColor = screenFlashColor;
+    screenFlash->otherScreenFlashColor = otherScreenFlashColor;
+    screenFlash->numFlashes = numFlashes;
+}
+
+static void EncounterEffect_ExecuteFlash(SysTask *task, void *param)
```
lhearachel, line 236: `EncounterEffect_FlashTask`

src/overlay005/encounter_effect.c
```diff
@@ -0,0 +1,1673 @@
...
+        SysTask_Done(task);
+        Heap_FreeToHeap(screenFlash);
+
+        return;
+    }
+
+    BrightnessFadeTask_Update(&screenFlash->otherScreenFadeTask);
+}
+
+BOOL EncounterEffect_GetHBlankFlag(EncounterEffect *param0)
```
lhearachel, line 293: name `param0`

src/overlay005/encounter_effect.c
```diff
@@ -0,0 +1,1673 @@
...
+    task->startValue = start;
+    task->delta = end - start;
+    task->numSteps = numSteps;
+    task->currentStep = 0;
+}
+
+BOOL LinearInterpolationTaskS32_Update(LinearInterpolationTaskS32 *task)
+{
+    int interpolated = task->delta *task->currentStep;
+    interpolated = interpolated / task->numSteps;
```
lhearachel, line 310: 1. put a space between both sides of a binary operator (here, the `*`)
2. does this match as the following?

```suggestion
    int interpolated = (task->delta * task->currentStep) / task->numSteps;
```

src/overlay005/encounter_effect.c
```diff
@@ -0,0 +1,1673 @@
...
+}
+
+void BrightnessFadeTask_ApplyBrightnessToScreen(int screen, int brightness)
+{
+    if (screen == 1) {
+        GX_SetMasterBrightness(brightness);
+    } else {
+        GXS_SetMasterBrightness(brightness);
+    }
+}
```
lhearachel, line 394: Does this match with a ternary?

src/overlay005/encounter_effect.c
```diff
@@ -0,0 +1,1673 @@
...
+
+    *(screenSplitEfx->done) = TRUE;
+
+    HBlankTask_Delete(screenSplitEfx->hBlankTask);
+    screenSplitEfx->hBlankTask = NULL;
+    SysTask_Done(screenSplitEfx->vBlankTask);
+    screenSplitEfx->vBlankTask = NULL;
+}
+
+void include_ov5_021DDBE8_rodata_funcptr(void)
```
lhearachel, line 656: This is a very strange construction; is there a way to simplify this somehow? This effectively looks like a no-op function to me.

lhearachel (commented):
As a general comment, I would advise against trying to document whole systems like this all at once. Focused reviews that start with the subsystems and work their way upwards are preferable.

include/field/field_system.h
```diff
@@ -54,7 +54,7 @@ typedef struct FieldSystem_t {
     int unk_18;
     Location * location;
     int unk_20;
-    UnkStruct_020203AC * unk_24;
+    UnkStruct_020203AC * camera;
```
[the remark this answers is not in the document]
Fexty12573: Because various encounter effects make use of it and modify its parameters. I tried messing around with the interpolation values and saw a clear difference in how the camera behaves. I.e. changing the distance it moves away from the player.

include/screen_scroll_manager.h
```diff
@@ -0,0 +1,59 @@
...
+} ScreenScrollManager;
+
+BufferManager *BufferManager_New(enum HeapId heapID, void *buffer1, void *buffer2);
+void BufferManager_Delete(BufferManager *bufferManager);
+void *BufferManager_GetWriteBuffer(const BufferManager *bufferManager);
+const void *BufferManager_GetReadBuffer(const BufferManager *bufferManager);
+void BufferManager_SetMode(BufferManager *bufferManager, u8 mode);
+void BufferManager_SwapBuffers(BufferManager *bufferManager);
+void BufferManager_StopDMA(void);
+void BufferManager_StartDMA(const void *src, void *dst, int size, enum BufferManagerDMAType dmaType);
```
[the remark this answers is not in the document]
Fexty12573: They are accessed separately outside of that context, so yeah having them in a separate header would make sense.

src/overlay005/encounter_effect.c
Fexty12573: Removing is not an option because it doesn't match without them. They aren't used anywhere though. Removing anything from inside the functions also leads to a mismatch. I'll just rename them for now

lhearachel (approved):

PR #224, opened 2024-06-15
Document common script commands and add script entry macro

ravepossum:
This PR names and partially documents some common script commands. It also adds a macro for script table entries (`ScriptEntry`). Some of the documentation is based off of information from pokeheartgold, DSPRE, and Mikelan98 and AdAstra's script command research table.

PR #225, opened 2024-06-15
Script command follow-up

ravepossum:
This is a follow-up to #224 to finish up some feedback. So far it just documents the `MessageVar` macro and addresses a few points of feedback (`destVarID` as argument name for macros that return the result in a var), but I can add on any further feedback that wasn't addressed in the last PR.


asm/macros/scrcmd.inc
```diff
@@ -14,7 +14,7 @@
     .short 2
     .endm
```
cbt6, line 16:
```suggestion
    .short 2
    .endm

```

asm/macros/scrcmd.inc
```diff
@@ -14,7 +14,7 @@
...
     
-    .macro WaitFrames frames, countdown
+    .macro WaitTime frames, countdown
```
cbt6, line 17:
```suggestion
    .macro WaitTime frames, countdownVarID
```

cbt6 (commented):
- `FieldSystem_GetVar` -> `FieldSystem_GetVarPointer`
- Use constants for args in `PlayFanfare`

ravepossum:
Should be ready for another look whenever now.

src/scrcmd.c
```diff
@@ -2291,50 +2291,43 @@ static BOOL ScrCmd_02F (ScriptContext * ctx)
...
-static BOOL sub_02040190 (ScriptContext * ctx)
+static BOOL ScriptContext_CheckABPress (ScriptContext * ctx)
 {
-    if (gCoreSys.pressedKeys & (PAD_BUTTON_A | PAD_BUTTON_B)) {
-        return 1;
-    }
-
-    return 0;
+    // this doesn't match using == TRUE or leaving off a comparison entirely
+    return (gCoreSys.pressedKeys & (PAD_BUTTON_A | PAD_BUTTON_B)) != FALSE;
```
lhearachel, line 2303: **general comment**: that makes sense to me, because this is a bitwise `AND`.

lhearachel (approved):

PR #226, opened 2024-06-17
Document Motion Blur

Fexty12573:
Documents motion_blur.h/.c. Almost exclusively used directly by `FieldMotionBlur`.
All of the currently undocumented functions inside `motion_blur.c` are documented by #219.


lhearachel (commented):
I don't have much commentary on the correctness of this code since this is some fairly low-level stuff. In general, it looks good, but a high-level description of what this code does and what it's used for would be good to include in the PR.

Fexty12573:
Ok sure I will add that

Fexty12573:
The `MotionBlur` system, is really just a wrapper around the DS's capture system, however it is exclusively used for blurring the screen.

You start the system using `MotionBlur_New` and stop it with `MotionBlur_Delete`. Once started, a "screenshot" will be made every screen and written to a buffer. This buffer can then be blended with data from another data.

As an example, `FieldMotionBlur` passes the following parameters to `MotionBlur_New`:
```c
MotionBlurParams motionBlurParams = {
    .displayMode        = GX_DISPMODE_VRAM_C,
    .bgMode             = GX_BGMODE_0,
    .bg0Mode            = GX_BG0_AS_3D,
    .captureSize        = GX_CAPTURE_SIZE_256x192,
    .captureMode        = GX_CAPTURE_MODE_AB,
    .captureSourceA     = GX_CAPTURE_SRCA_2D3D,
    .captureSourceB     = GX_CAPTURE_SRCB_VRAM_0x00000,
    .captureDestination = GX_CAPTURE_DEST_VRAM_C_0x00000,
    .blendCoeffA        = 0,
    .blendCoeffB        = 0,
    .heapID             = HEAP_ID_FIELD
};
```
Because the `captureMode` is set to `GX_CAPTURE_MODE_AB` that means each capture of `captureSourceA` is blended together with `captureSourceB` and stored at `captureDestination`. The destination address is also the same as capture source B, which leads to every new frame being blended together with the previously blended frame, resulting in a motion blur.

The blend coefficients affect the output color, and how each capture source is mixed in. The formula for this is:

$$ C_{OUT} = \cfrac{(C_A \times A_A \times EV_A)+(C_B \times A_B \times EV_B)}{16} $$

where
- $C_a / C_b$: The color from capture source A/B
- $A_a / A_b$: The alpha value of capture source A/B
- $EV_a / EV_b$: The blending coefficient of capture source A/B

So if you want to maintain the screen brightness, make sure $EV_a + EV_b = 16$.

Using `FieldMotionBlur` with $EV_A = EV_B = 8$ results in the following:

https://github.com/pret/pokeplatinum/assets/60443001/7e0bf256-d2ce-48f8-a0ae-9604d963a2d7

lhearachel (approved):
Thanks for the documentation; this is very well-detailed. Nice work digging into such a low-level system.

PR #228, opened 2024-06-18
Document 3D Camera

Fexty12573:
Documents the 3D Camera used by the game.

Below is a quick write-up of how the camera works. (Sorry for wall of text 😛)

### Camera Tracking
Using the `Camera_TrackTarget` function, a pointer to a `VecFx32` can be passed, which will be tracked by the camera. The main update function of the camera is `Camera_ComputeViewMatrix` which, as the name implies, calculates the cameras view matrix based on its position, target position, and angle.

It's important to note that by default the camera does *not* support rotation around the Z axis (i.e. "roll"). Z axis rotation is only taken into account when using `Camera_ComputeViewMatrixWithRoll`. By default, this function is only used when the player is in the distortion world. (It can however be enabled by setting `unk_20` in `FieldSystem` to 1. Also notable is that sprites don't get rotated (regardless of rotation axis).

Tracking can also be limited to only one or two axes using the `trackTarget[XYZ]` fields in the camera.

### Camera Position History
The camera has a "history" built in (has to be explicitly enabled using `Camera_InitHistory`) which allows following a target with a delay. It uses a circular buffer to store a history of camera position deltas, which are applied with a delay of any number of frames. The size of the history buffer is dependent on how many frames of delay are requested.
Just like the normal tracking, the delay can also only be applied to one or more axes.

### Free Move
The camera can also act on its own without being locked to a target. In this mode all camera operations must be manually performed, as opposed to tracking mode where the camera automatically moves relative to the target. Z axis rotation also requires using the dedicated function.

### Angles
Camera angles are specified as pitch, yaw, and roll. Typically held in a `CameraAngle` struct, each component represents an **index** into the Sin/Cos table. This means that a value of 0 is 0°, and a value of 65535 is 360°. This can lead to some confusion because passing `FX32_CONST(3.14159265358 / 2)` as an angle does not equate to 90°, it would instead have to be `FX_RAD_TO_IDX(FX32_CONST(3.14159265358 / 2))`.

Depending on which mode the camera is in, there are 2 different functions used for setting the camera angle:
- Tracking mode: `Camera_SetAngleAroundTarget`
- Free mode: `Camera_SetAngleAroundSelf` (not sure about that name yet)

### Projection
The `Camera_ComputeProjectionMatrix` function takes a `projection` parameter, which can be either `CAMERA_PROJECTION_PERSPECTIVE` or `CAMERA_PROJECTION_ORTHOGRAPHIC`, for perspective or orthographic projection respectively.

### Active Camera
For a camera to be used for rendering it must first be set as the active camera using `Camera_SetAsActive`, because `Camera_ComputeViewMatrix[WithRoll]` does not take any parameter. The camera can also be disabled using `Camera_ReleaseActive`.


include/camera.h
```diff
@@ -0,0 +1,95 @@
...
+
+enum CameraProjection {
+    CAMERA_PROJECTION_PERSPECTIVE = 0,
+    CAMERA_PROJECTION_ORTHOGRAPHIC,
+};
+
+typedef struct CameraLookAt {
+    VecFx32 position;
+    VecFx32 target;
+    VecFx32 up;
```
lhearachel, line 22: **question:** How does this field affect the camera? In my mind, the camera is at a position and pointed at a target, which defines the line of view between those two points.

include/data_02100844.h
```diff
@@ -3,6 +3,6 @@
 
 #include <nitro/gx.h>
 
-extern GXBufferMode Unk_02100844;
+extern GXBufferMode gBufferMode;
```
lhearachel, line 6: **suggestion:** Move this to `camera.h` since it is defined in `camera.c`

src/choose_starter/choose_starter_app.c
```diff
@@ -20,7 +20,7 @@
 #include "struct_decls/struct_02015214_decl.h"
 #include "struct_decls/struct_02018340_decl.h"
 #include "struct_decls/sys_task.h"
-#include "struct_decls/struct_020203AC_decl.h"
+#include "camera.h"
```
lhearachel, line 23: **quibble:** remove duplicate includes throughout the PR

src/unk_0201E190.c
```diff
@@ -2,9 +2,13 @@
 #include <string.h>
 #include <nnsys.h>
 
+#include "nitro/fx/fx_trig.h"
 #include "unk_0201D15C.h"
 #include "unk_0201E190.h"
 
+#define F32_PI                  ((f32)3.14159265358979323846)
+#define FX_F32_TO_IDX(f)	((u16)(FX_RAD_TO_IDX(FX32_CONST((f)))))
```
lhearachel, line 10: **polish:** align these defines

src/camera.c
```diff
@@ -0,0 +1,493 @@
...
+    if (camera->trackTargetZ == FALSE) {
+        deltaPos->z = 0;
+    }
+}
+
+static void Camera_UpdateHistory(Camera const *camera, const VecFx32 *inPos, VecFx32 *outPos)
+{
+    if (camera->history == NULL) {
+        *outPos = *inPos;
+    } else {
```
lhearachel, line 78: **suggestion:** Try using `camera->history == NULL` as a sentinel, e.g.:

```suggestion
    if (camera->history == NULL) {
        *outPos = *inPos;
        return;
    }
```

And then remove the `else` control.

src/camera.c
```diff
@@ -0,0 +1,493 @@
...
+
+        VEC_Subtract(sActiveCamera->targetPos, &sActiveCamera->prevTargetPos, &v0);
+
+        Camera_AdjustDeltaPos(sActiveCamera, &v0);
+        Camera_UpdateHistory(sActiveCamera, &v0, &v1);
+        Camera_Move(&v1, sActiveCamera);
+
+        sActiveCamera->prevTargetPos = *sActiveCamera->targetPos;
+
+        {
```
lhearachel, line 226: **quibble:** unnecessary scope block

src/camera.c
```diff
@@ -0,0 +1,493 @@
...
+    VecFx32 distanceVector;
+
+    Camera_Init(fovY, camera);
+
+    camera->lookAt.target = *target;
+    camera->lookAt.position = *position;
+    VEC_Subtract(position, target, &distanceVector);
+    camera->distance = VEC_Mag(&distanceVector);
+
+    {
```
lhearachel, line 325: **quibble:** unnecessary scope block

lhearachel (commented):
Nicely done on this one. Thanks for such in-depth documentation!

include/camera.h
Fexty12573: Well just a position and a target defines the X and Y rotation of how the camera is looking at the target, but it doesn't define a Z rotation. Given just a position and target the camera could be looking at the target upside down, sideways, etc. That's why you define an up vector.

Technically you can even use the up vector to simulate roll. E.g. If you add a `theta` field to the camera and put this inside `Camera_ComputeViewMatrix`:
```c
// Simulate "camera roll"
sActiveCamera->theta = (sActiveCamera->theta + 100) % 0x10000;
sActiveCamera->lookAt.up.x = FX_Mul(FX_SinIdx(sActiveCamera->theta), -FX32_ONE);
sActiveCamera->lookAt.up.y = FX_CosIdx(sActiveCamera->theta);
sActiveCamera->lookAt.up.z = 0;
```
You get the following:

https://github.com/pret/pokeplatinum/assets/60443001/755b2f7f-8814-42a6-8c52-a5c25a75e42c

Or if you simply define the up vector as `<0, -1, 0>` your game will simply be upside down:

https://github.com/pret/pokeplatinum/assets/60443001/e48ac462-bcb8-4f9c-8e03-8dcc819926ad

lhearachel: I assume then that this is how the Distortion World... "works," for lack of a better term. Neat! Very interesting. Thanks for providing the video evidence. :grin:

lhearachel (approved):
Will merge once conflicts are resolved.

PR #230, opened 2024-06-19
Document Task System

Fexty12573:
Documents the Task System and `SysTaskManager`. `TaskManager` is already taken so this was the next best thing I could come up with.

There are 4 main task managers:
- `mainTaskMgr`: The main task manager with the largest task capacity (160). Executed before VBlank.
- `vBlankTaskMgr`: Task submitted to this task manager get executed during VBlank. Up to 32 tasks
- `postVBlankTaskMgr`: Tasks get executed after VBlank, essentially the last thing that happens in the main loop. Up to 32 tasks.
- `unk_24`: Unsure what this one is for. It's executed right after the main one and can even execute twice per frame. Up to 4 tasks.

Each task manager manages its own pool of tasks which can be freely added to or removed from. All tasks also have a priority value. Tasks are executed in order from lowest to highest. The order of the tasts is dictated by a doubly linked list held by the manager.


src/sys_task_manager.c
```diff
@@ -0,0 +1,236 @@
...
+typedef struct SysTaskManager {
+    u16 maxTasks;
+    u16 stackPointer;
+    SysTask sentinelTask;
+    SysTask **taskStack;
+    SysTask *tasks;
+    BOOL isBeingModified;   // Whether the task list is being modified
+    SysTask *currentTask;   // The task that is currently being executed
+    SysTask *nextTask;      // The task that will be executed next
+} SysTaskManager;
```
lhearachel, line 33: **polish:** Move these into `sys_task_manager.h` and delete the other headers which duplicate the `typedef`.

src/sys_task_manager.c
```diff
@@ -0,0 +1,236 @@
...
+#include <string.h>
+
+#include "functypes/sys_task_func.h"
+#include "nitro/types.h"
+
+#include "sys_task_manager.h"
+
+enum TaskState {
+    TASK_STATE_ACTIVE = 0,
+    TASK_STATE_INACTIVE = 1,
```
lhearachel, line 11: **polish:** D.R.Y.; remove the `= 1`.

src/sys_task_manager.c
```diff
@@ -0,0 +1,236 @@
...
+};
+
+typedef struct SysTask {
+    SysTaskManager * manager;
+    SysTask * prevTask;
+    SysTask * nextTask;
+    u32 priority;
+    void * param;
+    SysTaskFunc callback;
+    u32 state;
```
lhearachel, line 21: **suggestion:** type this with `enum TaskState`

src/sys_task_manager.c
```diff
@@ -0,0 +1,236 @@
...
+    sysTaskMgr->sentinelTask.priority = 0;
+    sysTaskMgr->sentinelTask.param = NULL;
+    sysTaskMgr->sentinelTask.callback = NULL;
+
+    sysTaskMgr->currentTask = sysTaskMgr->sentinelTask.nextTask;
+}
+
+void SysTaskManager_ExecuteTasks(SysTaskManager *sysTaskMgr)
+{
+    if (sysTaskMgr->isBeingModified) {
```
lhearachel, line 125: **suggestion:** based on this behavior, maybe this member should be named `locked`

src/sys_task_manager.c
```diff
@@ -0,0 +1,236 @@
...
+
+    if (sysTaskMgr->currentTask->callback != NULL) {
+        // The task being added should only run starting from the next frame.
+        // If the priority of the task being added is greater than or equal to the current task's priority,
+        // it would be executed in the current frame, so it is set to inactive.
+        if (sysTaskMgr->currentTask->priority <= priority) {
+            task->state = TASK_STATE_INACTIVE;
+        } else {
+            task->state = TASK_STATE_ACTIVE;
+        }
```
lhearachel, line 176: **praise:** thank you for this comment; it is good context!

lhearachel (commented):
On the whole, this looks good. Just a few bits here and there.

lhearachel (approved):
Will merge after conflicts are addressed

cbt6:
> All tasks also have a priority value. Tasks are executed in order from lowest to highest.

This caught my eye because intuitively I would have assumed the other way round (that tasks with higher priority values are executed first). Do you think it is worth documenting this in the codebase too?

Fexty12573:
> > All tasks also have a priority value. Tasks are executed in order from lowest to highest.
> 
> This caught my eye because intuitively I would have assumed the other way round (that tasks with higher priority values are executed first). Do you think it is worth documenting this in the codebase too?

If so I would probably put this in the file where `SysTask_Start` and the like are, since that would likely be the first place people would be looking.

lhearachel:
> If so I would probably put this in the file where SysTask_Start and the like are, since that would likely be the first place people would be looking.

I also believe that this would be useful to add.

Also seems like the Camera PR introduced more conflicts to address. :sweat_smile:

include/sys_task_manager.h
```diff
@@ -0,0 +1,44 @@
...
+typedef struct SysTaskManager {
+    u16 maxTasks;
+    u16 stackPointer;
+    SysTask sentinelTask;
+    SysTask **taskStack;
+    SysTask *tasks;
+    BOOL locked;            // The task manager can't execute while a task is being added
+    SysTask *currentTask;   // The task that is currently being executed
+    SysTask *nextTask;      // The task that will be executed next
+} SysTaskManager;
```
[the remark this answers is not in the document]
lhearachel: We discussed this on Discord and decided that the alternatives are all quite a bit uglier:

* Have a single header split across two modules.
* Have a separate header containing only the structs and typedefs.
* Declare a typedef before an include in each of the headers.

PR #234, opened 2024-06-21
Document LibraryTV

zkitX:


src/library_tv/library_tv.c
```diff
@@ -0,0 +1,284 @@
...
+    return TRUE;
+}
+
+enum {
+    STATE_INIT = 0,
+    STATE_FALLTHROUGH = 1,
+    STATE_FADE_START = 2,
+    STATE_FADE_WAIT = 3,
+    STATE_MAIN = 4,
+    STATE_EXIT = 5
```
lhearachel, line 65: **polish:** You don't need to define these values explicitly; `enum` takes care of that for you.

src/library_tv/library_tv.c
```diff
@@ -0,0 +1,284 @@
...
+};
+
+#define LIBRARY_TV_DURATION 90 + 150
```
lhearachel, line 68: **polish:** Position `#define` constants as close to the top of the file as possible. Generally:

1. Includes
2. Local Defines
3. Local Structs and Enums
4. Static function signatures
5. Everything Else

src/library_tv/library_tv.c
```diff
@@ -0,0 +1,284 @@
...
+        SetMainCallback(LibraryTV_VBlank, (void *)ltv);
+        GXLayers_TurnBothDispOn();
+
+        ltv->waitTiming = 0;
+        *state = STATE_FALLTHROUGH;
+        break;
+    case STATE_FALLTHROUGH:
+        if (ltv->waitTiming) {
+            ltv->waitTiming--;
+        } else {
```
lhearachel, line 103: **question:** Is this block logically unreachable? If so, leave a comment regarding it. This also applies to the same block in `STATE_FADE_START`.

lhearachel (commented):
**question:** Is this the overlay that is utilized for, e.g., the TV-based intro that shows at the start of a new save file?

zkitX:
> **question:** Is this the overlay that is utilized for, e.g., the TV-based intro that shows at the start of a new save file?

This is the overlay that shows up when in the Canalave Library after the explosion at the lake.

lhearachel:
> This is the overlay that shows up when in the Canalave Library after the explosion at the lake.

I'm a little surprised that those don't share the same code, honestly. But, given that this overlay was originally 120, it makes sense that this is new to Platinum.

lhearachel (approved):

PR #235, opened 2024-06-21
Document game_records.c

lhearachel:
Brief diversion in here to also touch up some functions for encoding/decoding data streams using the built-in RNG functions.


src/game_records.c
```diff
@@ -0,0 +1,416 @@
...
+        }
+    } else if (id < MAX_RECORDS) {
+        if (sUsesHighLimit[id]) {
+            return HIGH_LIMIT_U16;
+        } else {
+            return LOW_LIMIT_U16;
+        }
+    }
+
+    GF_ASSERT(0);
```
cbt6, line 254:
```suggestion
    GF_ASSERT(FALSE);
```

src/game_records.c
```diff
@@ -0,0 +1,416 @@
...
+    u32 limit = GetRecordLimit(id);
+
+    DecodeGameRecords(records, id);
+    u32 new = SetRecordValueWithLimit(records, id, val, limit);
+    EncodeGameRecords(records, id);
+
+    return new;
+}
+
+u32 GameRecords_SetAndLimitRecordValue(GameRecords *records, int id, u32 val)
```
cbt6, line 335: I don't think this name is accurate? This function is different from `GameRecords_SetRecordValue` in that it sets the value only if the incoming value is greater than the current value, and both functions limit the value being set anyway.

cbt6 (commented):
What's the difference between a `RECORD_XXX` and a `SCORE_RECORD_XXX`?

lhearachel:
> What's the difference between a RECORD_XXX and a SCORE_RECORD_XXX?

Records are just counters for certain events in-game.

Score records are events which increase the trainer card score. Maybe they would better named as `SCORE_EVENT`, then?

src/game_records.c
lhearachel: `GameRecords_SetGreaterRecordValue`?

cbt6:
> Records are just counters for certain events in-game.

Are these the same records as in those for record mixing?

> Score records are events which increase the trainer card score. Maybe they would better named as `SCORE_EVENT`, then?

`TRAINER_SCORE_EVENT`?

lhearachel:
> Are these the same records as in those for record mixing?

Maybe, but I doubt it. These are counters stored in the save block for things like your step counter, your trainer score, the number of times you've used certain moves or fainted an enemy Pokemon, etc.

> `TRAINER_SCORE_EVENT`?

👍

PR #239, opened 2024-06-24
Documented SaveState struct and relevant functions

shricubed:
- struct_02025E0C -> SaveState
(not sure if this is the best name, other idea I had was just Player)
- Functions to initialize the struct, and to get each of the members

The SaveState struct stores the current options/settings for a game save, the current trainer info, number of coins, and current play time. Each of the functions take in a SaveData object, and use it to access the respective SaveState in the "save table".

Todo next:
- Document unk_02025CB0 (similar functions to SaveState) and the relevant structs (02025CCC and 02055BA8 - seems like it deals with the current datetime of the system?)


include/struct_decls/struct_save_state_decl.h
```diff
@@ -0,0 +1,6 @@
+#ifndef POKEPLATINUM_STRUCT_SAVE_STATE_DECL_H
+#define POKEPLATINUM_STRUCT_SAVE_STATE_DECL_H
+
+typedef struct SaveState_t SaveState;
```
lhearachel, line 4: **polish:** Pull this declaration and its definition into the primary header (and then delete this file); we use transparent structs in this project.

src/save_state.c
```diff
@@ -0,0 +1,58 @@
+#include <nitro.h>
+#include <string.h>
+
+#include "struct_decls/struct_save_state_decl.h"
+#include "savedata.h"
+
+#include "savedata.h"
```
lhearachel, line 7: **suggestion:** Since this module deals solely with save information, I think that it should be either shoved into the `savedata` folder (if we even want to keep that folder, since it's not an overlay) or keep this `save_` prefix on its name. In either case, I don't like calling this `save_state`, nor the encased `struct` `SaveState`. This is player-specific data, and the module name should reflect that.

lhearachel (changes requested):

src/save_player.c
```diff
@@ -0,0 +1,51 @@
...
+
+    Options_Init(&player->options);
+    TrainerInfo_Init(&player->info);
+    Coins_Init(&player->coins);
+    PlayTime_Init(&player->playTime);
+}
+
+TrainerInfo * SaveData_GetTrainerInfo (SaveData * savedata)
+{
+    PlayerSave * state = SaveData_SaveTable(savedata, 1);
```
ravepossum, line 31:
```suggestion
    PlayerSave * state = SaveData_SaveTable(savedata, SAVE_TABLE_ENTRY_PLAYER);
```
And the similar calls below as well.

src/save_player.c
```diff
@@ -0,0 +1,51 @@
...
+    return &state->info;
+}
+
+Options * SaveData_Options (SaveData * savedata)
```
ravepossum, line 35: We've mostly used `saveData` as the camelcasing elsewhere.

include/save_player.h
```diff
@@ -0,0 +1,23 @@
+#ifndef POKEPLATINUM_SAVE_PLAYER_H
+#define POKEPLATINUM_SAVE_PLAYER_H
+
+#include "play_time.h"
+#include "trainer_info.h"
+#include "game_options.h"
+#include "savedata.h"
+
+typedef struct PlayerSave_t {
```
ravepossum, line 9:
```suggestion
typedef struct PlayerSave {
```

lhearachel (approved):

PR #240, opened 2024-06-27
Document Resource Manager + Texture Resource Manager (Cell Actor Part 1)

Fexty12573:
Documents the `ResourceManager` and `TextureResourceManager` structs. 

## ResourceManager
Still debating on that name (opinions welcome), since it is less of a manager and more just a collection. It keeps an internal array of `Resource` objects, which can be "allocated" from and "deallocated" to at will. Potential name would be `ResourceCollection`.
Resources are stored with an ID and the raw data from the file.

## TextureResourceManager
This one manages textures exclusively (duh) and has some additional capabilities. It serves as a wrapper around `ResourceManager` but also allows storing textures in a different way where the actual texture data can be deallocated after it has been uploaded to VRAM (see `enum TextureResourceMode`), which saves a lot of memory. Normally this isn't possible with just the G2D API.

It also handles allocating and deallocating VRAM, keeping track of texture/palette keys, and performing the actual VRAM upload.

This is part 1 of a 3 (probably 4) part PR to document the Cell Actor/Sprite system.


include/resource_manager.h
```diff
@@ -0,0 +1,77 @@
...
+    ResourceManager *resMgr;
+    TextureResource *textures;
+} TextureResourceManager;
+
+ResourceManager *ResourceManager_New(s32 maxResources, enum HeapId heapID);
+void ResourceManager_Delete(ResourceManager *resMgr);
+BOOL ResourceManager_IsIDUnused(ResourceManager *resMgr, int id);
+Resource *ResourceManager_AddResource(ResourceManager *resMgr, void *data, int id);
+Resource *ResourceManager_AddResourceFromFile(ResourceManager *resMgr, const char *filename, int id, enum HeapId heapID);
+void ResourceManager_RemoveResource(ResourceManager *resMgr, Resource *resource);
```
lhearachel, line 47: **polish:** `Add`, `AddFromFile`, and `Remove` are sufficient, I think.

src/resource_manager.c
```diff
@@ -0,0 +1,507 @@
...
+
+NNSG3dResTex *TextureResource_GetUnderlyingResource(const TextureResource *texResource)
+{
+    GF_ASSERT(texResource);
+    return TextureResource_GetTexRes(texResource);
+}
+
+void TextureResource_UploadToVRam(TextureResource *texResource)
+{
+    NNSG3dResTex * texRes;
```
lhearachel, line 316: **question:** Does this variable need to be declared C89-style?

src/unk_0202298C.c
```diff
@@ -7,128 +7,128 @@
```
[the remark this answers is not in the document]
Fexty12573: Oh what the hell, yeah I just forgot to delete these 2 lol.

lhearachel (approved):

PR #241, opened 2024-06-27
Document Cell Actor (Collection) API (Cell Actor Part 2)

Fexty12573:
Documents the low level CellActor API (`cell_actor.h`). I am unfortunately not able to give a super detailed breakdown of how to use it yet because this system is basically never used directly. It is wrapped by a more abstract sprite API. Once I document that I will provide documentation.

Also apologies for the size of this PR but a lot of the types/functions in this PR are used in a ton of files. The actual changes are basically exclusively in `cell_actor.h/.c`)

There's 2 core types in this:
## CellActor
This is a pretty thin wrapper around the G2D API which supports
- Cells
- Cell Animations
- Multi Cells
- Multi Cell Animations
- VRAM Cell Animations

Both drawing and animation can be toggled on or off using the `CellActor_Set*Flag` functions.
It provides functionality for overwriting data provided by the OAM using `CellActor_SetExplicit*` functions.
There are also functions for performing affine transforms (translation, Z rotation, scale, See `CellActor_SetAffine*` functions). Scale is limited to x2 by the hardware.

It is important to note that these cell actors are *hardware sprites*. This is a relevant distinction because there is a concept of software sprites on the DS. Software sprites are basically just textures rendered onto quads using the 3D engine.

This API does *not* handle file loading/management. Cell Actors need to be initialized from already loaded cell banks and cell animation banks (See `CellActorInitParams[Ex]`)

## CellActorCollection
`CellActor`s are essentially always part of a collection. You will most likely not see individual CellActors being used. This is also the reason why there is a `CellActor_Delete` function, but no `CellActor_New`. You use `CellActorCollection_Add[Ex]` instead. `CellActorCollection` has space for a fixed number of cell actors, dictated by its initialization parameters (`CellActorCollectionParams`). Internally it keeps track of the active actors using a doubly linked list (similar to `SysTaskManager`).

A cell actor collection can:
- Add and remove elements
- Suspend its execution
- Update animations
- Draw actors

To remove an actor from a collection you use `CellActor_Delete`.

This is part 2 of a 4 part PR to document the sprite API. Full documentation on the system will be in part 4.


lhearachel:
As a general pre-comment, try to avoid commit descriptions like "Label some stuff"; the goal of each commit should be self-contained and buildable code. If necessary, do some rebasing before filing a PR.

include/cell_actor.h
```diff
@@ -0,0 +1,193 @@
...
+    u32 animData[CELL_ACTOR_ANIM_DATA_SIZE];
+    NNSG2dImageProxy imageProxy;
+    NNSG2dImagePaletteProxy paletteProxy;
+    u32 type;
+    u16 activeAnimID;
+    u8 explicitPriority;
+    u16 priority;
+    NNS_G2D_VRAM_TYPE vramType;
+    struct CellActor *prev;
+    struct CellActor *next;
```
lhearachel, line 83: **question:** Is this just a doubly-linked list? That also has a `head` pointer? Kind of odd.

src/cell_actor.c
```diff
@@ -0,0 +1,935 @@
...
+{
+    CellActor_UpdateAnim(actor, actor->animSpeed);
+}
+
+static void CellActor_UpdateAnimInternal_Stub(CellActor *actor)
+{
+    return;
+}
+
+static void CellActorCollection_Insert(CellActorCollection *collection, CellActor *actor)
```
lhearachel, line 860: **suggestion:** This code leads me to believe that this collection is effectively a priority queue. Does that make sense with helping to name it?

lhearachel (commented):
This is mostly straightforward, scoping things down to just `cell_actor`; my only big gripe is the name. It feels kind of lacking. It seems to me like it's just the thinnest GF wrapper around the Nitro 2D APIs... I think `Cell` makes sense since it refers to the `Cell`-based APIs from the Nitro APIs.

src/cell_actor.c
Fexty12573: Well yeah, it's pretty similar to the task system in that regard. I.. still don't know what to call it though. Personally at least I think Actor isn't bad, it's a pretty common term in gamedev. Maybe it needs something to indicate that it's a "visual thing". Idk. I'm open to changing it

lhearachel: Looking at this again, I think there is just so much going on in these structs that makes them messy and vague that any term we come up with is going to be purposefully and similarly vague. In that sense, I'm fine with `actor`.

lhearachel (approved):
I'll let this sit for a bit in case anyone else wants to chime in before it gets merged.

PR #244, opened 2024-07-06
Document Sprite Resource / Util (Cell Actor Part 3)

Fexty12573:
This is yet another resource "manager". This time specifically for sprite resources. A `SpriteResource` can hold any of
- Character Data (NCGR) -> Tiles
- Palette Data (NCLR) -> Palettes
- Cell Data (NCER) -> Sprites
- Cell Anim Data (NANR) -> Sprite Anims
- Multi Cell Data (NMCR) -> Multi Sprites
- Multi Cell Anim Data (NMAR) -> Multi Sprite Anims

It contains the raw file data and the unpacked data, obtained using the `NNS_G2dGetUnpacked*` functions, as well as some additional info in case of palettes and tiles.

Resources can (as usual...) be added to a collection: `SpriteResourceCollection`. A sprite resource collection can only hold resources of one type. I.e. only sprites or only palettes, etc. It provides methods for loading resources from NARCs

Another type in this is `SpriteResourceTable`, which is similar to a collection but it's a table deserialized directly from a file or a NARC. It, by itself, then contains further references to either other files or NARCs which contain the actual resources. All entries from such a table can be added directly to a sprite resource collection as well, via `SpriteResourceCollection_AddTable[Ex]`.

I now realized that this is most likely gonna be a 6 part PR, instead of 4. This file is mostly utility functions used by a more abstract system. However, there are two more files with utility functions like that. One is for handling VRAM transfers and the other I'm not sure yet.

I am also now debating going back to the cell actor PR and adjusting the terminology there (cell -> sprite) but I first need to see if there is a more concrete "Sprite" type elsewhere.


lhearachel:
Now that #176 has been merged, make sure to clean up your merge conflicts and run your code through the formatting utility.

src/sprite_resource.c
```diff
@@ -0,0 +1,884 @@
...
+
+typedef struct MultiSpriteAnimResourceData {
+    NNSG2dAnimBankData *multiSpriteAnimBank;
+} MultiSpriteAnimResourceData;
+
+typedef struct SpriteResourceTableBinary {
+    enum SpriteResourceType type;
+    union {
+        SpriteResourceTableEntryFile fileEntries[1];
+        SpriteResourceTableEntryNARC narcEntries[1];
```
lhearachel, line 51: **question:** Is the subscript necessary to match?

src/sprite_resource.c
```diff
@@ -0,0 +1,884 @@
...
+    spriteResources->resources = NULL;
+
+    Heap_FreeToHeap(spriteResources);
+    spriteResources = NULL;
+}
+
+SpriteResource *SpriteResourceCollection_AddFromTable(SpriteResourceCollection *spriteResources,
+    const SpriteResourceTable *table,
+    int index,
+    enum HeapId heapID)
```
lhearachel, line 107: **question:** Did `clang-format` mandate this?

src/sprite_resource.c
```diff
@@ -0,0 +1,884 @@
...
+            spriteRes,
+            narcEntry->narcIndex,
+            narcEntry->memberIndex,
+            narcEntry->compressed,
+            narcEntry->id,
+            narcEntry->vramType,
+            narcEntry->paletteIndex,
+            table->type,
+            heapID,
+            0);
```
lhearachel, line 143: **suggestion:** This call-pattern makes me question how meaningful `isNARC` is as a member-name. Maybe `loadFromNARC`?

src/sprite_resource.c
```diff
@@ -0,0 +1,884 @@
...
+{
+    int i = 0;
+    while (entries[i].narcIndex != NARC_INDEX_NONE) {
+        i++;
+    }
+
+    return i;
+}
+
+static void *SpriteUtil_ReadNARCMember(NARC *narc, u32 memberIdx, BOOL compressed, u32 heapID, u32 param4)
```
lhearachel, line 860: **suggestion:** Name `param4` (and the associated params in other functions); seems to me that it's a bool that designates whether or not an allocation should happen at the end of the heap or not.

src/sprite_resource.c
```diff
@@ -0,0 +1,884 @@
...
+int SpriteResourceCollection_AddTable(SpriteResourceCollection *spriteResources, const SpriteResourceTable *table, SpriteResourceList *outList, enum HeapId heapID)
+{
+    GF_ASSERT(spriteResources);
+    GF_ASSERT(table);
+
+    SpriteResoureCollection_AddTableEx(spriteResources, table, 0, table->count, outList, heapID);
+    return table->count;
+}
+
+void SpriteResoureCollection_AddTableEx(SpriteResourceCollection *spriteResources, const SpriteResourceTable *table, int first, int count, SpriteResourceList *outList, enum HeapId heapID)
```
lhearachel, line 412: **suggestion:** How about `SpriteResourceCollection_Extend` for this paradigm?

src/sprite_resource.c
Fexty12573: No, but in my opinion it makes the code that actually *uses* these fields nicer. Could probably do `[0]` as well, provided mwcc supports that.

src/sprite_resource.c
Fexty12573: It's a result of running clang-format, yes, but I already had some weird formatting there to begin with so idk if it just tried to respect that. Because if I put it all on one line and then run the formatter again it doesn't do this. Want me to change it?

src/sprite_resource.c
Fexty12573: I chose `isNARC` because it indicates the type of `entries` (being either `SpriteResourceTableEntryFile` or `SpriteResourceTableEntryNARC`, but I don't mind changing it

src/sprite_resource.c
cbt6:
```suggestion
        SpriteResourceTableEntryFile *fileEntries;
        SpriteResourceTableEntryNARC *narcEntries;
```

Fexty12573: That's not the same code tho. It's an inline array, not a pointer to an array.

src/sprite_resource.c
lhearachel: Yes, preferably. Function definitions should adhere to a single line, ideally.

lhearachel (approved):

PR #257, opened 2024-07-28
Match SPL

Fexty12573:


tools/cw/mwrap.c
```diff
@@ -490,6 +509,10 @@ int _tmain(int argc, _TCHAR *argv[])
 
     if (args.wrap_ver) tool_ver = args.wrap_ver;
 
+    if (_tcscmp(tool_ver, _T("1.2/base_b73")) == 0) {
+        remove_ipa_arguments(&new_argc, new_argv);
+    }
```
lhearachel, line 514: Is there a smoother way to do this? e.g., introduce a new arg that specifies to ignore all `ipa` arguments when passing them to the compiler?

Fexty12573: that's probably a better approach. I will add a `-noipa` argument

lib/spl/src/unk_0209C400.c
```diff
@@ -0,0 +1,434 @@
+#include "spl.h"
```
PikalaxALT, line 1: Remove ASM files if they're no longer part of the project

lib/spl/include/spl.h
```diff
@@ -357,60 +381,209 @@ typedef struct UnkSPLStruct1_t {
...
+    } etc;
+} SPLSimpleCollisionField;
+
+typedef struct SPLConvergence {
+    VecFx32 pos;
+    fx16 ratio;
+    u16 reserved0;
+} SPLConvergence;
+
+typedef struct SPLNode {
```
PikalaxALT, line 429: This is just an alias for an SPLEmitter

PikalaxALT (changes requested):

lib/spl/include/spl.h
Fexty12573: No it isn't. It's used for `SPLParticle` as well.

lib/spl/src/unk_0209DD54.c
```diff
@@ -0,0 +1,1029 @@
...
+        vel_dir = ptcl->unk_14;
+        VEC_Normalize(&vel_dir, &vel_dir);
+
+        dot = FX_MUL(vel_dir.x, -cmr->_02) + FX_MUL(vel_dir.y, -cmr->_12) + FX_MUL(vel_dir.z, -cmr->_22);
+        if (dot < 0) {
+            dot = -dot;
+        }
+
+        dot = FX_MUL(sclY, FX_MUL(FX32_ONE - dot, (fx32)mgr->unk_40.unk_00->p_res->unk_00->unk_48.unk_05_0) + FX32_ONE);
+        load._00 = FX_MUL(dir.x, sclX);
```
PikalaxALT, line 445: does it match if you use the `.m` accessor?

Fexty12573: Yeah

PikalaxALT (approved):

PR #259, opened 2024-08-03
Document SPL

Fexty12573:
This PR documents the particle system library used by the game.

Here is a quick breakdown of how the library works/how to use it:

### 1. Creating a Manager
To do anything with the library an instance of `SPLManager` must first be created. When creating one you specify a maximum number of simultaneous emitters and particles that the manager is able to create, as well as a range of polygon IDs to use:
```c
SPLManager *mgr = SPLManager_New(
    MyAllocFunc, // Memory allocation function that the manager will use
    10, // Maximum 10 emitters
    100, // Maximum 100 particles simultaneously
    5, // Fixed polygon ID
    6, // Mininum polygon ID
    63 // Maximum polygon ID
)
```

### 2. Loading Resources
Next the manager must be provided with data that defines what particles look like and how they behave. Usually this data comes from a file, but technically it can come from anywhere. The files are a custom binary format with the `.spa` extension. One such file can contain multiple resources. A resource defines parameters for an emitter, which then passes those parameters to particles it creates.
```c
void *data = NARC_AllocAndReadWholeMemberByIndexPair(...);
SPLManager_LoadResources(mgr, data);

// Both of these functions also have an Ex variant which allows you to specify a custom VRAM allocator
SPLManager_UploadTextures(mgr); // Uploads the textures to VRAM
SPLManager_UploadPalettes(mgr); // Uploads the palettes to VRAM

// Note that the data must be kept alive until the manager is destroyed
```

### 3. Creating Emitters
Now that the resources are loaded, emitters can be created. Each emitter is associated with a resource.
```c
VecFx32 pos = {0, 0, 0}; // Initial position of the emitter
SPLEmitter *emitter = SPLManager_CreateEmitter(mgr, 0, &pos);
```
Note that some emitters are self-maintaining, meaning they will automatically destroy themselves when they are done. Others are not and must be manually destroyed. In the case of the former, `SPLManager_CreateEmitter` (and similar functions) will return `NULL`.

If you still want to obtain a reference to a self-maintaining emitter, you can use `SPLManager_CreateEmitterWithCallback[Ex]` instead. This function will call a callback function after the emitter has been initialized, passing the emitter as an argument.

### 4. Updating Emitters
To update the emitters, call `SPLManager_Update`:
```c
SPLManager_Update(mgr);
```
This function should be called every frame. Also note that typically drawing is done *before* updating.

### 5. Drawing
To draw the particles, call `SPLManager_Draw`:
```c
NNS_G3dGlbFlush(); // Flush the 3D pipeline
const MtxFx43 *view = NNS_G3dGlbGetCameraMtx();
SPLManager_Draw(mgr, view);
```
Of course, you can pass any view matrix you want.

### 6. Destroying Emitters
For emitters that are not self-maintaining, you must manually destroy them when they are no longer needed:
```c
SPLManager_DeleteEmitter(mgr, emitter);
```

### Manually Emitting Particles
Normally, emitters will emit particles automatically. However, you can manually make an emitter emit particles:
```c
SPLManager_Emit(mgr, emitter);

// Or you can specify a position

VecFx32 pos = {0, 0, 0}; // Position of the emitter
SPLManager_EmitAt(mgr, emitter, &pos);
```

### Particle Capabilities
All particles have a set of common parameters:
- Position (relative to the emitter)
- Velocity
- Rotation
- Angular velocity
- Age
- Scale
- Color
- Texture ID
- Alpha

These should be self explanatory. However, there are additional behaviors that can be added to particles. Or rather, these behaviors can be added to emitters, which then pass them to particles. These behaviors include:
- **Gravity**: A constant acceleration in a given direction
- **Random acceleration**: A random acceleration, applied in a configurable interval
- **Magnetic force**: A force that attracts particles to a point
- **Spin**: A gradual revolution around the emitter
- **Collision Plane**: A plane that particles will either bounce off of or die when they collide with it
- **Convergence**: Particles will converge to a point, this acts directly on the position of the particles

Without any of these behaviors, particles will simply move in a straight line until they die. Their velocity is based on their initialization parameters. Behaviors are specified in the resource file.

Additionally, particles can also have the following parameters animated:
- Scale
- Color
- Alpha
- Texture ID

### Emitter Capabilities
Emitters are also not just static entities. They can move around and keep that of their own velocity. However, the velocity must be set manually. Emitter velocity directly affects the velocity of the particles it emits.

Emitters can also have a wide range of emission types:
- **Point**: Particles are emitted from a single point
- **Sphere Surface**: Particles are emitted on the surface of a sphere around the emitter
- **Circle Border**: Particles are emitted randomly on the border of a circle around the emitter
- **Circle Border Uniform**: Particles are emitted uniformly on the border of a circle around the emitter
- **Sphere**: Particles are emitted within a sphere around the emitter
- **Circle**: Particles are emitted within a circle around the emitter
- **Cylinder Surface**: Particles are emitted on the surface of a cylinder around the emitter
- **Cylinder**: Particles are emitted within a cylinder around the emitter
- **Hemisphere Surface**: Particles are emitted on the surface of a hemisphere around the emitter
- **Hemisphere**: Particles are emitted within a hemisphere around the emitter

### Child Particles
Particles can also spawn child particles. This behavior is restricted to a single level of recursion. Child particles also have less functionality than their parents. They are limited to Scale and Alpha animations, and they cannot have their own behaviors. Though they can optionally inherit the parent's behaviors.

To make a particle spawn child particles, the resource it stems from must have a Child resource defined.


lib/spl/src/spl_emit.c
```diff
@@ -0,0 +1,347 @@
...
+    ptclPos->y = FX_MUL(pos->x, emtr->crossAxis1.y) + FX_MUL(pos->y, emtr->crossAxis2.y) + FX_MUL(pos->z, vec.y);
+    ptclPos->z = FX_MUL(pos->x, emtr->crossAxis1.z) + FX_MUL(pos->y, emtr->crossAxis2.z) + FX_MUL(pos->z, vec.z);
+}
+
+void SPLEmitter_EmitParticles(SPLEmitter *emtr, SPLParticleList *list)
+{
+    SPLResource *res = emtr->resource;
+    SPLResourceHeader *header = res->header;
+
+    int i = 0;
```
lhearachel, line 81: **question:** Is this declaration required here? It doesn't seem to be used before the re-assignment in the header of the below for-loop.

lib/spl/src/spl_emit.c
```diff
@@ -0,0 +1,347 @@
...
+    int emitCountDec = emtr->emissionCount + FX32_CAST(emtr->emissionCountFractional);
+    int totalEmissions = emitCountDec >> FX32_SHIFT;
+    int emission = 0;
+    emtr->emissionCountFractional = emitCountDec & FX32_DEC_MASK;
+
+    enum SPLEmissionType emitType = header->flags.emissionType;
+    if (emitType == SPL_EMISSION_TYPE_CIRCLE_BORDER || emitType == SPL_EMISSION_TYPE_CIRCLE_BORDER_UNIFORM || emitType == SPL_EMISSION_TYPE_CIRCLE || emitType == SPL_EMISSION_TYPE_CYLINDER_SURFACE || emitType == SPL_EMISSION_TYPE_CYLINDER || emitType == SPL_EMISSION_TYPE_HEMISPHERE_SURFACE || emitType == SPL_EMISSION_TYPE_HEMISPHERE) {
```
lhearachel, line 88: **chore:** Format this disjunction-chain.

lib/spl/src/spl_draw.c
```diff
@@ -0,0 +1,953 @@
...
+typedef void (*DrawPlaneFunc)(fx16 s, fx16 t, fx16 offsetX, fx16 offsetY);
+
+void SPLDraw_Billboard(SPLManager *mgr, SPLParticle *ptcl);
+void SPLDraw_Child_Billboard(SPLManager *mgr, SPLParticle *ptcl);
+void SPLDraw_DirectionalBillboard(SPLManager *mgr, SPLParticle *ptcl);
+void SPLDraw_Child_DirectionalBillboard(SPLManager *mgr, SPLParticle *ptcl);
+void SPLDraw_Polygon(SPLManager *mgr, SPLParticle *ptcl);
+void SPLDraw_Child_Polygon(SPLManager *mgr, SPLParticle *ptcl);
+void SPLDraw_DirectionalPolygon(SPLManager *mgr, SPLParticle *ptcl);
+void SPLDraw_Child_DirectionalPolygon(SPLManager *mgr, SPLParticle *ptcl);
```
lhearachel, line 25: **question:** Aren't these included by the header?

lib/spl/src/spl_draw.c
```diff
@@ -0,0 +1,953 @@
...
+    mat->m[2][0] = Sp;
+    mat->m[3][0] = 0;
+    mat->m[0][1] = Sp;
+    mat->m[1][1] = C;
+    mat->m[2][1] = Sm;
+    mat->m[3][1] = 0;
+    mat->m[0][2] = Sm;
+    mat->m[1][2] = Sp;
+    mat->m[2][2] = C;
+    mat->m[3][2] = 0;
```
lhearachel, line 117: **question:** Does it match to roll this into a double-loop or to do an explicit array-set?

lib/spl/src/spl_emitter.c
```diff
@@ -0,0 +1,442 @@
...
+    BOOL loop;
+} FieldFunc8;
+
+static void SPLUtil_SetTexture(SPLTexture *tex); // spl_set_tex
+static void SPLUtil_SetTexture_Stub(SPLTexture *tex); // spl_set_tex_dummy
+static void SPLManager_DrawParticles(SPLManager *mgr);
+static void SPLManager_DrawChildParticles(SPLManager *mgr);
+void SPLEmitter_Init(SPLEmitter *emtr, SPLResource *res, const VecFx32 *param2);
+void SPLManager_DoDraw(SPLManager *mgr);
+void SPLEmitter_Update(SPLManager *mgr, SPLEmitter *emtr);
```
lhearachel, line 36: **question:** Duplicate declarations?

lib/spl/src/spl_emitter.c
```diff
@@ -0,0 +1,442 @@
...
+            fx32 emissionDelay = FX_MUL((fx32)ptcl->lifeTime << FX32_SHIFT, (fx32)child->misc.emissionDelay << FX32_SHIFT);
+
+            // The >> 8 here is a division by 256 because emissionDelay is a fraction of the particle's lifetime represented as a u8
+            fx32 diff = ((fx32)ptcl->age * FX32_ONE) - (emissionDelay >> 8);
+
+            if (diff >= 0) {
+                if ((diff >> FX32_SHIFT) % child->misc.emissionInterval == 0) {
+                    SPLEmitter_EmitChildren(ptcl, emtr, &mgr->inactiveParticles);
+                }
+            }
```
lhearachel, line 231: **suggestion:** Merge these into a single-if conjunction.

lib/spl/src/spl_manager.c
```diff
@@ -0,0 +1,429 @@
...
+#include "spl_emitter.h"
+#include "spl_internal.h"
+#include "spl_particle.h"
+
+#define DECODE_WH(X) ((u16)(1 << ((X) + 3)))
+#define EMITTER_SHOULD_TERMINATE(emtr, header)                                                                                         \
+    ((                                                                                                                                 \
+         (header->flags.selfMaintaining && header->emitterLifeTime != 0 && emtr->state.started && emtr->age > header->emitterLifeTime) \
+         || emtr->state.terminate)                                                                                                     \
+        && emtr->particles.count == 0 && emtr->childParticles.count == 0)
```
lhearachel, line 17: **suggestion:** Either format the parens in multi-line style or pull the second line here up to the first line.

lib/spl/include/spl_behavior.h
```diff
@@ -0,0 +1,66 @@
...
+
+// Applies a collision behavior to particles
+typedef struct SPLCollisionPlaneBehavior {
+    fx32 y; // The Y position of the collision plane
+    fx16 elasticity; // The elasticity of the collision, 1.0 being perfectly elastic and 0.0 being perfectly inelastic
+
+    struct {
+        u16 collisionType : 2;
+        u16 : 14;
+    } flags;
```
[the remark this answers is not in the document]
Fexty12573: The bit width is required but I am able to take it out of the struct, the anonymous 14 bits are also not required,

lib/spl/src/spl_emit.c
Fexty12573: I actually did but it seems clang-format changed it again to this

lib/spl/src/spl_emit.c
Fexty12573: Unfortunately it doesn't match if I declare it in the for loop

lib/spl/src/spl_draw.c
Fexty12573: How would that work?

lhearachel: Reading the code is hard, ignore me on this one

lhearachel (approved):
This looks good to me; will merge after the fix to the build after a merge from `main`.

PR #263, opened 2024-08-31
Decompile zukan_data NARC

h2o-DS:
Extracted data from application/zukanlist/zkn_data/zukan_list.narc

Individual pokemon data was appended to their data.json file

Body shapes were added as an enum in consts.pokemon

A make.py script was added to tool/scripts

Corresponding meson files were changed

prebuilt narc was deleted


adrienntindall:
zukan translates to pokedex, I forget how much of that is able to be translated file name wise but you should incorporate where you can

lhearachel:
> zukan translates to pokedex, I forget how much of that is able to be translated file name wise but you should incorporate where you can

Everything except for the final NARC name can be translated, essentially

h2o-DS:
I only used zukan in reference to filenames, like how shinzukan and pl_pokezukan are currently implemented.

I don't love the data.json variable names I came up with though, so if you have a better idea than "trainer_scale_f" let me know. Maybe they should be in a "pokedex_data" list or something so their application is more obvious.

lhearachel:
> `pokedex_data`

I think this would be fine. I also wouldn't shy away from using names that are descriptive as possible, in general. The more obvious data files are to end users, the better.

src/pokedex_data_index.c
```diff
@@ -0,0 +1,38 @@
...
+    if (param0 == 0) {
+        pokedex_data_narc_index = NARC_INDEX_APPLICATION__ZUKANLIST__ZKN_DATA__ZUKAN_DATA_GIRA;
+        weight_message_bank_index = 708;
+        height_message_bank_index = 710;
+    } else {
+        pokedex_data_narc_index = NARC_INDEX_APPLICATION__ZUKANLIST__ZKN_DATA__ZUKAN_DATA;
+        weight_message_bank_index = 707;
+        height_message_bank_index = 709;
+    }
+}
```
lhearachel, line 23: **suggestion:** Give `param0` a name, and also document names for these message banks.

src/pokedex_data_index.c
```diff
@@ -0,0 +1,38 @@
+#include "pokedex_data_index.h"
+
+#include <nitro.h>
+#include <string.h>
+
+#include "narc.h"
+
+static u32 pokedex_data_narc_index = NARC_INDEX_APPLICATION__ZUKANLIST__ZKN_DATA__ZUKAN_DATA;
+static u32 weight_message_bank_index = 707;
+static u32 height_message_bank_index = 709;
```
lhearachel, line 10: **polish:** The rest of the repo typically:

1. Uses `camelCase`.
2. Prefixes static data symbols with `s`, e.g. `sWeightMessageBankIndex`.

src/pokedex_heightweight.c
```diff
@@ -0,0 +1,174 @@
...
+    GF_ASSERT(hw_data);
+    GF_ASSERT(hw_data->pokemon_scale);
+    return hw_data->pokemon_scale[dex_number];
+}
+
+static int *Pokedex_Height_Array(NARC *pokedex_data_narc, int param1)
+{
+    void *height;
+
+    height = sub_0200723C(pokedex_data_narc, 0, 0, param1, 0);
```
lhearachel, line 140: **polish:** Prefer declaration+definition in a single statement where possible.

res/pokemon/abomasnow/data.json
```diff
@@ -124,15 +124,19 @@
...
+        "trainer_scale_f": 335,
+        "pokemon_scale_f": 256,
+        "trainer_scale_m": 328,
+        "pokemon_scale_m": 256,
+        "trainer_pos_f": 15,
+        "pokemon_pos_f": 4,
+        "trainer_pos_m": 15,
+        "pokemon_pos_m": 4,
+        "entry_text": "They appear when the snow flowers\\nbloom. When the petals fall, they\\nretreat to places unknown again.",
+        "category": "Frost Tree Pok\u00e9mon"
```
lhearachel, line 140: **polish:** Can these not be encoded using UTF-8 encoding on the files?

res/pokemon/abomasnow/data.json
```diff
@@ -124,15 +124,19 @@
...
+        "body_shape": "SHAPE_BIPEDAL_TAILED",
+        "trainer_scale_f": 335,
+        "pokemon_scale_f": 256,
+        "trainer_scale_m": 328,
+        "pokemon_scale_m": 256,
+        "trainer_pos_f": 15,
+        "pokemon_pos_f": 4,
+        "trainer_pos_m": 15,
+        "pokemon_pos_m": 4,
+        "entry_text": "They appear when the snow flowers\\nbloom. When the petals fall, they\\nretreat to places unknown again.",
```
lhearachel, line 139: **question:** Do `\n` characters need to be double-escaped?

lhearachel (changes requested):

lhearachel (approved):

PR #276, opened 2024-10-23
support GDB debugging on overlays via custom GDB build

joshua-smith-12:
this includes a `launch.json` for VSCode which could be removed if desired.

I added a `debug` target for `build.sh` which will build additional resources (`overlay.map`) which are required for GDB automatically. This target could also be updated to automatically define `GDB_DEBUGGING` instead of a manual define in `game_overlay.h`


include/game_overlay.h
```diff
@@ -13,4 +13,28 @@ void Overlay_UnloadByID(const FSOverlayID param0);
 int Overlay_GetLoadDestination(const FSOverlayID param0);
 BOOL Overlay_LoadByID(const FSOverlayID param0, int param1);
 
+// TODO: this should integrate into build system instead of a manual define here.
```
lhearachel, line 16: See: https://mesonbuild.com/howtox.html#set-extra-compiler-and-linker-flags-from-the-outside-when-eg-building-distro-packages

meson.build
```diff
@@ -35,6 +34,15 @@ c_args = [
     '-sym', 'on'
 ]
 
+if get_option('gdb_debugging')
+    c_args += [
+        '-O1,p',
+        '-inline', 'off'
```
joshua-smith-12, line 40: this is to help the debugger out a little, ideally this would be `-O0` but I had some issues with running the ROM with optimizations disabled completely

lhearachel (approved):
This looks good to me; pre-approving so that I can merge once CI is fixed.

PR #269, opened 2024-10-15
Decompile pl_batt_obj NARC, document related battle_icon.c

joshua-smith-12:
- Decompile pl_batt_obj.narc (NCLR/NCGR decompiled fully, NCER/NANR components remain prebuilt as nitrogfx can't currently handle all of these correctly)
- Implement Python scripts to build the NARC
- Rename unk_0207C908 -> battle_icon and document (some) fields

TODO:
- update any unfixed PNG files from this NARC with the correct width + palette

this is my first contribution, tried my best to keep it to a similar style as the rest of the code but I probably scuffed it a little :)

PR #279, opened 2024-10-25
Unpack resdat and cldat files, create C headers for offsets into palette resdats

joshua-smith-12:
Uncovered how the resdat files and palette offsets were being used while I was working on #269 so figured I would unpack the files + add a build step while it was fresh in the brain

shoutout to the txt files such as `res/data/tmapn_celact.txt` which were probably left in the filesystem by mistake and contain a direct explanation for the binary format for resdat/cldat files, lol

I'll probably extend this to do a bit more documentation in the files I've already modified, specifically around `unk_0200C6E4.c`. I need to check if the other gen4 projects already did work in this area (I think pokeheartgold does have a bunch of this file documented already)


res/data/plist_canm.json
```diff
@@ -0,0 +1,47 @@
+{
+    "type": "ANIM",
+    "contents": [
+        {
+            "narc": 20,
+            "member": 0,
```
joshua-smith-12, line 6: I think it would be nice to reference the NARC and member by name here and have the build step convert them into IDs on the fly, but need to figure out where the data mapping NARCs and members to indices would be available at this point in the build

lhearachel: You could move them from the C constants in `include/constants/narc.h` to be a generated multi-lang constant in `consts/`. Then, you'd have access to these in the Python script that builds the NARCs.

src/unk_0208EA44.c
```diff
@@ -430,7 +431,7 @@ void sub_0208F16C(PokemonSummaryApp *param0)
 static void sub_0208F194(PokemonSummaryApp *param0, u8 param1, u8 param2, u8 param3)
 {
     sub_0200D948(param0->renderer, param0->gfxHandler, sub_0207C944(), sub_0207C908(param3), 1, param2);
-    sub_0200D414(param0->unk_41C[param1], sub_0207C92C(param3) + 3);
+    sub_0200D414(param0->unk_41C[param1], sub_0207C92C(param3) + RESDAT_PST_PAL_NARC_27_MEMBER_74_1_OFFSET);
```
lhearachel, line 434: Should this be an `naix` value?

lhearachel (commented):
Is there any documentation on what these files actually represent? Are they just packs of resources that should always be loaded together?

joshua-smith-12:
> Is there any documentation on what these files actually represent? Are they just packs of resources that should always be loaded together?

yes! if I understand it right,

- each `resdat` file contains a list of resources of 1 type (e.g. cells, tiles, anims, palettes) which are all loaded as `SpriteResource` types into a `SpriteResourceCollection`
- each `cldat` file contains a list of indices into these collections which can be used to construct a `CellActorResourceData` type which contains data for each resource type (`NNSG2dImageProxy`, `NNSG2dCellAnimBankData`, etc)

so all of these `resdat` files get loaded together, and then the `cldat` file constructs what is essentially templates of `CellActorResourceData` from the data in the resdats which can be used for constructing a `CellActor`

include/sprite_renderer.h
```diff
@@ -0,0 +1,179 @@
...
+    s16 z;
+    u16 anim;
+    int priority;
+    int palette;
+    int vramType;
+    int unk_18;
+    int unk_1C;
+    int unk_20;
+    int unk_24;
+} CellActorPropsTemplate;
```
lhearachel, line 62: **question**: How does this differ from the existing `SpriteTemplate`?

include/sprite_renderer.h
```diff
@@ -0,0 +1,179 @@
+#ifndef POKEPLATINUM_UNK_0200C6E4_H
+#define POKEPLATINUM_UNK_0200C6E4_H
```
lhearachel, line 2: **polish**: Modify the include guards (and the comment at the end of this file).

res/meson.build
```diff
@@ -68,3 +68,4 @@ subdir('items')
 subdir('pokemon')
 subdir('sound')
 subdir('trainers')
+subdir('data')
```
lhearachel, line 71: **polish**: Keep this list sorted in lexicographical order, unless a specific processing order is needed.

src/applications/pokemon_summary_screen/sprite.c
```diff
@@ -7,8 +7,7 @@
 
 #include "applications/pokemon_summary_screen/main.h"
 #include "applications/pokemon_summary_screen/subscreen.h"
-#include "overlay007/struct_ov7_0224F2EC.h"
-#include "overlay007/struct_ov7_0224F358.h"
+#include "data/pst_pal.naix"
```
lhearachel, line 10: **polish**: `#include "res/data/pst_pal.naix"` will sort this to the bottom of the includes, which I think is preferable.

src/sprite_renderer.c
```diff
@@ -0,0 +1,1277 @@
...
+
+    sub_0201E994();
+    sub_0201F8E4();
+
+    return TRUE;
+}
+
+BOOL SpriteRenderer_CreateCellActorList(SpriteRenderer *renderer, SpriteGfxHandler *gfxHandler, int numCells)
+{
+    if ((renderer == NULL) || (gfxHandler == NULL)) {
```
lhearachel, line 109: **polish**: Unnecessary parens.

src/sprite_renderer.c
```diff
@@ -0,0 +1,1277 @@
...
+    {
+        UnkStruct_ov22_022559F8 v0;
+
+        v0.unk_00 = param2->unk_00;
+        v0.unk_04 = param2->unk_04;
+        v0.unk_08 = param2->unk_08;
+        v0.unk_0C = renderer->heapID;
+
+        sub_0201E88C(&v0, param2->unk_0C, param2->unk_10);
+    }
```
lhearachel, line 90: **polish**: I'll leave it up to you if you want to break down this unnecessary scope-block or not; since the inner struct here is undocumented, I can let it slide.

src/sprite_renderer.c
```diff
@@ -0,0 +1,1277 @@
...
+
+            resourceList->resources[i] = NULL;
+            resourceList->count--;
+
+            return TRUE;
+        }
+    }
+
+    return FALSE;
+}
```
lhearachel, line 721: **polish**: Can we name these functions?

res/data/btower_canm.json
```diff
@@ -0,0 +1,11 @@
+{
+    "kind": "anim",
+    "graphics": [
+        {
+            "narc": 123,
+            "file": 3,
+            "compressed": false,
+            "id": 0
+        }
```
lhearachel, line 9: **suggestion**: I think that, in general, I would rather we get these to map to NARCs and NAIX files. Maybe this means that the conversion utility should be in C code so that we can easily import the necessary header files.

lhearachel (commented):
Some cursory questions and basic polish comments.

lhearachel:
Code changes herein are superseded by #367.

The unpacking for the actual `resdat` files will be dependent on unpacking the actual NARCs that they reference, and so is out-of-scope for now.

PR #293, opened 2024-11-05
Document field_battle_data_transfer.c

lhearachel:
This documents the data structure and code used for transferring data between the `field` and `battle` processes. Some additional files are documented herein to make the transfer code easier to understand.


src/map_tile_behavior.c
```diff
@@ -1,697 +1,736 @@
...
-    0x0,
-    0x0,
-    0x0
+#define TILE_BEHAVIOR_FLAG_NONE      0
+#define TILE_BEHAVIOR_FLAG_SURFABLE  (1 << 0)
+#define TILE_BEHAVIOR_FLAG_ENCOUNTER (1 << 1)
+
+#define TILE_BEHAVIOR_FLAG_SURFABLE_ENCOUNTER (TILE_BEHAVIOR_FLAG_ENCOUNTER | TILE_BEHAVIOR_FLAG_SURFABLE)
+
+static const u8 sTileBehaviorFlags[] = {
```
cbt6, line 9: Use designated initializers here (and use the consts in the various `TileBehavior_IsX` functions)

lhearachel: I can specify hexadecimal bytes since we don't have constants yet; I opted not to because I wasn't certain how useful that was, but it's probably useful enough with an array this large.

lhearachel: Done in most recent commit.

PR #294, opened 2024-11-06
Collapse shell scripts into a Makefile; update INSTALL and CONTRIBUTING docs

mid-kid:
Supersedes https://github.com/pret/pokeplatinum/pull/274


lhearachel (approved):
image

PR #297, opened 2024-11-10
Decompile encdata_ex and Move Pokedex Narcs

h2o-DS:
Created res/field/encounters_ex to store encdata_ex info
Moved pokedex narcs out of prebuilt
Added pokemon/*/data.json as input for relevant narcs


res/text/meson.build
```diff
@@ -742,6 +742,511 @@ gmm_to_bin_gen = generator(msgenc_exe,
     output: '@BASENAME@'
 )
 
+# Process mons in national dex order
+pokemon_subdirs = [
```
h2o-DS, line 746: I don't like duplicating a list like this, especially such a long one. If possible, pokemon_subdirs may want to be a file.

res/field/encounters_ex/honey_tree_common.json
```diff
@@ -0,0 +1,10 @@
+{
```
lhearachel, line 1: **suggestion**: Combine the honey tree encounters into a single file:

```json
{
    "common": [ ... ],
    "uncommon": [ ... ],
    "rare": [ ... ]
}

File: res/field/encounters_ex/great_marsh_dailies_natdex.json [hunk 3]
```diff
@@ -0,0 +1,36 @@
+{
```
lhearachel, line 1: **suggestion**: Combine the Great Marsh daily lookout encounters into a single file, including the coordinates that they map to, if that is known.

res/field/encounters_ex/mt_coronet_lake.json
```diff
@@ -0,0 +1,537 @@
+{
+    "species": "SPECIES_FEEBAS",
+    "map_dimensions": [
+        228,
+        300
+    ],
+    "tiles": [
```
lhearachel, line 7: **question**: Is this a bounding-box? Can we just compute it from a set of rectangle-definitions?

res/field/encounters_ex/trophy_garden_dailies.json
```diff
@@ -0,0 +1,20 @@
+{
```
lhearachel, line 1: **suggestion**: Store this data in the trophy garden's encounter file.

res/field/meson.build
```diff
@@ -13,3 +14,213 @@ tutorable_moves_h = custom_target('tutorable_moves.h',
         '--output-dir', meson.current_build_dir(),
     ]
 )
+
+zukan_enc_platinum_srcs = files(
```
lhearachel, line 18: **suggestion**: If you get rid of all the files in `encounters_ex`, then move all of this build logic into the `encounters` folder.

res/text/meson.build
lhearachel: Is it maybe preferable to generate these from the `pokemon` directory and then output the files into `text`? I have an idea of how I would want this to work in a refactored build-system.

lhearachel (changes requested):

lhearachel:
This PR should be merged/re-reviewed after #299.

res/text/meson.build
h2o-DS: Moving these messages to the ```pokemon``` directory would certainly make generating them easier. What was your envisioned implementation?

res/field/encounters_ex/mt_coronet_lake.json
h2o-DS: Upon further investigation it appears that these tiles are essentially a result of `tile = (v14 * 32 * y) + x` where I am guessing v14 is the height of the map or something like that. Not sure why it is multiplied by 32. Either way, if I can debug v14 during runtime I could turn these tile numbers into (x,y) coordinates and then it would be easier to make a bounding box, if possible.

h2o-DS:
I have figured out how to pass the list of files from meson to python, so I have removed globbing in the scripts relevant for this PR. I have made it so that files are handled exclusively in meson.

tools/scripts/make_encdata_ex.py
```diff
@@ -35,63 +47,74 @@
 
 NUM_FILES = 12
 
-numDigits = len(str(NUM_FILES))
+NUM_DIGITS = len(str(NUM_FILES))
 
-def WriteBin(binData, fileNum):
-    target_fname = str(private_dir / output_name) + f'_{fileNum:0{numDigits}}.bin'
+def Write_Binary(bin_data, fileNum):
```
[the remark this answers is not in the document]
h2o-DS: Seeing as this is a script that creates and processes the binaries I didn't see the point in naming them. Especially because the files do not coincide 1:1 with the input and would require their own names. However, I do see the value in making identifying the files easier, so I can add an order file. Would it be reasonable to generate that within this script?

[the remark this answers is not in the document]
lhearachel: I would sooner say that you should list it as a static file, so that we can mandate its inclusion and verify its contents within the repo. Then, you can remove this binary-mapping code from your script and directly process to target file-names built from the input stems.

lhearachel (approved):

PR #298, opened 2024-11-11
Unpack summary screen graphics

ravepossum:
Unpacks all the graphics in the summary screen graphics NARC `pl_pst_gra.narc`.

There is an issue with NCER generation in nitrogfx that has to be addressed in order for the .jsons in this PR to match, so this will stay as a draft until [the relevant nitrogfx PR](https://github.com/red031000/nitrogfx/pull/18) is merged. It should be ready to review though.


include/constants/narc.h
```diff
@@ -42,7 +42,7 @@ enum {
     NARC_INDEX_GRAPHIC__FIELD_BOARD,
     NARC_INDEX_FIELDDATA__ENCOUNTDATA__D_ENC_DATA,
     NARC_INDEX_GRAPHIC__PL_WINFRAME,
-    NARC_INDEX_GRAPHIC__PL_PST_GRA,
```
lhearachel, line 45: **issue**: We usually keep these aligned with the actual path of the NARC; this would be a marked shift away from that standard. I can be convinced that that's the right direction moving ahead, but we should file a follow-up issue if we decide as-such.

res/graphics/pokemon_summary_screen/a_button.png
lhearachel, line 1: **suggestion**: Align this (and other graphics) as recognizable squares, if possible, using `nitrogfx` options for `mwidth` and `mheight`.

res/graphics/pokemon_summary_screen/move_info.bin
lhearachel, line 1: **polish**: I assume that this is a tilemap (`.NSCR`); if so, then we should save it with the `.NSCR` extension so that Tilemap Studio (and other programs) can immediately recognize it.

lhearachel (changes requested):

include/constants/narc.h
ravepossum: I was thinking it might be nice for it to have an actual name since the file can't have one, but the original name does more clearly convey that it's associated with that NARC file. I switched it back.

res/graphics/pokemon_summary_screen/a_button.png
ravepossum: There's unfortunately a few stubborn ones that can't be made to look nice (pokerus icon, move cursor, ribbon cursor, tabs) but the rest should be fixed now.

lhearachel (approved):

PR #300, opened 2024-11-12
Refactor generation of event data files and packing of zone_event.narc

lhearachel:


tools/json2bin/event.py
```diff
@@ -0,0 +1,99 @@
...
+    ])
+
+def parse_bg_events(bg_events: list[dict]) -> bytes:
+    return b''.join([
+        u32(len(bg_events)),
+        *[parse_bg_event(bg) for bg in bg_events],
+    ])
+
+def parse_object_event(obj: dict, i: int) -> bytes:
+    obj_data = obj['data'] + [0] * (3 - len(obj['data'])) # pad up to length 3
```
lhearachel, line 26: Just making a note that I *did* test this in the Python REPL by trying to pad with a negative result size, and Python gracefully refused to alter it:

```
>>> l = []
>>> l += [0] * 3
>>> l
[0, 0, 0]
>>> l += [0] * -5
>>> l
[0, 0, 0]
```

tools/json2bin/event.py
```diff
@@ -0,0 +1,99 @@
...
+    assert len(obj_data) <= 3, f"more than 3 values found in object event data ({obj_data})"
```
cbt6, line 27: Raise explicit error instead of doing assertion.

PR #301, opened 2024-11-13
Specify exact commit in subproject files

mid-kid:
- **Create script to "lock" subprojects by commit**
- **Update subprojects**


subprojects/update.py
lhearachel, line 1: Should this be added to `Makefile` in some fashion?

mid-kid: No, this is for us maintainers to run when relevant. Users shouldn't really touch it.

PR #302, opened 2024-11-15
Rename UnkStruct_ov16_0226AC98 to MoveDisplayName and rename loops in battle_display.c and ov16_0226871C.c

Viperio19:
`UnkStruct_ov16_0226AC98` is now `MoveDisplayName` with struct members `move`, `curPP` and `maxPP`.
Variables used for loops in `battle_display.c` and `ov16_0226871C.c` are now `i` or `j`, and loops that go over all moves with `... < 4` now use `... < LEARNED_MOVES_MAX`.


include/battle/move_display_info.h
```diff
@@ -0,0 +1,10 @@
+#ifndef POKEPLATINUM_MOVE_DISPLAY_INFO
+#define POKEPLATINUM_MOVE_DISPLAY_INFO
+
+typedef struct MoveDisplayInfo {
+    u16 move[4];
```
lhearachel, line 5: **polish**: Add `#include "constants/moves.h"` at the top so that you can use `LEARNED_MOVES_MAX` in these array declarations.

lhearachel (commented):
This looks generally fine to me. There are some extra `4` values that I think are tied to `LEARNED_MOVES_MAX`, but don't sweat looking through those.

include/battle/move_display_info.h
Viperio19: good suggestion, I've added that

lhearachel (approved):

PR #304, opened 2024-11-16
Document encounter.c

lhearachel:
`StartTrainerBattle` scripting commands are still in need of decoration with constants; that will be a future task after the trainer data build is refactored and generates an NAIX header from semantic names.


include/encounter.h
```diff
@@ -8,20 +8,19 @@
...
-void sub_02051450(FieldSystem *fieldSystem, FieldBattleDTO *param1);
-void sub_02051480(FieldTask *param0, int param1, int param2, int *param3);
-void sub_02051590(FieldTask *param0);
-void sub_020515CC(FieldTask *param0, int param1, int param2, int param3, int param4, int *param5);
-void sub_0205167C(FieldTask *param0, const u8 *param1, int param2);
-void sub_020516F4(FieldTask *param0, int param1, int param2, int param3);
-void sub_020517E8(FieldSystem *fieldSystem, const u8 *param1, int param2);
-void sub_0205184C(FieldSystem *fieldSystem, const Party *param1, int param2);
-void sub_02051ABC(FieldTask *param0, u16 param1, u8 param2, int *param3, BOOL param4);
+void Encounter_NewVsWild(FieldSystem *fieldSystem, FieldBattleDTO *dto);
```
ravepossum, line 11: **thought:** I know we adhere pretty closely to the `Module_Function()` format but I wonder here if we would gain any clarity by introducing something like `Module_Function_Type()`. `Encounter_New_VsWild` and `Encounter_New_VsTrainer` for example. It is a bit snake case-y, but it reads more clearly to my brain at least.

include/unk_02048614.h
```diff
@@ -8,8 +8,8 @@ BOOL ScrCmd_0B7(ScriptContext *param0);
 BOOL ScrCmd_0B8(ScriptContext *param0);
 BOOL ScrCmd_0B9(ScriptContext *param0);
 BOOL ScrCmd_0E4(ScriptContext *param0);
-BOOL ScrCmd_0E5(ScriptContext *param0);
-BOOL ScrCmd_2A0(ScriptContext *param0);
+BOOL ScrCmd_StartTrainerBattle(ScriptContext *param0);
```
ravepossum, line 11: **nitpick**: 
```diff
- ScrCmd_StartTrainerBattle(ScriptContext *param0);
+ ScrCmd_StartTrainerBattle(ScriptContext *ctx);
```

src/scrcmd.c
```diff
@@ -4914,20 +4916,18 @@ static BOOL ScrCmd_0F5(ScriptContext *ctx)
...
+static BOOL ScrCmd_StartLinkBattle(ScriptContext *ctx)
 {
-    PartyManagementData *v0;
-    void **v1;
-
-    v1 = FieldSystem_GetScriptMemberPtr(ctx->fieldSystem, 19);
-    v0 = *v1;
+    PartyManagementData *partyManagementData;
+    void **partyManagementDataPtr = FieldSystem_GetScriptMemberPtr(ctx->fieldSystem, SCRIPT_MANAGER_PARTY_MANAGEMENT_DATA);
+    partyManagementData = *partyManagementDataPtr;
```
ravepossum, line 4923: **nitpick:**
```diff
-    PartyManagementData *partyManagementData;
     void **partyManagementDataPtr = FieldSystem_GetScriptMemberPtr(ctx->fieldSystem, SCRIPT_MANAGER_PARTY_MANAGEMENT_DATA);
-    partyManagementData = *partyManagementDataPtr;
+    PartyManagementData *partyManagementData = *partyManagementDataPtr;
```

src/encounter.c
```diff
@@ -54,982 +55,951 @@
...
 
-        sub_020518B0(fieldSystem, v1->unk_10);
-        sub_02051988(fieldSystem, v1->unk_10);
+        UpdateGameRecords(fieldSystem, encounter->dto);
+        UpdateJournal(fieldSystem, encounter->dto);
 
         if (GetRadarChainActive(fieldSystem->chain)) {
             if (sub_02069798(fieldSystem->chain)) {
-                if ((!(v1->unk_10->resultMask == BATTLE_RESULT_WIN)) && (!(v1->unk_10->resultMask == BATTLE_RESULT_CAPTURED_MON))) {
+                if ((!(encounter->dto->resultMask == BATTLE_RESULT_WIN)) && (!(encounter->dto->resultMask == BATTLE_RESULT_CAPTURED_MON))) {
```
ravepossum, line 415: **polish:**
```diff
-    if ((!(encounter->dto->resultMask == BATTLE_RESULT_WIN)) && (!(encounter->dto->resultMask == BATTLE_RESULT_CAPTURED_MON))) {
+    if (encounter->dto->resultMask != BATTLE_RESULT_WIN && encounter->dto->resultMask != BATTLE_RESULT_CAPTURED_MON) {
```

src/encounter.c
```diff
@@ -54,982 +55,951 @@
...
             }
         } else {
-            PCBoxes *v7 = SaveData_PCBoxes(fieldSystem->saveData);
-            Party *v8 = Party_GetFromSavedata(fieldSystem->saveData);
+            PCBoxes *boxes = SaveData_PCBoxes(fieldSystem->saveData);
+            Party *party = Party_GetFromSavedata(fieldSystem->saveData);
 
-            if ((PCBoxes_FirstEmptyBox(v7) == 18) && (Party_GetCurrentCount(v8) == 6)) {
-                ScriptManager_Start(taskMan, 8822, NULL, NULL);
+            if (PCBoxes_FirstEmptyBox(boxes) == 18 && Party_GetCurrentCount(party) == MAX_PARTY_SIZE) {
```
ravepossum, line 522: **suggestion**: Might not hurt to go ahead and add a `MAX_PC_BOXES` constant or something to `include/unk_020797C8.h`, so there's less stuff like this to hunt down later.

src/encounter.c
```diff
@@ -54,982 +55,951 @@
...
-    FieldBattleDTO *v1;
-    FieldSystem *fieldSystem = FieldTask_GetFieldSystem(taskMan);
+    Encounter *encounter;
+    FieldBattleDTO *dto;
+    FieldSystem *fieldSystem = FieldTask_GetFieldSystem(task);
 
-    v1 = FieldBattleDTO_NewCatchingTutorial(11, fieldSystem);
-    v0 = sub_02050ACC(v1, EncEffects_CutInEffect(v1), EncEffects_BGM(v1), NULL);
+    dto = FieldBattleDTO_NewCatchingTutorial(HEAP_ID_FIELDMAP, fieldSystem);
+    encounter = NewEncounter(dto, EncEffects_CutInEffect(dto), EncEffects_BGM(dto), NULL);
```
ravepossum, line 731: **polish:**
```diff
-    Encounter *encounter;
-    FieldBattleDTO *dto;
-    FieldSystem *fieldSystem = FieldTask_GetFieldSystem(task);
-
-    dto = FieldBattleDTO_NewCatchingTutorial(HEAP_ID_FIELDMAP, fieldSystem);
-    encounter = NewEncounter(dto, EncEffects_CutInEffect(dto), EncEffects_BGM(dto), NULL);
+    FieldSystem *fieldSystem = FieldTask_GetFieldSystem(task);
+    FieldBattleDTO *dto = FieldBattleDTO_NewCatchingTutorial(HEAP_ID_FIELDMAP, fieldSystem);
+    Encounter *encounter = NewEncounter(dto, EncEffects_CutInEffect(dto), EncEffects_BGM(dto), NULL);
```

And so on for the similar functions.

src/encounter.c
```diff
@@ -54,982 +55,951 @@
...
         v0 = (UnkEnum_0202F510_14);
-    } else if (param1 & 0x2) {
+    } else if (battleType & BATTLE_TYPE_DOUBLES) {
         v0 = (UnkEnum_0202F510_07);
     } else {
         v0 = (UnkEnum_0202F510_00);
     }
 
-    if (v1 != 0xff) {
+    if (v1 != 0xFF) {
```
ravepossum, line 796: 🙏 all my homies hate lowercase hex

src/encounter.c
```diff
@@ -54,982 +55,951 @@
...
+    int introEffectID;
+    int battleBGM;
+    int *resultMaskPtr;
+    FieldBattleDTO *dto;
+} WildEncounter;
+
+static Encounter *NewEncounter(FieldBattleDTO *dto, int introEffectID, int battleBGM, int *resultMaskPtr);
+static WildEncounter *NewWildEncounter(FieldBattleDTO *dto, int introEffectID, int battleBGM, int *resultMaskPtr);
+static void FreeEncounter(Encounter *encounter);
+static void FreeWildEncounter(WildEncounter *encounter);
```
ravepossum, line 77: **question:** Should these still follow the `Module_Function()` format even though they're static, for clarity/uniformity?

src/encounter.c
```diff
@@ -54,982 +55,951 @@
...
-
-    if (param3 == 0) {
-        v4 = (0x4 | 0x1);
-        v2 = FieldBattleDTO_New(11, (0x4 | 0x1));
+    FieldSystem *fieldSystem = FieldTask_GetFieldSystem(task);
+    Encounter *encounter;
+    FieldBattleDTO *dto;
+    int battleType, recordingResultCode, v5;
+
+    if (wifiBattleType == 0) {
```
ravepossum, line 810: **question:** Is this something we have enough information about to make descriptive constants at this time?

ravepossum (commented):
I didn't have too much of substance here - just a few thoughts and nitpicks.

One (semi-related) thought I had while looking at this is that at some point we may want to revisit the naming of `encounter_effect.c` and `enc_effects.c`. From those names alone, it's unclear what the distinction is.

src/encounter.c
lhearachel: This doesn't match for all of these similar functions, unfortunately, which is why I kept them this way for overall consistency.

src/encounter.c
lhearachel: We don't really have a rule for that for `static` functions; I find that the simpler format here is a bit more readable, and clearly communicates that these are `static`.

src/encounter.c
lhearachel: Not really, in my opinion, which is unfortunate. But, this function is only invoked in one place, so it's relatively easy to track down whenever that calling function is documented.

include/encounter.h
lhearachel: How would this extend for functions that have varying parameters, e.g. `Encounter_NewVsLink`, `Encounter_NewVsLinkWithRecording`, and `Encounter_NewVsLinkWithRecordingAndParty`?

ravepossum (approved):

PR #307, opened 2024-11-18
Renaming variables and functions related to overworld poison damage and FieldOverworldState

Viperio19:
First commit: Renaming `sub_02054B04` to `Pokemon_DoPoisonDamage` and renamng all the variables in that function and related functions.
Second and third commit: Renaming various functions and variables in `field_overworld_state.c` and related files.
Fourth commit: Renaming variables in `unk_02054884.c`.

There's one function name in `field_overworld_state.c` and some in `unk_02054884.c` which I wasn't sure what to name, so I left those as they are. Also, let me know if this PR was too all over the place, I just started from somewhere and then one renamed variable lead me to other variables to be renamed.


src/overlay005/field_control.c
```diff
@@ -905,14 +905,14 @@ static BOOL Field_UpdatePoison(FieldSystem *fieldSystem)
         return FALSE;
     }
 
-    switch (sub_02054B04(party, MapHeader_GetMapLabelTextID(fieldSystem->location->mapId))) {
+    switch (Pokemon_DoPoisonDamage(party, MapHeader_GetMapLabelTextID(fieldSystem->location->mapId))) {
     case 0:
```
ravepossum, line 909: **suggestion:** It would be good to add descriptive constants for these cases (and the corresponding return values in `Pokemon_DoPoisonDamage()`.

src/overlay005/fieldmap.c
```diff
@@ -400,24 +400,24 @@ static BOOL ov5_021D119C(FieldSystem *fieldSystem)
 static BOOL FieldMap_ChangeZone(FieldSystem *fieldSystem)
 {
     u32 v0;
-    u32 v1;
-    int v2, v3;
-    FieldOverworldState *v4;
+    u32 mapId;
```
ravepossum, line 403: **polish:** We're working on cleaning up C89-style declarations to be more concise and declared when needed, as long as it matches. For example:

```diff
-    u32 mapId;
...
-    mapId = fieldSystem->location->mapId;
+    u32 mapId = fieldSystem->location->mapId;
```
I think these should match.

src/overlay005/fieldmap.c
```diff
@@ -426,10 +426,10 @@ static BOOL FieldMap_ChangeZone(FieldSystem *fieldSystem)
     }
 
     {
```
ravepossum, line 428: **polish:** We're also working on cleaning up unnecessary scope blocks like these.

src/overlay005/fieldmap.c
```diff
@@ -426,10 +426,10 @@ static BOOL FieldMap_ChangeZone(FieldSystem *fieldSystem)
...
-        int v5 = MapHeaderData_GetNumObjectEvents(fieldSystem);
-        const ObjectEvent *v6 = MapHeaderData_GetObjectEvents(fieldSystem);
+        int npcCount = MapHeaderData_GetNumObjectEvents(fieldSystem);
```
ravepossum, line 429: **suggestion:** I see we do use NPC in a few places in the repo right now, but it feels a bit inaccurate to me. I feel like it doesn't encapsulate the full range of object events, and it's not uniform with the struct field it retrieves. I think I'd go with `objEventCount` or something, for example.

src/overlay023/ov23_022521F0.c
```diff
@@ -937,17 +937,17 @@ void ov23_022534A0(FieldSystem *fieldSystem)
...
-        v10 = MapHeaderData_GetObjectEvents(v4->fieldSystem);
+        npcCount = MapHeaderData_GetNumObjectEvents(v4->fieldSystem);
+        npcList = MapHeaderData_GetObjectEvents(v4->fieldSystem);
 
-        for (v9 = 0; v9 < v8; v9++) {
-            if ((v10[v9].x == v1) && (v10[v9].z == v2)) {
-                v11 = MapObjMan_LocalMapObjByIndex(v4->fieldSystem->mapObjMan, v9);
-                MapObject_SetDir(v11, v3);
+        for (i = 0; i < npcCount; i++) {
+            if ((npcList[i].x == v1) && (npcList[i].z == v2)) {
```
ravepossum, line 948: **polish:** The inner sets of parentheses here can be discarded.

src/unk_0206AFE0.c
```diff
@@ -455,10 +455,10 @@ static u8 sub_0206B4A4(VarsFlags *param0)
 
 void sub_0206B514(SaveData *param0)
 {
-    u32 v0;
+    u32 unused;
```
ravepossum, line 458: **polish:** You should just be able to discard this variable.

src/unk_02054884.c
```diff
@@ -175,41 +175,41 @@ void Party_GiveChampionRibbons(Party *party)
     }
 }
 
-int sub_02054B04(Party *param0, u16 param1)
+int Pokemon_DoPoisonDamage(Party *party, u16 param1)
```
ravepossum, line 178: **suggestion:** `param1` here looks like `mapLabelTextID`, looking at the invocation in `Field_UpdatePoison()` as well as what it eventually gets compared to (met location) in `Pokemon_UpdateFriendship()`.

src/unk_02054884.c
```diff
@@ -175,41 +175,41 @@ void Party_GiveChampionRibbons(Party *party)
...
 
-    for (v2 = 0; v2 < v3; v2++) {
-        v4 = Party_GetPokemonBySlotIndex(param0, v2);
+    for (i = 0; i < partyCount; i++) {
+        mon = Party_GetPokemonBySlotIndex(party, i);
 
-        if (Pokemon_CanBattle(v4)) {
-            if (Pokemon_GetValue(v4, MON_DATA_STATUS_CONDITION, NULL) & (MON_CONDITION_TOXIC | MON_CONDITION_POISON)) {
-                u32 v5 = Pokemon_GetValue(v4, MON_DATA_CURRENT_HP, NULL);
+        if (Pokemon_CanBattle(mon)) {
```
ravepossum, line 190: **suggestion:** I think you can collapse these two `if` statements into one with a conjunction and it'll still match.

src/unk_020553DC.c
```diff
@@ -119,23 +119,23 @@ void sub_020553DC()
     return;
 }
 
-void sub_020553F0(FieldSystem *fieldSystem, u16 param1)
+void Sound_SetSpecialBGM(FieldSystem *fieldSystem, u16 param1)
```
ravepossum, line 122: **suggestion:** I think `param1` here is a SDAT ID. For all the invocations of this function that pass a constant, you should be able to find a matching constant in the SDAT consts file. I would recommend just ctrl+Fing for the number in the generated file `build/consts/sdat.h`.

src/unk_020553DC.c
```diff
@@ -119,23 +119,23 @@ void sub_020553DC()
...
 {
-    u16 *v0 = sub_0203A748(SaveData_GetFieldOverworldState(fieldSystem->saveData));
+    u16 *v0 = FieldOverworldState_GetSpecialBGM(SaveData_GetFieldOverworldState(fieldSystem->saveData));
 
     *v0 = param1;
     return;
```
ravepossum, line 127: **polish:** This `return` can be removed.

src/unk_02054884.c
```diff
@@ -31,55 +31,55 @@ BOOL Pokemon_CanBattle(Pokemon *mon)
...
 {
     int v0;
-    BOOL v1;
-    TrainerInfo *v2 = SaveData_GetTrainerInfo(param1);
-    Party *v3 = Party_GetFromSavedata(param1);
-    Pokemon *v4 = Pokemon_New(32);
+    BOOL result;
+    TrainerInfo *trainerInfo = SaveData_GetTrainerInfo(param1);
+    Party *party = Party_GetFromSavedata(param1);
+    Pokemon *mon = Pokemon_New(32);
```
ravepossum, line 67: **polish:** The `32` here is a heap ID - `HEAP_ID_FIELD_TASK`.

src/unk_02054884.c
```diff
@@ -31,55 +31,55 @@ BOOL Pokemon_CanBattle(Pokemon *mon)
 
 BOOL sub_020548B0(int param0, SaveData *param1, u16 param2, u8 param3, u16 param4, int param5, int param6)
```
ravepossum, line 32: **thought:** I wonder if this and the function below are documented enough to name at this point. Looks like this particular one is only invoked in the script command `ScrCmd_096`, which DSPRE has named `GivePokemon`. The one below still has a bit more mystique in how it's invoked, so maybe just this one.

ravepossum (commented):
A few thoughts on names, and a variety of small style stuff we're working on cleaning up. For the style matters, I only commented on the first occurrence of each that I noticed. Thanks for another contribution :)

src/unk_02054884.c
Viperio19: just wondering in general, is there a difference between a `mapId` and a `mapLabelTextId`?

lhearachel: I personally prefer `ID`, but it's not a huge deal in either direction

include/field_map_change.h
```diff
@@ -13,23 +13,23 @@ void sub_02053494(FieldSystem *fieldSystem);
 void FieldSystem_SetLoadNewGameSpawnTask(FieldSystem *fieldSystem);
 void FieldSystem_SetLoadSavedGameMapTask(FieldSystem *fieldSystem);
 void FieldSystem_StartLoadMapFromErrorTask(FieldSystem *fieldSystem);
-void FieldSystem_StartChangeMapTask(FieldTask *param0, const Location *param1);
-void FieldTask_ChangeMapByLocation(FieldTask *param0, const Location *param1);
-void FieldTask_ChangeMapToLocation(FieldTask *param0, int param1, int param2, int param3, int param4, int param5);
-void FieldTask_StartMapChangeFull(FieldTask *param0, int param1, int param2, int param3, int param4, int param5);
+void FieldSystem_StartChangeMapTask(FieldTask *taskMan, const Location *param1);
```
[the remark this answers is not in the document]
lhearachel: It's being normalized to `task` for `FieldTask`, because there's no "management" aspect to it. There is a light call-stack, but that doesn't totally qualify it as a "manager" to me.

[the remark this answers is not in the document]
Viperio19: should I change all of them to `task` then? I can just immediately do it in other files as well, if this is something that you want to normalize across the whole repo

src/unk_02054884.c
ravepossum: I believe `mapID` is the unique numeric identifier for the map (so basically what it sounds like), while `mapLabelTextID` refers to the index of the text entry that corresponds to its name in `message_bank_location_names.gmm`.

I agree with Lhea, I prefer and have been using `ID` rather than `Id`.

src/overlay005/fieldmap.c
ravepossum: Looks like there's still a couple pesky extra scope blocks in this function.

Viperio19: Should be taken care of now

ravepossum (approved):

PR #313, opened 2024-11-30
Document some map object functions

cbt6:


include/struct_decls/struct_02061AB4_decl.h
```diff
@@ -1,6 +1,6 @@
 #ifndef POKEPLATINUM_STRUCT_02061AB4_DECL_H
 #define POKEPLATINUM_STRUCT_02061AB4_DECL_H
 
-typedef struct UnkStruct_02061AB4_t MapObject;
+typedef struct MapObject MapObject;
```
lhearachel, line 4: **question**: Is this struct too tangled to extract into `map_object.h` at present?

lhearachel (approved):
This looks good to me. Thanks for decorating all of those function invocations with appropriate constants. 🙏 

My question below is non-blocking.

include/struct_decls/struct_02061AB4_decl.h
cbt6: Yes, at least for me. There are some circular dependency going on with it and its fields, where the struct definitions are spread across multiple different headers, and personally I'm not too interested in spending much more effort into untangling it.

PR #315, opened 2024-12-08
Add base-devel to MSYS2 install

h2o-DS:


lhearachel (approved):

mid-kid:
base-devel is intended for the purposes of developing msys2 packages themselves, and includes a bunch of dependencies that aren't necessary. What was missing?
If base-devel is to be kept, I'd suggest removing some dependencies from that list that are now made redundant by base-devel.

h2o-DS:
Make was missing.

PR #318, opened 2024-12-13
Document some movement action functions

cbt6:


src/unk_020655F4.c
```diff
@@ -19,68 +22,66 @@
...
     s8 unk_0F;
-} UnkStruct_02065E0C;
+} JumpMovementData;
 
-typedef struct {
-    int unk_00;
-} UnkStruct_020662AC;
+typedef struct DelayMovementData {
+    int delay;
+} DelayMovementData;
```
lhearachel, line 60: **question**: Would the code look/feel cleaner if we decayed this `struct` to a `typedef`?

lhearachel (commented):
There is a lot here, but it largely looks good to me. I just have the one lingering question.

src/unk_020655F4.c
lhearachel: Resolved on discord; we will keep this as a `struct`.

lhearachel (approved):

PR #321, opened 2024-12-25
Document pokedex_data

h2o-DS:
unk_0202631 -> pokedex_data

This is the file that manages all form data encountered, including genders and languages.


src/pokedex_data.c
```diff
@@ -0,0 +1,1240 @@
...
+    u32 magic;
+    u32 caughtPokemon[DEX_SIZE_U32];
+    u32 seenPokemon[DEX_SIZE_U32];
+    u32 recordedGenders[2][DEX_SIZE_U32];
+    u32 spindaForms;
+    u8 shellosForms;
+    u8 gastrodonForms;
+    u8 burmyForms;
+    u8 wormadamForms;
+    u8 unownForms[UNOWN_COUNT];
```
cbt6, line 36: Do these keep track of the forms being seen or caught?

src/pokedex_data.c
```diff
@@ -0,0 +1,1240 @@
...
+    }
+
+    return ReadBit_3Forms(formArray, formIndex);
+}
+
+static BOOL CountsForDexCompletion_National(u16 species)
+{
+    int i;
+    BOOL included;
+    static const u16 mythicals[NUM_MYTHICALS_NATIONAL] = {
```
cbt6, line 697:
```suggestion
    static const u16 excludedMons[NUM_EXCLUDED_FROM_NATIONAL] = {
```

src/pokedex_data.c
```diff
@@ -0,0 +1,1240 @@
...
+#include "constants/species.h"
+
+#include "heap.h"
+#include "inlines.h"
+#include "pokemon.h"
+#include "savedata.h"
+#include "unk_020986CC.h"
+
+#define DEX_SIZE_U32           ((int)((NATIONAL_DEX_COUNT - 1) / 32) + 1) // default 16
+#define MAGIC_NUMBER           3203386110
```
cbt6, line 15:
```suggestion
#define MAGIC_NUMBER           0xbeefcafe
```

src/pokedex_data.c
```diff
@@ -0,0 +1,1240 @@
...
+
+    for (species = 1; species <= NATIONAL_DEX_COUNT; species++) {
+        if (PokedexData_HasSeenSpecies(pokedexData, species) == TRUE) {
+            u32 localDexNum = Pokemon_SinnohDexNumber(species);
+
+            if (localDexNum != 0) {
+                if (CountsForDexCompletion_Local(species) == TRUE) {
+                    numCaught++;
+                }
+            }
```
cbt6, line 869:
```suggestion
            if (Pokemon_SinnohDexNumber(species) != 0
                && CountsForDexCompletion_Local(species) == TRUE) {
                numCaught++;
            }
```

src/pokedex_data.c
h2o-DS: This is great, thank you.

src/pokedex_data.c
h2o-DS: I think it's forms seen

src/pokedex_data.c
```diff
@@ -0,0 +1,1235 @@
...
+        return TRUE;
+    } else {
+        return FALSE;
+    }
+}
+
+static inline BOOL ReadBit_2Forms(const u8 *array, u16 bitIndex)
+{
+    bitIndex--;
+    return 0 != (array[bitIndex >> 0x3] & (1 << (bitIndex & 0x7)));
```
lhearachel, line 83: **style**: Use decimal values for bitshift operands; applies throughout the file.

```suggestion
    return 0 != (array[bitIndex >> 3] & (1 << (bitIndex & 0x7)));
```

src/pokedex_data.c
```diff
@@ -0,0 +1,1235 @@
...
+    if (isSeen == FALSE) {
+        SetBit_2Forms((u8 *)pokedexData->recordedGenders[1], gender, bitIndex);
+    }
+
+    SetBit_2Forms((u8 *)pokedexData->recordedGenders[isSeen], gender, bitIndex);
+}
+
+static void UpdateGender(PokedexData *pokedexData, u8 gender, u8 isSeen, u16 bitIndex)
+{
+    GF_ASSERT(gender <= 2);
```
lhearachel, line 136: **suggestion**: `2` here should be `GENDER_NONE`.

src/pokedex_data.c
```diff
@@ -0,0 +1,1235 @@
...
+    case SPECIES_GIRATINA:
+        form = Pokemon_GetValue(pokemon, MON_DATA_FORM, NULL);
+        UpdateForms_TwoForms(pokedexData, species, form);
+        break;
+    case SPECIES_ROTOM:
+        form = Pokemon_GetValue(pokemon, MON_DATA_FORM, NULL);
+        UpdateForms_Rotom(pokedexData, species, form);
+        break;
+    default:
+        break;
```
lhearachel, line 591: **suggestion**: Remove this if it is unnecessary to match.

src/pokedex_data.c
```diff
@@ -0,0 +1,1235 @@
...
+    }
+}
+
+static void UpdateLanguage(PokedexData *pokedexData, u16 species, u32 language)
+{
+    int bitIndex = species;
+    int languageIndex = sub_020986CC(language);
+
+    if (languageIndex == 6) {
```
lhearachel, line 600: **question**: Should we mark this as an unused constant in `include/global/pm_version.h`?

h2o-DS: Strangely, it appears that language 2 is set to 6 in #323. I am not 100% sure, but based on the dex entries it looks like 2==diamond and 6==platinum, at least for the dex. Perhaps a holdover from Pearl?

lhearachel: Interesting... I would think that 2 would be Platinum, and 6 would be Pearl, given the context that most of the version-specific code checks for `DIAMOND || PLATINUM`, treating `PEARL` as a separate entity. Can you confirm?

h2o-DS: It would likely be more clear in this specific instance if `include/global/pm_version.h` had a value for this unused language. Would that be within the scope of this PR?

lhearachel: I think so, yes.

h2o-DS: Actually, upon further investigation I was correct the first time. This is languageIndex, so that language variable is not relevant. This is resolved in #323 where the `6` is changed to `NUM_LANGUAGES`

src/pokedex_data.c
```diff
@@ -11,17 +11,34 @@
...
+    SPECIES_DARKRAI,
+    SPECIES_SHAYMIN,
+    SPECIES_ARCEUS
+};
+static const u16 sExcludedMons_local[] = {};
+
+#define DEX_SIZE_U32          ((int)((NATIONAL_DEX_COUNT - 1) / 32) + 1) // default 16
+#define MAGIC_NUMBER          0xbeefcafe
+#define NUM_EXCLUDED_NATIONAL ((int)(sizeof(sExcludedMons_national) / sizeof(u16)))
+#define NUM_EXCLUDED_LOCAL    0 //((int)(sizeof(sExcludedMons_local) / sizeof(u16)))
```
lhearachel, line 32: **question**: Can this not be uncommented?

src/pokedex_data.c
```diff
@@ -11,17 +11,34 @@
...
+    SPECIES_CELEBI,
+    SPECIES_JIRACHI,
+    SPECIES_DEOXYS,
+    SPECIES_PHIONE,
+    SPECIES_MANAPHY,
+    SPECIES_DARKRAI,
+    SPECIES_SHAYMIN,
+    SPECIES_ARCEUS
+};
+static const u16 sExcludedMons_local[] = {};
```
lhearachel, line 27: **polish**: Format these names according to repo style:

```suggestion
static const u16 sExcludedMonsNational[] = {
    SPECIES_MEW,
    SPECIES_LUGIA,
    SPECIES_HO_OH,
    SPECIES_CELEBI,
    SPECIES_JIRACHI,
    SPECIES_DEOXYS,
    SPECIES_PHIONE,
    SPECIES_MANAPHY,
    SPECIES_DARKRAI,
    SPECIES_SHAYMIN,
    SPECIES_ARCEUS
};
static const u16 sExcludedMonsLocal[] = {};
```

src/pokedex_data.c
```diff
@@ -11,17 +11,34 @@
...
+#define NUM_EXCLUDED_LOCAL    0 //((int)(sizeof(sExcludedMons_local) / sizeof(u16)))
+#define NATIONAL_DEX_GOAL     (NATIONAL_DEX_COUNT - NUM_EXCLUDED_NATIONAL)
+#define LOCAL_DEX_GOAL        (LOCAL_DEX_COUNT - NUM_EXCLUDED_LOCAL)
+#define UNOWN_COUNT           28
+#define DEOXYS_COUNT          4
+#define ROTOM_COUNT           6
+#define TERMINAL_4BITS        0x7
+#define TERMINAL_BYTE         0xf
+#define TERMINAL_U8           0xff
+#define TERMINAL_U32          0xffffffff
```
lhearachel, line 41: **polish**: I do not think that these are necessary to keep as constants (and they are not kept as-such throughout this PR anyways). I would just remove them.

The names here are misleading, regardless; `0x7` is not the 4 least-significant bits, it is the 3 least-significant bits. `0xF` has the 4th bit, and is not a whole byte; it is a [nibble](https://en.wikipedia.org/wiki/Nibble).

In either case, throughout the PR:

1. Capitalize references to hex values, e.g. `0xFF` instead of `0xff` (note the lowercase `x`, which should be kept)
2. Left-pad hexadecimal values with `0` to the nearest byte, e.g.  `0x0F` instead of `0xF`, `0x0FFF` instead of `0xFFF`, etc.

lhearachel (commented):
Just a few more lingering comments. The rest looks great to me. :+1:

src/pokedex_data.c
h2o-DS: I was getting a `data type incomplete` error. I guess an empty array is just undefined instead of having a size of 0.

lhearachel: Understood. I think that this is a comment worth keeping, then.

lhearachel (approved):

PR #322, opened 2024-12-26
mods folder and fast_gauge.md

DylanDoesProgramming664:
Adds ``docs/mods`` folder, ``docs/mods/fast_gauge.md`` to explain a possible edit to ``src/battle/healthbar.c`` for faster exp and hp bars, and added a link to the file in ``docs/index.md``.


docs/index.md
```diff
@@ -11,3 +11,7 @@ For more detailed information about the project as a whole, please refer to its
 
 - [2D Graphics](2d_rendering.md)
 - [3D Graphics](3d_rendering.md)
+
+## Mods
+
+- [Fast Guage](mods/fast_guage.md)
```
lhearachel, line 17: **typo**: `Guage` -> `Gauge` (applies to the file-name as well)

docs/mods/fast_guage.md
```diff
@@ -0,0 +1,18 @@
+# Fast Guage
+
+This is a quick tutorial on how to modify the guage of exp and health bars of Plat so that the bar visually updates in subpixels per second instead of points per second.
+
+This mod is a backport of the healthbar fix made by the contributors of [pret/pokeheartgold](https://github.com/pret/pokeheartgold) for the heartgold/soulsilver decomp.
```
lhearachel, line 5: **style**: Provide a perma-link to the original fix in `pokeheartgold`.

DylanDoesProgramming664: I fixed the typos and provided a direct link to the line that's changed in pret/pokeheartgold.

docs/mods/fast_gauge.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Edit
+
+1. Go to [/src/battle/healthbar.c](/src/battle/healthbar.c)
+2. Go to the ``UpdateGauge`` function.
+3. Edit the function like so:
+```diff
+if (max < corrected) {
+-   ratio = max * 0x100 / corrected;
++   ratio = (max >> 8) / corrected;
```
lhearachel, line 17: Can you show a video of the difference here, by-chance?

docs/mods/fast_gauge.md
```diff
@@ -0,0 +1,18 @@
+# Fast Gauge
+
+This is a quick tutorial on how to modify the gauge of exp and health bars of Plat so that the bar visually updates in subpixels per second instead of points per second.
+
+This mod is a backport of the healthbar fix made by the contributors of [pret/pokeheartgold](https://github.com/pret/pokeheartgold/blob/src/battle/battle_hp_bar.c#L1497) for the heartgold/soulsilver decomp.
```
lhearachel, line 5: I believe that [this](https://github.com/pret/pokeheartgold/blob/077d9ee9e86da74953c319dcdbc43b23c40e5758/src/battle/battle_hp_bar.c#L1496) is the correct link.

DylanDoesProgramming664: The ``USE_SUBPIXEL_TEST`` definition is just an abstraction of a common comparison in the pokeheartgold code. The main difference in pokeplatinum is that we don't use the same constant names in all cases where that comparison is used. Regardless, that comparison is still in the pokeplatinum codebase when dealing with guages. However, the difference is is that pokeplat calculates adjusts the bar at a constant rate of 0x100, while the pokeheartgold uses bit shifting in order to standardize the rate in which health bars increase and decrease, regardless of size.

docs/mods/fast_gauge.md
DylanDoesProgramming664: Will do. Also noticed that I made a typo. Will fix that.

docs/mods/fast_gauge.md
lhearachel: The bitshift in `pokeheartgold` is in the opposite direction (leftward instead of the rightward shift listed in this document below), and is functionally equivalent to multiplying by `0x100`. To illustrate:

```
15 * 0x100 = 0xF * 0x100
           = 0xF00
           = 3840

15 << 8    = 3840
```

So, I guess that I am confused... how does this mod work? What change does it institute? It may be worthwhile to further document the tiling system used by the HP bar as part of this PR.

scbroede:
in my (admittedly not perfect) understanding of how the code works, healthbars deplete at a constant rate of 1 hit point per frame, and the `max < corrected` checks are in place to prevent scenarios where this results in two pixels of the healthbar disappearing at once (`corrected` is always 48 for healthbars, and the healthbar has 48 pixels in it). as far as i can tell, the change here would only kick in for pokemon with low max hp, a case where the healthbar is already decently fast.

in terms of speeding it up, some options i've tested are to pass in something higher than 1 for the `fillOffset` (such as a higher constant or copying what the exp gauge does), and/or setting `gSystem.frameCounter` to 1 somewhere in that flow to temporarily make the game run at 60fps instead of 30

lhearachel:
Closing, as this PR is stale and the author is inactive

PR #323, opened 2024-12-26
Document Pokedex Text and Languages

h2o-DS:
unk_020986CC -> pokedex_language
ov21_021D5600 -> pokedex_text

Files to access Pokedex text. Includes access to all six available languages.
To be merged after #321

PR #324, opened 2024-12-28
Generate Foreign dex entries

h2o-DS:
Added foreign pokedex text to pokemon data files
Added foreign pokedex message files to tools/scripts/make_pokedex_message_banks.py
Renamed message banks 700-723

To be merged after #323


res/pokemon/abomasnow/data.json
```diff
@@ -136,7 +136,34 @@
...
+        "spanish": {
+            "name": "Abomasnow",
+            "category": "Pokémon Árbol Nieve",
+            "entry_text": "Es el abominable hombre de las\nnieves. Produce ventiscas en las\nmontañas cubiertas de nieve."
+        },
+        "japanese": {
+            "name": "ユキノオ－",
+            "category": "じゅひょうポケモン",
+            "entry_text": "まんねんゆきの　つもる　やまで\nくらし　ブリザ－ドを　まきおこす。\nゆきおとこの　しょうたいだ。"
+        }
```
lhearachel, line 167: **suggestion**: I think that we should consider moving Pokedex entries to a data-markup format that permits multi-line strings. These raw `\n` characters are quite ugly, in my opinion.

h2o-DS: I will look into yaml. Should probably move all Pokedex data into this separate file.

h2o-DS: I have added a pokedex.yaml to each species folder.

zukan_data.narc needs to read from both data.json and pokedex.yaml due to the name and type sorts.

tools/scripts/make_pokedex_data.py
```diff
@@ -1,13 +1,16 @@
 #!/usr/bin/env python3
 import argparse
 import json
+import yaml
```
lhearachel, line 4: Is this `pyyaml`?

h2o-DS: Yes. How should I clarify that?

lhearachel: If it's a new dependency, then we need to both clarify it in the `INSTALL.md`, provide a way to standardize what version we are pulling (e.g., via a `requirements.txt` for `pip`), and set up the `Makefile` to enforce virtual environments so that users don't have to worry about cross-pollinating project dependencies with their system packages.

lhearachel: It looks like `apt` has packaging for `pyyaml` by way of `python3-yaml`:

- [Ubuntu](https://launchpad.net/ubuntu/+source/pyyaml)
- [Debian](https://packages.debian.org/search?searchon=sourcenames&keywords=pyyaml)

So we can add `python3-yaml` to our `apt install` command to ensure that a user has it on their system.

lhearachel: Scratch the above; Mac OS would be left in the dust with that approach. [Homebrew](https://formulae.brew.sh/formula/pyyaml) has disabled the formula, and [the official recommendation is to enforce virtual environments](https://github.com/orgs/Homebrew/discussions/5707). So, we do need to move towards enforcing virtual environments via the Makefile.

h2o-DS:
`pyyaml` integration into INSTALL.md is dependent on changes in #329

PR #329, opened 2025-01-02
Enforce Python venv usage, install meson and ninja via venv pip

lhearachel:
Ubuntu 24.04 ships with [Python 3.12](https://packages.ubuntu.com/noble/python3) as the default system installation, which enforces [PEP 668](https://peps.python.org/pep-0668/). The present installation instructions for the project do not wholly account for this nor make it easy for non-savvy users to decide on how to resolve the `externally-managed-environment` error. The easiest way to resolve it which many users will come across is to run the `pip install` command with the `--break-system-packages` switch, which is ill-advised for a whole host of reasons.

The correct path forward, instead, is to enforce usage of `meson` and `ninja` via a virtual environment, per recommendations in PEP 668. This change-set:

1. Adds a new target which will create the virtual environment, defaulting to `.venv`,
2. Creates dependencies for the `.venv` on the targets which build `mwrap` and `$(BUILD)/build.ninja`, and
3. Wraps all invocations of `meson` and `ninja` in an implicit `venv activate`.

As a secondary benefit to this procedure, we now have the full breadth of PyPI at our disposal, should we wish to make use of external dependencies from it in our Python-based tooling. However, this *must* be met with scrutiny, and we should establish clear guidelines on what dependencies will be permitted as extensions to Python's standard library.


requirements.txt
```diff
@@ -0,0 +1,2 @@
+meson >= 1.3.*
```
lhearachel, line 1: I would like for us to evaluate pinning this version requirement to a more recent minor-version, now that we are enforcing installation via a `venv` managed by the `Makefile`. I think it may also make sense to have a `venv-update` target which performs a `pip install --update` on the `requirements.txt` list.

It may even make sense for us to have a `make update` target which handles both `pip` updates and `subproject` updates.

lhearachel:
Incorporated @mid-kid's proposed changes to clone `meson` rather than setting it up as a subproject. There is one bug to fix where `make` believes that the `meson` target must be rebuilt on every invocation, which results in some ugly output:

```console
> make
MWCONFIG=/Users/rachel/code/github.com/lhearachel/pokeplatinum/build/.mwconfig subprojects/meson-1.7.0/meson.py setup \
                 --wrap-mode=nopromote \
                 --native-file=meson/native_macos.ini \
                 --native-file=build/root.ini \
                 --cross-file=meson/cross_unix.ini \
                 --cross-file=build/root.ini \
                 -- build
Directory already configured.

Just run your build command (e.g. ninja) and Meson will regenerate as necessary.
Run "meson setup --reconfigure" to force Meson to regenerate.

If build failures persist, run "meson setup --wipe" to rebuild from scratch
using the same options as passed when configuring the build.
subprojects/meson-1.7.0/meson.py configure build -Dgdb_debugging=false
ninja -C build data
ninja: Entering directory `build'
ninja: no work to do.
ninja -C build pokeplatinum.us.nds
ninja: Entering directory `build'
ninja: no work to do.
subprojects/meson-1.7.0/meson.py test -C build
ninja: Entering directory `/Users/rachel/code/github.com/lhearachel/pokeplatinum/build'
ninja: no work to do.
1/3 SBIN Checksums              OK              0.02s
2/3 ROM Checksum                OK              0.08s
3/3 Filesystem Checksums        OK              0.08s

Ok:                 3
Expected Fail:      0
Fail:               0
Unexpected Pass:    0
Skipped:            0
Timeout:            0

Full log written to /Users/rachel/code/github.com/lhearachel/pokeplatinum/build/meson-logs/testlog.txt
```

lhearachel:
> There is one bug to fix where make believes that the meson target must be rebuilt on every invocation, which results in some ugly output:

This bug has been fixed by changing the `meson` target to an `order-only` dependency on the targets which require it and using variable-substitutions for the rule instead of a wildcard-rule.

lhearachel:
8c0efd7 refactors the existing tooling a bit for `make update`; it now leverages the `update.py` wrapper that we have for manual git updates. While this is slower than using `meson subprojects --update`, it should be more reliable in the case of moving tooling around amongst various subprojects.

mid-kid:
Is meson not able to update the subprojects after changing the revision in the wrap files?
I'd rather offload the subproject management to meson instead of implementing that ourselves.

PR #333, opened 2025-01-07
Map object documentation

Viperio19:
Documentation centered around `map_object.c` and `map_object_save.c`


src/map_object.c
```diff
@@ -933,22 +896,22 @@ static void sub_02062604(MapObject *mapObj)
...
+    MapObject_SetStatusFlagOff(mapObj, MAP_OBJ_STATUS_SHOW_SHADOW | MAP_OBJ_STATUS_HIDE_SHADOW | MAP_OBJ_STATUS_26 | MAP_OBJ_STATUS_24);
 }
 
 static void sub_02062628(MapObject *mapObj)
 {
     (void)0;
 }
 
-static int sub_0206262C(FieldSystem *fieldSystem, int param1)
+static int FieldSystem_GetGraphicsID(FieldSystem *fieldSystem, int graphicsID)
```
cbt6, line 907: Revert renaming this function for now. There's more that meets the eye in `sub_0203F164`

src/map_object.c
```diff
@@ -1176,15 +1135,15 @@ FieldSystem *MapObjectMan_FieldSystem(const MapObjectManager *mapObjMan)
     return mapObjMan->fieldSystem;
 }
 
-void sub_0206289C(MapObjectManager *mapObjMan, NARC *param1)
+void MapObjectMan_GetNARC(MapObjectManager *mapObjMan, NARC *narc)
```
cbt6, line 1138:
```suggestion
void MapObjectMan_SetNARC(MapObjectManager *mapObjMan, NARC *narc)
```

src/map_object.c
```diff
@@ -1176,15 +1135,15 @@ FieldSystem *MapObjectMan_FieldSystem(const MapObjectManager *mapObjMan)
...
 {
-    mapObjMan->unk_14 = param1;
+    mapObjMan->narc = narc;
 }
 
 NARC *sub_020628A0(const MapObjectManager *mapObjMan)
```
cbt6, line 1143:
```suggestion
NARC *MapObjectMan_GetNARC(const MapObjectManager *mapObjMan)
```

src/map_object_move.c
```diff
@@ -970,22 +966,22 @@ void VecFx32_StepDirection(int dir, VecFx32 *vec, fx32 val)
...
+    int movementType = MapObject_GetMovementType(mapObj);
 
-    if ((v0 == 0x33) || (v0 == 0x34) || (v0 == 0x35) || (v0 == 0x36)) {
+    if (movementType == 0x33 || movementType == 0x34 || movementType == 0x35 || movementType == 0x36) {
         sub_02062B14(mapObj);
     }
 }
 
-static const int Unk_020EE79C[] = {
+static const int sOppositeDirections[] = {
```
cbt6, line 984:
```c
static const int sOppositeDirections[] = {
    [DIR_NORTH] = DIR_SOUTH,
    [DIR_SOUTH] = DIR_NORTH,
    [DIR_WEST] = DIR_EAST,
    [DIR_EAST] = DIR_WEST,
};
```

src/map_object_move.c
```diff
@@ -994,56 +990,56 @@ static const int Unk_020EE79C[] = {
 
 int Direction_GetOpposite(int dir)
 {
-    return Unk_020EE79C[dir];
+    return sOppositeDirections[dir];
 }
 
-int sub_02064488(int param0, int param1, int param2, int param3)
+int sub_02064488(int x, int z, int xPrev, int zPrev)
```
cbt6, line 996: Are you sure about these parameter names? They don't look quite right.

src/map_object.c
```diff
@@ -1141,24 +1100,24 @@ UnkStruct_ov5_021ED0A4 *sub_0206285C(const MapObjectManager *mapObjMan)
...
-    return mapObjMan->unk_124;
+    return mapObjMan->mapObj;
 }
 
-MapObject *sub_02062878(const MapObjectManager *mapObjMan)
+MapObject *MapObjectMan_GetMapObject(const MapObjectManager *mapObjMan)
 {
-    return mapObjMan->unk_124;
+    return mapObjMan->mapObj;
 }
```
cbt6, line 1121: GF moment

src/map_object_move.c
Viperio19: It seems like I focussed a bit too much on one specific instance of it being used:
```c
static int sub_0206A034(MapObject *mapObj, UnkStruct_02069F48 *param1)
{
    int v0 = MapObject_GetX(mapObj);
    int v1 = MapObject_GetZ(mapObj);
    int v2 = MapObject_GetX(param1->unk_08);
    int v3 = MapObject_GetZ(param1->unk_08);
    int v4 = MapObject_GetXPrev(param1->unk_08);
    int v5 = MapObject_GetZPrev(param1->unk_08);
    int v6;

    if ((v0 == v2) && (v1 == v3)) {
        return 0;
    }

    v6 = sub_02064488(v0, v1, v4, v5);
```
It varies quite a bit tough, so you're right in that it doesn't really make sense. I don't think there's really descriptive names I could give them that covers all usages of the function besides just `x1`, `z1`, `x2` and `z2`, do you think I should do that, or revert them back to `paramX` names?

cbt6: Keep as `paramX` if unsure.

Viperio19: I am sure that the params are two sets of coordinates, so `x1`, `z1`, `x2` and `z2` would be true, but they're not very descriptive besides that it's two sets of coordinates. The issue is that I don't think there's a more descriptive option because of this function being used in different ways, so the param names can't be more specific I think. Sorry I should've made that a bit more clear, I'm just not sure if it's then still worth it to give the names `x1`, `z1`, `x2` and `z2`.

cbt6: If you're unsure then it's okay to leave this function as it is for now and not rename anything. We can always come back again when the other functions (that invoke this function) themselves get documented.

Viperio19: Okay, I have changed it back and committed that together with fixes for your other comments, thanks for all the feedback!

src/map_object.c
```diff
@@ -2357,37 +2312,35 @@ int ObjectEvent_GetZ(const ObjectEvent *objectEvent)
...
 
-static int sub_020631D8(const ObjectEvent *objectEvent)
+static int ObjectEvent_HasNoScript(const ObjectEvent *objectEvent)
 {
-    u16 v0 = (u16)ObjectEvent_GetScript(objectEvent);
+    u16 script = (u16)ObjectEvent_GetScript(objectEvent);
 
-    if (v0 == 0xffff) {
-        return 1;
+    if (script == 0xffff) {
```
lhearachel, line 2334: **non-blocking**: Making a note that we should probably establish a constant for this and start using it, since it seems to be the marker for "no such script".

src/map_object.c
```diff
@@ -2453,48 +2406,43 @@ static const UnkStruct_ov5_021FB0F0 *sub_02063244(u32 param0)
...
-    } while (v0);
+        mapObj++;
+        maxObjects--;
+    } while (maxObjects);
 
     return NULL;
 }
 
-void sub_020632D4(MapObject *mapObj, const VecFx32 *param1, int param2)
+void sub_020632D4(MapObject *mapObj, const VecFx32 *pos, int param2)
```
lhearachel, line 2432: **question**: Is this the inverse of `SetPosDir`...?

src/map_object.c
```diff
@@ -2504,23 +2452,23 @@ void sub_020632D4(MapObject *mapObj, const VecFx32 *param1, int param2)
...
 
-void MapObject_SetPosDir(MapObject *mapObj, int param1, int param2, int param3, int param4)
+void MapObject_SetPosDir(MapObject *mapObj, int x, int y, int z, int dir)
 {
-    VecFx32 v0;
+    VecFx32 pos;
 
-    v0.x = (((param1) << 4) * FX32_ONE) + ((16 * FX32_ONE) >> 1);
-    MapObject_SetX(mapObj, param1);
+    pos.x = ((x << 4) * FX32_ONE) + ((16 * FX32_ONE) >> 1);
```
lhearachel, line 2459: **suggestion**: I would think about formalizing these into an inline function or a basic macro, if possible.

src/map_object_move.c
```diff
@@ -363,149 +363,159 @@ static void sub_02063A30(MapObject *mapObj, u8 param1, u8 param2, const UnkStruc
...
     if (param3->unk_04_4 == 0) {
         return;
     }
 
-    if ((TileBehavior_IsTallGrass(param1) == 1) || (TileBehavior_IsVeryTallGrass(param1) == 1) || (sub_0206406C(mapObj, param1) == 1) || (TileBehavior_IsPuddle(param1) == 1) || (TileBehavior_IsShallowWater(param1) == 1) || (MapObject_IsOnSnow(mapObj, param1) == 1) || (TileBehavior_IsMud(param1) == 1) || (TileBehavior_IsMudWithGrass(param1) == 1) || TileBehavior_IsReflective(param1)) {
+    if (TileBehavior_IsTallGrass(currTileBehavior) == TRUE || TileBehavior_IsVeryTallGrass(currTileBehavior) == TRUE
+        || MapObject_IsOnWater(mapObj, currTileBehavior) == TRUE || TileBehavior_IsPuddle(currTileBehavior) == TRUE
+        || TileBehavior_IsShallowWater(currTileBehavior) == TRUE || MapObject_IsOnSnow(mapObj, currTileBehavior) == TRUE
+        || TileBehavior_IsMud(currTileBehavior) == TRUE || TileBehavior_IsMudWithGrass(currTileBehavior) == TRUE
+        || TileBehavior_IsReflective(currTileBehavior)) {
```
lhearachel, line 392: **suggestion**: I would just make these one-per-line, though this list is quite long. So maybe this arrangement is okay.

src/map_object_move.c
```diff
@@ -813,21 +809,21 @@ static const int DATA_GPosY_Dir4AddTbl[] = {
     0
 };
 
-static const int sMapObjectDyDir[] = {
+static const int sMapObjectDzDir[] = {
     -1,
     0x1,
     0x0,
     0x0
 };
```
lhearachel, line 817: Similar suggestion applies to `sMapObjectDxDir` above.

```suggestion
static const int sMapObjectDzDir[] = {
    [DIR_NORTH] = -1,
    [DIR_SOUTH] = 1,
    [DIR_WEST] = 0,
    [DIR_EAST] = 0,
};
```

src/map_object.c
Viperio19: It doesn't seem to be the inverse, it's just this one uses a `VecFx32` as param, and `SetPosDir` uses coordinates (`int x, int y, int z,`), but they seem to do the exact same thing.

Viperio19: I could maybe rename this one to `MapObject_SetPosDirFromVec` and the other one to `MapObject_SetPosDirFromCoords`.

lhearachel: I think that would be fine, yeah.

src/map_object_move.c
Viperio19:
```static const int DATA_GPosY_Dir4AddTbl[] = {
    0,
    0,
    0,
    0
};```
This is in the middle of those two, and it is unused. What's the best way to go about that? Remove it, rename it to something like `UNUSED_GPosY_Dir4AddTbl`, or add a comment like `// unused`?

src/map_object.c
Viperio19: This is something that appears in a lot of files and isn't really specific to or centered around map objects, it's just a general conversion from regular coords to however they are stored in `VecFx32`, so I think it may be better for a different PR, but it is a good point. Something like `#define X_OR_Z_TO_VECFX32(coord) ((((coord) << 4) * FX32_ONE) + ((16 * FX32_ONE) >> 1))` maybe.

src/map_object_move.c
lhearachel: Something like that would be fine, yeah!

Viperio19: Done!

lhearachel (approved):

PR #334, opened 2025-01-09
Update required Meson version to 1.5.0

lhearachel:
Changes to the field scripting compiler to depend on `trdata.naix` use a `CustomTargetIndex` in the `generator`'s `depends` kwarg. This support was added in 1.5.0, as per tags on mesonbuild/meson PR #12956.


lhearachel:
Source PR: https://github.com/pret/pokeplatinum/pull/332

PR #337, opened 2025-01-15
Add FreeBSD compatibility

SetiQyu:
Changed 3 files to make the project build on FreeBSD (and possibly the other BSDs, requires gmake).

The meson.build in the project root has the most questionable change that should be improved, since the new second custom_target call outputs the file "junk.txt" that serves no purpose other than the fact that meson does not allow an empty output for the custom_target method. For whatever reason FreeBSD does not like some of the shell commands that are in the meson.build files custom_target methods, requiring some of them to be tweaked a little to build successfully.

The issue is that the wine call with makerom_exe did not work when sandwiched in there so there had to be some separate execution step with the added custom_target running after pokeplatinum_nds for it work. 

The repo with this patch builds fine on my Debian and FreeBSD machines.

I would love any suggestions as to how to improve this very bad solution, I'll test them!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,13 @@ ifneq (,$(findstring Darwin,$(UNAME_S)))
 NATIVE := native_macos.ini
 CROSS := cross_unix.ini
 else
+ifneq (,$(findstring BSD, $(UNAME_S)))
+NATIVE := native_unix.ini 
+CROSS := cross_unix.ini
+else
 NATIVE := native.ini
 CROSS := cross.ini
+endif
 endif
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -224,9 +224,6 @@ pokeplatinum_nds = custom_target('pokeplatinum.us.nds',
         nitrofs_files # Make sure this is always listed last
     ],
     command : [
-        'sh', '-c',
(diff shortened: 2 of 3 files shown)


SetiQyu:
To clarify the issue, when the command to run wine (makerom_exe) was part of pokeplatinum_nds command array, it would fail. But the wine command and the second command after && in the original meson.build had no trouble running individually, so the "solution" was to only have makerom_exe run in pokeplatinum_nds, then create a second custom_target that depended on pokeplatinum_nds so that makerom_exe and the second command would still be ran sequentially but as part of separate shell commands.

SetiQyu:
I fixed my hacky solution now, now there is no junk file being created! Tested on my FreeBSD and Debian machine.

lhearachel:
Changes build on my Arch Linux machine. I will pull onto my Macbook later today and verify there as well.

lhearachel:
Build succeeds on my M3 Macbook, as well, so I'll call this one good.

lhearachel (approved):

mid-kid:
Are you sure `sh -c 'printf %s\\n' -- a b c` doesn't work on FreeBSD?
The FreeBSD man page implies it does: https://man.freebsd.org/cgi/man.cgi?query=sh

> A "--" or plain `-' will stop  option  processing	and  will  force the remaining words on	the command line to be treated as arguments.

PR #339, opened 2025-01-17
document poffin.c and poffin case struct

fakuzatsu:
As per description.

Only undocumented function in `poffin.c` now is `sub_0202A9E4` which seems to be related in some way to the cooking minigame in which you make poffins.

Renames some headers appropriately and notably includes the `poffin.h` header directly into `struct_defs/poffin_case.h` to use the `MAX_POFFINS` macro as the array size.


src/poffin.c
```diff
@@ -3,21 +3,23 @@
...
-#include "struct_defs/struct_0202A93C.h"
-#include "struct_defs/struct_0202AB28.h"
+#include "struct_defs/poffin.h"
+#include "struct_defs/poffin_case.h"
 
 #include "heap.h"
 #include "math.h"
 #include "savedata.h"
 
+#define FLAVOUR_NONE    30
```
Viperio19, line 13: Very small nitpick, "flavour" is spelled without the "u" everywhere else, so I'd change this to `FLAVOR_NONE` for consistency.

fakuzatsu: true. UK brain strikes again. changed.

include/struct_defs/poffin_case.h
```diff
@@ -0,0 +1,10 @@
+#ifndef POKEPLATINUM_STRUCT_POFFIN_CASE_H
+#define POKEPLATINUM_STRUCT_POFFIN_CASE_H
+
+#include "poffin.h"
+
+typedef struct PoffinCase {
```
lhearachel, line 6: **polish**: In general, we try to zip these structures up into the owning module's header file. `struct`-only headers are reserved for special cases when needed for untangling circular dependencies or sharing definitions with build-tooling.

include/struct_defs/poffin.h
```diff
@@ -1,5 +1,5 @@
-#ifndef POKEPLATINUM_STRUCT_0202A93C_H
-#define POKEPLATINUM_STRUCT_0202A93C_H
+#ifndef POKEPLATINUM_STRUCT_POFFIN_H
+#define POKEPLATINUM_STRUCT_POFFIN_H
 
 typedef struct {
```
lhearachel, line 4: **polish**: As with the above comment (and also for the `struct_decls/poffin_case_decl.h` file, as well):

1. Zip this definition into `include/poffin.h`
2. Replace all the inclusions of `struct_defs/poffin.h` with `poffin.h`
3. Delete this file (and `struct_decls/poffin_case_decl.h`).

src/poffin.c
```diff
@@ -26,7 +28,7 @@ BOOL sub_0202A914(Poffin *poffin)
 
 void Poffin_clear(Poffin *poffin)
```
lhearachel, line 29:
```suggestion
void Poffin_Clear(Poffin *poffin)
```

Similarly (github will not let me do this on the associated lines):

```diff
-Poffin *Poffin_malloc(int heapID)
+Poffin *Poffin_New(int heapID)
```

```diff
-void Poffin_copy(Poffin *src, Poffin *dest)
+void Poffin_Copy(Poffin *src, Poffin *dest)
```

src/poffin.c
```diff
@@ -234,144 +236,149 @@ int Poffin_SaveSize(void)
...
-        v2 = sub_0202ABB0(param0, v3);
+        // found an empty slot, proceed
+        emptySlotNum = i;
+        nextValidSlotNum = Poffin_GetFirstValidPoffin(poffinCase, emptySlotNum); // find the next valid poffin
 
-        if (v2 == 0xFFFF) {
+        if (nextValidSlotNum == POFFIN_NONE) {
             break;
-        }
+        } // skips rest of execution, reached end of poffin case
```
lhearachel, line 316: **polish**: Remove this comment (unnecessary).

lhearachel (changes requested):

src/poffin.c
fakuzatsu: for this one I felt it important to document that this can only occur when reaching the end of the poffin case, although this may be more clear following some other changes I made subsiquent.

include/poffin.h
```diff
@@ -1,11 +1,31 @@
...
+            u8 dryness;
+            u8 sweetness;
+            u8 bitterness;
+            u8 sourness;
+            u8 val1_06;
+            u8 dummy;
+        };
+        u8 attributes[8];
+    };
+} Poffin;
```
lhearachel, line 23: Would something like this work?

```suggestion
typedef struct {
    union {
        struct PoffinInner {
            u8 flavor;
            u8 spiciness;
            u8 dryness;
            u8 sweetness;
            u8 bitterness;
            u8 sourness;
            u8 val1_06;
            u8 dummy;
        };
        u8 attributes[sizeof(struct PoffinInner)];
    };
} Poffin;
```

This way, the size of `attributes` is always tied to the size of the inner struct used in the union.

fakuzatsu: We're about to find out

fakuzatsu:
update, it did in fact work

lhearachel (approved):

PR #341, opened 2025-01-18
TRAINER_NAME_LEN location. Naming conventions for ov6_02240C9C.

DylanDoesProgramming664:
Gave names to the vars and functions involved in encounter slot chosen for wild encounters.


src/overlay006/ov6_02240C9C.c
```diff
@@ -810,131 +810,131 @@ static BOOL ov6_022418DC(FieldSystem *fieldSystem, u32 encounterRate)
     }
 }
 
-static u8 ov6_02241904(void)
+static u8 GetGrassEncounterSlot(void)
```
lhearachel, line 813: **quibble**: `GetGroundEncounterSlot` (this also applies to cave floors)

src/overlay006/ov6_02240C9C.c
```diff
@@ -810,131 +810,131 @@ static BOOL ov6_022418DC(FieldSystem *fieldSystem, u32 encounterRate)
...
-    u8 v0 = inline_020564D0(100);
+    u8 roll = inline_020564D0(100); // a common rng function
 
-    if (v0 < 20) {
+    if (roll < 20) {
         return 0;
     }
 
-    if (v0 >= 20 && v0 < 40) {
+    if (roll >= 20 && roll < 40) {
```
lhearachel, line 821: **question**: Do these branches match if you condense them into an `if-else` tree and remove the `>=` operations? e.g.

```c
if (roll < 20) {
    return 0;
} else if (roll < 40) {
    return 1;
} else if (roll < 50) {
    return 2;
} // etc.
```

DylanDoesProgramming664: yes. though, using a const rate table and for loop works too.

lhearachel: It does?! What does the table look like? If it's a 0-99 table, I wouldn't want that, but if it's more compact, then I would much rather represent it as a table.

DylanDoesProgramming664:
```c
static const u8 GroundEncounterSlotRates[MAX_GRASS_ENCOUNTERS] = {
    20,
    20,
    10,
    10,
    10,
    10,
    10,
    10,
    5,
    5,
    4,
    4,
    1,
    1
};

// code

static u8 GetGroundEncounterSlot(void)
{
    u8 roll = inline_020564D0(100);
    
    u8 encSlot = 0;
    u8 hit = 0;
    for (encSlot = 0; encSlot < MAX_GRASS_ENCOUNTERS; encSlot++) {
        hit += GroundEncounterSlotRates[encSlot];
        if (roll < hit) {
            return encSlot;
        }
    }

    return encSlot;
}
```

lhearachel: If that really matches, just prefix the table name with `s` and ship it.

DylanDoesProgramming664: It doesn't work. I reverted the code and am trimming it to compact if-else statements.

DylanDoesProgramming664:
Fixed. Linter complained about whitespace earlier.

src/overlay006/ov6_02240C9C.c
```diff
@@ -810,131 +810,105 @@ static BOOL ov6_022418DC(FieldSystem *fieldSystem, u32 encounterRate)
...
-    u8 v0 = inline_020564D0(100);
+    u8 roll = inline_020564D0(100); // a common rng function
 
-    if (v0 < 20) {
+    if (roll < 20) {
         return 0;
-    }
-
-    if (v0 >= 20 && v0 < 40) {
+    } else if (roll >= 20 && roll < 40) {
```
lhearachel, line 819: Are the `>=` checks on these necessary?

DylanDoesProgramming664: Unfortunately yes. Failing to do so results in the checksum being incorrect. Don't know why.

DylanDoesProgramming664: Actually, I'm wrong. It was the linter that complained that time. I'll do it again, and use the formatter to make it tidy.

DylanDoesProgramming664: And build failed now.

lhearachel (approved):

PR #342, opened 2025-01-18
Documenting `main.c`, `map_header_data.c` and `map_header.c`

Viperio19:
Small PR with some documenting for `main.c`, `map_header_data.c` and `map_header.c` (naming variables and using constants)


src/map_header.c
```diff
@@ -219,88 +221,88 @@ BOOL MapHeader_IsAzureFluteAllowed(const u32 headerID)
...
+        if (headerID == pokecenters2F[i]) {
+            return TRUE;
         }
     }
 
-    return 0;
+    return FALSE;
 }
 
 u32 MapHeader_GetMapEvolutionMethod(u32 headerID)
```
lhearachel, line 255: **question**: Does this match if you return the enum directly instead of a `u32`?

src/map_header_data.c
```diff
@@ -196,52 +196,44 @@ static void MapHeaderData_ParseEvents(MapHeaderData *data)
...
-        }
-        events += sizeof(BgEvent) * data->numBgEvents;
+    data->numBgEvents = *(u32 *)events;
+    events += sizeof(u32);
+    if (data->numBgEvents != 0) {
+        data->bgEvents = (const BgEvent *)events;
+    } else {
+        data->bgEvents = NULL;
     }
+    events += sizeof(BgEvent) * data->numBgEvents;
```
lhearachel, line 206: With this macro defined somewhere (probably just above this function):

```c
#define CONSUME_EVENTS(T, dataTEvents, dataNumTEvents) \
    do {                                               \
        (dataNumTEvents) = *(u32 *)events;             \
        events += sizeof(u32);                         \
        if ((numEvents) != 0) {                        \
            (dataTEvents) = (const T *)events;         \
        } else {                                       \
            (dataTEvents) = NULL;                      \
        }                                              \
        events += sizeof(T) * (dataNumTEvents);        \
    } while (0)
```

You can replace this whole block:

```suggestion
    CONSUME_EVENTS(BgEvent, data->bgEvents, data->numBgEvents);
```

And also do similarly for the other sections:

```c
    CONSUME_EVENTS(ObjectEvent, data->objectEvents, data->numObjectEvents);
    CONSUME_EVENTS(WarpEvent, data->warpEvents, data->numWarpEvents);
    CONSUME_EVENTS(CoordEvent, data->coordEvents, data->numCoordEvents);
```

Viperio19: Great suggestion! Does do the do-while part matter? It seems to at least match without it as well, but I might be missing some reason for it.

lhearachel: It enforces the existence of a semicolon to make it semantically resemble a function in the code which uses it. If you get rid of the semicolon with the `do while` loop still in place, then it should fail to compile the file at all, much less match.

src/map_header.c
Viperio19: Took a bit to realize you were talking about `MapHeader_GetMapEvolutionMethod`... Anyways, it still matches with the enum!

lhearachel (approved):

PR #349, opened 2025-01-24
Switch metang to python implementation as temp-fix for native Windows builds

lhearachel:
Some native Windows users have reported seg-faults from `metang` when generating headers; this is a temporary hot-switch to a Python implementation until I have enough the time+means to investigate these errors.


lhearachel:
Build from this branch has been confirmed by some Windows users, so marking this one as good.

Issue #325, opened 2024-12-29
Good first documentation efforts for new contributors


ravepossum:
Interested in contributing to pokeplatinum? Documenting the following smaller-scope files may be a good starting point:

- `unk_0205D8CC` - Interaction with field elements that spawns text boxes (signboards, NPCs, etc.)
- `unk_020562F8` - Pal Park library
- `unk_02071CFC` and `ov5_021DD6FC` - Map name pop-ups
- `unk_020986CC` and  `ov21_021D5600` - Language-related Pokedex functionality

This list is a work-in-progress. If you tackle one of these items, please mention this issue in your PR so we know to update the list.

lhearachel:
`ov6_02246184` is responsible for loading the data structure used for in-game trade Pokemon (refer to https://bulbapedia.bulbagarden.net/wiki/In-game_trade#Generation_IV).

`ScrCmd_226` is the script command responsible for loading this data into memory during an in-game trade event.

lhearachel:
`ov5_021DC018.c` is used to display menu boxes in the field for player interaction, e.g., for multi-choice menus (where all options are displayed at once) and list-menus (where the number of options is a scrollable list). These functions are used in various script-commands, as well.

lhearachel:
Documentation for #374 uncovered the following:

## Distribution Events

`unk_0204B830.c` contains code for handling the distribution-events (Darkrai, Shaymin, Arceus, and the Rotom forms). Specifically, the table `Unk_020EBE94` contains structs of subroutines used to handle these Event Pokemon.

## Jubilife Lottery

`unk_020480A8.c` contains scripting commands for handling the Jubilife Lottery. Only `ScrCmd_24F` lacks documentation; it checks the trainer ID of all owned Pokemon in the player's party and PC boxes for a matching trainer ID input.

PR #352, opened 2025-01-28
MapNamePopUp Documentation

estellarc:
- Renamed unk_02071CFC -> map_header_util
- Renamed ov5_021DD6FC -> map_name_popup

I'm unsure how I should name `sub_02071D10` and about the most variables in `MapNamePopUp_DrawWindowFrame`

#325


src/map_header_util.c
```diff
@@ -0,0 +1,34 @@
...
+{
+    u32 mapLabelTextID;
+    MessageLoader *msgLoader;
+
+    msgLoader = MessageLoader_Init(MESSAGE_LOADER_NARC_HANDLE, NARC_INDEX_MSGDATA__PL_MSG, message_bank_location_names, heapID);
+    mapLabelTextID = MapHeader_GetMapLabelTextID(headerID);
+
+    MapHeader_GetStringWidth(msgLoader, mapLabelTextID, strbuf); // It should be MessageLoader_GetStrbuf()
+    MessageLoader_Free(msgLoader);
+}
```
lhearachel, line 34: The first of these two functions seems to be more generic than just `MapHeader`; rather, it seems like it's used to pull any given string from any given text bank, then return its width. I would suggest naming these functions as:

- `MapHeader_GetStringWidth` -> `MapHeader_LoadString`
- `sub_02071D10` -> `MapHeader_LoadName`

src/map_header_util.c
```diff
@@ -0,0 +1,34 @@
...
+{
+    u32 width;
+
+    MessageLoader_GetStrbuf(msgLoader, entryID, strbuf);
+    width = Font_CalcStrbufWidth(FONT_SYSTEM, strbuf, 0);
+
+    return width;
+}
+
+void sub_02071D10(int headerID, u32 heapID, Strbuf *strbuf)
```
lhearachel, line 24: **polish**: Try replacing `headerID` here with `enum MapHeader` from `generated/map_headers.h` after merging from `main`.

src/map_header_util.c
```diff
@@ -0,0 +1,34 @@
+#include "map_header_util.h"
+
+#include <nitro.h>
+#include <string.h>
+
+#include "field/field_system_sub2_t.h"
+#include "text/pl_msg.naix"
```
lhearachel, line 7: **chore**: Replace this with `generated/text_banks.h` after merging from `main`.

src/map_header_util.c
```diff
@@ -0,0 +1,34 @@
...
+
+#include "font.h"
+#include "map_header.h"
+#include "message.h"
+#include "strbuf.h"
+
+u32 MapHeader_GetStringWidth(MessageLoader *msgLoader, u32 entryID, Strbuf *strbuf)
+{
+    u32 width;
```
lhearachel, line 16: **polish**: Try to declare + define variables inline where possible, e.g., prefer

```c
u32 width = Font_CalcStrbufWidth(FONT_SYSTEM, strbuf, 0);
```

vs.

```c
u32 width;

// other code

width = Font_CalcStrbufWidth(FONT_SYSTEM, strbuf, 0);
```

src/overlay005/map_name_popup.c
```diff
@@ -0,0 +1,327 @@
...
+static void MapNamePopUp_Reset(MapNamePopUp *mapPopUp);
+static void MapNamePopUp_SetBgConfig(MapNamePopUp *mapPopUp, BgConfig *bgConfig);
+static void SysTask_MapNamePopUpWindow(SysTask *task, void *param);
+static void MapNamePopUp_DrawWindowFrame(MapNamePopUp *mapPopUp, const s32 strWidth);
+static void MapNamePopUp_StartSlideOut(MapNamePopUp *mapPopUp);
+static void MapNamePopUp_PrintMapName(MapNamePopUp *mapPopUp, const Strbuf *strbuf);
+
+static void MapNamePopUp_LoadPalette(void *src, u16 size, u16 offset)
+{
+    DC_FlushRange((void *)src, size * 32);
```
lhearachel, line 57: **question**: Is this `void *` cast necessary?

src/overlay005/map_name_popup.c
```diff
@@ -0,0 +1,327 @@
...
+    GX_LoadBGPltt((const void *)src, offset * 32, size * 32);
```
lhearachel, line 58: **polish**: For these multiplications by 32, `palette.h` includes some functional-macros that you can use instead. e.g., you can replace `offset * 32` with `PLTT_OFFSET(offset)` and have some extra semantic meaning.

src/overlay005/map_name_popup.c
```diff
@@ -0,0 +1,327 @@
...
+}
+
+static void MapNamePopUp_CreateWindow(MapNamePopUp *mapPopUp)
+{
+    Window_Add(mapPopUp->bgConfig, &mapPopUp->window, BG_LAYER_MAIN_3, 0, 0, 32, 5, 7, ((((1024 - (18 + 12) - 9 - (32 * 8)) - (18 + 12 + 24)) - (27 * 4)) - (32 * 5)));
```
lhearachel, line 63: **suggestion**: It might be worth playing around with the values used in this long math expression to get a sense for what each of them means.

src/overlay005/map_name_popup.c
```diff
@@ -0,0 +1,327 @@
...
+
+static void MapNamePopUp_LoadAreaGfx(MapNamePopUp *mapPopUp, u8 bgLayer, u16 tileStart, u8 offset, u8 unused)
+{
+    void *ptr;
+
+    NNSG2dPaletteData *paletteData;
+    u8 narcMemberIdx;
+
+    // I'm confused by this
+    narcMemberIdx = mapPopUp->windowID * 2;
```
lhearachel, line 74: Looking at the structure of the NARC itself (`arc/area_win_gra.narc`), it looks like the window ID is just which window type is being used for this pop-up. There are 2 files for each pop-up type: a set of tiles, and a palette. So, multiplying the type ID by 2 yields the member index of the tile-data.

src/overlay005/map_name_popup.c
```diff
@@ -0,0 +1,327 @@
...
+    mapPopUp->state = MAP_NAME_POPUP_STATE_SLIDE_OUT;
+    mapPopUp->timer = 0;
+}
+
+MapNamePopUp *MapNamePopUp_Create(BgConfig *bgConfig)
+{
+    MapNamePopUp *mapPopUp;
+
+    mapPopUp = Heap_AllocFromHeap(4, sizeof(MapNamePopUp));
+    mapPopUp->strbuf = Strbuf_Init(22, 4);
```
lhearachel, line 236: **polish**: Replace these 4s with `HEAP_ID_FIELD` from `constants/heap.h`

src/overlay005/map_name_popup.c
```diff
@@ -0,0 +1,327 @@
...
+{
+    MessageLoader_Free(mapPopUp->msgLoader);
+    Window_Remove(&mapPopUp->window);
+    Strbuf_Free(mapPopUp->strbuf);
+    Heap_FreeToHeap(mapPopUp);
+
+    mapPopUp = NULL;
+}
+
+void MapNamePop_Show(MapNamePopUp *mapPopUp, s32 mapLabelTextID, s32 mapLabelWindowID)
```
lhearachel, line 255: **polish**:
```suggestion
void MapNamePopUp_Show(MapNamePopUp *mapPopUp, s32 mapLabelTextID, s32 mapLabelWindowID)
```

src/overlay005/map_name_popup.c
```diff
@@ -0,0 +1,327 @@
...
+    int width;
+    int v1;
+    int xOffset;
+    int v3;
+    int v4;
+    int v5;
+    u8 v6;
+    u8 v7;
+
+    v3 = (strWidth + 8) / 8 * 8;
```
lhearachel, line 96: This should be the total number of pixels used by the string. `8` here is the width of a tile in pixels.

src/map_header_util.c
```diff
@@ -0,0 +1,34 @@
...
+
+    return width;
+}
+
+void sub_02071D10(int headerID, u32 heapID, Strbuf *strbuf)
+{
+    u32 mapLabelTextID;
+    MessageLoader *msgLoader;
+
+    msgLoader = MessageLoader_Init(MESSAGE_LOADER_NARC_HANDLE, NARC_INDEX_MSGDATA__PL_MSG, message_bank_location_names, heapID);
```
lhearachel, line 29: After you merge from `main`, `message_bank_location_names` will become `TEXT_BANK_LOCATION_NAMES`.

lhearachel (commented):
This looks great overall! Thanks for your contribution. I have some feedback and collaborative thoughts scattered throughout.

You'll also want to do a pull from `main` before continuing, as some generated headers are being shifted around. I tried to call those out where I saw them.

src/overlay005/map_name_popup.c
estellarc: Apparently not, it gives a match, the same with the `const void *`on `GX_LoadBGPltt`

lhearachel: That's fairly typical. 😁 An explicit `void *` cast is almost never necessary, and the compiler will often do it for you.

src/overlay005/map_name_popup.c
estellarc: Silly me
35429ff19

src/overlay005/map_name_popup.c
estellarc: 👍
8ad93df3d

src/overlay005/map_name_popup.c
estellarc: At first glance, it doesn't seem to do anything. In any case most calls to `Window_Add` are left with similar expressions. To be honest, I thought it was left that way on purpose, so that a macro could be used later

lhearachel: You're right, it was left this way for that reason. I recall that this is the base tile parameter for the window, which should be what tile number the window treats as "tile 0" in VRAM.

estellarc: With this in mind, the `baseTile` value might be used to clear the VRAM in this case while hiding the window in `MapNamePopUp_Hide`. Since `Window_ClearAndCopyToVRAM` might use `baseTile` depending on the configuration

lhearachel: Mmh! That's an interesting observation. And it would follow logically from the window being created before being filled with tile data.

There is also this sample below when the graphics data is being loaded:

```
1024 - (18 + 12) - 9 - (32 * 8)
```

Which is a key component of the expression here. That should offer some clues. `32 * 5` would cleanly refer to the width and height of the window being defined here, but I'm not so sure where `18 + 12 + 24` or `27 * 4` come into play...

estellarc: `18 + 12 + 24 `and `27 * 4` also does appear in most `Window_Add` calls just like `1024 - (18 + 12) - 9 - (32 * 8)` also does, the only difference is the `32 * 5`. So, you are probably right, and it is the size.

But figuring out the rest of the magic numbers might be out of scope by now, but I'm not gonna lie, I also want to know the answer.

lhearachel: I would say that it's out of scope for this PR and can be revisited in the future. :grin:

lhearachel (approved):
I'm fine to call this mergeable, after the individual comments (e.g. `I'm confused by this`) are stripped.

PR #354, opened 2025-01-30
Document some math helpers

Gudf:
Item "`unk_0201E3BC` and `unk_0201E190` - Math-related helper functions" from #325 

I'm still unsure what to rename the files? Any input is appreciated.

It looks like the first param of CalcLinearFov is actually half of what's typically called the field of view, i.e half of the angle between the top and bottom frustum planes. This is consistent with the rest of the codebase so I stuck with calling it the FoV here, but it might be worth at least documenting, or renaming every similarly incorrect instance of FoV throughout the codebase to something like half FoV?


src/unk_0201E190.c
```diff
@@ -49,39 +49,35 @@ fx32 VEC_AngleBetween(const VecFx32 *a, const VecFx32 *b)
...
-    MTX_RotX33(param0, FX_SinIdx((u16)param1->x), FX_CosIdx((u16)param1->x));
-    MTX_RotY33(&v0, FX_SinIdx((u16)param1->y), FX_CosIdx((u16)param1->y));
-    MTX_Concat33(param0, &v0, param0);
-    MTX_RotZ33(&v0, FX_SinIdx((u16)param1->z), FX_CosIdx((u16)param1->z));
-    MTX_Concat33(param0, &v0, param0);
+    MTX_RotX33(outRot, FX_SinIdx((u16)angles->x), FX_CosIdx((u16)angles->x));
+    MTX_RotY33(&tmp, FX_SinIdx((u16)angles->y), FX_CosIdx((u16)angles->y));
+    MTX_Concat33(outRot, &tmp, outRot);
+    MTX_RotZ33(&tmp, FX_SinIdx((u16)angles->z), FX_CosIdx((u16)angles->z));
+    MTX_Concat33(outRot, &tmp, outRot);
```
lhearachel, line 60: **question**: Do these calls match without the explicit casts?

src/unk_0201E3BC.c
```diff
@@ -4,14 +4,11 @@
...
+u32 CalcDistance2D(s32 x0, s32 y0, s32 x1, s32 y1)
 {
-    s32 v0, v1;
-    u32 v2;
 
-    v0 = param0 - param2;
-    v1 = param1 - param3;
-    v2 = FX_Sqrt(((v0 * v0) + (v1 * v1)) << FX32_SHIFT) >> FX32_SHIFT;
+    s32 deltaX = x0 - x1;
+    s32 deltaY = y0 - y1;
```
lhearachel, line 11: **polish**: I would name these `dx` and `dy` -- or just `a` and `b` to make the relationship to the Pythagorean theorem more obvious. Your choice.

lhearachel (commented):
This looks great so far! :clap:

> It looks like the first param of CalcLinearFov is actually half of what's typically called the field of view, i.e half of the angle between the top and bottom frustum planes. This is consistent with the rest of the codebase so I stuck with calling it the FoV here, but it might be worth at least documenting, or renaming every similarly incorrect instance of FoV throughout the codebase to something like half FoV?

I would prefer we tend towards accuracy, but it may be worth tabling this until we find evidence of any usage of full-FoV.

Some thoughts on naming the modules:

1. I took a glance at our linkerscript template (`platinum.us/main.lsf`) and noticed that `unk_0201E190` and `unk_0201E3BC` are neighbors in the compiled binary. This means that, if we so choose, we can merge these two modules into one, which might make naming easier.
2. Extending from the above... to my eye, this looks like a simple set of linear algebra extensions. `fx32` values are just 20.12 fixed-point rationals, and the rest of the module builds off of that as a baseline, providing some basic computation routines for matrices and vectors.
3. Presuming that we merge the two modules into one (extracting `CalcDistance2D`'s definition, appending it to the end of `unk_0201E190.c`, and deleting the `unk_0201E3BC` translation unit), a suitable name might be `fx_util`, or `linear_alg`, or some-such. Those might feel too high-level, but we could always revisit them, if we decide that we prefer a different name.

src/unk_0201E190.c
Gudf: They do not.

Gudf:
The issue I'd see with merging them and naming the resulting module either `linear_alg` or `fx_util` is that min and max aren't really linear algebra, and CalcDistance2D uses integers... I'm thinking of not merging them and renaming `unk_0201E190` to `fx_util` and `unk_0201E3BC` to `int_distance` or simply `distance`.
Alternatively, at a quick glance, CalcDistance2D seems to be used for some 2D rendering stuff? So maybe something along the lines of `rendering_util` might make sense if we merge them.

lhearachel:
> Alternatively, at a quick glance, CalcDistance2D seems to be used for some 2D rendering stuff? So maybe something along the lines of `rendering_util` might make sense if we merge them.

I think that I would sooner not merge them than use a name like that... it sounds a little too vague for what the contents are.

lhearachel (approved):

PR #357, opened 2025-02-03
Document VRAM transfer managers for sprite components

lhearachel:
This wound up being a bigger PR than I wanted it to be, but I felt like I should just document all of these related components in one fell swoop rather than leave any of them lingering.


src/char_transfer.c
```diff
@@ -0,0 +1,1033 @@
...
+    }
+
+    if (allocation->vramType & NNS_G2D_VRAM_TYPE_2DSUB) {
+        int count = CalcBlockMaximum(allocation->size, sTaskManager->blockSizeSub);
+        int start = CalcBlockMaximum(allocation->offset - sTaskManager->freeSizeSub, sTaskManager->blockSizeSub);
+        ClearTransferRange(start, count, sTaskManager->bufSub);
+    }
+}
+
+void *CharTransfer_PopTaskManager(void)
```
ravepossum, line 436: question: I'm only familiar with push/pop in the context of stacks - is this another way it's used? Without looking at these functions, I would be inclined to think it's a stack, unless I'm just unfamiliar with this usage of that terminology.

src/char_transfer.c
```diff
@@ -0,0 +1,1033 @@
...
+        LoadImageMappingForScreen(task, task->vramType);
+    } else {
+        LoadImageMappingForScreen(task, NNS_G2D_VRAM_TYPE_2DMAIN);
+        LoadImageMappingForScreen(task, NNS_G2D_VRAM_TYPE_2DSUB);
+    }
+}
+
+static void LoadImageMappingForScreen(CharTransferTask *task, NNS_G2D_VRAM_TYPE vramType)
+{
+    int mappingType = task->useHardwareMappingType
```
ravepossum, line 660: question: Is this our preferred style or is this a clang-format moment? (Personally I don't like this syntax for ternaries)

src/char_transfer.c
```diff
@@ -0,0 +1,1033 @@
...
+
+static void UpdateVramCapacities(void)
+{
+    switch (GX_GetBankForOBJ()) {
+    case GX_VRAM_OBJ_NONE:
+        sTaskManager->vramSizeMain = 0;
+        break;
+    case GX_VRAM_OBJ_16_F:
+    case GX_VRAM_OBJ_16_G:
+        sTaskManager->vramSizeMain = 16 * 1024;
```
ravepossum, line 729: question: Is there a descriptive constant we can use for the `1024`s here?

src/pltt_transfer.c
```diff
@@ -0,0 +1,502 @@
...
+    task->data = template->data;
+
+    if (PlttTransfer_HasTask(template->resourceID) == TRUE) {
+        GF_ASSERT(FALSE);
+        return FALSE;
+    }
+
+    task->resourceID = template->resourceID;
+    task->vramType = template->vramType;
+    task->hasWork = 1;
```
ravepossum, line 263: suggestion: I think this would make more sense as `TRUE` even if the type isn't a `BOOL` (which I assume doesn't match)

src/char_transfer.c
```diff
@@ -0,0 +1,1033 @@
...
+    NNS_G2D_VRAM_TYPE vramType;
+    u8 done;
+    u32 resourceID;
+    NNSG2dImageProxy imageProxy;
+    u32 baseAddrMain;
+    u32 baseAddrSub;
+    u8 state;
+    BOOL useHardwareMappingType;
+    BOOL atEnd;
+    BOOL haveRange;
```
ravepossum, line 31: very nitpicky nitpick: Should this be `hasRange`?

src/pltt_transfer.c
```diff
@@ -0,0 +1,502 @@
...
+
+typedef struct PlttTransferTask {
+    NNSG2dPaletteData *data;
+    NNS_G2D_VRAM_TYPE vramType;
+    u32 numPalettes;
+    u32 resourceID;
+    NNSG2dImagePaletteProxy paletteProxy;
+    u32 baseAddrMain;
+    u32 baseAddrSub;
+    u8 hasWork;
```
ravepossum, line 25: thought: It's not immediately clear to me what `work` means in this context. I wonder if we could come up with a better name for this (even if it's just something like `hasData` or `hasResource`). Maybe even `initialized`?

ravepossum (commented):
Just a small handful of nitpicks and names I think we could refine.

src/char_transfer.c
lhearachel: Push/Pop can apply to any collection or concept where an underlying reference changes as a result of the operation. That definitely applies here (the private task manager changes pointer changes).

src/char_transfer.c
lhearachel: Either works. I went with the newlines by default, but I will revert.

src/pltt_transfer.c
lhearachel: I just missed this one, good catch. :grin:

src/pltt_transfer.c
lhearachel: I think `initialized` is fine.

src/char_transfer.c
```diff
@@ -0,0 +1,1033 @@
...
+    }
+    if (task->vramType & NNS_G2D_VRAM_TYPE_2DSUB) {
+        VramTransfer_Request(NNS_GFD_DST_2D_OBJ_CHAR_SUB, task->baseAddrSub, data->pRawData, data->szByte);
+    }
+}
+
+void CharTransfer_ResetTask(int resourceID)
+{
+    CharTransferTask *task;
+    BOOL repeat = TRUE;
```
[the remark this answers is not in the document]
lhearachel: It's really just a linear search for an active task. I decided on `stillSearching`.

src/char_transfer.c
lhearachel: It's just a kilobyte.

ravepossum (approved):
Looks good to me!

PR #358, opened 2025-02-03
Documenting code for the Stats Judge

justusaac:
I renamed `ScrCmd_2EE` to `ScrCmd_JudgeStats` and documented it. That function is basically just for the stats judge (scientist guy in the battle tower) to decide what to say.


asm/macros/scrcmd.inc
```diff
@@ -4077,7 +4077,7 @@
     .short 749
     .endm
 
-    .macro ScrCmd_2EE arg0, arg1, arg2, arg3
+    .macro ScrCmd_JudgeStats arg0, arg1, arg2, arg3
```
lhearachel, line 4080: **polish**: For the ASM macro only, change the name to `JudgeStats` (so that scripts can reference it as such).

src/overlay005/ov5_021F6454.c
```diff
@@ -436,57 +436,57 @@ static void ov5_021F6AD4(UnkStruct_ov5_021F6704 *param0)
...
 
-static const u16 Unk_ov5_02200C84[] = {
-    0x7A,
-    0x7B,
-    0x7C,
-    0x7F,
-    0x7D,
-    0x7E
+// See message_bank_battle_tower.gmm
+static const u16 highestIVMessageIndices[] = {
```
lhearachel, line 440: **polish**: `sHighestIVMessageIndices`

src/overlay005/ov5_021F6454.c
```diff
@@ -436,57 +436,57 @@ static void ov5_021F6AD4(UnkStruct_ov5_021F6704 *param0)
...
+    122,
```
lhearachel, line 441: **polish**: Replace hard-coded indices with the generated constants from message encoding. (Check out `build/res/text/bank/battle_tower.h` to see what I mean; you can include that file like any other C header via `#include "res/text/bank/battle_tower.h"`)

lhearachel (approved):

PR #359, opened 2025-02-03
Document map matrix

Kuruyia:
This documents code related to the map matrix NARC (mainly, what was `unk_02039C80.c`/`unk_02039C80.h`).

For reference, here are related files in other projects:
- pokediamond: [map_matrix.c](https://github.com/pret/pokediamond/blob/master/arm9/src/map_matrix.c), [map_matrix.h](https://github.com/pret/pokediamond/blob/master/include/map_matrix.h)
- pokeheartgold: [map_matrix.c](https://github.com/pret/pokeheartgold/blob/master/src/map_matrix.c), [map_matrix.h](https://github.com/pret/pokeheartgold/blob/master/include/map_matrix.h)


src/map_matrix.c
```diff
@@ -0,0 +1,271 @@
...
+struct MapMatrix_t {
+    u8 width;
+    u8 height;
+    u8 matrixId;
+    MapMatrixData data;
+};
+
+typedef struct MainMapMatrixData_t {
+    u16 mapHeaderIds[MAP_MATRIX_MAX_SIZE];
+} MainMapMatrixData;
```
lhearachel, line 33: **polish**: We tend to expose structs as fully-transparent when they need to have public accessibility. Which is to say: move these and their `typedef` declarations to `include/map_matrix.h` and delete the `struct_decl` headers. :grin:

src/map_matrix.c
```diff
@@ -0,0 +1,271 @@
...
+const u16 MapMatrix_GetLandDataIdByIndex(const int index, const MapMatrix *mapMatrix)
+{
+    u16 v0;
+    int x = index % mapMatrix->width;
+    int y = index / mapMatrix->width;
+
+    GF_ASSERT(mapMatrix != NULL);
+
+    v0 = MapMatrix_GetLandDataIdByIndex2(index, mapMatrix);
+    return v0;
```
lhearachel, line 230:
```suggestion
    return MapMatrix_GetLandDataIdByIndex2(index, mapMatrix);
```

src/map_matrix.c
```diff
@@ -0,0 +1,271 @@
...
+    GF_ASSERT(0 <= y && y < mapMatrixHeight);
+
+    return mapMatrix->data.mapHeaderIds[x + y * mapMatrixWidth];
+}
+
+u16 MapMatrix_GetMapHeaderIDAtIndex(const MapMatrix *mapMatrix, const int index)
+{
+    int mapSize = mapMatrix->width * mapMatrix->height;
+
+    GF_ASSERT(0 <= index && index < mapSize);
```
lhearachel, line 158:
```suggestion
    GF_ASSERT(0 <= index && index < mapMatrix->width * mapMatrix->height);
```

src/map_matrix.c
```diff
@@ -0,0 +1,271 @@
...
+    return v0;
+}
+
+void MapMatrix_RevealSpringPath(MapMatrix *mapMatrix)
+{
+    u16 *mapMatrixLandDataIds = mapMatrix->data.landDataIds;
+    u8 *mapMatrixAltitudes = mapMatrix->data.altitudes;
+    int mapMatrixHeight = mapMatrix->height;
+    int mapMatrixWidth = mapMatrix->width;
+    int v4, v5;
```
lhearachel, line 239: **polish**: Remove unused vars

src/map_matrix.c
```diff
@@ -0,0 +1,271 @@
...
+
+int MainMapMatrixData_GetMapHeaderIDAtCoords(const MainMapMatrixData *mainMapMatrixData, const u32 x, const u32 y)
+{
+    return mainMapMatrixData->mapHeaderIds[x + y * MAP_MATRIX_MAX_WIDTH];
+}
+
+int MapMatrixData_GetMapHeaderIDAtCoords(const int mapMatrixId, const u16 x, const u16 y)
+{
+    MapMatrixData mapMatrixData;
+    u16 v1, v2;
```
lhearachel, line 211: **polish**: Remove unused vars.

src/map_matrix.c
```diff
@@ -0,0 +1,271 @@
...
+
+    mapMatrixAltitudes[23 + mapMatrixWidth * 21] = 2;
+    mapMatrixAltitudes[24 + mapMatrixWidth * 21] = 2;
+    mapMatrixAltitudes[23 + mapMatrixWidth * 22] = 2;
+    mapMatrixAltitudes[24 + mapMatrixWidth * 22] = 2;
+}
+
+void MapMatrix_RevealSeabreakPath(MapMatrix *mapMatrix)
+{
+    int v0, v1;
```
lhearachel, line 258: **polish**: Remove unused vars

src/map_matrix.c
Kuruyia: Understood, looks cleaner that way!

My only problem with that is that it increases the amount of dependencies there is on a single header file (in this case, `map_matrix.h`, which is now required by `field_system.h`, which in its turn is required by a ton of other headers), which may greatly increase the amount of source file to recompile in case the header is modified. Is that an acceptable tradeoff?

Kuruyia:
Thanks for the review!

src/map_matrix.c
lhearachel: Yep, that's a known and acceptable trade-off to us. Thanks for double-checking. 😁

src/map_matrix.c
lhearachel: Unfortunate. Ah, well.

lhearachel (approved):

PR #361, opened 2025-02-05
Document sprite_util.c

lhearachel:
This file is a little bit on the awkward side; it doesn't really fit anywhere else, but it can't easily be merged into the neighboring translation units, either. So, here it stays.

Next on my list is `unk_0200C6E4`, which should finally have sufficient external context to be able to be documented itself.


src/render_view.c
```diff
@@ -0,0 +1,72 @@
...
+    const NNSG2dCellBoundingRectS16 *boundingRect = NNS_G2dGetCellBoundingRect(cellData);
+    const fx32 boundingRadius = NNS_G2dGetCellBoundingSphereR(cellData);
+    const fx32 originX = objectCoords->_20 - viewRect->posTopLeft.x;
+    const fx32 originY = objectCoords->_21 - viewRect->posTopLeft.y;
+
+    fx32 minY, maxY;
+    fx32 minX, maxX;
+    fx32 tmp;
+
+    if (NNS_G2dCellHasBR(cellData) == 1) {
```
ravepossum, line 42: polish: `TRUE` here

src/sprite_util.c
```diff
@@ -0,0 +1,212 @@
...
+#include "pltt_transfer.h"
+#include "render_oam.h"
+#include "render_view.h"
+#include "sprite_resource.h"
+#include "sprite_transfer.h"
+#include "sprite_util.h"
+
+#define RESDAT_TABLE_TERMINATOR 0xFFFFFFFE
+
+void SpriteResourcesHeader_Init(CellActorResourceData *resourceHeader, int charResourceID, int plttResourceID, int cellResourceID, int animResourceID, int mcellResourceID, int manimResourceID, int vramTransfer, int priority, SpriteResourceCollection *charResources, SpriteResourceCollection *plttResources, SpriteResourceCollection *cellResources, SpriteResourceCollection *animResources, SpriteResourceCollection *mcellResources, SpriteResourceCollection *manimResources)
```
ravepossum, line 18: question: Does this match with `int vramTransfer` replaced with `BOOL vramTransfer`?

src/sprite_util.c
```diff
@@ -0,0 +1,212 @@
...
+
+    default:
+        GF_ASSERT(FALSE);
+        break;
+    }
+}
+
+void ReserveSlotsForWirelessIconPalette(NNS_G2D_VRAM_TYPE vramType)
+{
+    PlttTransfer_MarkReservedSlots((1 << 14 | 1 << 15), vramType);
```
ravepossum, line 211: question: Are these masks something we should make constants for?

include/sprite_util.h
```diff
@@ -0,0 +1,64 @@
...
+    u32 cellResourceID;
+    u32 animResourceID;
+    u32 mcellResourceID;
+    u32 manimResourceID;
+    u32 vramTransfer;
+    u32 priority;
+} ResdatTableEntry;
+
+typedef struct G2dRenderer {
+    NNSG2dRendererInstance renderer;
```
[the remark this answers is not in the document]
lhearachel: Likely, yes. We could call it something different, or more clearly delineate that `renderer` is the encapsulated hardware renderer.

src/render_oam.c
```diff
@@ -0,0 +1,160 @@
...
+    int trueMainOamStart;
+    int trueMainOamCount;
+    int trueMainAffineOamStart;
+    int trueMainAffineOamCount;
+
+    // The first 4 OAM managers are reserved.
+    if (mainOamStart < 4) {
+        trueMainOamStart = 4;
+
+        if (mainOamCount > 124) {
```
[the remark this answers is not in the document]
lhearachel: Aha, there are definitions from the SDK!

src/sprite_util.c
lhearachel: At some point, yes, but that will be up to the communication code. Out of scope for now, I think.

src/sprite_util.c
lhearachel: Sure does.

ravepossum (approved):

include/sprite_util.h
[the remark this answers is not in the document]
lhearachel: We determined that this is a non-issue; no one outside of this module seems to actually interface with the SDK renderer directly, instead going through the `CellActorCollection` that `SpriteList_InitRendering` returns.

PR #363, opened 2025-02-06
Document BDHC util

Kuruyia:
This documents the `bdhc_util.c` file, which mainly contains the function used to calculated an object's height based on its 2D position and the BDHC of the map its currently on.


Kuruyia:
Still not sure about the `objectY` naming convention. I like the consistency with `objectX`/`objectZ`, but `objectHeight` might be more understandable.

include/overlay005/bdhc_util.h
```diff
@@ -0,0 +1,11 @@
+#ifndef POKEPLATINUM_BDHC_UTIL_H
+#define POKEPLATINUM_BDHC_UTIL_H
+
+#include <nitro/fx/fx.h>
+#include <nitro/types.h>
+
+#include "overlay005/bdhc.h"
+
+BOOL BDHCUtil_CalculateObjectHeight(const fx32 objectY, const fx32 objectX, const fx32 objectZ, const BDHC *bdhc, fx32 *newObjectY);
```
lhearachel, line 9: **suggestion**: I'm definitely more in favor of `objectHeight` or `objectH` for clarity.

include/overlay005/bdhc_util.h
```diff
@@ -0,0 +1,11 @@
+#ifndef POKEPLATINUM_BDHC_UTIL_H
```
lhearachel, line 1: **question**: Given how small this is, how would we feel combining this with `bdhc_loader`?

src/overlay005/bdhc_util.c
```diff
@@ -0,0 +1,205 @@
...
+
+    // Simple binary search.
+    low = 0;
+    high = stripsSize - 1;
+    mid = high / 2;
+
+    do {
+        stripLowerBound = strips[mid].lowerBound;
+
+        if (stripLowerBound > targetLowerBound) {
```
lhearachel, line 93: **question**: Does this match as:

```suggestion
        if (strips[mid].lowerBound > targetLowerBound) {
```

after removing the `stripLowerBound` declaration?

Kuruyia: Indeed

include/overlay005/bdhc_util.h
Kuruyia: Modified every `objectY` to `objectHeight`

include/overlay005/bdhc_util.h
Kuruyia: Sounds good to me, please let me know if the final file size is alright. Merged `bdhc_loader` and `bdhc_util` into `bdhc`.

include/overlay005/bdhc.h
```diff
@@ -32,4 +35,15 @@ typedef struct BDHC {
     int stripsSize;
 } BDHC;
 
+BDHCUtil_CalculateObjectHeight(const fx32 objectHeight, const fx32 objectX, const fx32 objectZ, const BDHC *bdhc, fx32 *newObjectHeight);
```
lhearachel, line 38:
```suggestion
BOOL BDHC_CalculateObjectHeight(const fx32 objectHeight, const fx32 objectX, const fx32 objectZ, const BDHC *bdhc, fx32 *newObjectHeight);
```

Let's also remove the BDHCUtil prefix from the static functions in the C file, or at least change them to just `BDHC`.

include/overlay005/bdhc.h
```diff
@@ -32,4 +35,15 @@ typedef struct BDHC {
...
+
+void BDHCLoader_Load(NARC *narc, const int bdhcSize, BDHC *bdhc, u8 *buffer);
+SysTask *BDHCLoader_StartTask(NARC *landDataNARC, const int param1, BDHC *bdhc, int *param3, u8 **buffer, int *param5);
+void BDHCLoader_ForceExitTask(SysTask *sysTask);
+void BDHCLoader_MarkBDHCNotLoaded(BDHC *bdhc);
```
lhearachel, line 43: I think that we can simplify these names now:

```suggestion
void BDHC_Load(NARC *narc, const int bdhcSize, BDHC *bdhc, u8 *buffer);
SysTask *BDHC_LazyLoad(NARC *landDataNARC, const int param1, BDHC *bdhc, int *param3, u8 **buffer, int *param5);
void BDHC_KillLoad(SysTask *sysTask);
void BDHC_MarkNotLoaded(BDHC *bdhc);
```

Looking at the undocumented fields in `LazyLoad`, I also see that...

- `param1` (and `unk_E4`) is unused internally.
- `param3` (and `unk_DC`) is effectively unused (it might have been intended to be a communication mechanism to the client?).
- `param5` (and `unk_E8`) is effectively unused (it acts as a delayed kill-switch in the task's `Prepare` stage).

I think that we should update those names while we're in this module, just to capstone them.

Kuruyia: Agreed for `unk_E4`.

For `unk_E8`, it looks like it is used. Referenced [here](https://github.com/pret/pokeplatinum/blob/f1ac42f3875bd4345d3a985dd296d4bc458f1756/src/overlay005/ov5_021E779C.c#L503) (L503) as `param6->unk_10.unk_00`, increased right [above that line](https://github.com/pret/pokeplatinum/blob/f1ac42f3875bd4345d3a985dd296d4bc458f1756/src/overlay005/ov5_021E779C.c#L497) (L497) and passed right after (L498) to a function that starts a systask.  
This function + the systask itself are pretty much identical to our Load/LazyLoad; pretty certain that this is loading the NSBMD seeing the NNSG3d* types in the function parameters and seeing that it's started before the BDHC systask.
Anyway, `param6->unk_10.unk_00` is [now](https://github.com/pret/pokeplatinum/blob/f1ac42f3875bd4345d3a985dd296d4bc458f1756/src/overlay005/ov5_021E779C.c#L1939) (L1939) `v1->unk_20`, and in the actual systask routine, it's updated [here](https://github.com/pret/pokeplatinum/blob/f1ac42f3875bd4345d3a985dd296d4bc458f1756/src/overlay005/ov5_021E779C.c#L1917) (L1917) at the last subtask. (and just thought about it when I was writing this last sentence - should have probably called them "task steps")
So it looks like `unk_E8` is some sort of synchonization variable that forces the NSBMD(?) to be loaded before the BDHC.

Didn't bother checking if the same was true for `unk_DC`, but it may be another sync variable (they're [both in the same struct](https://github.com/pret/pokeplatinum/blob/f1ac42f3875bd4345d3a985dd296d4bc458f1756/src/overlay005/ov5_021E779C.c#L29-L32) in `ov5_021E779C`)

Kuruyia: And another question - what do you usually call unused variables? I guess `unusedX`?

lhearachel: For parameters, just `unusedX` is fine. For struct fields, typically we use `dummyX` (where `X` is a byte-address within the struct, for clarity).

include/overlay005/bdhc.h
Kuruyia: Normalized the prefixes to `BDHC` everywhere, looks way cleaner ^^

include/overlay005/bdhc.h
Kuruyia: Renamed the unused internal variables. Leaving `unk_DC`/`unk_E8` (and related) untouched for now as we'll need to explore what those pointers are shared with.

lhearachel (approved):

PR #364, opened 2025-02-08
Document boot.c and system.c

lhearachel:


src/system.c
```diff
@@ -0,0 +1,463 @@
...
+    gSystem.touchHeld = 0;
+    gSystem.touchAutoSampling = FALSE;
+
+    TP_Init();
+
+    TPCalibrateParam touchCalibration;
+    if (TP_GetUserInfo(&touchCalibration) == TRUE) {
+        TP_SetCalibrateParam(&touchCalibration);
+    } else {
+        touchCalibration.x0 = 0x02AE;
```
ravepossum, line 272: question: Do these magic numbers warrant constants? I don't immediately see any out-of-the-box ones in the SDK, so I guess those specific values are just something GF came up with?

ravepossum (commented):
I didn't have much to comment on here - significant quantity of it is just renaming noise lol.

src/gx_layers.c
```diff
@@ -98,7 +98,7 @@ void GXLayers_TurnBothDispOn(void)
 
 void GXLayers_SwapDisplay(void)
 {
-    if (gCoreSys.unk_65 == 0) {
+    if (gSystem.whichScreenIs3D == 0) {
```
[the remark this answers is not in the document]
lhearachel: Caught the rest of them while I was at it.

src/system.c
lhearachel: Yeah, they are just GF-values. I'm not sure what to call them. It would require a deep-dive into how the touch calibration system works. Feels out-of-scope.

ravepossum (approved):

PR #367, opened 2025-02-09
Document sprite_system.c + update sprite terminology

lhearachel:
The size of this PR is a bit misleading due to the number of core symbols renamed. The main files of focus are `sprite.h`, `sprite.c`, `sprite_system.h`, and `sprite_system.c`. All other files touched are affected primarily by mass-renames.


include/struct_decls/sprite_decl.h
```diff
@@ -1,6 +1,6 @@
 #ifndef POKEPLATINUM_SPRITE_DECL_H
 #define POKEPLATINUM_SPRITE_DECL_H
 
-typedef struct Sprite Sprite;
+typedef struct PokemonSprite PokemonSprite;
```
ravepossum, line 4: question: Did you leave this file in because we have a dependency web, or can we yeet this? If we are keeping it around, we might want to rename the file.

include/struct_defs/sprite_template.h
```diff
@@ -3,20 +3,4 @@
...
-    s16 z; //!< Starting Z-coordinate of the sprite
-    u16 animIdx; //!< Index in the animation file to use
-    int priority; //!< Priority layer to draw the sprite against
-    int plttIdx; //!< Index in the palette file to use
-    enum NNS_G2D_VRAM_TYPE vramType; //!< VRAM type which will own any loaded resources
-    int resources[MAX_SPRITE_RESOURCES]; //!< Set of resource IDs to register as loaded into VRAM
-    int bgPriority;
-    BOOL transferToVRAM;
-} SpriteTemplate;
-
```
ravepossum, line 21: question: Is there any reason for this to be a file at this point instead of just including `g2d_Image.h` directly?

src/sprite.c
```diff
@@ -0,0 +1,929 @@
...
+    sprite->list = NULL;
+    memset(sprite, 0, sizeof(Sprite));
+
+    NNS_G2dInitImageProxy(&sprite->imageProxy);
+    NNS_G2dInitImagePaletteProxy(&sprite->paletteProxy);
+
+    sprite->explicitOamMode = GX_OAM_MODE_NORMAL;
+}
+
+Sprite *SpriteList_AddEx(const AffineSpriteListTemplate *params)
```
ravepossum, line 175: suggestion: We should probably rename this something like `SpriteList_AddAffine` now that we've renamed the parameter struct.

src/sprite.c
```diff
@@ -0,0 +1,929 @@
...
+{
+    return &sprite->imageProxy;
+}
+
+NNSG2dImagePaletteProxy *Sprite_GetPaletteProxy(Sprite *paletteProxy)
+{
+    return &paletteProxy->paletteProxy;
+}
+
+void Sprite_SetPixelated(Sprite *sprite, BOOL pixelated)
```
ravepossum, line 544: thought: I know this name predates this PR, so it could be out of scope, but calling this `Pixelated` instead of `Mosaic` feels odd to me since that's a more canonical name for the effect (and also what's used by the SDK and the struct).

src/sprite_system.c
```diff
@@ -0,0 +1,1249 @@
...
+            unownedResources->resources[i] = NULL;
+            unownedResources->count--;
+            return TRUE;
+        }
+    }
+
+    return FALSE;
+}
+
+static BOOL UnregisterLoadedCharResource(SpriteResourceCollection *ownedResources, SpriteResourceList *unownedResources, int toUnload)
```
ravepossum, line 695: suggestion: Maybe this would be a bit clearer as something like `idToUnload` instead of just `toUnload`? It is annoying how the casing ends up being incongruent though 🤔

src/sprite_system.c
```diff
@@ -0,0 +1,1249 @@
...
+{
+    return Sprite_GetAnimFrame(sprite);
+}
+
+u16 ManagedSprite_GetAnimationFrame(ManagedSprite *managedSprite)
+{
+    return GetSpriteAnimationFrame(managedSprite->sprite);
+}
+
+void Sprite_SetDrawFlag2(Sprite *sprite, BOOL draw)
```
ravepossum, line 825: why is GF the way they are

src/sprite_system.c
```diff
@@ -0,0 +1,1249 @@
...
+void ManagedSprite_SetAffineTranslation(ManagedSprite *managedSprite, s16 x, s16 y)
+{
+    VecFx32 translation;
+    translation.x = x << FX32_SHIFT;
+    translation.y = y << FX32_SHIFT;
+    translation.z = 0;
+    Sprite_SetAffineTranslation(managedSprite->sprite, &translation);
+}
+
+void ManagedSprite_SetPixelated(ManagedSprite *managedSprite, BOOL pixelated)
```
ravepossum, line 1121: thought: if we do end up renaming `Pixelated`->`Mosaic`, we'd want to do so here too of course

include/sprite_system.h
```diff
@@ -0,0 +1,188 @@
...
+    s16 z;
+    u16 animIdx;
+    int priority;
+    int plttIdx;
+    enum NNS_G2D_VRAM_TYPE vramType;
+    int dummy18;
+    int dummy1C;
+    int dummy20;
+    int dummy24;
+} SpriteTemplateFromResourceHeader;
```
ravepossum, line 62: thought: This feels a little too verbose, but it does seem like a more accurate name than what I had proposed (`SpriteListEntryTemplate`). That being said...I don't immediately have a better proposal. I don't know if it's worth fretting too much given how scarcely it's used, but if a flash of inspiration for a concise name hits you (or someone else) maybe it'd be worth discussing further.

include/applications/pokemon_summary_screen/sprites.h
```diff
@@ -1,5 +1,5 @@
-#ifndef POKEPLATINUM_POKEMON_SUMMARY_SCREEN_SPRITE_H
-#define POKEPLATINUM_POKEMON_SUMMARY_SCREEN_SPRITE_H
+#ifndef POKEPLATINUM_POKEMON_SUMMARY_SCREEN_SPRITES_H
+#define POKEPLATINUM_POKEMON_SUMMARY_SCREEN_SPRITES_H
```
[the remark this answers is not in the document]
lhearachel: It's required to compile with our current `-I` paths, which prefer current directory above all other possible paths. So, this creates a collision with `include/sprite.h`.

include/battle/healthbar.h
```diff
@@ -65,8 +62,8 @@ enum HealthbarGaugeType {
  * @param palette
  * @param healthbarType
  */
-void Healthbar_LoadResources(SpriteRenderer *renderer, SpriteGfxHandler *gfxHandler, NARC *narc, PaletteData *palette, enum HealthbarType healthbarType);
-CellActorData *Healthbar_LoadCellActor(SpriteRenderer *renderer, SpriteGfxHandler *handler, int type);
+void Healthbar_LoadResources(SpriteSystem *renderer, SpriteManager *gfxHandler, NARC *narc, PaletteData *palette, enum HealthbarType healthbarType);
+ManagedSprite *Healthbar_LoadCellActor(SpriteSystem *renderer, SpriteManager *handler, int type);
```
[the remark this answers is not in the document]
lhearachel: I am personally planning to revisit the healthbar code (all the battle code, really).

As for the others, I think that we should leave it to the various owners to clean things up, since there is so much of it.

include/struct_decls/sprite_decl.h
lhearachel: I didn't try to remove it; I only did this rename to allow renaming `CellActor` to `Sprite`. I can try to remove it.

include/struct_defs/sprite_template.h
lhearachel: Probably not. Will take a look at fixing.

src/sprite.c
lhearachel: I didn't invest much effort into renaming the functions in this file; would you like for me to take up that work, or do you want to take care of it as the spearhead for the terminology updates?

src/sprite.c
lhearachel: I agree, and it should be updated.

include/battle/healthbar.h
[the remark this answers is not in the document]
ravepossum: I only had in mind a relatively simple find/replace which is why I was specifically calling out the super generically named ones - definitely agree that anything more specific than that should be handled separately. I don't think it's a dealbreaker or anything but would be nice to have - if it proves messy I don't see a problem with handling it as we go.

src/sprite.c
ravepossum: Honestly I thought nearly all of them still seemed appropriate post-renames, this and the mosaic one were really the only ones that stuck out to me. I don't have any qualms with revisiting them after the fact myself though.

src/sprite_system.c
lhearachel: Settled on `resourceID` for the parameter and `memberResourceID` for the local variable.

lhearachel:
Addressed feedback after resolving merge conflicts.

ravepossum (approved):

PR #368, opened 2025-02-10
Document map 3D objects

Kuruyia:
This mainly documents the `ov5_021E15F4.c` file, which is in charge of reading 3D map objects from the land data NARC, managing them, and rendering them.

A "3D map object" is defined here as a 3D model that is separate from the map 3D model, and that can be positioned, rotated, and scaled independently. For instance, they are usually used to place buildings on the various maps.


Kuruyia:
Still not sure about the `Map3DObject` naming scheme, may be too close to the already-existing `MapObject`

Also, maybe `LoadedMap3DObjects` is worthy of being called `Map3DObjectManager`?

include/constants/map_3d_object.h
```diff
@@ -0,0 +1,6 @@
+#ifndef POKEPLATINUM_CONSTANTS_MAP_3D_OBJECT_H
+#define POKEPLATINUM_CONSTANTS_MAP_3D_OBJECT_H
+
+#define MAX_LOADED_3D_OBJECTS 32
```
lhearachel, line 4: **suggestion**: I think that this can just live in the translation unit's header file.

include/field/field_system.h
```diff
@@ -109,7 +109,7 @@ typedef struct FieldSystem_t {
     BagCursor *unk_98;
     JournalEntry *journalEntry;
     UnkStruct_ov5_021EF300 *unk_A0;
-    UnkStruct_ov5_021E1608 *unk_A4;
+    LoadedMap3DObjects *unk_A4;
```
lhearachel, line 112: **polish**: Let's name this field, since we know what it is.

include/overlay005/honey_tree.h
```diff
@@ -14,7 +14,7 @@ u16 HoneyTree_GetTreeSlatherStatus(FieldSystem *fieldSystem);
 void HoneyTree_SlatherTree(FieldSystem *fieldSystem);
 void HoneyTree_StopShaking(FieldSystem *fieldSystem);
 int HoneyTree_GetSpecies(FieldSystem *fieldSystem);
-void ov5_021F0030(void *param0, const int param1, UnkStruct_ov5_021E1608 *const param2);
+void ov5_021F0030(void *param0, const int param1, LoadedMap3DObjects *const param2);
```
lhearachel, line 17: **question**: Is `param2` here just `treeObject`?

src/overlay005/map_3d_object.c
```diff
@@ -0,0 +1,355 @@
...
+                v4 = ov5_021EFAB4(param1);
+
+                if (param2 == TRUE) {
+                    ov5_021D5948(param3, loadedObject->resMdl, (1 | 1 << 1 | 1 << 4 | 1 << 5 | 1 << 8 | 1 << 9 | 1 << 10 | 1 << 11));
+                }
+
+                ov5_021D5790(loadedObject->id, v4, &v5);
+
+                if (v5 == 0) {
+                    Easy3D_DrawRenderObj(&(loadedObject->renderObj), &(position), &rotationMatrix, &(loadedObject->scale));
```
lhearachel, line 171: **polish**: Remove unnecessary parens.

include/overlay005/map_3d_object.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include "narc.h"
+
+typedef struct Map3DObject {
+    int id;
+    BOOL loaded;
+    BOOL hidden;
+    BOOL applyRotation;
+    NNSG3dRenderObj renderObj;
+    NNSG3dResMdl *resMdl;
```
lhearachel, line 21: **suggestion**: It was difficult for me to understand what `Mdl` meant at first. Maybe this should be `modelResource`?

lhearachel (commented):
> Still not sure about the `Map3DObject` naming scheme, may be too close to the already-existing `MapObject`.

I believe that these are just called `Building`s in the binary space.

> Also, maybe `LoadedMap3DObjects` is worthy of being called `Map3DObjectManager`?

I agree.

Kuruyia:
> I believe that these are just called `Building`s in the binary space.

Didn't know about that, makes sense. Then let's go with `Building`/`BuildingFile`/`BuildingManager` for the structs here

include/overlay005/honey_tree.h
Kuruyia: `param2` here would be all the currently loaded 3D objects. But if we go a bit further, we can see that the tree object is seemingly [retrieved from those loaded 3D objects](https://github.com/pret/pokeplatinum/blob/0e45deb3f6606501cfebb896753f3efe075cdab4/src/overlay005/honey_tree.c#L368), and then [an animation is added to its render object](https://github.com/pret/pokeplatinum/blob/b4fc5c1a79f8215d04fa70e2bbd46f14d57ce708/src/overlay005/ov5_021D37AC.c#L349) (certainly the shaking animation if the function names are to be believed).

As a side note, it seems that by searching all calls to `LoadedMap3DObjects_FindLoadedObjectById`, we can build a list of "well-known" 3D object IDs. For now, we can only easily deduce that 26 is the ID of the tree object because other calls to that function are in undocumented territory.

include/overlay005/map_3d_object.h
Kuruyia: I've renamed it to simply `model` as it is also the name used in `easy3d_object.h`. Specifying a "Resource" suffix would be redundant if multiple `NNSG3dResX` variables are present at the same place.

lhearachel:
> Note: it may make less sense for non-building objects such as the honey tree you mentioned in another comment

I think that this is a fine trade-off. In the context of mapping, it's easier to refer to them as `buildings` as a unifying term, and I believe that the `nsbmd` extension used in the SDK types implicitly refers to `building` as a term.

lhearachel (dismissed):

src/overlay005/building.c
```diff
@@ -0,0 +1,354 @@
...
+
+static void BuildingManager_InitRenderObj(const int loadedObjectId, UnkStruct_ov5_021EF76C *const param1, NNSG3dRenderObj *renderObj, NNSG3dResMdl **resMdl)
+{
+    NNSG3dResFileHeader **nsbmdFile = ov5_021EF9E8(loadedObjectId, param1);
+
+    *resMdl = NNS_G3dGetMdlByIdx(NNS_G3dGetMdlSet(*nsbmdFile), 0);
+    NNS_G3dRenderObjInit(renderObj, *resMdl);
+}
+
+static void ov5_021E1944(const NNSG3dResMdl *resMdl, VecFx32 *position, MtxFx33 *rotation, VecFx32 *scale, const UnkStruct_ov5_021D5778 *param4, const int param5)
```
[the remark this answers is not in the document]
Kuruyia: We'll probably want to document the [`ov5_021D56BC.c`](https://github.com/pret/pokeplatinum/blob/d634ed467ee3373e5b48321d3f10cf08d2700976/src/overlay005/ov5_021D56BC.c) file as this is the one in which `UnkStruct_ov5_021D5778`, `ov5_021D579C()` and `ov5_021D57B4()` are defined. Looks like it may be a pretty quick one to document.

Also, if you have an idea of what `NNS_G3dDraw1Mat1Shp()` does, that would help me. Looking at the function name and definition wasn't really helpful. Something about material and shape?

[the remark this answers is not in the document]
lhearachel: > Also, if you have an idea of what `NNS_G3dDraw1Mat1Shp()` does, that would help me. Looking at the function name and definition wasn't really helpful. Something about material and shape?

Yep, that's my reading on it: sends the requested material and shape to VRAM for rendering.

lhearachel:
Coming back to this, per discussion on Discord, let's walk back on using the binary terminology and generalize the name to something like `MapProp` or `Map3DObject`. There were some good points raised in that discussion.

Kuruyia:
> Coming back to this, per discussion on Discord, let's walk back on using the binary terminology and generalize the name to something like `MapProp` or `Map3DObject`. There were some good points raised in that discussion.

`MapProp` sounds good to me, renamed everything to that. Updated the commit message to reflect that as well.

Also added a `MapPropId` enum for gathering all well-known map prop IDs in a single place. Related to [this comment](https://github.com/pret/pokeplatinum/pull/368#discussion_r1950238144) about the (now) `MapPropManager_FindLoadedPropById` function.

lhearachel (approved):
The new rename looks good! Thanks for working through all the churn of renaming. :sweat_smile:

PR #373, opened 2025-02-13
Document npc_trade.c

cbt6:


res/text/npc_trade_names.gmm
```diff
@@ -1,35 +1,35 @@
 <?xml version="1.0"?>
 <body language="English">
 	<key value="4839" />
-	<row id="pl_msg_00000370_00000" index="0">
+	<row id="npc_trades_names_kazza" index="0">
```
lhearachel, line 4: **nitpick**: `npc_trade_names_...` to match the filename (or vice-versa, your call).

res/field/trades/meson.build
```diff
@@ -0,0 +1,32 @@
...
+    'npc_trade_charap_chatot.json',
+    'npc_trade_gaspar_haunter.json',
+    'npc_trade_foopa_magikarp.json',
+)
+
+npc_trades_order = files('npc_trades.order')
+npc_trades_narc = custom_target('fld_trade.narc',
+    output: [
+        'fld_trade.narc',
+        'fld_trade.naix',
```
lhearachel, line 18: **question**: Is there any reason to generate this `naix` rather than use `generated/npc_trades.txt`?

res/field/trades/meson.build
```diff
@@ -0,0 +1,32 @@
...
+    arguments: [ '@INPUT@', '@OUTPUT@', ],
+    output: '@BASENAME@'
+)
+
+npc_trades_files = files(
+    'npc_trade_kazza_abra.json',
+    'npc_trade_charap_chatot.json',
+    'npc_trade_gaspar_haunter.json',
+    'npc_trade_foopa_magikarp.json',
+)
```
lhearachel, line 12: **question**: Can we tie this to `generated/npc_trades.txt` like we do for `res/pokemon/meson.build` or `res/trainers/meson.build`?

src/overlay006/npc_trade.c
```diff
@@ -0,0 +1,152 @@
...
+u32 NpcTrade_GetRequestedSpecies(const NpcTradeData *data)
+{
+    return data->npcTradeMon->requestedSpecies;
+}
+
+void ov6_02246234(FieldSystem *fieldSystem, NpcTradeData *data, int slot)
+{
+    sub_0207A128(Party_GetFromSavedata(fieldSystem->saveData), slot, data->mon);
+    sub_0202F180(fieldSystem->saveData, data->mon);
+}
```
lhearachel, line 76: **suggestion**: This looks to me like it should be `NpcTrade_SetSlotToTradedMon` or some-such:

- `sub_0207A128` sets a given party slot to a `Pokemon` struct
- `sub_0202F180` updates all the relevant flags if the species is newly-acquired

src/overlay006/npc_trade.c
```diff
@@ -0,0 +1,152 @@
...
+
+    int timeOfDay = FieldSystem_GetTimeOfDay(fieldSystem);
+    if (timeOfDay == TIMEOFDAY_MORNING || timeOfDay == TIMEOFDAY_DAY) {
+        param3->unk_0C = 0;
+    } else if (timeOfDay == TIMEOFDAY_TWILIGHT) {
+        param3->unk_0C = 1;
+    } else {
+        param3->unk_0C = 2;
+    }
+}
```
lhearachel, line 103: **suggestion**: I think that this is the routine for initializing all the data for the actual trade cutscene, since it contains:

- The two mons involved in the trade
- The options (implies to me that it wants the message frame and text printing speed, since none of the others make sense)
- The NPC's trainer info (it would want their name)

I am not sure what relevance the time of day has... did the trade cutscene in Platinum react to time of day?

lhearachel (commented):
Great work! Some collaborative thoughts below.

src/overlay006/npc_trade.c
cbt6: Gonna leave this function as it is for now, it can be renamed more suitably when `UnkStruct_ov6_02246254` gets documented. And yes the time of day determines the background used in the trade cutscene.

res/field/trades/meson.build
cbt6: No I just left that in and forgot to remove it.

lhearachel (approved):

PR #378, opened 2025-02-13
Pre poketch display

lucie-sharpe:
I am not sure if i have handled the always true function correctly. It is called in multiple if conditionals. Also i used some macros to define NARC parameters is this the correct way to do this? Should the overlay24/ folder be renamed?


platinum.us/main.lsf
```diff
@@ -864,7 +864,7 @@ Overlay overlay23
 Overlay overlay24
```
lhearachel, line 864: **polish**: Since this overlay is wholly documented, let's give it a name. You can replace this specific symbol `overlay24` with a semantic name and all the occurrences of the same token in the code. Standard C rules for an identifier apply.

Maybe `pre_poketch_subscreen`?

include/overlay024/pre_poketch_display.h
```diff
@@ -0,0 +1,10 @@
+#ifndef POKEPLATINUM_PRE_POKETCH_DISPLAY_H
```
lhearachel, line 1: **quibble**: I'd call this `pre_poketch_subscreen`, since it explicitly lives on the subscreen.

(Note that `main` and `sub` screens are different; there's no hard rule about which physical display is which at any given moment.)

src/overlay024/pre_poketch_display.c
```diff
@@ -0,0 +1,64 @@
...
+#include "graphics.h"
+#include "heap.h"
+#include "narc.h"
+
+#define PRE_POKETCH_DISPLAY_HEAP_SIZE    0x18000
+#define PRE_POKETCH_DISPLAY_NARC_HEAP_ID 8
+
+#define PRE_POKETCH_DISPLAY_NARC_TILES_IDX   10
+#define PRE_POKETCH_DISPLAY_NARC_TILEMAP_IDX 11
+#define PRE_POKETCH_DISPLAY_NARC_PALETTE_IDX 12
```
lhearachel, line 15: **thought**: These are fine for now, but we will want to replace them with an `naix` generated by re-packing this NARC once we unpack them into the resource tree.

src/overlay024/pre_poketch_display.c
```diff
@@ -0,0 +1,64 @@
...
+
+#include "constants/heap.h"
+
+#include "bg_window.h"
+#include "graphics.h"
+#include "heap.h"
+#include "narc.h"
+
+#define PRE_POKETCH_DISPLAY_HEAP_SIZE    0x18000
+#define PRE_POKETCH_DISPLAY_NARC_HEAP_ID 8
```
lhearachel, line 11: **polish**: The former of these constants belong sin `constants/heap.h` as `HEAP_SIZE_PRE_POKETCH_DISPLAY`; the latter is equivalent to `HEAP_ID_POKETCH_APP`, so we should just use that, even though this is technically not a Poketch App.

**thought**: Though, one could argue that it is the `NULL` app. :slightly_smiling_face:

src/overlay024/pre_poketch_display.c
```diff
@@ -0,0 +1,64 @@
...
+        1,
+        GX_BG_COLORMODE_16,
+        GX_BG_SCRBASE_0x6000,
+        GX_BG_CHARBASE_0x00000,
+        GX_BG_EXTPLTT_01,
+        0,
+        0,
+        0,
+        0
+    };
```
lhearachel, line 33: **polish**: Our compiler is C99 compliant; use designated initializers here for clarity.

```suggestion
    static const BgTemplate template = {
        .x = 0,
        .y = 0,
        .bufferSize = 0x800,
        .baseTile = 0,
        .screenSize = BG_SCREEN_SIZE_256x256,
        .colorMode = GX_BG_COLORMODE_16,
        .screenBase = GX_BG_SCRBASE_0x6000,
        .charBase = GX_BG_CHARBASE_0x00000,
        .bgExtPltt = GX_BG_EXTPLTT_01,
        .priority = 0,
        .areaOver = 0,
        .dummy = 0,
        .mosaic = FALSE,
    };
```

(Ideally, remove the `dummy` initializer, if it still matches.)

include/overlay024/pre_poketch_display.h
```diff
@@ -0,0 +1,10 @@
...
+#define POKEPLATINUM_PRE_POKETCH_DISPLAY_H
+
+#include "bg_window.h"
+
+void PrePoketchDisplay_DisplayBg(BgConfig *bgConfig);
+void PrePoketchDisplay_FreeBg(BgConfig *bgConfig);
+BOOL PrePoketchDisplay_AlwaysTrue(BgConfig *bgConfig);
```
lhearachel, line 8: **suggestions**:

- `PrePoketchSubscreen_Init`
- `PrePoketchSubscreen_Exit`
- `PrePoketchSubscreen_IsDone`

My logic here is that these routines are always used immediately after the overlay is loaded or immediately before it is unloaded. The latter two, especially, only ever seem to be called in the `Exit` -> `IsDone`, so we should imply some kind of relationship between them, however loosely.

src/overlay024/pre_poketch_display.c
```diff
@@ -0,0 +1,64 @@
...
+
+    Heap_Create(HEAP_ID_APPLICATION, HEAP_ID_POKETCH_APP, PRE_POKETCH_DISPLAY_HEAP_SIZE);
+
+    GXS_SetGraphicsMode(GX_BGMODE_0);
+    GX_SetBankForSubBG(GX_VRAM_SUB_BG_32_H);
+    GX_SetBankForSubOBJ(GX_VRAM_SUB_OBJ_16_I);
+    GXS_SetOBJVRamModeChar(GX_OBJVRAMMODE_CHAR_1D_32K);
+
+    Bg_InitFromTemplate(bgConfig, BG_LAYER_SUB_0, &template, BG_TYPE_STATIC);
```
lhearachel, line 42: **praise**: Thanks for labeling these routines with the right constants. :pray:

lhearachel (commented):
This looks good, thanks for digging into this! :+1: I have some collaborative thoughts below.

lhearachel (approved):

PR #382, opened 2025-02-14
Use a shell script for the final ROM target

lhearachel:
Verified that this `OK`s on my Arch and macOS machines, and that the dependency detection still works when modifying source files.


lhearachel:
Closing this in favor of the native `makerom` reimplementation that I'm working on.

PR #384, opened 2025-02-15
Document map prop material and shape

Kuruyia:
This documents the `ov5_021D56BC.c` file, which is in charge of reading the `build_model_matshp.dat` file from the ROM.

This file contains the material and shape IDs necessary to render the map props that use the 1Mat1Shp rendering path from NitroSystem.

---

This is a follow-up to [this comment](https://github.com/pret/pokeplatinum/pull/368#discussion_r1953128078) on #368.


Kuruyia:
As said in the commit message, this code reads the `build_model_matshp.dat` file from the ROM. Because we decided to use `MapProp` instead of the binary terminology, and because this PR is closely related to the map props system, the naming used here reflects that.

include/overlay005/map_prop_mat_shp.h
```diff
@@ -0,0 +1,28 @@
+#ifndef POKEPLATINUM_OV5_MAP_PROP_MATSHP_H
+#define POKEPLATINUM_OV5_MAP_PROP_MATSHP_H
+
+#include <nitro/types.h>
+
+typedef struct MapPropMatShpIdsLocator {
```
ravepossum, line 6: suggestion: Personally I think the tradeoff of clarity vs. brevity isn't worth it for `Shape` -> `Shp` so I would consider writing out `Shape` (even if `Shp` is what the SDK uses). If we're concerned about uniformity/grep-ability I could be convinced otherwise though. Would be curious to hear other opinions on this (including yours).

Additionally, I think we prefer `ID` over `Id` but we're very not uniform about that in the repo so don't consider that blocking at all.

include/overlay005/map_prop_mat_shp.h
```diff
@@ -0,0 +1,28 @@
...
+
+typedef struct MapPropMatShpIds {
+    u16 materialId;
+    u16 shapeId;
+} MapPropMatShpIds;
+
+typedef struct MapPropMatShp {
+    MapPropMatShpIdsLocator *matShpIdsLocators;
+    MapPropMatShpIds *matShpIds;
+} MapPropMatShp;
```
ravepossum, line 19: thought: Unless I'm missing something. this is quite a bit of abstraction for something with very little usage in the codebase. Serious GF moment.

src/overlay005/map_prop_mat_shp.c
```diff
@@ -0,0 +1,79 @@
...
+}
+
+void MapPropMatShp_Free(MapPropMatShp *mapPropMatShp)
+{
+    Heap_FreeToHeap(mapPropMatShp->matShpIds);
+    Heap_FreeToHeap(mapPropMatShp->matShpIdsLocators);
+    Heap_FreeToHeap(mapPropMatShp);
+}
+
+void MapPropMatShp_GetMatShpIdsCountOfMapProp(const int mapPropId, const MapPropMatShp *mapPropMatShp, u16 *matShpIdsCount)
```
ravepossum, line 65: thought: I wonder if this (and the function below) would be more concise and read better as something like `MapProp_GetMapShpIdsCount`. While it would be a little out of place with the other module prefixes here, we have done similar stuff elsewhere, and a name like that is much clearer to me.

src/overlay005/map_prop.c
```diff
@@ -155,21 +153,21 @@ void MapPropManager_Render(const VecFx32 *positionOffset, const UnkStruct_ov5_02
             position.z += positionOffset->z;
 
             if (sub_0201CED8(loadedProp->model, &position, &rotationMatrix, &loadedProp->scale)) {
-                const UnkStruct_ov5_021D5778 *v4;
-                u16 v5;
+                const MapPropMatShp *mapPropMatShp;
+                u16 matShpIdsCount;
 
-                v4 = ov5_021EFAB4(param1);
+                mapPropMatShp = ov5_021EFAB4(param1);
```
ravepossum, line 159: question: Does this match like this?
```c
  const MapPropMatShp *mapPropMatShp = ov5_021EFAB4(param1);
  u16 matShpIdsCount;
```

src/overlay005/map_prop.c
```diff
@@ -334,20 +332,20 @@ void MapPropManager_Render2(MapPropManager *mapPropManager, UnkStruct_ov5_021EF7
                 continue;
             }
 
-            const UnkStruct_ov5_021D5778 *v3;
-            u16 v4;
+            const MapPropMatShp *mapPropMatShp;
```
ravepossum, line 335: question: Same question as above - does this match with the declaration and assignment combined?

include/overlay005/map_prop_mat_shp.h
Kuruyia: > suggestion: Personally I think the tradeoff of clarity vs. brevity isn't worth it for `Shape` -> `Shp` so I would consider writing out `Shape` (even if `Shp` is what the SDK uses). If we're concerned about uniformity/grep-ability I could be convinced otherwise though. Would be curious to hear other opinions on this (including yours).

Oh, I totally agree with you. Myself as a newcomer I had a bit of a hard time understanding what `Shp` meant, so I'd be more than willing to replace that with `Shape`.
As you mentioned, I did it mainly for consistency with the SDK naming. I also don't know whether people are used to the `Shp` name or not at all. I'd value more feedback about this.

> Additionally, I think we prefer `ID` over `Id` but we're very not uniform about that in the repo so don't consider that blocking at all.

My bad, I'll correct that. I should have asked at some point.

include/overlay005/map_prop_mat_shp.h
Kuruyia: Well, actually it seems to be used quite a lot! Commenting [this line](https://github.com/pret/pokeplatinum/blob/ccdafdb554329e1439ad36e4c601f5cf70d11cb4/src/overlay005/map_prop.c#L275) and going into random cities (I tried Jubilife, Veilstone, Sunyshore), you can see a lot (not all) of buildings not rendering anymore. Looks like the render path with 1Mat1Shp is quite popular.

For the record, in `build_model_matshp.dat`, there are 590 `matShpIdsLocators`, so it's certainly as much as unique map props rendered with that. They probably did this structure to be able to share a single material/shape with multiple map props. I didn't take the time to check how effective that is in the end.
I won't judge whether this is over-engineering, I'll just say that it reminded me a lot about BDHC strips :)

src/overlay005/map_prop_mat_shp.c
Kuruyia: Sounds good to me. Do we want to do the same thing with `MapPropMatShp_GetMatShpIdsLocatorOfMapProp`?

src/overlay005/map_prop.c
Kuruyia: This does match, indeed

src/overlay005/map_prop.c
Kuruyia: As does this one

include/overlay005/map_prop_mat_shp.h
Gudf: > suggestion: Personally I think the tradeoff of clarity vs. brevity isn't worth it for `Shape` -> `Shp` so I would consider writing out `Shape` (even if `Shp` is what the SDK uses). If we're concerned about uniformity/grep-ability I could be convinced otherwise though. Would be curious to hear other opinions on this (including yours).

In my opinion this should also apply to `Mat`. While `Material` *is* significantly longer, I feel `Mat` might get confused for `Matrix`, especially in a 3D rendering context.

src/overlay005/map_prop_mat_shp.c
ravepossum: Yup, I think it would make sense for both.

include/overlay005/map_prop_mat_shp.h
ravepossum: Well, I more so meant having the 3 different structs and associated functionality for it, only to be called in a few places - but you're correct, I am indeed just the peanut gallery 20 years later without deadlines to hit :)

include/overlay005/map_prop_mat_shp.h
ravepossum: A couple additional thoughts based on what y'all have said:
- Is the `Map` part necessary to keep if we're trying to keep it concise? My gut says no, but then we lose the assocation with `MapProp` which doesn't seem ideal so I'm a little torn. I think maybe not but I thought I'd at least throw that out there.
- Would it make sense to keep the full names for the structs, but shorten any parameters or local variables? So we end up with something like:
```c
void Foo(const MapPropMaterialShape *mapPropMatShp);
```
or
```c
void Foo(const MapPropMaterialShape *propMatShp);
```
Could help cut down on verbosity a little bit.

lhearachel: Is there any reason we can't just call this a `MapPropResources` or some-such?

Kuruyia: > * Is the `Map` part necessary to keep if we're trying to keep it concise? My gut says no, but then we lose the assocation with `MapProp` which doesn't seem ideal so I'm a little torn. I think maybe not but I thought I'd at least throw that out there.

> Is there any reason we can't just call this a `MapPropResources` or some-such?

I'd like to have the name reflect that it is related to both map props and the 1Mat1Shp rendering in some way.

I'm worried that `MapPropResources` would be too generic, not related enough with the fact that those structs are used for rendering, and that it may make more sense for something later on to be called this name.

> * Would it make sense to keep the full names for the structs, but shorten any parameters or local variables?

Sure, makes total sense to me. Anyone that'd be confused by the `MatShp` in the name could refer to the full name of the struct.

lhearachel: > I'm worried that `MapPropResources` would be too generic, not related enough with the fact that those structs are used for rendering, and that it may make more sense for something later on to be called this name.

Good shout. I think `MapPropMaterialShape` for the struct and `propMatShp` for members/parameters sounds good. :grin:

Kuruyia: Thanks you all for your comments. Just pushed the changes that resulted from that.

ravepossum (approved):
Looks good to me!

PR #388, opened 2025-02-17
Poketch digital clock app

lucie-sharpe:
Documented OV26 which is the digital clock app on the poketch. Some more work could be done on the function `PoketchDigitalClock_CopyDigitTilemap` in `poketch_digital_clock_graphics.c`


src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,239 @@
...
+static void PoketchDigitalClock_SetAppTaskFunction(AppData *appData, u32 taskID);
+static BOOL PoketchDigitalClock_LoadApp(AppData *appData);
+static BOOL PoketchDigitalClock_UpdateApp(AppData *appData);
+static BOOL PoketchDigitalClock_UnloadApp(AppData *appData);
+
+static const TouchScreenHitTable digital_clock_buttons[] = {
+    { BACKLIGHT_BUTTON_MIN_Y, BACKLIGHT_BUTTON_MAX_Y, BACKLIGHT_BUTTON_MIN_X, BACKLIGHT_BUTTON_MAX_X }
+};
+
+static void NitroStaticInit(void)
```
cbt6, line 59: Is this function never used/called?

src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,239 @@
...
+    PoketchSystem_SetAppFunctions(PoketchDigitalClock_Init, PoketchDigitalClock_Shutdown);
+}
+
+static BOOL PoketchDigitalClock_Init(void **appData, PoketchSystem *poketchSys, BgConfig *bgConfig, u32 appID)
+{
+    AppData *newAppData = (AppData *)Heap_AllocFromHeap(HEAP_ID_POKETCH_APP, sizeof(AppData));
+
+    if (newAppData != NULL) {
+        if (PoketchDigitalClock_SetupAppData(newAppData, poketchSys, bgConfig, appID)) {
+            if (SysTask_Start(PoketchDigitalClock_AppTaskRunner, newAppData, 1) != NULL) {
```
cbt6, line 70:
```suggestion
        if (PoketchDigitalClock_SetupAppData(newAppData, poketchSys, bgConfig, appID)
        && SysTask_Start(PoketchDigitalClock_AppTaskRunner, newAppData, 1)) {
```

src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,239 @@
...
+    if (appData->appTasktaskID < NELEMS(funcArray)) {
+        PoketechSystem_UpdateButtonManager(appData->poketchSys, appData->buttonManager);
+
+        if (funcArray[appData->appTasktaskID](appData)) {
+            PoketchDigitalClock_FreeAppData(appData);
+            SysTask_Done(sysTask);
+            PoketchSystem_NotifyAppUnloaded(appData->poketchSys);
+        }
+    } else {
+    }
```
cbt6, line 144: Is this needed for matching or ... ?

src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,239 @@
...
+#define APP_TASK_LOAD_APP   0
+#define APP_TASK_UPDATE_APP 1
+#define APP_TASK_UNLOAD_APP 2
+
+typedef struct {
+    u8 appTasktaskID;
+    u8 taskStage;
+    u8 shutdown;
+    u8 unk_03;
+    u8 unk_04;
```
cbt6, line 30:
```suggestion
    u8 unused1; // value is set but never read from
    u8 unused2;
```

src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,239 @@
...
+    u8 hour;
+    u8 minute;
+    u8 backlightChange;
+    u32 unk_08;
+    u32 unk_0C;
```
cbt6, line 35:
```suggestion
    u8 unused3[8];
```

src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,239 @@
...
+#define BACKLIGHT_BUTTON_MIN_X 0x10
+#define BACKLIGHT_BUTTON_MAX_X 0xCF
+
+#define APP_TASK_LOAD_APP   0
+#define APP_TASK_UPDATE_APP 1
+#define APP_TASK_UNLOAD_APP 2
+
+typedef struct {
+    u8 appTasktaskID;
+    u8 taskStage;
```
cbt6, line 27:
```suggestion
    u8 state;
```
To be consistent across the repo

include/poketch_digital_clock/poketch_digital_clock_graphics.h
```diff
@@ -0,0 +1,24 @@
+#ifndef POKEPLATINUM_POKETCH_DIGITAL_CLOCK_GRAPHICS_H
+#define POKEPLATINUM_POKETCH_DIGITAL_CLOCK_GRAPHICS_H
+
+#include "bg_window.h"
+
+#define SYS_TASK_FUNC_SETUP_BACKGROUND         0
+#define SYS_TASK_FUNC_UPDATE_CLOCK_DIGITS      1
+#define SYS_TASK_FUNC_TOGGLE_BACKLIGHT_PALETTE 2
+#define SYS_TASK_FUNC_FREE_BACKGROUND          3
```
cbt6, line 9: Move this to the source file since they are only ever used there.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
```diff
@@ -0,0 +1,251 @@
...
+#define POKETCH_DIGIT_TILEMAP_WIDTH      (POKETCH_CLOCK_DIGIT_WIDTH * POKETCH_NUM_CLOCK_DIGITS)
+#define POKETCH_DIGIT_TILEMAP_HEIGHT     POKETCH_CLOCK_DIGIT_HEIGHT
+#define POKETCH_DIGIT_TILEMAP_SIZE_BYTES (POKETCH_DIGIT_TILEMAP_WIDTH * POKETCH_DIGIT_TILEMAP_HEIGHT)
+
+struct DisplayData_t {
+    const ClockData *clockData;
+    BgConfig *bgConfig;
+    u32 taskList[2 + NUM_TASK_SLOTS];
+    u16 digitsTilemap[POKETCH_DIGIT_TILEMAP_SIZE_BYTES];
+};
```
cbt6, line 41:
```suggestion
typedef struct DisplayData {
    const ClockData *clockData;
    BgConfig *bgConfig;
    u32 taskList[2 + NUM_TASK_SLOTS];
    u16 digitsTilemap[POKETCH_DIGIT_TILEMAP_SIZE_BYTES];
} DisplayData;
```
and define this in the header file instead and remove the typedef there.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
```diff
@@ -0,0 +1,251 @@
...
+#include "overlay025/ov25_02255090.h"
+
+#include "bg_window.h"
+#include "graphics.h"
+#include "heap.h"
+#include "sys_task_manager.h"
+
+#define NUM_TASK_SLOTS 8
+
+#define POKETCH_CLOCK_DIGITS_TILEMAP_IDX 25
```
cbt6, line 16: This should probably be defined somewhere else together for all indexes of files in `graphic/poketch.narc`, but idk where would be appropriate.

cbt6 (commented):
- Use "digital watch" instead of "digital clock" since that's what it's called in the game
- Remove outdated files (e.g. `overlay026/ov26_022561C0.c`, `overlay026/ov26_02256404.c`, etc.)

src/poketch_digital_clock/poketch_digital_clock_manager.c
lhearachel: There are a number of these functions that are scattered throughout the codebase. My understanding is that they are invoked when the overlay is loaded as an automatic process, but I don't remember where I read that.

lucie-sharpe: This loads the two entry point functions into the main code when the overlay is loaded. No other functions in ov26 are called from outside.

src/poketch_digital_clock/poketch_digital_clock_manager.c
lucie-sharpe: I can changed the nested if statement but left `!= NULL` as the function returns a pointer and explicitly the NULL macro on an error. Removing it implies a BOOL return. What do you think? It has no real effect on the code just interpretation.

include/poketch_digital_clock/poketch_digital_clock_graphics.h
lucie-sharpe: These are used in both `*_manager.c` and `*_graphics.c` so should be left here.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
lucie-sharpe: @lhearachel commented on #378 "These are fine for now, but we will want to replace them with an naix generated by re-packing this NARC once we unpack them into the resource tree."

I am not sure about the process of unpacking NARCs into the resource tree.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
lucie-sharpe: I can move this but then these macros will need to be moved into the header even though they are not used outside of `*_graphics.c`

```
#define POKETCH_NUM_CLOCK_DIGITS 10

#define POKETCH_CLOCK_DIGIT_WIDTH  4
#define POKETCH_CLOCK_DIGIT_HEIGHT 9

#define POKETCH_DIGIT_TILEMAP_WIDTH      (POKETCH_CLOCK_DIGIT_WIDTH * POKETCH_NUM_CLOCK_DIGITS)
#define POKETCH_DIGIT_TILEMAP_HEIGHT     (POKETCH_CLOCK_DIGIT_HEIGHT)
#define POKETCH_DIGIT_TILEMAP_SIZE_BYTES (POKETCH_DIGIT_TILEMAP_WIDTH * POKETCH_DIGIT_TILEMAP_HEIGHT)
```

src/poketch_digital_clock/poketch_digital_clock_manager.c
lhearachel: I don't mind the implied `BOOL` return, personally, if it still matches. This is a pretty common C construct.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
lhearachel: That's generally fine; you can define a sister-header for the constants, if you'd like. If you do, store it as `include/constants/poketch/digital_watch.h`.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
lhearachel: Unpacking is a bit of a process unto itself, but probably one that we should do sooner rather than later, since there are multiple hands touching the Poketch subsystem at the moment. A lot of it is going to be manual work and tinkering to figure out what values align with what using a previewer like Tinke or some-such.

platinum.us/main.lsf
```diff
@@ -878,11 +878,11 @@ Overlay overlay25
 	Object main.nef.p/src_overlay025_ov25_02255DBC.c.o
 }
 
-Overlay overlay26
+Overlay poketch_digital_clock
```
lhearachel, line 881: **thought**: I'm making a note for myself whenever I get back onto source-tree reorganization that the `poketch` is sufficiently large as to deserve its own top-level folder (similar to `field`, `battle`, and `contest`).

include/poketch_digital_clock/poketch_digital_clock_graphics.h
```diff
@@ -0,0 +1,24 @@
...
+
+typedef struct DisplayData_t DisplayData;
+
+typedef struct {
+    RTCTime time;
+    u32 backlightActive;
+} ClockData;
```
lhearachel, line 16: **question**: Presuming we rename this using `watch` instead of `clock` (which, IMO, we should), what do you think about naming this `WatchState`? Is that too synonymous with a `SysTask` state?

**polish**: Copy the name to the `struct` tag.

include/poketch_digital_clock/poketch_digital_clock_graphics.h
lhearachel: They could go in the aforementioned `constants` header, maybe, if that gets made.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
```diff
@@ -0,0 +1,249 @@
...
+#define PRE_POKETCH_DISPLAY_NARC_TILES_IDX   23
+#define PRE_POKETCH_DISPLAY_NARC_TILEMAP_IDX 24
+
+#define POKETCH_CLOCK_DIGIT_WIDTH     4
+#define POKETCH_CLOCK_DIGIT_HEIGHT    9
+#define POKETCH_CLOCK_DIGIT_GAP       1
+#define POKETCH_CLOCK_HOURS_DIGIT_X   3
+#define POKETCH_CLOCK_HOURS_DIGIT_Y   7
+#define POKETCH_CLOCK_MINUETS_DIGIT_X 15
+#define POKETCH_CLOCK_MINUETS_DIGIT_Y 7
```
lhearachel, line 28: **polish**: `MINUTES`

src/poketch_digital_clock/poketch_digital_clock_graphics.c
```diff
@@ -0,0 +1,249 @@
...
+#define POKETCH_DIGIT_TILEMAP_HEIGHT     (POKETCH_CLOCK_DIGIT_HEIGHT)
+#define POKETCH_DIGIT_TILEMAP_SIZE_BYTES (POKETCH_DIGIT_TILEMAP_WIDTH * POKETCH_DIGIT_TILEMAP_HEIGHT)
+
+static void PoketchDigitalClock_CopyDigitTilemap(const u16 *rawScreenData, u16 *dst);
+static void PoketchDigitalClock_EndTask(PoketchTaskManager *taskManager);
+static void PoketchDigitalClock_SetupBackground(SysTask *sysTask, void *taskManager);
+static void PoketchDigitalClock_UpdateClockDigits(SysTask *sysTask, void *taskManager);
+static void PoketchDigitalClock_ToggleBacklightPalette(SysTask *sysTask, void *taskManager);
+static void PoketchDigitalClock_FreeBackground(SysTask *sysTask, void *taskManager);
+static void PoketchDigitalClock_DrawClockDigits(DisplayData *displayData);
```
lhearachel, line 40: **note**: Feel free to drop the module-level prefix on `static` functions, if it makes the code semantics nicer.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
```diff
@@ -0,0 +1,249 @@
...
+struct DisplayData_t {
+    const ClockData *clockData;
+    BgConfig *bgConfig;
+    u32 taskList[2 + NUM_TASK_SLOTS];
+    u16 digitsTilemap[POKETCH_DIGIT_TILEMAP_SIZE_BYTES];
+};
+
+BOOL PoketchDigitalClock_SetupDisplayData(DisplayData **displayData, const ClockData *clockData, BgConfig *bgConfig)
+{
+    DisplayData *newDisplayData = (DisplayData *)Heap_AllocFromHeap(HEAP_ID_POKETCH_APP, sizeof(DisplayData));
```
lhearachel, line 51: **polish**: Remove the explicit cast; the compiler will handle casts from `void *` for you automatically.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
```diff
@@ -0,0 +1,249 @@
...
+    }
+}
+
+static const PoketchTask appTasks[] = {
+    { SYS_TASK_FUNC_SETUP_BACKGROUND, PoketchDigitalClock_SetupBackground, 0x0 },
+    { SYS_TASK_FUNC_UPDATE_CLOCK_DIGITS, PoketchDigitalClock_UpdateClockDigits, 0x0 },
+    { SYS_TASK_FUNC_TOGGLE_BACKLIGHT_PALETTE, PoketchDigitalClock_ToggleBacklightPalette, 0x0 },
+    { SYS_TASK_FUNC_FREE_BACKGROUND, PoketchDigitalClock_FreeBackground, 0x0 },
+    { 0x0, NULL, 0x0 }
+};
```
lhearachel, line 106: **polish**:

1. As a general rule, decompose hexadecimal values to decimal equivalents, unless referring to a type where a hexadecimal value is more expressive (e.g., a byte-offset or a bitmask value).
2. Static values should be prefixed with `s` (and global `extern`s similarly with `g`).

```suggestion
static const PoketchTask sAppTasks[] = {
    { SYS_TASK_FUNC_SETUP_BACKGROUND, PoketchDigitalClock_SetupBackground, 0 },
    { SYS_TASK_FUNC_UPDATE_CLOCK_DIGITS, PoketchDigitalClock_UpdateClockDigits, 0 },
    { SYS_TASK_FUNC_TOGGLE_BACKLIGHT_PALETTE, PoketchDigitalClock_ToggleBacklightPalette, 0 },
    { SYS_TASK_FUNC_FREE_BACKGROUND, PoketchDigitalClock_FreeBackground, 0 },
    { 0, NULL, 0 }
};
```

src/poketch_digital_clock/poketch_digital_clock_graphics.c
```diff
@@ -0,0 +1,249 @@
...
+    Bg_CopyToTilemapRect(displayData->bgConfig,
+        BG_LAYER_SUB_2,
+        POKETCH_CLOCK_HOURS_DIGIT_X,
+        POKETCH_CLOCK_HOURS_DIGIT_Y,
+        POKETCH_CLOCK_DIGIT_WIDTH,
+        POKETCH_CLOCK_DIGIT_HEIGHT,
+        displayData->digitsTilemap,
+        POKETCH_CLOCK_DIGIT_WIDTH * tensDigitOffset,
+        0,
+        (POKETCH_CLOCK_DIGIT_WIDTH * POKETCH_NUM_CLOCK_DIGITS),
```
lhearachel, line 208: **polish**: Remove unnecessary parens.

src/poketch_digital_clock/poketch_digital_clock_graphics.c
```diff
@@ -0,0 +1,249 @@
+#include "poketch_digital_clock/poketch_digital_clock_graphics.h"
```
lhearachel, line 1: **question**: Could this module be more concisely named? e.g., `poketch_digital_watch/display.c`

src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,236 @@
+#include <nitro.h>
```
lhearachel, line 1: **question**: Similarly, could we tighten this name? e.g., `poketch_digital_watch/main.c`?

src/poketch_digital_clock/poketch_digital_clock_manager.c
lhearachel: Yep, that tracks to me, since the apps are (sub?)subsystems of the parent Poketch subsystem.

src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,236 @@
...
+};
+
+static void NitroStaticInit(void)
+{
+    PoketchSystem_SetAppFunctions(PoketchDigitalClock_Init, PoketchDigitalClock_Shutdown);
+}
+
+static BOOL PoketchDigitalClock_Init(void **appData, PoketchSystem *poketchSys, BgConfig *bgConfig, u32 appID)
+{
+    AppData *newAppData = (AppData *)Heap_AllocFromHeap(HEAP_ID_POKETCH_APP, sizeof(AppData));
```
lhearachel, line 65: **polish**: Remove explicit cast.

src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,236 @@
...
+}
+
+static void PoketchDigitalClock_FreeAppData(AppData *appData)
+{
+    PoketchDigitalClock_FreeDisplayData(appData->displayData);
+    PoketchButtonManager_Free(appData->buttonManager);
+    Heap_FreeToHeap(appData);
+}
+
+static void PoketchDigitalClock_AppTaskRunner(SysTask *sysTask, void *callbackData)
```
lhearachel, line 123: **suggestion**: I think that this function is important enough that it is worthy of bearing the `Main` designation, since it is what controls the app's loop. It would also align the function-sequence with the paradigm used by `OverlayTemplate` (`Init`, `Main`, and `Exit`). What do you think?

src/poketch_digital_clock/poketch_digital_clock_manager.c
```diff
@@ -0,0 +1,236 @@
...
+
+        if (funcArray[appData->appTasktaskID](appData)) {
+            PoketchDigitalClock_FreeAppData(appData);
+            SysTask_Done(sysTask);
+            PoketchSystem_NotifyAppUnloaded(appData->poketchSys);
+        }
+    }
+}
+
+static void PoketchDigitalClock_Shutdown(void *appData)
```
lhearachel, line 144: **suggestion**: Rename this to `Exit`, (prefix optional) to align it with the usual paradigm for `OverlayTemplate`s.

lucie-sharpe:
I have made the changes suggested. I did not feel a constants file was needed so i have just used the local header file. Tell me what you think of naming conventions?

platinum.us/main.lsf
```diff
@@ -881,8 +881,8 @@ Overlay overlay25
 Overlay poketch_digital_clock
```
lhearachel, line 881: **polish**: Rename the overlay, too, just to synonymize it with the code folder.

lhearachel (approved):
Just one little quibble below; otherwise looks great!

PR #389, opened 2025-02-18
Document `ov5_021DC018` -> `field_menu`

Gudf:
#325: multi-choice menus

Not much to say at this point, quite similar structure to what's in `overlay_005/scrcmd_move_tutors`, except for the script commands themselves that are in `scrcmd` instead.

This also handles a number of specific menus and currency displays, it seems. Might be worth looking into splitting the module?

Text bank `unk_0543` contains strings related to merchants/traders, I'll probably rename it to `merchants` if that's fine.

As always, name suggestions are welcome!


include/overlay005/ov5_021DC018.h
```diff
@@ -8,18 +8,18 @@
...
-UnkStruct_ov5_021DC1A4 *ov5_021DC150(FieldSystem *fieldSystem, u8 param1, u8 param2, u8 param3, u8 param4, u16 *param5, StringTemplate *param6, Window *param7, MessageLoader *param8);
-void ov5_021DC1A4(UnkStruct_ov5_021DC1A4 *param0, u32 param1, u32 param2);
-void ov5_021DC1AC(UnkStruct_ov5_021DC1A4 *param0);
-void ov5_021DC424(UnkStruct_ov5_021DC1A4 *param0);
-UnkStruct_ov5_021DC1A4 *ov5_021DC48C(FieldSystem *fieldSystem, u8 param1, u8 param2, u8 param3, u8 param4, u16 *param5, StringTemplate *param6, Window *param7, MessageLoader *param8);
-void ov5_021DC4B0(UnkStruct_ov5_021DC1A4 *param0, u32 param1, u32 param2, u32 param3);
-void ov5_021DC4B8(UnkStruct_ov5_021DC1A4 *param0);
-void ov5_021DC528(UnkStruct_ov5_021DC1A4 *param0, u16 param1);
-void ov5_021DC600(UnkStruct_ov5_021DC1A4 *param0, u16 *param1, u16 *param2);
+GenericMenuManager *ov5_021DC150(FieldSystem *fieldSystem, u8 param1, u8 param2, u8 param3, u8 param4, u16 *param5, StringTemplate *param6, Window *param7, MessageLoader *param8);
```
lhearachel, line 11: **suggestion**: Could this be `FieldMenuManager`? Or is that too synonymous with the existing `FieldMenu`? (the latter should probably be `StartMenu`...)

src/overlay005/ov5_021DC018.c
```diff
@@ -27,152 +32,137 @@
 #include "unk_0200F174.h"
 #include "unk_0202D05C.h"
 
-struct UnkStruct_ov5_021DC1A4_t {
+#include "res/text/bank/common_strings_2.h"
+#include "res/text/bank/unk_0543.h"
+
+#define GENERIC_MENU_ENTRIES_MAX 28
+
+struct GenericMenuManager_t {
```
lhearachel, line 40: **note**: Make sure to pull this up into the owning header.

lhearachel (commented):
I don't have much to comment on here, since much of this is just basic renaming.

include/overlay005/ov5_021DC018.h
Gudf: `FieldMenuManager` would probably be a better name, but yeah I fear it'd be confusing unless we rename `FieldMenu`. I am unsure about naming it `StartMenu` though, since that sounds like it's the menu just past the title screen IMO.

Gudf: As discussed on the pret Discord, renaming `FieldMenu` is out-of-scope for this PR. I'll still go ahead and rename `GenericMenuManager` here to `FieldMenuManager`, and rebase after the `FieldMenu` -> `StartMenu` rename.

Gudf:
I'm unsure how I should name the module... `field_menu_manager` wouldn't make sense with the elevator floor and currencies windows, but I think `field_menu`(or `field_menus`) might be confusing if I then go and name the functions acting on the struct `FieldMenu(s)_Foo`. 
Would naming the module `field_menu(s)` while keeping these functions prefixed with the struct name be fine? Then we would have `FieldMenu(s)_ShowMoneyWindow` or similar along the `FieldMenuManager_*` functions.

ravepossum:
> I'm unsure how I should name the module... `field_menu_manager` wouldn't make sense with the elevator floor and currencies windows, but I think `field_menu`(or `field_menus`) might be confusing if I then go and name the functions acting on the struct `FieldMenu(s)_Foo`. Would naming the module `field_menu(s)` while keeping these functions prefixed with the struct name be fine? Then we would have `FieldMenu(s)_ShowMoneyWindow` or similar along the `FieldMenuManager_*` functions.

I've definitely intermingled prefixes within a module in several places, and I feel like the repo generally makes sense with the struct prefixes given how pseudo-OOP a lot of it is. So I don't see any issue with having function prefixes that don't match the module if it would be clearer that way. The PR I'm working on right now even does something like that.

lhearachel (commented):
Excellent work! :pray: Thanks for sticking this one out. I have a few suggestions for potential improvement, but all are optional.

src/overlay005/field_menu.c
```diff
@@ -0,0 +1,862 @@
...
+
+    if (menuManager->anchorRight) {
+        menuManager->anchorX -= menuWidth;
+    }
+
+    if (menuManager->anchorBottom) {
+        menuManager->anchorY -= menuManager->optionsCount * 2;
+    }
+
+    Window_Add(menuManager->fieldSystem->bgConfig, &menuManager->menuWindow, BG_LAYER_MAIN_3, menuManager->anchorX, menuManager->anchorY, menuWidth, menuManager->optionsCount * 2, 13, ((1 + (10 * 4)) + (10 * 2)));
```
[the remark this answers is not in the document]
Gudf: I think it'd make more sense to use an enum, since I'd expect at least some of them to have a consistent usage? I'll have a look.

src/overlay005/field_menu.c
```diff
@@ -0,0 +1,862 @@
...
+    {
+        Strbuf *fmtString = Strbuf_Init((40 * 2), HEAP_ID_FIELD);
+
+        MessageLoader_GetStrbuf(menuManager->messageLoader, entryID, fmtString);
+        StringTemplate_Format(menuManager->stringTemplate, menuManager->choicesStringsBuffers[menuManager->optionsCount], fmtString);
+        menuManager->listMenuChoicesStrings[menuManager->optionsCount].entry = (const void *)menuManager->choicesStringsBuffers[menuManager->optionsCount];
+        Strbuf_Free(fmtString);
+    }
+
+    if (entryIndex == 0xfa) {
```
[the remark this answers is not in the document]
Gudf: Ah I already encountered that in `scrcmd_move_tutors` and couldn't figure out what it was, but I just looked through the usages of `ScrCmd_046` (the script command used to add entries to the menu in scripts), and apparently that's the index they use to signal a header, except it's written as `250` in scripts — and the 0xfffffffd on the line below is in fact `LIST_HEADER` (-3) from `list_menu.h`. I'll go through the script commands that use this module and their usages next, so I'll just add a constant to `constants/scrcmd.h` for this for now, as something like `LIST_MENU_BUILDER_HEADER`.

src/overlay005/field_menu.c
```diff
@@ -0,0 +1,862 @@
...
+    FieldMenuManager *menuManager = FieldMenuManager_New(fieldSystem, anchorX, anchorY, 0, FALSE, selectedOptionPtr, stringTemplate, NULL, NULL);
+
+    Window_Add(menuManager->fieldSystem->bgConfig, &menuManager->menuWindow, BG_LAYER_MAIN_3, menuManager->anchorX, menuManager->anchorY, SHARD_COST_WINDOW_WIDTH, 16, 13, ((1 + (10 * 4)) + (10 * 2)));
+    LoadStandardWindowGraphics(menuManager->fieldSystem->bgConfig, BG_LAYER_MAIN_3, 1024 - (18 + 12) - 9, 11, STANDARD_WINDOW_SYSTEM, HEAP_ID_FIELD);
+    Window_DrawStandardFrame(&menuManager->menuWindow, TRUE, 1024 - (18 + 12) - 9, 11);
+    Window_FillRectWithColor(&menuManager->menuWindow, 15, 0, 0, (SHARD_COST_WINDOW_WIDTH * 8), (16 * 8));
+
+    FieldMenuManager_PrintString(menuManager, pl_msg_00000361_00273, 0, 0);
+    StringTemplate_SetNumber(stringTemplate, 0, redCost, 3, PADDING_MODE_SPACES, CHARSET_MODE_EN);
+    FieldMenuManager_PrintString(menuManager, pl_msg_00000361_00277, 0, (0 + 16));
```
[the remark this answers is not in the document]
Gudf: Good catch. I figured this was some sort of line height but didn't know how it related to the glyphs' height itself.

lhearachel (approved):

PR #399, opened 2025-02-25
Use various constants in field events

cbt6:


lhearachel (approved):
I am going to trust that the constants are correct based on the status of the build. The code changes look fine to me.

PR #400, opened 2025-02-25
Document model attributes

Kuruyia:
This documents the `ov5_021D5878.c` file, which manages a data structure containing various attributes used when rendering 3D models.

Those attributes directly map to polygon attributes/parameters that are exposed by the NDS 3D hardware.


include/overlay005/model_attributes.h
```diff
@@ -0,0 +1,89 @@
...
+    BOOL enableSpecularReflectShininessTable;
+    int lightEnableFlags;
+    GXPolygonMode polygonMode;
+    GXCull cullMode;
+    int polygonID;
+    int alpha;
+    int miscFlags;
+} ModelAttributes;
+
+enum ModelAttributesShift {
```
ravepossum, line 26: suggestion: I feel like these constants (and the ones in `enum ModelAttributesBit` below) are too verbose, and there's a bit of redundancy as well. I wonder if we need to specify that it's a model *attribute* in the constants themselves at all, and if we need to specify that the attribute bits are bits either. Here's a few examples of what I mean by that:

```diff
- MODEL_ATTRIBUTES_LIGHT_VECTORS_ATTR_SHIFT
+ MODEL_LIGHT_VECTORS_SHIFT
```
```diff
- MODEL_ATTRIBUTES_LIGHT_VECTOR_1_ATTR_BIT
+ MODEL_LIGHT_VECTOR_1
```

include/overlay005/model_attributes.h
```diff
@@ -0,0 +1,89 @@
...
+typedef struct ModelAttributes {
+    VecFx16 lightVectors[GX_LIGHTS_COUNT];
+    GXRgb lightColors[GX_LIGHTS_COUNT];
+    GXRgb diffuseReflectColor;
+    GXRgb ambientReflectColor;
+    GXRgb specularReflectColor;
+    GXRgb emissionColor;
+    BOOL diffuseReflectColorAsVertexColor;
+    BOOL enableSpecularReflectShininessTable;
+    int lightEnableFlags;
```
ravepossum, line 18: thought: I wonder if it might make more sense to explicitly call this a `Mask` instead of `Flags`. I also think specifying light*s* plural makes what it does a bit more clear. Something like `enabledLightsMask` maybe?

include/overlay005/model_attributes.h
```diff
@@ -0,0 +1,89 @@
...
+#include "constants/gx.h"
+
+typedef struct ModelAttributes {
+    VecFx16 lightVectors[GX_LIGHTS_COUNT];
+    GXRgb lightColors[GX_LIGHTS_COUNT];
+    GXRgb diffuseReflectColor;
+    GXRgb ambientReflectColor;
+    GXRgb specularReflectColor;
+    GXRgb emissionColor;
+    BOOL diffuseReflectColorAsVertexColor;
```
ravepossum, line 16: suggestion: Maybe something like `setDiffuseColorAsVertexColor`? I don't think it's the end of the world if this one has to be a little verbose to be clear, but I do think I like it better with an action verb at the beginning.

include/overlay005/model_attributes.h
```diff
@@ -0,0 +1,89 @@
...
+    BOOL enableSpecularReflectShininessTable;
```
ravepossum, line 17: question: Is this just as clear without `Table` at the end?

src/overlay005/model_attributes.c
```diff
@@ -0,0 +1,208 @@
...
+    return modelAttrs;
+}
+
+void ModelAttributes_Free(ModelAttributes **modelAttrs)
+{
+    Heap_FreeToHeapExplicit(HEAP_ID_FIELD, *modelAttrs);
+    *modelAttrs = NULL;
+}
+
+void ModelAttributes_ApplyToGlobal(ModelAttributes *modelAttrs, int attributes)
```
ravepossum, line 25: nitpick: Admittedly this is splitting hairs, but I wonder if this comes across a little bit more elegantly as `ModelAttributes_ApplyGlobal`.

src/overlay005/model_attributes.c
```diff
@@ -0,0 +1,208 @@
...
+    if (attributes & MODEL_ATTRIBUTES_MISC_FAR_CLIPPING_ATTR_BIT) {
+        NNS_G3dMdlSetMdlFarClipAll(model, modelAttrs->miscFlags & GX_POLYGON_ATTR_MISC_FAR_CLIPPING);
+    }
+
+    if (attributes & MODEL_ATTRIBUTES_MISC_XLU_DEPTH_UPDATE_ATTR_BIT) {
+        NNS_G3dMdlSetMdlXLDepthUpdateAll(model, modelAttrs->miscFlags & GX_POLYGON_ATTR_MISC_XLU_DEPTH_UPDATE);
+    }
+}
+
+void ModelAttributes_SetLightVector(ModelAttributes *modelAttrs, int lightID, fx16 x, fx16 y, fx16 z)
```
ravepossum, line 111: suggestion: I tend to prefer patterns like this as `lightIndex` (or even simply `index` since light is in the function name). My thinking is that it distinguishes it from things like `speciesID` that refer to a unique constant value rather than a "slot" of sorts.

src/overlay005/model_attributes.c
```diff
@@ -0,0 +1,208 @@
...
+    NNS_G3dGlbLightVector(lightID, modelAttrs->lightVectors[lightID].x, modelAttrs->lightVectors[lightID].y, modelAttrs->lightVectors[lightID].z);
+}
+
+void ModelAttributes_SetLightColor(ModelAttributes *modelAttrs, int lightID, GXRgb lightColor)
+{
+    modelAttrs->lightColors[lightID] = lightColor;
+    NNS_G3dGlbLightColor(lightID, modelAttrs->lightColors[lightID]);
+}
+
+void ModelAttributes_SetDiffuseReflection(ModelAttributes *modelAttrs, GXRgb diffuseReflectColor, BOOL diffuseReflectColorAsVertexColor, BOOL applyToGlobal)
```
ravepossum, line 126: Same thoughts on the naming of the parameters here as on the struct member for `diffuseReflectColorAsVertexColor` and `applyToGlobal` above (and in the other similar instances below).

src/overlay005/fieldmap.c
```diff
@@ -100,7 +99,7 @@ static void ov5_021D1444(BgConfig *bgl);
 static void ov5_021D1524(BgConfig *bgl);
 static void ov5_021D154C(void);
 static void ov5_021D1570(void);
-static void ov5_021D1578(UnkStruct_ov5_021D5894 *param0);
+static void FieldMap_InitModelAttributes(ModelAttributes *param0);
```
ravepossum, line 102: polish: `*modelAttributes` to match function definition (although I think maybe using `modelAttrs` to match the other functions may be nice for consistency. I'm not sure it matters enough to fret about though).

ravepossum (commented):
Re: the GBATEK mappings, I'm not sure we need to include those in the code itself, but I think it could be valuable info for the in-repo docs like the existing 2D and 3D rendering ones.

include/constants/gx.h
```diff
@@ -0,0 +1,6 @@
+#ifndef POKEPLATINUM_CONSTANTS_GX_H
+#define POKEPLATINUM_CONSTANTS_GX_H
+
+#define GX_LIGHTS_COUNT 4
```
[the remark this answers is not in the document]
Kuruyia: As it is not a `model_attributes`-specific constant, I wanted that to live in its own constants file in case some other place in the pokeplatinum code would need it as well. (imo, this should/could have been something provided by the Nitro SDK, but couldn't find anything)

include/overlay005/model_attributes.h
```diff
@@ -0,0 +1,89 @@
...
+    MODEL_ATTRIBUTES_CULL_MODE_ATTR_SHIFT = 14,
+    MODEL_ATTRIBUTES_POLYGON_ID_ATTR_SHIFT = 15,
+    MODEL_ATTRIBUTES_ALPHA_ATTR_SHIFT = 16,
+    MODEL_ATTRIBUTES_MISC_FOG_ATTR_SHIFT = 17,
+    MODEL_ATTRIBUTES_MISC_DEPTHTEST_DECAL_ATTR_SHIFT = 18,
+    MODEL_ATTRIBUTES_MISC_DISP_1DOT_ATTR_SHIFT = 19,
+    MODEL_ATTRIBUTES_MISC_FAR_CLIPPING_ATTR_SHIFT = 20,
+    MODEL_ATTRIBUTES_MISC_XLU_DEPTH_UPDATE_ATTR_SHIFT = 21,
+    MODEL_ATTRIBUTES_ATTR_LAST = 22
+};
```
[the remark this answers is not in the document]
Kuruyia: No specific reason; only forgot to clean that up

include/overlay005/model_attributes.h
Kuruyia: I can only answer that in terms of meaning in plain English; to me, that'd change the meaning of this member var.

I'd understand `enableSpecularReflectShininess` something along the lines of "whether to enable shininess for specular reflection", whereas the meaning I'd like to convey is something like "whether to use custom shininess values for specular reflection".

src/overlay005/model_attributes.c
Kuruyia: Agreed, I just used the SDK name here.

include/overlay005/model_attributes.h
ravepossum: I don't see an issue with keeping the `ATTRIBUTES` on this one, but just in general I don't think we need to adhere to strictly unique prefixes for constants as long as there's no overlap between them. I.E. just because we prefix these with `MODEL` doesn't mean we can't prefix another set of constants with `MODEL` as long as they don't both have a value it would make sense to call (for example) `MODEL_LIGHT_VECTOR_1`. Curious to hear others' thoughts on this though.

lhearachel: Agree; simplicity in naming is valuable.

include/overlay005/model_attributes.h
```diff
@@ -0,0 +1,89 @@
+#ifndef POKEPLATINUM_OV5_MODEL_H
+#define POKEPLATINUM_OV5_MODEL_H
```
ravepossum, line 2: polish: Looks like your enum rename snagged the header guards.

include/overlay005/model_attributes.h
```diff
@@ -0,0 +1,89 @@
...
+    MODEL_POLYGON_MODE_SHIFT,
+    MODEL_CULL_MODE_SHIFT,
+    MODEL_POLYGON_ID_SHIFT,
+    MODEL_ALPHA_SHIFT,
+    MODEL_MISC_FOG_SHIFT,
+    MODEL_MISC_DEPTHTEST_DECAL_SHIFT,
+    MODEL_MISC_DISP_1DOT_SHIFT,
+    MODEL_MISC_FAR_CLIPPING_SHIFT,
+    MODEL_MISC_XLU_DEPTH_UPDATE_SHIFT,
+    MODEL_LAST_SHIFT
```
ravepossum, line 44: suggestion: Sorry for not specifying this earlier but I think this particular entry (and the corresponding `Bit` enum entry) makes more sense with `ATTRIBUTE` included in the name since it's very generic, and `LAST_SHIFT` specifically refers to the *attributes* last shift - it's not really an attribute in an of itself.

ravepossum (commented):
Couple of last little things then this should be good to go :)

Kuruyia:
Oops, I didn't check what changed after fixing the prefixes thinking it was a small change. Should be good now :)

ravepossum (approved):

PR #401, opened 2025-02-26
Document Pokemon Storage System

CharlesFolz4:
Most functions of the box system in unk_020797C8 have been documented, but ran into blockers on figuring out the rest, so creating this as a draft PR.


src/unk_020797C8.c
```diff
@@ -3,24 +3,26 @@
...
 typedef struct PCBoxes {
-    u32 unk_00;
-    BoxPokemon unk_04[18][30];
-    u16 unk_10E4[18][20];
+    u32 currentBoxIndex;
+    BoxPokemon boxedPokemonArray[MAX_PC_BOXES][30];
+    u16 boxNames[MAX_PC_BOXES][20];
     u8 unk_13B4[18];
-    u8 unk_13C6;
+    u8 unk_13C6; // This seems like some sort of bit flag? but I don't know for what
```
ravepossum, line 18: note: I think this may be a bitmask that represents the secret wallpapers the player has unlocked.

src/unk_020797C8.c
```diff
@@ -29,94 +31,95 @@ u32 PCBoxes_SaveSize(void)
...
+    for (boxIndex = 0; boxIndex < MAX_PC_BOXES; boxIndex++) {
+        for (monPosInBox = 0; monPosInBox < MAX_MONS_PER_BOX; monPosInBox++) {
+            BoxPokemon_Init(&(pcBoxes->boxedPokemonArray[boxIndex][monPosInBox]));
         }
     }
 
-    for (v0 = 0, v1 = 0; v0 < 18; v0++) {
-        param0->unk_13B4[v0] = v1++;
+    // Pretty sure that monPosInBox isn't being used in the same way here as in the previous loop,
+    // so the variable's name no longer accurately reflects what it's doing in this loop
```
ravepossum, line 47: suggestion: Don't be afraid to use names like `i` and `j` for loop counters, especially in these sorts of scenarios where one single name may not be apt. Well documented constants can fill in some of the blanks - e.g. if you're looping until `MAX_PC_BOXES` it's pretty clear the counter represents a box index.

src/unk_020797C8.c
```diff
@@ -29,94 +31,95 @@ u32 PCBoxes_SaveSize(void)
...
 }
 
-static void sub_020797DC(PCBoxes *param0)
+// InitPokemonStorageSystem?
+static void sub_020797DC(PCBoxes *pcBoxes)
 {
-    u32 v0, v1;
-    MessageLoader *v2;
+    u32 boxIndex, monPosInBox;
+    MessageLoader *messageLoader;
```
ravepossum, line 38: polish: Where possible we're cleaning up variables declared C89 style (at the top of the scope) to instead be declared as needed as long as it still matches. I tried it myself in this instance - if you declare `messageLoader` as it's initialized below, the function still matches.

src/unk_020797C8.c
```diff
@@ -29,94 +31,95 @@ u32 PCBoxes_SaveSize(void)
...
 
-    for (v0 = 0; v0 < 18; v0++) {
-        for (v1 = 0; v1 < (5 * 6); v1++) {
-            BoxPokemon_Init(&(param0->unk_04[v0][v1]));
+    for (boxIndex = 0; boxIndex < MAX_PC_BOXES; boxIndex++) {
+        for (monPosInBox = 0; monPosInBox < MAX_MONS_PER_BOX; monPosInBox++) {
+            BoxPokemon_Init(&(pcBoxes->boxedPokemonArray[boxIndex][monPosInBox]));
```
ravepossum, line 42: polish: We could remove the extra parens here:
```diff
-            BoxPokemon_Init(&(pcBoxes->boxedPokemonArray[boxIndex][monPosInBox]));
+            BoxPokemon_Init(&pcBoxes->boxedPokemonArray[boxIndex][monPosInBox]);
```
(unless we think it diminishes the clarity)

src/unk_020797C8.c
```diff
@@ -29,94 +31,95 @@ u32 PCBoxes_SaveSize(void)
...
-    v2 = MessageLoader_Init(1, 26, 18, 0);
+    pcBoxes->unk_13C6 = 0;
+    messageLoader = MessageLoader_Init(MESSAGE_LOADER_NARC_HANDLE, NARC_INDEX_MSGDATA__PL_MSG, TEXT_BANK_POKEMON_STORAGE_SYSTEM, 0);
 
-    if (v2) {
-        for (v0 = 0; v0 < 18; v0++) {
-            MessageLoader_Get(v2, 6 + v0, param0->unk_10E4[v0]);
+    if (messageLoader) {
+        for (boxIndex = 0; boxIndex < MAX_PC_BOXES; boxIndex++) {
+            MessageLoader_Get(messageLoader, BOX1_TEXT_BANK_ENTRYID + boxIndex, pcBoxes->boxNames[boxIndex]);
```
ravepossum, line 61: suggestion: We should probably use the generated constant for this text entry instead of creating a preprocessor one. (`pl_msg_00000018_00006` as of right now, but presumably you would document it and the other related ones)

src/unk_020797C8.c
```diff
@@ -29,94 +31,95 @@ u32 PCBoxes_SaveSize(void)
...
-    for (v0 = 0, v1 = 0; v0 < 18; v0++) {
-        param0->unk_13B4[v0] = v1++;
+    // Pretty sure that monPosInBox isn't being used in the same way here as in the previous loop,
+    // so the variable's name no longer accurately reflects what it's doing in this loop
+    for (boxIndex = 0, monPosInBox = 0; boxIndex < MAX_PC_BOXES; boxIndex++) {
+        pcBoxes->unk_13B4[boxIndex] = monPosInBox++;
 
-        if (v1 >= 16) {
-            v1 = 0;
+        if (monPosInBox >= 16) {
```
ravepossum, line 51: note: There's 16 unique standard wallpapers - I think `unk_13B4` may represent the wallpaper assigned to a particular box (if so, then the conditional loops it back around to ensure only valid wallpapers get assigned).

src/unk_020797C8.c
```diff
@@ -29,94 +31,95 @@ u32 PCBoxes_SaveSize(void)
...
-        MessageLoader_Free(v2);
+        MessageLoader_Free(messageLoader);
     }
 
-    param0->unk_00 = 0;
+    pcBoxes->currentBoxIndex = 0;
 }
 
-BOOL sub_02079868(PCBoxes *param0, BoxPokemon *param1)
+BOOL StorePokemonInPc(PCBoxes *pcBoxes, BoxPokemon *boxedPokemon)
```
ravepossum, line 70: suggestion: For functions like this that have a chance of failure and indicate success via the returned boolean, I like to have `Try` in the name to make that obvious from the name alone.

src/unk_020797C8.c
```diff
@@ -29,94 +31,95 @@ u32 PCBoxes_SaveSize(void)
...
         }
-    } while (v0 != param0->unk_00);
+    } while (boxIndex != pcBoxes->currentBoxIndex);
 
-    return 0;
+    return FALSE;
 }
 
-BOOL sub_020798A0(PCBoxes *param0, u32 param1, BoxPokemon *param2)
+BOOL StorePokemonInBox(PCBoxes *pcBoxes, u32 boxIndex, BoxPokemon *boxedPokemon)
```
ravepossum, line 90: suggestion: I think `box` probably suffices for the index parameter here, but I don't feel strongly about that.

src/unk_020797C8.c
```diff
@@ -29,94 +31,95 @@ u32 PCBoxes_SaveSize(void)
...
 }
 
-BOOL sub_02079914(PCBoxes *param0, u32 param1, u32 param2, BoxPokemon *param3)
+BOOL StorePokemonInPcAt(PCBoxes *pcBoxes, u32 param1, u32 param2, BoxPokemon *param3)
 {
-    sub_0207896C(param3);
+    RestoreBoxedPokemonPP(param3);
     BoxPokemon_SetShayminForm(param3, 0);
 
     if (param1 == 0xffffffff) {
```
ravepossum, line 117: polish: `-1` as you did in the function above. `param1` also looks like a box index.

src/unk_020797C8.c
```diff
@@ -126,287 +129,287 @@ BOOL sub_02079914(PCBoxes *param0, u32 param1, u32 param2, BoxPokemon *param3)
...
 {
-    if (param1 == 0xffffffff) {
-        param1 = param0->unk_00;
+    if (boxIndex == -1) {
+        boxIndex = pcBoxes->currentBoxIndex;
     }
 
-    if ((param2 < (5 * 6)) && (param1 < 18)) {
-        BoxPokemon_Init(&(param0->unk_04[param1][param2]));
+    if ((monPosInBox < MAX_MONS_PER_BOX) && (boxIndex < MAX_PC_BOXES)) {
```
ravepossum, line 138: polish: The extra parens here can be dropped:
```diff
- if ((monPosInBox < MAX_MONS_PER_BOX) && (boxIndex < MAX_PC_BOXES)) {
+ if (monPosInBox < MAX_MONS_PER_BOX && boxIndex < MAX_PC_BOXES) {
```

src/unk_020797C8.c
```diff
@@ -126,287 +129,287 @@ BOOL sub_02079914(PCBoxes *param0, u32 param1, u32 param2, BoxPokemon *param3)
...
 
-    v0 = param0->unk_00;
+    boxIndex = pcBoxes->currentBoxIndex;
 
     while (TRUE) {
-        for (v1 = 0; v1 < (5 * 6); v1++) {
-            if (BoxPokemon_GetValue((BoxPokemon *)(&(param0->unk_04[v0][v1])), MON_DATA_SPECIES_EXISTS, NULL) == 0) {
-                return v0;
+        for (monPosInBox = 0; monPosInBox < MAX_MONS_PER_BOX; monPosInBox++) {
+            if (BoxPokemon_GetValue((BoxPokemon *)(&(pcBoxes->boxedPokemonArray[boxIndex][monPosInBox])), MON_DATA_SPECIES_EXISTS, NULL) == 0) {
```
ravepossum, line 159: polish: We can remove the unnecessary cast here.

src/unk_020797C8.c
```diff
@@ -126,287 +129,287 @@ BOOL sub_02079914(PCBoxes *param0, u32 param1, u32 param2, BoxPokemon *param3)
...
-        param0->unk_13B4[param1] = param2;
+        pcBoxes->unk_13B4[boxIndex] = param2;
         SaveData_SetFullSaveRequired();
     } else {
         GF_ASSERT(0);
     }
 }
 
-void sub_02079AF4(const PCBoxes *param0, u32 param1, Strbuf *param2)
+void CopyBoxName(const PCBoxes *pcBoxes, u32 boxIndex, Strbuf *copyDestination)
```
ravepossum, line 265: suggestion: Something like `destBuf` or even `dst`/`dest` is probably fine for the `Strbuf` parameter here.

src/unk_020797C8.c
```diff
@@ -126,287 +129,287 @@ BOOL sub_02079914(PCBoxes *param0, u32 param1, u32 param2, BoxPokemon *param3)
...
+    if (boxIndex == -1) {
+        boxIndex = pcBoxes->currentBoxIndex;
     }
 
-    if (param1 < 18) {
-        int v0;
-        u32 v1 = 0;
+    if (boxIndex < MAX_PC_BOXES) {
+        int monPosInBox;
+        u32 pokemonCount = 0;
```
ravepossum, line 298: suggestion: `count` probably suffices since the thing being counted is in the function name, and there's nothing else being counted in this function. I don't feel terribly strongly about that though. If you do keep the `pokemon` it can probably just be `mon`.

src/unk_020797C8.c
```diff
@@ -126,287 +129,287 @@ BOOL sub_02079914(PCBoxes *param0, u32 param1, u32 param2, BoxPokemon *param3)
...
+    if (boxIndex == -1) {
+        boxIndex = pcBoxes->currentBoxIndex;
     }
 
-    if (param1 < 18) {
-        int v0;
-        u32 v1 = 0;
+    if (boxIndex < MAX_PC_BOXES) {
+        int monPosInBox;
+        u32 eggCount = 0;
```
ravepossum, line 322: suggestion: Same comment as above with regard to `eggCount` vs. `count`.

ravepossum (commented):
Ran out of time for tonight so I didn't quite get to everything, but especially since this is a draft PR I figured I'd go ahead and get this feedback out to you.

Do note there's a handful of things I commented on that have similar occurrences elsewhere, but I didn't comment on every single one.

src/unk_020797C8.c
CharlesFolz4: I do prefer the parens there for readability.  Maybe that's an indication that I'm not familiar enough with C, but I'd prefer to keep them unless there are strong opinions otherwise.

src/unk_020797C8.c
```diff
@@ -3,24 +3,26 @@
...
 #include "savedata.h"
 #include "strbuf.h"
 
 typedef struct PCBoxes {
-    u32 unk_00;
-    BoxPokemon unk_04[18][30];
-    u16 unk_10E4[18][20];
+    u32 currentBoxIndex;
+    BoxPokemon boxedPokemonArray[MAX_PC_BOXES][30];
+    u16 boxNames[MAX_PC_BOXES][20];
```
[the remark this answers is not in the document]
CharlesFolz4: Defining 20 as `MAX_BOX_NAME_LENGTH` or something similar might be a bit misleading, because while technically correct from a data perspective, the effective max length is 8 because that's the max of the text input screen when renaming in this context.

src/unk_020797C8.c
CharlesFolz4: I couldn't figure out how to do this.  Could you give some direction on that, and/or an example to take a look at?

CharlesFolz4:
unk_020797C8 has been documented as the pokemon_storage_system_manager.  Further documentation of related overlay 19 will take place in a later PR.

src/unk_020797C8.c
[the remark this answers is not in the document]
ravepossum: Huh, yeah. I'm not sure what I was smoking with that suggestion lol. I do still think we should get a constant for that though. `PC_BOX_NAME_BUFFER_LEN` or something.

src/overlay019/ov19_021D0D80.c
```diff
@@ -65,27 +65,27 @@
...
 FS_EXTERN_OVERLAY(overlay84);
 
-static const TouchScreenHitTable Unk_ov19_021DFDF6[] = {
-    { 0xFE, 0x0, 0xB8, 0x28 },
-    { 0xFE, 0xFF, 0xB8, 0x28 },
-    { 0xFF, 0x0, 0x0, 0x0 }
+static const TouchScreenHitTable mainPcButtons[] = {
+    { 254, LEFT_MAIN_PC_BUTTON_X, MAIN_PC_BUTTON_Y, MAIN_PC_BUTTON_RAD },
+    { 254, RIGHT_MAIN_PC_BUTTON_X, MAIN_PC_BUTTON_Y, MAIN_PC_BUTTON_RAD },
+    { 255, 0, 0, 0 }
```
ravepossum, line 71: thought: Not in scope for this PR, so just thinking out loud, but I wonder if we should make some kind of `TOUCHSCREEN_REGION_END` or `TOUCHSCREEN_REGION_DUMMY` macro for all of the times this pattern appears.

src/overlay019/ov19_021D0D80.c
```diff
@@ -65,27 +65,27 @@
...
 
-static const TouchScreenHitTable Unk_ov19_021DFE12[] = {
-    { 0xFE, 0x98, 0xA8, 0x10 },
-    { 0xFE, 0xA3, 0x86, 0x10 },
-    { 0xFE, 0xB8, 0x68, 0x10 },
-    { 0xFE, 0xD8, 0x54, 0x10 },
-    { 0xFE, 0xF0, 0x34, 0x10 },
-    { 0xFE, 0xF0, 0x10, 0x10 },
-    { 0xFF, 0x0, 0x0, 0x0 }
+static const TouchScreenHitTable pokemonSymbolButtons[] = {
```
ravepossum, line 81: suggestion: Same comment as on the constants re: markings instead of symbol.

ravepossum (commented):
Thanks for addressing the feedback so far - got a fresh batch of comments for ya!

include/overlay019/ov19_021D0D80.h
```diff
@@ -14,6 +14,29 @@
 #include "pokemon.h"
 #include "string_template.h"
 
+#define LEFT_MAIN_PC_BUTTON_X  0
+#define RIGHT_MAIN_PC_BUTTON_X 255
+#define MAIN_PC_BUTTON_Y       184
+#define MAIN_PC_BUTTON_RAD     40
```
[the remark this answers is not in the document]
CharlesFolz4: These changes were from when I thought that I could document the overlay at the same time as the manager in one PR, so these are basically temporary placeholder names.  Same goes for basically all the changes in the ov19_* files here that aren't renamed functions from `pokemon_storage_system_manager`.  Should these types of changes be reverted, and be entirely isolated to the following overlay PR, or is it fine to leave them like this until that next planned PR?

include/overlay019/ov19_021D0D80.h
```diff
@@ -14,6 +14,29 @@
...
+#define PC_SYMBOL_BUTTON2_Y          134
+#define PC_SYMBOL_BUTTON3_X          184
+#define PC_SYMBOL_BUTTON3_Y          104
+#define PC_SYMBOL_BUTTON4_X          216
+#define PC_SYMBOL_BUTTON4_Y          84
+#define PC_SYMBOL_BUTTON5_X          240
+#define PC_SYMBOL_BUTTON5_Y          52
+#define PC_SYMBOL_BUTTON6_X          240
+#define PC_SYMBOL_BUTTON6_Y          16
+#define POKEMON_PC_SYMBOL_BUTTON_RAD 16
```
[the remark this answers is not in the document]
CharlesFolz4: It's possible to scroll the list, so that would only be accurate for the default positions.
Initial positions:
Screenshot_20250228_231939
Scrolled over a few ticks:
Screenshot_20250228_232206

include/overlay019/ov19_021D0D80.h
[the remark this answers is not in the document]
ravepossum: If you have sufficient info and context to conclude the names are appropriate, keeping them in this PR seems fine to me.

include/overlay019/ov19_021D0D80.h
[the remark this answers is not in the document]
ravepossum: Ahhh that makes sense, the regions stay in one place, just the gfx move. Gotcha.

src/pokemon_storage_system_manager.c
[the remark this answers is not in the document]
CharlesFolz4: The official name is the [Pokemon Storage System](https://bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_Storage_System), but in this context, maybe it would make sense to name it something closer to the code like `pc_boxes_manager`?

CharlesFolz4:
Merge conflicts resolved, ready for re-review.

res/text/pokemon_storage_system.gmm
```diff
@@ -30,7 +30,7 @@
 		<attribute name="window_context_name">used</attribute>
 		<language name="English">{STRVAR_1 11, 0, 0}</language>
 	</row>
-	<row id="BOX1_NAME_ID" index="6">
+	<row id="box1_name_id" index="6">
```
ravepossum, line 33: suggestion: I think `id` is implicit in this being a text bank entry constant. I think `box1_name` would suffice.

include/overlay019/ov19_021D0D80.h
```diff
@@ -14,6 +14,29 @@
...
+#define PC_MARKINGS_BUTTON2_Y          134
+#define PC_MARKINGS_BUTTON3_X          184
+#define PC_MARKINGS_BUTTON3_Y          104
+#define PC_MARKINGS_BUTTON4_X          216
+#define PC_MARKINGS_BUTTON4_Y          84
+#define PC_MARKINGS_BUTTON5_X          240
+#define PC_MARKINGS_BUTTON5_Y          52
+#define PC_MARKINGS_BUTTON6_X          240
+#define PC_MARKINGS_BUTTON6_Y          16
+#define POKEMON_PC_MARKINGS_BUTTON_RAD 16
```
ravepossum, line 38: question: Is there a reason the prefix on this one is different than the x/y ones?
suggestion: I'm still not a fan of `RAD` here either.

include/overlay019/ov19_021D0D80.h
```diff
@@ -14,6 +14,29 @@
...
 #include "string_template.h"
 
+#define LEFT_MAIN_PC_BUTTON_X  0
+#define RIGHT_MAIN_PC_BUTTON_X 255
+#define MAIN_PC_BUTTON_Y       184
+#define MAIN_PC_BUTTON_RAD     40
+
+#define COMPARE_POKEMON_PC_BUTTON_X   128
+#define COMPARE_POKEMON_PC_BUTTON_Y   178
+#define COMPARE_POKEMON_PC_BUTTON_RAD 12
```
ravepossum, line 24: suggestion: If we're concerned about length here, I think `MON` is just as clear as `POKEMON` in this context.

include/pokemon_storage_system_manager.h
```diff
@@ -0,0 +1,49 @@
...
+#define MAX_DEFAULT_WALLPAPERS    16
+#define MAX_UNLOCKABLE_WALLPAPERS 8
+#define MAX_WALLPAPERS            (MAX_DEFAULT_WALLPAPERS + MAX_UNLOCKABLE_WALLPAPERS)
+#define PC_BOX_NAME_BUFFER_LEN    20
+
+#define USE_CURRENT_BOX -1
+#define CheckUseCurrentBox(pcBoxes, box) \
+    if (box == USE_CURRENT_BOX) {        \
+        box = pcBoxes->currentBox;       \
+    }
```
ravepossum, line 22: suggestion: I like what you're going for here, but I'm actually not really a fan of macros like this - I'm not sure the tradeoff of lines saved is worth the reduction in clarity, since it's a relatively short snippet. I prefer macros for things like data entries or larger repeated chunks of code. Curious to hear others' thoughts on this.

EIther way we land on the macro, the casing standard for macros is `UPPERCASE_SNAKE_CASE`.

src/pokemon_storage_system_manager.c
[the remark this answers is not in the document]
ravepossum: I'd be fine with simply `pc_boxes`. This isn't a _really_ a manager in the way we currently use the term IMO (see something like `script_manager.c` which houses the struct of the same name), so I wouldn't include that. `pc_boxes` would also match the name of the struct (like you mentioned) which is nice.

src/pokemon_storage_system_manager.c
```diff
@@ -0,0 +1,382 @@
...
+            break;
+        }
+    }
+
+    return MAX_PC_BOXES;
+}
+
+BOOL PCBoxes_TryGetNextAvailableSpace(const PCBoxes *pcBoxes, int *boxIndexDest, int *monPosInBoxDest)
+{
+    if (*boxIndexDest == -1) {
```
ravepossum, line 167: polish: `USE_CURRENT_BOX`

src/pokemon_storage_system_manager.c
```diff
@@ -0,0 +1,382 @@
...
+    GF_ASSERT(monPosInBox < MAX_MONS_PER_BOX);
+
+    CheckUseCurrentBox(pcBoxes, box);
+
+    return BoxPokemon_GetValue((&pcBoxes->boxMons[box][monPosInBox]), pokemonData, dest);
+}
+
+void PCBoxes_SetBoxMonData(PCBoxes *pcBoxes, u32 box, u32 monPosInBox, enum PokemonDataParam pokemonData, void *value)
+{
+    GF_ASSERT(box < MAX_PC_BOXES || box == -1);
```
ravepossum, line 337: polish: `USE_CURRENT_BOX`

src/pokemon_storage_system_manager.c
```diff
@@ -0,0 +1,382 @@
...
+
+    CheckUseCurrentBox(pcBoxes, box);
+
+    BoxPokemon_SetValue((&pcBoxes->boxMons[box][monPosInBox]), pokemonData, value);
+    SaveData_SetFullSaveRequired();
+}
+
+BoxPokemon *PCBoxes_GetBoxMonAt(const PCBoxes *pcBoxes, u32 box, u32 monPosInBox)
+{
+    GF_ASSERT(box < MAX_PC_BOXES || box == -1);
```
ravepossum, line 348: polish: `USE_CURRENT_BOX`

src/unk_0204CFFC.c
```diff
@@ -438,26 +438,26 @@ BOOL ScrCmd_CountAliveMonsExcept(ScriptContext *ctx)
 
 BOOL ScrCmd_19C(ScriptContext *param0)
```
ravepossum, line 439: suggestion: I don't want to belabor this PR so feel free to just address in a future one (since I think you mentioned you had your eyes on more PC stuff), but if this script command is fully documented we should probably go ahead and give it a name. `v5` here is used to pass through the script var that receives the resulting count. We usually refer to this as `destVar`.

In case this helps, DSPRE refers to it as `CountTotalAlivePokemon` and pokeheartgold refers to it as `count_alive_mons_and_pc`. The actual function itself should retain the `ScrCmd_` prefix, but the corresponding macro in `asm/macros/scrcmd.inc` and in the scripts themselves should be named the same but drop the prefix.

src/pokemon_storage_system_manager.c
```diff
@@ -0,0 +1,382 @@
...
+    Init(pcBoxes);
+    SaveData_SetFullSaveRequired();
+}
+
+u32 PCBoxes_SaveSize(void)
+{
+    return sizeof(PCBoxes);
+}
+
+static void Init(PCBoxes *pcBoxes)
```
ravepossum, line 37: suggestion: I missed this on the last go-around, but this name feels just a tad bit too generic (as opposed to `InitBoxes` or something even). It feels slightly off, even for a static function. If others disagree I don't feel super strongly about that since it's confined to this module.

ravepossum (commented):
Thank you for the patience! A few last suggestions and clean-up.

include/pokemon_storage_system_manager.h
lhearachel: I personally don't mind shifting to PascalCase for functional macros like these, but I otherwise agree that this specific macro feels a bit unnecessary.

src/pokemon_storage_system_manager.c
CharlesFolz4: Couldn't think of anything more fitting to call it.  Would have liked `PCBoxes_Init`, but that's already taken by the function that calls this one.

include/overlay019/ov19_021D0D80.h
```diff
@@ -14,6 +14,29 @@
 #include "pokemon.h"
 #include "string_template.h"
 
+#define MAIN_PC_LEFT_BUTTON_X  0
+#define MAIN_PC_RIGHT_BUTTON_X 255
+#define MAIN_PC_BUTTON_Y       184
+#define MAIN_PC_BUTTON_R       40
```
ravepossum, line 20: Sorry if my suggestion before was unclear - I meant that I thought using the full term `RADIUS` was preferable.

ravepossum (approved):

PR #403, opened 2025-02-26
App storage

lucie-sharpe:
Not sure about the file name?


Kuruyia:
About the name, something like that makes me think of `PoketchScratchBuffer` or `PoketchWorkBuffer`.

lhearachel (approved):

Issue #405, opened 2025-02-26
Documentation Overhaul


lhearachel:
This is a bit of a kitchen-sink issue so that we don't lose track of ideas for easy-win improvements to documentation and process. In particular, our top-level documentation -- `INSTALL.md` and `CONTRIBUTING.md` -- have gone without touch-ups for some time and could do with some attention to standardize their language and structure and harden them against potential misreadings.

Some additional thoughts:

1. Should we centralize these documents in the existing `docs` folder, or does that make them less visible? Are they visible as-is among the sea of other files in the repository root?
2. Should we adopt a [Code of Conduct](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-code-of-conduct-to-your-project)?
3. Is a license file in-scope? The typical ones are likely *very* out of the picture, given the nature of this work. We may want to discuss this with the `pret` mods.
4. Is a [citation file](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) in-scope?
5. Should we adopt a Pull Request template to standardize some basic pre-PR procedures? I am thinking:
    1. A space to list key files and functions.
    2. A space to raise any concerns that the author has with the PR before a reviewer comes to it.
    3. A space to pre-document some systems that may be initially confusing.
    4. A check-list of expected pre-actions (e.g., `make format`).
    5. A free-form space for any other pre-review notes.

CharlesFolz4:
For point 1, `INSTALL.md` is already linked in the `README.md`, that makes it very visible from a standard starting point without any searching needed.  However,  `CONTRIBUTING.md` isn't linked anywhere that I could find, and it would make it more discoverable if it were linked from one of the other two documents.  That way, everything required to get started would be encountered just by reading the `README`, which I think is a reasonable expectation.

I also think that it may be beneficial for new contributors to have a list of basic information that is generally useful to know, such as a link to the [overlay ID mappings](https://gist.github.com/lhearachel/7983d1e594ff21300941b94bc2fbbb62). If it's useful enough and non-obvious enough to have documentation outside of the code, it's probably also useful and non-obvious enough to have that documentation be as easily discoverable as possible.

lhearachel:
[...]

PR #406, opened 2025-02-27
Overhaul CONTRIBUTING.md + integrate pre-commit

lhearachel:
This is part of a documentation overhaul, as per #405. The goal here is to:

1. Make the `CONTRIBUTING.md` clearer, friendlier, and more thorough in helping a newcomer get set up with their first contribution. The existing `CONTRIBUTING.md` -- while technically sound -- was terse and direct. I think that opening it up to a friendlier air will have many benefits. As always, it's a living document, and we should expect to update it as things change.
2. Make the development environment feel (closer to) seamless. We've had a lot of issues with the home-spun `clang-format` git hook in the past; this is part of the motivation behind `pre-commit` (the tool, see below).

## `pre-commit` (The Tool)

For those who are unfamiliar, `pre-commit` has loads of documentation on [their website](https://pre-commit.com).

The `.pre-commit-config.yaml` file in this PR does most of the work. It effectively acts as a direct representation of all of the actions that are taken whenever a contributor makes a commit to the repository. Existing contributors making use of the home-spun script in `.githooks` will need to update their `git` configuration:

```bash
git config --unset core.hooksPath
```

Then, follow the instructions listed in the `CONTRIBUTING.md` changes below:

```bash
pip install pre-commit
pre-commit install
```

Now, the actions listed in the dotfile will execute as needed when making a commit:

```console
> git commit -m "Add some vanilla pre-commit hooks"
clang-format.........................................(no files to check)Skipped
check that executables have shebangs.................(no files to check)Skipped
check for case conflicts.................................................Passed
check illegal windows names..........................(no files to check)Skipped
check yaml...............................................................Passed
[contributing-md 73579ed12] Add some vanilla pre-commit hooks
 1 file changed, 7 insertions(+)
```

By default, actions *only* run on files that have changes. If you want to run the set of actions against all files (for whatever reason), you can do that with:

```bash
pre-commit run --all-files
```

Eventually, I would like to explore using [`pre-commit.ci`](https://pre-commit.ci/) as an auto-resolution tool whenever a newcomer fails to update commits according to expectations.


CONTRIBUTING.md
```diff
@@ -1,74 +1,291 @@
...
+startup. Visual Studio Code users will need to take extra steps to configure
+their environment:
+
+1. Open the `C/C++: Edit Configurations (UI)` view.
+2. Scroll down to `Advanced Settings` and expand the section.
+3. Scroll down to `Compile commands` and enter the following in the textbox:
+
+```text
+compile_commands.json
+```
```
Fexty12573, line 82: Given that there already is a `.vscode/launch.json` this part could be included inside a `.vscode/c_cpp_properties.json` with something like this:
```json
{
    "configurations": [
        {
            "name": "Linux",
            "includePath": [
                "${workspaceFolder}/**"
            ],
            "defines": [],
            "compilerPath": "/usr/bin/gcc",
            "cStandard": "c99",
            "compileCommands": [
                "compile_commands.json"
            ]
        }
    ],
    "version": 4
}
```

Fexty12573:
Some other thoughts: maybe of note is that the intellisense of Microsofts C/C++ extension initially takes a bit to build up its symbol db (can take a minute or two), during which it is pretty much completely unresponsive. Might be worth noting.

An alternative for VSCode is also the `clangd` extension (which will prompt the user to install the clangd LSP automatically). It is initially faster with being responsive because builds its symbol db incrementally. However, as a result sometimes it misses references or can't find certain definitions if they're in some obscure location. Memory footprint also seems higher on occasion from what I could tell.

*If* you use clangd instead, you have to either uninstall the MS extension, or disable the intellisense engine.

CONTRIBUTING.md
lhearachel: That's a good shout; historically, we have tried to be editor-agnostic, but I think that maintaining integrations with popular editors is worth it for easing newcomers in.

Do we need to update the compiler-path specified here to use `arm-none-eabi-gcc` as specified in `compile_commands.json`, or is that automatically overwritten?

lhearachel:
> Some other thoughts: maybe of note is that the intellisense of Microsofts C/C++ extension initially takes a bit to build up its symbol db (can take a minute or two), during which it is pretty much completely unresponsive. Might be worth noting.

Definitely worth noting.

> An alternative for VSCode is also the `clangd` extension (which will prompt the user to install the `clangd` LSP automatically). It is initially faster with being responsive because builds its symbol db incrementally. However, as a result sometimes it misses references or can't find certain definitions if they're in some obscure location. Memory footprint also seems higher on occasion from what I could tell.

I think that we can note both options and the trade-offs and let users decide which they prefer. Thanks for the feedback! ❤️

CONTRIBUTING.md
lhearachel: I opted instead here to provide a template `c_cpp_properties.json` file within `CONTRIBUTING.md`. I am still wary about providing specific editor configurations, as I don't want us to deal with the headaches internally of remembering not to commit changes to the file.

ravepossum:
I don't have any proposed changes for `CONTRIBUTING.MD` - looks like a big improvement to me.

lhearachel:
I've updated the original state of this branch to provide setup-guides for Neovim and VS Code as "popular editors" in our in-repo docs. The vision here is that users can file pull requests to provide integration guides for other popular editors (e.g., Sublime, Emacs, whatever) that other maintainers may not be 100% familiar with.

PR #391, opened 2025-02-19
Document Shop Menu-related code

mudskipper13:
This PR does the following:
- Document `overlay007/ov7_0224CD28` -> `shop_menu`: Shop Menu interface itself.
- Document `unk_0209AC14` -> `shop_misc`: Some kind of helper function used after going back from the buy menu to the context list menu.
- Document `unk_02046AD4` -> `scrcmd_shop`: Script Commands related to Shop Menu, and also their usage as well.
- Unpack `shop_gra.narc`

PR #407, opened 2025-02-27
Document Accessory Shop (ov7_0224BE9C)

mudskipper13:
Follow-up to the Shop Menu PR (https://github.com/pret/pokeplatinum/pull/391).

This PR does the following:
- Document `ov7_0224BE9C` -> `accessory_shop`: Accessory Shop interface itself.
- Document `ScrCmd_257` -> `ShowAccessoryShop`: Script Command for opening the Accessory Shop.


include/overlay007/accessory_shop.h
```diff
@@ -0,0 +1,136 @@
...
+    enum HeapId heapID;
+} AccessoryShop_YesNoChoice;
+
+typedef struct AccessoryShop_DataPointers {
+    enum HeapId heapID;
+    BgConfig *bgConfig;
+    AccessoryShop_DescriptionBox *descBox;
+    const AccessoryShop_ItemList *itemList;
+    Bag *bag;
+} AccessoryShop_DataPointers;
```
ravepossum, line 110: suggestion: I think `Ptrs` would be a fine abbreviation since the struct name is already so long.

include/overlay007/accessory_shop.h
```diff
@@ -0,0 +1,136 @@
...
+    StringTemplate *strTemplate;
+    Strbuf *tempStrbuf;
+    u16 cursorPos;
+    u16 unused;
+    void *unk_130;
+    AccessoryShop_UpdateFunc updateFunc;
+    SpriteResourceCollection *spriteResCollections[ACCESSORY_SHOP_RES_IDX_COUNT];
+    SpriteResource *spriteResources[ACCESSORY_SHOP_RES_IDX_COUNT];
+    Sprite *sprites[ACCESSORY_SHOP_SPRITE_COUNT];
+} AccessoryShop_ItemListMenu;
```
ravepossum, line 96: suggestion: If you do rename the item list struct, would it make more sense to just call this one the `AccessoryShopItemList`? I think that would result in more concise access patterns and avoid stuff like `itemListMenu->listMenu`.

src/overlay007/accessory_shop.c
```diff
@@ -0,0 +1,767 @@
...
+{
+    SpriteTransfer_ResetCharTransfer(itemListMenu->spriteResources[ACCESSORY_SHOP_RES_IDX_TILES]);
+    SpriteTransfer_ResetPlttTransfer(itemListMenu->spriteResources[ACCESSORY_SHOP_RES_IDX_PALETTE]);
+
+    for (int i = 0; i < 4; i++) {
+        SpriteResourceCollection_Delete(itemListMenu->spriteResCollections[i]);
+    }
+}
+
+static void AccessoryShop_UpdateScrollArrowDrawFlag(AccessoryShop_ItemListMenu *itemListMenu)
```
ravepossum, line 652: suggestion: maybe something like `AccessoryShop_UpdateScrollArrowsVisibility`?

include/overlay007/accessory_shop.h
```diff
@@ -0,0 +1,136 @@
...
+    BOOL active;
+    Menu *yesNoChoice;
+    enum HeapId heapID;
+} AccessoryShop_YesNoChoice;
+
+typedef struct AccessoryShop_DataPointers {
+    enum HeapId heapID;
+    BgConfig *bgConfig;
+    AccessoryShop_DescriptionBox *descBox;
+    const AccessoryShop_ItemList *itemList;
```
ravepossum, line 108: suggestion: If you did rename the struct to `AccessoryShopItem`, this could probably just be something like `items`.

include/overlay007/accessory_shop.h
```diff
@@ -0,0 +1,136 @@
...
+    Strbuf *strbuf;
+} AccessoryShop_MessageBox;
+
+typedef struct AccessoryShop_DescriptionBox {
+    BOOL active;
+    Window *window;
+    StringTemplate *strTemplate;
+    Strbuf *strbuf;
+    Strbuf *fmtString[3];
+} AccessoryShop_DescriptionBox;
```
ravepossum, line 78: suggestion: I feel like `AccessoryShopDescBox` would still be descriptive enough.

src/overlay007/accessory_shop.c
```diff
@@ -0,0 +1,767 @@
...
+            }
+        }
+        break;
+    case ACCESSORY_SHOP_STATE_FREE_INTERFACE:
+        AccessoryShop_DeactivateMsgBox(&accessoryShop->msgbox);
+        AccessoryShop_DeactivateDescBox(&accessoryShop->descBox);
+        AccessoryShop_DeactivateItemListMenu(&accessoryShop->itemListMenu);
+        AccessoryShop_DeactivateYesNoChoice(&accessoryShop->yesNoChoice);
+        return TRUE;
+    default:
```
ravepossum, line 264: polish: I think this should match without this unnecessary default case.

src/overlay007/accessory_shop.c
```diff
@@ -0,0 +1,767 @@
...
+
+BOOL AccessoryShop_Main(AccessoryShop *accessoryShop)
+{
+    Strbuf *strbuf;
+    u32 input;
+
+    switch (accessoryShop->state) {
+    case ACCESSORY_SHOP_STATE_LOAD_GREET_MSG:
+        AccessoryShop_ActivateMsgBox(&accessoryShop->msgbox, accessoryShop->bgConfig, accessoryShop->heapID, accessoryShop->renderDelay);
+        strbuf = MessageLoader_GetNewStrbuf(accessoryShop->msgLoader, pl_msg_00000572_00004);
```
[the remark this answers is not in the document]
mudskipper13: Would it be okay if I only name the ids connected to/used in the shop? There were some other ids inside the .gmm file that is not used by the shop. ^^

mudskipper13:
Rebased to update upstream and apply review.

src/overlay007/accessory_shop.c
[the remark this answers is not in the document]
ravepossum: Definitely, I wouldn't worry about it if it's still unclear what it's used for or totally unrelated to the rest of the PR.

src/overlay007/accessory_shop.c
```diff
@@ -0,0 +1,767 @@
...
+}
+
+static void AccessoryShop_UpdateDescBox(AccessoryShop_DescriptionBox *descBox, u32 berryId, u32 totalAmount, u32 inBagAmount)
+{
+    StringTemplate_SetItemName(descBox->strTemplate, 0, berryId + FIRST_BERRY_IDX);
+    StringTemplate_SetNumber(descBox->strTemplate, 1, totalAmount, 3, PADDING_MODE_SPACES, CHARSET_MODE_EN);
+    StringTemplate_SetNumber(descBox->strTemplate, 2, inBagAmount, 3, PADDING_MODE_SPACES, CHARSET_MODE_EN);
+    Window_FillTilemap(descBox->window, 15);
+
+    for (int i = 0; i < 3; i++) {
```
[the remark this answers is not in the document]
ravepossum: hmmmm `ACCESSORY_SHOP_FORMAT_STRS` or something?

src/overlay007/accessory_shop.c
```diff
@@ -0,0 +1,764 @@
...
+    if (cursorPos < ACCESSORY_SHOP_ITEM_LIST_COUNT) {
+        inBagAmount = Bag_GetItemQuantity(dataPtr->bag, dataPtr->items[cursorPos].itemBerryID + FIRST_BERRY_IDX, dataPtr->heapID);
+        AccessoryShop_UpdateDescBox(dataPtr->descBox, dataPtr->items[cursorPos].itemBerryID, dataPtr->items[cursorPos].totalAmount, inBagAmount);
+    } else {
+        Bg_FillTilemapRect(dataPtr->bgConfig, BG_LAYER_MAIN_3, 0, 0, 10, 16, 8, 17);
+        Bg_CopyTilemapBufferToVRAM(dataPtr->bgConfig, BG_LAYER_MAIN_3);
+    }
+}
+
+static void AccessoryShop_ActivateYesNoChoice(AccessoryShopMenu *yesNoChoice, BgConfig *bgConfig, enum HeapId heapID)
```
ravepossum, line 691: polish: Don't forget to update the `AccessoryShopMenu` parameter here for the new struct name (same with the couple of functions below).

src/overlay007/accessory_shop.c
```diff
@@ -0,0 +1,764 @@
...
+    }
+
+    if (listPos >= (itemList->maxListItems - MAX_ITEM_SHOWN)) {
+        Sprite_SetDrawFlag(itemList->sprites[ACCESSORY_SHOP_SPRITE_SCROLL_ARROW_BOTTOM], FALSE);
+    } else {
+        Sprite_SetDrawFlag(itemList->sprites[ACCESSORY_SHOP_SPRITE_SCROLL_ARROW_BOTTOM], TRUE);
+    }
+}
+
+static void AccessoryShop_SetDataPointers(AccessoryShopDataPtrs *dataPtr, AccessoryShopDescBox *descBox, const AccessoryShopItem *itemList, Bag *bag, enum HeapId heapID, BgConfig *bgConfig)
```
ravepossum, line 668: suggestion: I would rename the `AccessoryShopItem *itemList` local vars and parameters to `*items` like the struct member so they don't get confused with the `AccessoryShopItemList` struct (same with a few below).

src/overlay007/accessory_shop.c
```diff
@@ -0,0 +1,764 @@
...
+static void AccessoryShop_SetDataPointers(AccessoryShopDataPtrs *dataPtr, AccessoryShopDescBox *descBox, const AccessoryShopItem *itemList, Bag *bag, enum HeapId heapID, BgConfig *bgConfig);
+static void AccessoryShop_Update(void *ptr, u32 cursorPos);
+static BOOL AccessoryShop_HasAllAccessories(const UnkStruct_02029D04 *param0, const AccessoryShopItem *itemList, u32 maxItems);
+static BOOL AccessoryShop_HasEnoughBerries(Bag *bag, const AccessoryShopItem *itemList, u32 idx, enum HeapId heapID);
+static BOOL ov7_0224CCE4(const UnkStruct_02029D04 *param0, const AccessoryShopItem *itemList, u32 idx);
+static void AccessoryShop_DoPurchase(UnkStruct_02029D04 *param0, Bag *bag, const AccessoryShopItem *itemList, u32 idx, enum HeapId heapID);
+
+AccessoryShop *AccessoryShop_New(enum HeapId heapID, SaveData *saveData, BgConfig *bgConfig)
+{
+    AccessoryShop *accessoryShop = Heap_AllocFromHeap(heapID, sizeof(AccessoryShop));
```
ravepossum, line 92: suggestion: I think this function and the handful below would also benefit from the brevity of `AccessoryShop *shop` for the parameters/local vars since the shop we're dealing with is in the function name (and no other shops).

res/text/flower_shop.gmm
```diff
@@ -17,67 +17,67 @@
 		<attribute name="window_context_name">used</attribute>
 		<language name="English">Use that Sprayduck to water the soil\nwhere you’ve planted Berries.\rYou should also water Berry plants\nwhile they are growing.</language>
 	</row>
-	<row id="pl_msg_00000572_00004" index="4">
+	<row id="pl_msg_accessory_shop_greet" index="4">
```
ravepossum, line 20: todo: You can drop the `pl_msg_` on these IDs, we don't need them once they've been documented.

res/text/flower_shop.gmm
```diff
@@ -17,67 +17,67 @@
...
 		<attribute name="window_context_name">used</attribute>
 		<language name="English">YES</language>
 	</row>
-	<row id="pl_msg_00000572_00015" index="15">
+	<row id="pl_msg_accessory_shop_no" index="15">
 		<attribute name="window_context_name">used</attribute>
 		<language name="English">NO</language>
 	</row>
-	<row id="pl_msg_00000572_00016" index="16">
+	<row id="pl_msg_accessory_shop_berrynamebuf" index="16">
```
ravepossum, line 68: suggestion: Since this doesn't really refer to a buffer, I would name it slightly differently. I don't think we have a solid naming standard for these types of text entries yet - I've been referring to them as `template`s (see `summary_template_item_name` and friends in `pokemon_summary_screen.gmm`) but I could also see a name that mentions `format`/`fmt` in some way. Curious to hear some opinions.

ravepossum (commented):
Thanks for responding to everything so far - just a few more comments and then this should be good to go!

mudskipper13:
Rebased to update upstream and apply review once again.

res/text/flower_shop.gmm
lhearachel: The current in-repo term is a `StringTemplate` for the struct which contains this string and a store of variable-values; `StringTemplate_Format` performs the expansion itself into a renderable piece of text.

I personally don't mind if we choose to name the raw data loaded from a text bank as `fmt`, `fstring`, etc., but we should pick one and be consistent. Might require a light sweep of other instances already named.

mudskipper13:
Review applied.
(I thought I've pushed these before the review oops)

ravepossum (approved):

PR #408, opened 2025-02-27
Pal Park documentation - Part 1

AsparagusEduardo:


include/constants/gba/items.h
```diff
@@ -0,0 +1,383 @@
+#ifndef POKEPLATINUM_CONSTANTS_GBA_ITEMS_H
```
lhearachel, line 1: **note**: You're free to use an `enum` for these; we standardize the use of them throughout the repository where possible, and generate them instead if the values need to be used by an assembler (e.g., in scripting).

include/constants/gba/items.h
```diff
@@ -0,0 +1,383 @@
...
+#define GBA_ITEM_RAINBOW_PASS   368
+#define GBA_ITEM_TEA            369
+#define GBA_ITEM_MYSTIC_TICKET  370
+#define GBA_ITEM_AURORA_TICKET  371
+#define GBA_ITEM_POWDER_JAR     372
+#define GBA_ITEM_RUBY           373
+#define GBA_ITEM_SAPPHIRE       374
+#define GBA_ITEM_MAGMA_EMBLEM   375
+#define GBA_ITEM_OLD_SEA_MAP    376
+#define ITEMS_COUNT             377
```
lhearachel, line 381: **polish**: I would call this `GBA_ITEM_MAX` or `GBA_ITEM_COUNT`.

src/overlay097/ov97_02233B18.c
```diff
@@ -548,19 +562,19 @@ static void ov97_022340FC(AffineSpriteListTemplate *param0, UnkStruct_ov97_02234
     param0->heapID = 78;
 }
 
-static int ov97_02234124(UnkStruct_ov97_02234A2C *param0, int param1, int param2)
+static int GetGBABoxMonSpeciesAtBox(UnkStruct_ov97_02234A2C *param0, int boxId, int boxPosition)
```
lhearachel, line 565: **suggestion**: I wonder if the semantics of `InBox` vs `AtBox` are better here. Thoughts?

src/overlay097/ov97_02233B18.c
```diff
@@ -595,90 +609,98 @@ static void ov97_022341EC(u32 param0, NNSG2dCharacterData **param1, void *param2
...
-    u8 v0;
+#define NUM_UNOWN_FORMS 28
 
-    v0 = 0;
+#define GET_UNOWN_LETTER_FROM_PERSONALITY(personality) ((                                          \
+                                                            (((personality) & 0x03000000) >> 18)   \
+                                                            | (((personality) & 0x00030000) >> 12) \
+                                                            | (((personality) & 0x00000300) >> 6)  \
+                                                            | (((personality) & 0x00000003) >> 0)) \
+    % NUM_UNOWN_FORMS)
```
lhearachel, line 619: **thought**: Oh, I'm very surprised that it formatted this this way. Maybe something we need to tinker with in clang-format.

src/overlay097/ov97_02233B18.c
```diff
@@ -812,296 +834,295 @@ static void ov97_0223468C(UnkStruct_ov97_02234A2C *param0)
...
+            return TRUE;
         }
     }
 
-    return 0;
+    return FALSE;
 }
 
-static BOOL ov97_02234828(UnkStruct_ov97_02234A2C *param0, int param1)
+static BOOL BoxMonGBAIsInvalidSpecies(UnkStruct_ov97_02234A2C *param0, int boxPosition)
```
lhearachel, line 1039: **suggestion**: `IsBoxMonGBAInvalidSpecies`

src/overlay097/ov97_02236380.c
```diff
@@ -14,148 +14,150 @@
 #include "pokemon.h"
 #include "system.h"
 
+#define LAST_GBA_MOVE MOVE_PSYCHO_BOOST
```
lhearachel, line 17: **suggestion**: This may be better placed in `constants/moves.h`, but if it feels specific-enough to this file, then I'm fine to have it live here.

src/overlay097/ov97_02236380.c
```diff
@@ -14,148 +14,150 @@
...
+    { GBA_SPECIES_KYOGRE, SPECIES_KYOGRE },
+    { GBA_SPECIES_GROUDON, SPECIES_GROUDON },
+    { GBA_SPECIES_RAYQUAZA, SPECIES_RAYQUAZA },
+    { GBA_SPECIES_LATIAS, SPECIES_LATIAS },
+    { GBA_SPECIES_LATIOS, SPECIES_LATIOS },
+    { GBA_SPECIES_JIRACHI, SPECIES_JIRACHI },
+    { GBA_SPECIES_DEOXYS, SPECIES_DEOXYS },
+    { GBA_SPECIES_CHIMECHO, SPECIES_CHIMECHO },
+    { GBA_SPECIES_EGG, 0x19C },
     { 0x19D, 0x19D },
```
lhearachel, line 161: **thought**: Are these the `BAD_EGG` + form placeholders from the GBA?

src/overlay097/ov97_02236380.c
```diff
@@ -1352,271 +1351,271 @@ static int ov97_02236E28(BoxPokemonGBA *boxMonGBA, BoxPokemon *boxMon)
...
+    BoxPokemon_SetValue(param1, MON_DATA_OT_ID, (u8 *)&value);
 
-    v1 = GetGBABoxMonData(param0, 25, NULL);
-    BoxPokemon_SetValue(param1, MON_DATA_EXP, (u8 *)&v1);
+    value = GetGBABoxMonData(param0, GBA_MON_DATA_EXP, NULL);
+    BoxPokemon_SetValue(param1, MON_DATA_EXP, (u8 *)&value);
 
-    v1 = 70;
-    BoxPokemon_SetValue(param1, MON_DATA_FRIENDSHIP, (u8 *)&v1);
+    value = 70;
```
lhearachel, line 1386: **thought**: This may be worth a constant in `constants/pokemon.h`, e.g., `BASE_FRIENDSHIP_VALUE`

src/overlay097/ov97_02236380.c
```diff
@@ -1352,271 +1351,271 @@ static int ov97_02236E28(BoxPokemonGBA *boxMonGBA, BoxPokemon *boxMon)
...
             (void)0;
         }
 
-        BoxPokemon_SetValue(param1, MON_DATA_MOVE1 + v2, (u8 *)&v1);
+        BoxPokemon_SetValue(param1, MON_DATA_MOVE1 + i, (u8 *)&value);
 
-        v1 = GetGBABoxMonData(param0, 21, NULL);
-        v1 = (v1 & (0x3 << (v2 * 2))) >> (v2 * 2);
+        value = GetGBABoxMonData(param0, GBA_MON_DATA_PP_BONUSES, NULL);
+        value = (value & (0x3 << (i * 2))) >> (i * 2);
```
lhearachel, line 1444: **question**: Is there a bitmask that we can borrow from `pokeemerald` for this?

lhearachel (commented):
Great work! I have some comments/questions for polish, but the rest looks wonderful. ❤️

src/overlay097/ov97_02233B18.c
AsparagusEduardo: Yeah, makes sense :)
I'll update them

src/overlay097/ov97_02236380.c
AsparagusEduardo: I wanted to, but I don't know how to add it via `generated/moves.txt`

lhearachel: No need to add it to the generated constants, I think; there is a separate `include/constants/moves.h`, which exists to easily pull all of the move-related generated enums in, plus add some extra stuff that doesn't really fit as any of those enums, e.g.:

```c
#define LEARNED_MOVES_MAX 4
#define MOVE_NOT_SELECTED 0xFF

#define NUM_MOVES MOVE_SHADOW_FORCE
```

I would say that it fits well in this batch. 😁

src/overlay097/ov97_02236380.c
AsparagusEduardo: I'm not sure. After `SPECIES_EGG` on pokeemerald, there's 26 species IDs for Unown B-QMark, but that's way more than what's here. Would you prefer them to be added just in case?

AsparagusEduardo:
Forgot to say, but I addressed everything I could

src/overlay097/ov97_02236380.c
lhearachel: I would rather we not document things that we don't have strong answers for; a wrong answer is worse than no answer, IMO.

Especially since this set of species IDs is so... seemingly random...? `0x19C` is Burmy, and it follows the chain from there:

1. Wormadam
2. Mothim
3. Combee
4. Vespiquen
5. Pachirisu
6. Buizel
7. Floatzel
8. Cherubi

lhearachel (approved):

PR #409, opened 2025-02-28
Document area light

Kuruyia:
This documents the `ov5_021D521C.c` file, which is in charge of managing lighting according to the currently loaded map area and the current wall-clock time.


Kuruyia:
For reference, here's how the 4 lights available in the NDS 3D engine are used:

| Light # | Screenshot | Used on (non-exhaustive) |
|--------|--------|--------|
| Light 0 | light0 | 2D sprites, base 3D model of the map, most of the polygons of map props |
| Light 1 | N/A | Unused, as far as I can tell (this light is always set to invalid in the `area_light.narc` archive) |
| Light 2 | light2 | Building windows | 
| Light 3 | light3 | Lamp post lights, building lights and doors (only glass doors: PokéCenter, PokéMart, Gyms...) | 

(corresponding lights are in red in each screenshot)

src/overlay005/area_light.c
```diff
@@ -0,0 +1,286 @@
...
+    ModelAttributes_SetEmissionColor(modelAttrs, params->emissionColor, TRUE);
+}
+
+static void AreaLightManager_ApplyActiveParamsToAreaModelAttrs(AreaLightManager *areaLightMan)
+{
+    AreaLightParameters *activeParams = &areaLightMan->paramsList[areaLightMan->activeParamsIndex];
+    AreaLightParameters_ApplyToModelAttributes(activeParams, areaLightMan->areaModelAttrs);
+}
+
+void AreaLight_UseGlobalModelAttributes(NNSG3dResMdl *model)
```
Kuruyia, line 125: This function may be too generic to have the `AreaLight_` prefix, but those 4 parameters still come from the area light NARC, so it's not entirely unrelated. Thoughts?

src/overlay005/area_light.c
```diff
@@ -0,0 +1,286 @@
+#include "overlay005/area_light.h"
+
+#include <nitro.h>
+#include <string.h>
+
+#include "constants/gx.h"
```
lhearachel, line 6: **thought**: I wonder if we want a different name for this constants file. Maybe just `constants/graphics.h`? Is that too general?

src/overlay005/area_light.c
```diff
@@ -0,0 +1,286 @@
...
+static u32 AreaLightParameters_New(u32 archiveID, AreaLightParameters **paramsList);
+static void AreaLightParameters_Free(AreaLightParameters **params);
+static char *AreaLightParameters_ParseLightParams(char *fileIter, GXRgb *lightColor, VecFx16 *lightVector);
+static char *AreaLightParameters_ParseColor(char *fileIter, GXRgb *color);
+
+AreaLightManager *AreaLightManager_New(ModelAttributes *areaModelAttrs, const u8 archiveID)
+{
+    int i;
+    AreaLightManager *areaLightMan;
+    int currentTime;
```
lhearachel, line 31: **polish**: Compress declarations and definitions to C99-style (inline definition with declaration) where possible.

src/overlay005/area_light.c
lhearachel: I would keep the prefix, knowing that these come from that NARC and that NARC only.

src/overlay005/area_light.c
```diff
@@ -0,0 +1,286 @@
...
+            }
+
+            if (areaLightMan->applyToAreaModelAttrs) {
+                AreaLightManager_ApplyActiveParamsToAreaModelAttrs(areaLightMan);
+            }
+        }
+    }
+}
+
+void AreaLightParameters_ApplyToModelAttributes(const AreaLightParameters *params, ModelAttributes *modelAttrs)
```
lhearachel, line 99: **suggestion**: The usage of this `Parameters` structure seems to reflect what we typically refer to as a `Template` in-repo (see: `BgTemplate`, `SpriteTemplate`, etc.). I think that this should be renamed to follow along with that paradigm.

lhearachel (commented):
Fantastic work, and thank you so much for the illustrative documentation. :heart: 

> For reference, here's how the 4 lights available in the NDS 3D engine are used

I think that this would be a great candidate for an entry somewhere in `docs`. We should start hammering down on how we want to structure the external-facing documentation, so that we can communicate to future users how the various subsystems of the game work.

Kuruyia:
> I think that this would be a great candidate for an entry somewhere in docs. We should start hammering down on how we want to structure the external-facing documentation, so that we can communicate to future users how the various subsystems of the game work.

Oh absolutely :) I think we're almost ready to document how maps work in `docs/`, I just wanted to tackle as much maps-related NARC as I could beforehand. (the last one should be `land_data.narc` in `ov5_021E779C.c`)

I was thinking having a big `docs/maps/file_specifications.md` file that would be huge tables of the fields found in those various NARCs, and then maybe having `docs/maps/x.md` that would be a plain-English description of a maps "sub-subsystem", which may correspond (but not necessarily) to data found in a single NARC.

src/overlay005/area_light.c
Kuruyia: I named it that way to have some kind of link to Nitro SDK's `<nitro/gx.h>`, if that matters.

lhearachel: That's what I presumed. I'd consider it non-blocking.

lhearachel:
> I was thinking having a big `docs/maps/file_specifications.md` file that would be huge tables of the fields found in those various NARCs, and then maybe having `docs/maps/x.md` that would be a plain-English description of a maps "sub-subsystem", which may correspond (but not necessarily) to data found in a single NARC.

I agree, I think that having all of the technical specifications separated from the plain-English explanations is worthwhile. Most of our users I expect will be some degree of non-technical, or at least not care much for all the gritty details.

lhearachel (approved):

PR #410, opened 2025-03-01
Pal Park Documentation - Part 2

AsparagusEduardo:

PR #411, opened 2025-03-02
BoxPokemon documentation

AsparagusEduardo:
~~Requires #410 to be merged to master first. Once that's done, I can merge master to this PR and the specific commits will disappear from it to make reviewing easier.
In the meantime, here are the differences between it and this PR: https://github.com/AsparagusEduardo/pokeplatinum/compare/document4...AsparagusEduardo:document5~~


src/overlay005/ov5_021E622C.c
```diff
@@ -2138,23 +2138,23 @@ u16 ov5_021E64F8(Party *param0, StringTemplate *param1, UnkStruct_02026310 *para
...
 }
 
-int ov5_021E6520(BoxPokemon *param0, u32 param1)
+int BoxPokemon_GiveExperience(BoxPokemon *boxMon, u32 givenExp)
 {
-    Pokemon *v0 = Pokemon_New(4);
-    BoxPokemon *v1 = Pokemon_GetBoxPokemon(v0);
-    int v2;
-    u32 v3;
+    Pokemon *mon = Pokemon_New(4);
```
lhearachel, line 2143: **polish**:

```suggestion
    Pokemon *mon = Pokemon_New(HEAP_FIELD);
```

src/overlay019/ov19_021D0D80.c
```diff
@@ -2299,26 +2299,24 @@ static void ov19_021D3978(SysTask *param0, void *param1)
...
+static BOOL BoxPokemon_HasMove(BoxPokemon *boxMon, u16 move)
 {
-    BOOL v0, v1;
+    BOOL hasMove = FALSE;
+    BOOL reencrypt = BoxPokemon_EnterDecryptionContext(boxMon);
 
-    v0 = 0;
-    v1 = BoxPokemon_EnterDecryptionContext(param0);
+    if (BoxPokemon_GetValue(boxMon, MON_DATA_EGG_EXISTS, NULL) == FALSE) {
+        int i;
```
lhearachel, line 2308: **question**: Can this be declared inline in the loop?

src/overlay094/ov94_0223D0C4.c
```diff
@@ -540,8 +540,8 @@ void ov94_0223D910(MessageLoader *param0, MessageLoader *param1, StringTemplate
 
     ov94_02245900(&param3[0], v4, 0, 0, 0, TEXT_COLOR(15, 2, 0));
 
-    if (v6 != 3) {
-        ov94_02245900(&param3[0], v1, 70, 0, 0, Unk_ov94_022467FC[v6]);
+    if (gender != 3) {
```
lhearachel, line 543: **thought**: Odd... shouldn't this be impossible?

src/overlay094/ov94_0223E46C.c
```diff
@@ -1123,7 +1125,11 @@ static void ov94_0223F9FC(Window *param0, Window *param1, MessageLoader *param2)
...
 static int ov94_0223FB0C(const UnkStruct_ov94_0223BA88_sub3 *param0, const UnkStruct_ov94_0223BA88_sub3 *param1, int param2, int param3)
 {
-    if ((param0->unk_00 == param1->unk_00) && (param0->unk_02 == param1->unk_02) && (param0->unk_03 == param1->unk_03) && (param0->unk_04 == param1->unk_04) && (param2 == param3)) {
+    if ((param0->species == param1->species)
+        && (param0->gender == param1->gender)
+        && (param0->level == param1->level)
+        && (param0->level2 == param1->level2)
+        && (param2 == param3)) {
         return 1;
     }
```
lhearachel, line 1135: **question**: I wonder if this matches as just `return (conditional)`, since it looks to be `BOOL`-like.

src/overlay094/ov94_0223FB48.c
```diff
@@ -1227,83 +1227,72 @@ static int ov94_0224121C(Party *param0, PCBoxes *param1, int param2, int param3)
...
+    if (count) {
+        return TRUE;
     }
 
-    return 0;
+    return FALSE;
 }
 
-static int ov94_02241278(BoxPokemon *param0)
+static int ov94_02241278(BoxPokemon *boxMon)
```
lhearachel, line 1263: **question**: Can we not name this function? I suppose we may want to wait to see how it's used, since it's not just "Pokemon with mechanically-distinct forms" (lacking Deoxys, Wormadam).

In either case, I think we can change the signature to return `BOOL`.

src/overlay094/ov94_022414B8.c
```diff
@@ -598,15 +598,15 @@ int ov94_02241B80(UnkStruct_ov94_0223BA88_sub3 *param0, int param1)
 {
     switch (param1) {
     case 0:
-        param0->unk_02 = 0 + 1;
+        param0->gender = 0 + 1;
```
lhearachel, line 601: **polish**: Replace the left-operand on these with constants from `generated/genders.h`.

include/overlay094/struct_ov94_0223BA88_sub3.h
```diff
@@ -2,10 +2,10 @@
...
 
 typedef struct {
-    s16 unk_00;
-    s8 unk_02;
-    s8 unk_03;
-    s8 unk_04;
+    s16 species;
+    s8 gender;
+    s8 level;
+    s8 level2;
```
lhearachel, line 8: **question**: I'm not sure what's up with this field. Maybe something that will be illuminated later...?

src/overlay094/ov94_0223FB48.c
AsparagusEduardo: At the time, I could only identify that it listed forms added in Platinum, but couldn't confirm it was that.
However, I now checked its usage, and seeing that its used alongside `ov94_022412C8` (which only lists Griseous Orb), confirms that suspicion. I'll update both.

src/overlay094/ov94_0223D0C4.c
AsparagusEduardo: I would think that this is `GENDER_NONE + 1`, as `UnkStruct_ov94_0223BA88_sub2`'s gender field gets value set + 1 in multiple places. I'll update this to use that

include/overlay094/struct_ov94_0223BA88_sub3.h
AsparagusEduardo: My suspicion is that maybe it's met level, but I'm not 100% sure :/
It's definitely some sort of level, based on comparisons in `ov94_02242970`

AsparagusEduardo:
Ready for re-review

src/overlay094/ov94_0223FB48.c
lhearachel: Ah! So this would be "forms that can't be sent over to Diamond and Pearl," gotcha. 😄

lhearachel (approved):

PR #415, opened 2025-03-05
Same-line variable declarations and definitions

AsparagusEduardo:
Requires #414 
https://github.com/AsparagusEduardo/pokeplatinum/compare/document7...AsparagusEduardo:document8


lhearachel (approved):
This is fine to merge as-is, but (as a future note) we generally just leave this sort of clean-up work to the documentation of the file itself.

PR #421, opened 2025-03-07
Document land data

Kuruyia:
This documents the `ov5_021E779C.c` file, which is in charge of loading map data from the `land_data.narc` archive, dynamically load and unload maps as the player moves, and render the map base models and their props.

Also, for the viewer's pleasure, it includes a ton of Distortion World-specific functions.

This also documents the `ov5_021EEAC8.c` file, which is in charge of managing buffer allocation for the map base model and BDHC data.


include/constants/field/map_object_move.h
```diff
@@ -0,0 +1,8 @@
+#ifndef POKEPLATINUM_CONSTANTS_MAP_FIELD_OBJECT_MOVE_H
+#define POKEPLATINUM_CONSTANTS_MAP_FIELD_OBJECT_MOVE_H
+
+#include <nitro/fx/fx_const.h>
+
+#define MAP_OBJECT_MOVE_TILE_SIZE (16 * FX32_ONE)
```
lhearachel, line 6: **suggestion**: I think that these are closely-related enough that we can merge them into `constants/field/map.h`. Thoughts?

include/overlay005/land_data_manager_decl.h
```diff
@@ -0,0 +1,6 @@
+#ifndef POKEPLATINUM_STRUCT_OV5_LAND_DATA_MANAGER_DECL_H
+#define POKEPLATINUM_STRUCT_OV5_LAND_DATA_MANAGER_DECL_H
+
+typedef struct LandDataManager_t LandDataManager;
```
lhearachel, line 4: **question**: Can this be merged with its definition somehow? I know the `FieldSystem` struct is notorious for how many forward-declares it does. Maybe we should untangle a system for that now so that we can offer more direct guidance. Perhaps, for example, we could supply forward declarations in `field_system.h`...? Or we could forward-declare `FieldSystem` and allow requisite structs to include that.

src/overlay005/fieldmap.c
```diff
@@ -404,8 +405,8 @@ static BOOL FieldMap_ChangeZone(FieldSystem *fieldSystem)
         return FALSE;
     }
 
-    x = (Player_GetXPos(fieldSystem->playerAvatar) - ov5_021EA6AC(fieldSystem->unk_28)) / 32;
-    y = (Player_GetZPos(fieldSystem->playerAvatar) - ov5_021EA6B4(fieldSystem->unk_28)) / 32;
+    x = (Player_GetXPos(fieldSystem->playerAvatar) - LandDataManager_GetOffsetTileX(fieldSystem->landDataMan)) / 32;
+    y = (Player_GetZPos(fieldSystem->playerAvatar) - LandDataManager_GetOffsetTileY(fieldSystem->landDataMan)) / 32;
```
lhearachel, line 409: **polish**:

```suggestion
    x = (Player_GetXPos(fieldSystem->playerAvatar) - LandDataManager_GetOffsetTileX(fieldSystem->landDataMan)) / MAP_TILES_COUNT_X;
    y = (Player_GetZPos(fieldSystem->playerAvatar) - LandDataManager_GetOffsetTileY(fieldSystem->landDataMan)) / MAP_TILES_COUNT_X;
```

src/overlay005/land_data.c
```diff
@@ -0,0 +1,2590 @@
...
+#include "heap.h"
+#include "map_matrix.h"
+#include "narc.h"
+#include "sys_task.h"
+#include "sys_task_manager.h"
+#include "system.h"
+#include "unk_020366A0.h"
+
+#define INVALID_LAND_DATA_ID       0xFFFF
+#define INVALID_TERRAIN_ATTRIBUTES -1
```
lhearachel, line 30: **question**: Should this be part of `constants/field/map_tile_behaviors`?

src/overlay005/land_data.c
```diff
@@ -0,0 +1,2590 @@
...
+        }
+
+        landDataMan->loadedMaps[i]->mapMatrixIndex = INVALID_MAP_MATRIX_INDEX;
+    }
+}
+
+static void LandDataHeader_Load(NARC *landDataNARC, const int landDataID, LandDataHeader *landDataHeader)
+{
+    void *buffer;
+    int bufferSize = sizeof(int) * 4;
```
lhearachel, line 475: **question**: Does `sizeof(LandDataHeader)` not match?

src/overlay005/land_data.c
```diff
@@ -0,0 +1,2590 @@
...
+{
+    if (mapMatrixIndex < 0 || mapMatrixIndex >= mapMatrixWidth * mapMatrixHeight) {
+        return;
+    }
+
+    int landDataID = MapMatrix_GetLandDataIdByIndex(mapMatrixIndex, mapMatrix);
+
+    if (landDataID == INVALID_LAND_DATA_ID) {
+        return;
+    }
```
lhearachel, line 627: **thought**: I see this same validation-suite over these various loaders. Maybe this is worth a macro?

src/overlay005/land_data.c
```diff
@@ -0,0 +1,2590 @@
...
+            }
+        } else {
+            GF_ASSERT(FALSE);
+        }
+
+        LandDataManager_QueueLazyLoadMapPair(firstMapMatrixIndexToLoad, secondMapMatrixIndexToLoad, firstQuadrantToLoad, secondQuadrantToLoad, trackedTargetDirection, landDataMan);
+        break;
+
+    case DIRECTION_EAST:
+        // The tracked target is going east, and is currently in one of the left quadrants within its map,
```
lhearachel, line 1101: **praise**: Thanks so much for this documentation. :pray:

src/overlay005/land_data.c
```diff
@@ -0,0 +1,2590 @@
...
+    landDataMan->lazyLoaderSubTask = LAZY_LOADER_SUBTASK_GRACEFUL_SHUTDOWN;
+    landDataMan->ending = TRUE;
+    landDataMan->lazyLoaderTrackedTarget.tracking = FALSE;
+
+    for (i = 0; i < QUADRANT_COUNT; i++) {
+        MI_CpuFillFast(landDataMan->loadedMaps[i]->terrainAttributes, INVALID_TERRAIN_ATTRIBUTES, sizeof(u16) * MAP_TILES_COUNT_X * MAP_TILES_COUNT_Y);
+    }
+}
+
+void LandDataManager_DistortionWorldPrepareGoingToNextFloor(MapMatrix *mapMatrix, AreaDataManager *areaDataMan, LandDataManager *src, LandDataManager *dest, const int trackedTargetTileX, const int trackedTargetTileY)
```
lhearachel, line 2240: **suggestion**: I think that this can do with a shorter name. `PrepareNextFloor`?

src/overlay005/land_data.c
Kuruyia: Introduced those macros:
```c
#define LAND_DATA_LOADER_ASSERT_MAP_MATRIX_INDEX(index, matrixWidth, matrixHeight) \
    if (index < 0 || index >= matrixWidth * matrixHeight) {                        \
        return;                                                                    \
    }

#define LAND_DATA_LOADER_ASSERT_LAND_DATA_ID(id) \
    if (id == INVALID_LAND_DATA_ID) {            \
        return;                                  \
    }
```

And done those modifications:
```diff
     int mapMatrixIndex = loader->mapMatrixIndexes[index];
-
-    if (mapMatrixIndex < 0 || mapMatrixIndex >= mapMatrixWidth * mapMatrixHeight) {
-        return;
-    }
+    LAND_DATA_LOADER_ASSERT_MAP_MATRIX_INDEX(mapMatrixIndex, mapMatrixWidth, mapMatrixHeight);

     int landDataID = MapMatrix_GetLandDataIdByIndex(mapMatrixIndex, mapMatrix);
-
-    if (landDataID == INVALID_LAND_DATA_ID) {
-        return;
-    }
+    LAND_DATA_LOADER_ASSERT_LAND_DATA_ID(landDataID);
```

Please let me know if you had something else in mind.

src/overlay005/land_data.c
Kuruyia: Indeed it does, just didn't cross my mind ^^'

src/overlay005/land_data.c
Kuruyia: Sounds like a good idea. As part of the `TileBehavior` enum?

```diff
 enum TileBehavior {
+    TILE_BEHAVIOR_INVALID = -1,
     TILE_BEHAVIOR_NONE = 0,
     TILE_BEHAVIOR_UNUSED_x01,
```

Also, that begs the question, is `terrainAttributes` correctly named? I see mentions of "attributes", "behavior" and "collision" for those and the difference is fuzzy to me.

I also see that I forgot to rename "collision" to "attribute" in functions, I'll fix that.

include/overlay005/land_data_manager_decl.h
Kuruyia: Not sure this answers your question, but a big problem we also have here is that there is a circular dependency:
`LandDataManager --> MapLoadFunctions --> (InitialLoadMapsFn | LazyLoadMapsFn) --> LandDataManager`

include/constants/field/map_object_move.h
Kuruyia: Sounds good to me. Also renamed to `MAP_OBJECT_TILE_SIZE` as I think it's not specific to object movement (although it's going to be used in a lot of math related to object movement).

include/overlay005/land_data_manager_decl.h
lhearachel: I see. That would still be handled by having the forward-declaration being in `field_system.h`, but I get that we have a bit of indeterminate state here on how we want to handle untangling all of these circular references.

I would say that, for now, let's adhere to the usual naming scheme for `typedef`s:

```suggestion
typedef struct LandDataManager LandDataManager;
```

And then, for the definition of the `struct` itself, trim the actual definition line to adhere to [the ISO C99 standard (PDF download)](https://www.dii.uchile.cl/~daespino/files/Iso_C_1999_definition.pdf):

> **§6.2.2/6**
> The following identifiers have no linkage: an identifier declared to be anything other than an object or a function; an identifier declared to be a function parameter; a block scope identifier for an object declared without the storage-class specifier `extern`.
>
> **§6.7/3**
> If an identifier has no linkage, there shall be no more than one declaration of the identifier (in a declarator or type specifier) with the same scope and in the same name space, except for tags as specified in 6.7.2.3.

In plainer terms, change this in `land_data_manager.c`:

```c
typedef struct LandDataManager_t {
    // snip
} LandDataManager;
```

to this to avoid the non-standard duplicate `typedef`:

```c
struct LandDataManager {
    // snip
};
```

src/overlay005/land_data.c
lhearachel: A `collision` in my mind should be as simple as "passable" / "impassable". A behavior defines additional patterns applicable to a tile. I would say that the sum of these two aspects (which are different, and stored in different files) forms the tile's attributes.

Kuruyia: Understood, thanks for the clarification. I can indeed also see that in `unk_02054D00.c`:

https://github.com/pret/pokeplatinum/blob/006523df6b56581272c92587bf661c78d8a86c2c/src/unk_02054D00.c#L264-L295

In that case, I don't think `INVALID_TERRAIN_ATTRIBUTES` should be part of the map tile behaviors consts, as it represents both an invalid tile behavior + an invalid collision.

Of course, in the file linked above, an invalid tile behavior is also represented by -1 (line 294), so this is just a matter of semantics. Let me know if you have a strong opinion towards moving that const.

lhearachel (approved):

PR #423, opened 2025-03-09
Documenting UnkStruct_ov19_021D5BAC as PCMonPreview

CharlesFolz4:
Documenting struct UnkStruct_ov19_021D5BAC as PCMonPreview

This struct is specifically for the data backing this preview:
Screenshot_20250308_194117


res/text/box_messages.gmm
```diff
@@ -81,7 +81,7 @@
 		<attribute name="window_context_name">used</attribute>
 		<language name="English">Deposit in which Box?</language>
 	</row>
-	<row id="pl_msg_00000019_00020" index="20">
+	<row id="no_item" index="20">
```
lhearachel, line 84: **suggestion**: Let's keep the `box_messages` prefix here, as `no_item` could mean many things:

```suggestion
	<row id="box_message_no_item" index="20">
```

Or, if we have a better name in mind for this file, then that would also work.

src/overlay019/ov19_021D0D80.c
```diff
@@ -3938,57 +3940,57 @@ static void ov19_021D584C(UnkStruct_ov19_021D5DF8 *param0)
...
-    v0->unk_0D = BoxPokemon_GetValue(boxMon, MON_DATA_TYPE_2, NULL);
-
-    if ((v0->unk_0F == 0) && BoxPokemon_GetValue(boxMon, MON_DATA_NIDORAN_HAS_NICKNAME, NULL)) {
-        v0->unk_0E = SpeciesData_GetGenderOf(speciesData, v0->species, BoxPokemon_GetValue(boxMon, MON_DATA_PERSONALITY, NULL));
+    preview->mon = boxMon;
+    preview->species = BoxPokemon_GetValue(boxMon, MON_DATA_SPECIES, NULL);
+    preview->heldItem = BoxPokemon_GetValue(boxMon, MON_DATA_HELD_ITEM, NULL);
+    preview->dexNum = GetDexNumber(SaveData_GetDexMode(param2->unk_11C), preview->species);
+    preview->isEgg = BoxPokemon_GetValue(boxMon, MON_DATA_EGG_EXISTS, NULL);
+    speciesData = SpeciesData_FromMonSpecies(preview->species, 9);
```
lhearachel, line 3961: **question**: Is this `9` a heap ID?

src/overlay019/ov19_021D0D80.c
```diff
@@ -3938,57 +3940,57 @@ static void ov19_021D584C(UnkStruct_ov19_021D5DF8 *param0)
...
+    preview->level = SpeciesData_GetLevelAt(speciesData, preview->species, BoxPokemon_GetValue(boxMon, MON_DATA_EXP, NULL));
+    preview->markings = BoxPokemon_GetValue(boxMon, MON_DATA_MARKS, NULL);
+    preview->type1 = BoxPokemon_GetValue(boxMon, MON_DATA_TYPE_1, NULL);
+    preview->type2 = BoxPokemon_GetValue(boxMon, MON_DATA_TYPE_2, NULL);
+
+    if ((preview->isEgg == 0) && BoxPokemon_GetValue(boxMon, MON_DATA_NIDORAN_HAS_NICKNAME, NULL)) {
+        preview->gender = SpeciesData_GetGenderOf(speciesData, preview->species, BoxPokemon_GetValue(boxMon, MON_DATA_PERSONALITY, NULL));
     } else {
-        v0->unk_0E = 255;
+        preview->gender = 255;
```
lhearachel, line 3970: **polish**: This looks like `GENDER_RATIO_NO_GENDER` from `generated/gender_ratios.h`.

src/overlay019/ov19_021D0D80.c
```diff
@@ -3938,57 +3940,57 @@ static void ov19_021D584C(UnkStruct_ov19_021D5DF8 *param0)
...
     } else {
-        Strbuf_Copy(v0->unk_18, v0->unk_14);
-        Strbuf_Clear(v0->unk_14);
+        Strbuf_Copy(preview->speciesName, preview->nickname);
+        Strbuf_Clear(preview->nickname);
     }
 
-    if (v0->unk_06 != 0) {
-        Item_LoadName(v0->unk_1C, v0->unk_06, 9);
+    if (preview->heldItem != 0) {
```
lhearachel, line 3982: **polish**:

```suggestion
    if (preview->heldItem != ITEM_NONE) {
```

src/overlay019/ov19_021D0D80.c
```diff
@@ -3235,22 +3237,22 @@ static void ov19_021D4E7C(UnkStruct_ov19_021D4F5C *param0)
...
 
-static void ov19_021D4E88(UnkStruct_ov19_021D5BAC *param0)
+static void PCMonPreviewInit(PCMonPreview *param0)
 {
-    param0->unk_14 = Strbuf_Init(12, HEAP_ID_9);
-    param0->unk_18 = Strbuf_Init(12, HEAP_ID_9);
-    param0->unk_1C = Strbuf_Init(18, HEAP_ID_9);
-    param0->unk_20 = Strbuf_Init(12, HEAP_ID_9);
-    param0->unk_24 = Strbuf_Init(16, HEAP_ID_9);
+    param0->nickname = Strbuf_Init(12, HEAP_ID_9);
```
lhearachel, line 3242: **thought**: Maybe worth updating `HEAP_ID_9` to `HEAP_ID_PC_BOXES`?

lhearachel (commented):
This looks pretty well-contained, good job! I would ordinarily be more of a stickler about getting rid of the `ov19` prefix, but I think that it's fine to keep for now, since we are still documenting this application.

src/overlay019/ov19_021D0D80.c
CharlesFolz4: I saw that, but found it very strange because the other branch of the if pulls from `genders.h`, which has GENDER_NONE as an option, so using `GENDER_RATIO_NO_GENDER` didn't quite seem right.

src/overlay019/ov19_021D0D80.c
CharlesFolz4: That's entirely possible, but I've also seen a lot of `HEAP_ID_10` in the box overlays too.  I think it's worth waiting to know what specifically each does here first before that rename.

src/overlay019/ov19_021D0D80.c
lhearachel: Hm, you're right. Contextually, this should match with `-1`, since it's a `u8`. We could clarify that as `GENDER_INVALID` within this source file; what do you think?

CharlesFolz4: Seems reasonable; didn't see this pattern appearing anywhere else that I could find.

lhearachel (approved):

PR #424, opened 2025-03-09
Document PokemonSprite and related structs

ravepossum:
Documents `PokemonSprite` and several related and sub-structs. `PokemonSprite`s, unlike regular `Sprite`s, do not use hardware sprites and are instead billboarded onto a polygon using 3D graphics. All of the rendering is handled by the `PokemonSpriteManager`. They have a variety of software transformations provided by `PokemonSpriteTransforms` for things like affine scale/rotation, color changes, and drawing partial sprites.

Notes:
- This started as a transfer of documentation from pokeheartgold but diverged pretty significantly as I uncovered more info, so it's a good bit more than just an alignment transfer.
- I didn't document constants/bools/etc in usages of `PokemonSprite_Get/Set/AddAttribute()` in functions that weren't documented yet since there were so many and I figured it'd be appropriate to document those with the function instead.
- This also adds a .naix file for `pl_otherpoke.narc` so the pokemon shadow NCGR and NCLR files could have their entry IDs referred to with constants.

Remaining to-do:
- There's a handful of references to `PokeSprite` and `ArchivedPokeSprite` in other files that I'm planning to revisit and name in a more uniform way.


include/pokemon_sprite.h
```diff
@@ -0,0 +1,237 @@
+#ifndef POKEPLATINUM_POKEMON_SPRITE_H
+#define POKEPLATINUM_POKEMON_SPRITE_H
+
+#include <nnsys.h>
+
+#include "constants/heap.h"
+
+#include "struct_defs/sprite_animation_frame.h"
+
+#define MAX_MON_SPRITES 4
```
lhearachel, line 10: **question**: Is this just the number of faces times the number of genders? Or the number that can be displayed at one time? I wonder if we could expose that more clearly.

include/pokemon_sprite.h
```diff
@@ -0,0 +1,237 @@
...
+} PokemonSpriteShadow;
+
+struct PokemonSprite {
+    u32 active : 1;
+    u32 polygonID : 6;
+    u32 needReloadChar : 1;
+    u32 needReloadPltt : 1;
+    u32 padding_00_9 : 23;
+    PokemonSpriteTemplate template;
+    PokemonSpriteTemplate templateBackup;
```
lhearachel, line 143: **thought**: Is this for handling Transform?

include/pokemon_sprite.h
```diff
@@ -0,0 +1,237 @@
...
+void PokemonSpriteManager_SetExcludeIdentity(PokemonSpriteManager *monSpriteMan, int value);
+BOOL PokemonSprite_IsActive(PokemonSprite *monSprite);
+void PokemonSpriteManager_SetHideShadows(PokemonSpriteManager *monSpriteMan, u32 value);
+void PokemonSpriteManager_ClearHideShadows(PokemonSpriteManager *monSpriteMan, u32 value);
+void PokemonSprite_DrawSpindaSpots(u8 *rawCharData, u32 personality, BOOL isAnimated);
+void PokemonSprite_DecryptPt(u8 *rawCharData);
+void PokemonSprite_DecryptDP(u8 *rawCharData);
+void PokemonSprite_Decrypt(u8 *rawCharData, int narcID);
+
+inline void PokemonSpriteManager_HideShadows(PokemonSpriteManager *monSpriteMan)
```
lhearachel, line 227: **praise**: I love this idea. Great addition.

src/pokemon_sprite.c
```diff
@@ -0,0 +1,1619 @@
...
+#include "res/pokemon/pl_otherpoke.naix"
+
+#define MAX_SPINDA_SPOTS       4
+#define SPINDA_SPOT_COORDS_END 0xFF
+
+#define MAX_MON_SHADOWS         2
+#define MAX_MON_SPRITE_PALETTES (MAX_MON_SPRITES + MAX_MON_SHADOWS)
+
+#define MON_SPRITE_WIDTH_TILES  32
+#define MON_SPRITE_HEIGHT_TILES 32
```
lhearachel, line 26: **thought**: Interesting. I think that we might want to rename these, because these are definitely not the actual dimensions of the in-game sprites. Those are 20x10 (two 80x80 sprites).

src/pokemon_sprite.c
```diff
@@ -0,0 +1,1619 @@
...
+#define MON_AFFINE_SHIFT    8
+#define MON_AFFINE_SCALE(i) (i << MON_AFFINE_SHIFT)
+
+// one particular usage of the PLTT_OFFSET macro in this file doesn't match without the cast
+#define PLTT_OFFSET_CAST(i) ((i) * (u16)PALETTE_SIZE_BYTES)
+
+typedef struct {
+    u8 x;
+    u8 y;
+} SpindaSpotCoords;
```
lhearachel, line 63: **thought**: Do we have a generic U8Coords struct somewhere? This feels a bit too specific.

src/pokemon_sprite.c
```diff
@@ -0,0 +1,1619 @@
...
+    { 45, 41 },
+    { 46, 41 },
+    { 47, 41 },
+    { 43, 42 },
+    { 44, 42 },
+    { 45, 42 },
+    { SPINDA_SPOT_COORDS_END, SPINDA_SPOT_COORDS_END }
+};
+
+static const SpindaSpotCoords *sSpindaSpotCoords[MAX_SPINDA_SPOTS] = {
```
lhearachel, line 310: **suggestion**: A double-array on the type would be more appropriate here than a pointer-array, if it's possible to make it match.

src/pokemon_sprite.c
```diff
@@ -0,0 +1,1619 @@
...
+{
+    return monSprite->transforms.fadeActive == TRUE;
+}
+
+void PokemonSprite_CalcScaledYOffset(PokemonSprite *monSprite, int height)
+{
+    monSprite->transforms.yOffset = ((MON_SPRITE_HEIGHT / 2) - height) - ((((MON_SPRITE_HEIGHT / 2) - height) * monSprite->transforms.scaleY) >> 8);
+}
+
+static inline void RunPokemonSpriteTaskAnim(u8 *active, u8 *currSpriteFrame, u8 *currAnimFrame, u8 *frameDelay, u8 *loopTimers, const SpriteAnimationFrame *animFrames)
```
lhearachel, line 1139: **suggestion**: I think that `Tick` is better verbiage here than `Run`; thoughts?

lhearachel (commented):
I don't have much in the way of feedback about the overall structure. Despite how many commits are here, this is pretty well-contained. Thanks for sticking with this one for so long. :pray:

include/pokemon_sprite.h
ravepossum: This is the max that can be handled by the manager at one time, capping out at the max number of battlers. `MAX_MON_SPRITE_MAN_SPRITES` or something maybe? That's kind of a clunky name, admittedly.

include/pokemon_sprite.h
ravepossum: At a glance, it appears like it's used for both Transform and Substitute.

src/pokemon_sprite.c
ravepossum: They correspond to the dimensions of the char data (the `H` and `W` members of `NNSG2dCharacterData`), so `MON_SPRITE_CHAR_BUF_TILES_H`/`MON_SPRITE_CHAR_BUF_TILES_W` or something maybe?

src/pokemon_sprite.c
ravepossum: I can't quite get something like that to match. pokeheartgold has it written like this:
```c
static const u8 sSpindaSpot1Coords[][2] = {

...

static const u8 (*sSpindaSpotCoords[MAX_SPINDA_SPOTS])[2] = {
    sSpindaSpot0Coords,
    sSpindaSpot1Coords,
    sSpindaSpot2Coords,
    sSpindaSpot3Coords,
};
```
Which does match for us as well, but I think it makes less clear to access that way:
```c
    const u8(*spotCoords)[2];

...

        while (spotCoords[j][0] != 0xFF) {
            x = spotCoords[j][0] + ((pid & 0x0F) - 8);
            y = spotCoords[j][1] + (((pid & 0xF0) >> 4) - 8);
...
```

src/pokemon_sprite.c
ravepossum: Not that I can find, but I could certainly add one with this PR. Just not sure exactly where it should live. My first thought was `sprite_util.h` but it's not inherently sprite-related in any way, and it doesn't really fit with the other contents of that file.

src/pokemon_sprite.c
lhearachel: I like those names. Let's go with those.

src/pokemon_sprite.c
lhearachel: Hm, I would almost say that it can go in `global`, but that's probably too much. Let's leave it here for now and we can reassess later if we bump into another need for such a struct.

ravepossum:
All feedback should be addressed now.

lhearachel (approved):

PR #431, opened 2025-03-13
Document Egg Moves

AsparagusEduardo:
Using pokeemerald's structure, I didn't feel confident on how to apply it to species data json files ^^;


AsparagusEduardo:
(force pushed to fix format)

lhearachel (approved):
I don't mind keeping this data in C for the time being, but we do need to come up with a plan to migrate it to the species files. But, that doesn't seem like it will be too hard. Documented is better than not. :slightly_smiling_face:

PR #432, opened 2025-03-13
Sound_PlayEffect literals to constants

AsparagusEduardo:
Note: I removed redundant calls to `#include "generated/sdat.h"`, as it's being added in `include/unk_02005474.h`


lhearachel (approved):
Quick scan through the diffed files didn't yield anything suspicious-looking, so I trust the build here, since this is all constants-replacement.

PR #434, opened 2025-03-15
Start Markdown documentation of the maps system

Kuruyia:
This starts a human-readable documentation of the various subsystems used in the game's map system. The following files are included in this commit:

- `maps.md`: a general overview of maps
- `bdhc.md`: an overview of the BDHC subsystem, which is used to provide height information on maps
- `loading_maps.md`: an overview of how maps are loaded by the game
- `file_format_specifications.md`: specs for the maps data structures

The `file_format_specifications.md` is meant to be a purely technical document that contains a precise description of the data structures used by the game's map system.

The other files are meant to be semi-technical documentation, where some of the nitty-gritty details have been either simplified or left out entirely.

Some of the maps subsystems were purposefully left out of this commit, such as map objects and map scripts. They will need to be addressed in future updates to those documents.


docs/maps/README.md
```diff
@@ -0,0 +1,54 @@
+# Map documentation
+
+This section of the documentation aims to provide an overview of the different
+map-related subsystems, the interaction between them, and how they load their
+data from the ROM so that, in the end, the game can seamlessly render a world
+in which the player evolves.
```
lhearachel, line 6: **quibble**: `in which the player moves.`?

docs/maps/README.md
```diff
@@ -0,0 +1,54 @@
...
+- **Map header**: a structure that contains metadata about a map, such as its
+  music, its name, whether cycling is allowed, etc. Each map is associated with
+  a map header, and has a specific, well-known ID.
+- **Map matrix**: a 2D array of maps that form a playable area. This system allows
+  the game to dynamically load and unload maps as the player moves around the world.
+  The overworld is one such example of a map matrix, containing 30x30 maps.
+- **Map object**: an entity that is placed on the map. Examples include the player
+  and NPCs.
+- **Map prop**: a 3D model that is placed on the map. An example is the various
+  buildings in the game.
```
lhearachel, line 24: **suggestion**: It may be worthwhile listing other things which are props that may surprise a player, e.g., interior furniture, honey trees, etc.

docs/maps/README.md
```diff
@@ -0,0 +1,54 @@
...
+  (think archive as in a `.tar` file, a file that contains files)
+- **NSBCA/NSBTA**: a file format used to store animations for 3D models for the
+  Nintendo DS. See [apicula's documentation](https://raw.githubusercontent.com/scurest/nsbmd_docs/master/nsbmd_docs.txt)
+  for more information.
+- **NSBMD**: a file format used to store 3D models, textures and palettes for the
+  Nintendo DS. See [apicula's documentation](https://raw.githubusercontent.com/scurest/nsbmd_docs/master/nsbmd_docs.txt)
+  for more information.
+- **NSBTX**: a file format used to store textures and palettes for the
+  Nintendo DS. See [apicula's documentation](https://raw.githubusercontent.com/scurest/nsbmd_docs/master/nsbmd_docs.txt)
+  for more information.
```
lhearachel, line 35: **suggestion**: Long-term, I think that it would be good for us to mirror this documentation to ensure its longevity. The worst thing that can happen for our docs is for them to wind up with broken links. Obviously, a Credits / Acknowledgements section is an exception to this.

docs/maps/bdhc.md
```diff
@@ -0,0 +1,268 @@
+# BDHC
+
+This document aims to provide an overview of the BDHC subsystem, which is used to provide height information on maps.
```
lhearachel, line 3: **suggestion**: I feel that the `Introduction` section below serves a better overview of the contents of this document, and that this line is superfluous. What do you think?

docs/maps/bdhc.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Introduction
+
+The BDHC subsystem is used to calculate the height of the terrain. It is used for
+collision detection and the map objects height, for example.
+
+It is basically Pokémon's way of adding height to an otherwise 2D engine (at
+least, from the perspective of the map).
+
+BDHC data is stored alongside each map data.
```
lhearachel, line 13: **suggestion**:

```suggestion
The BDHC subsystem is used to calculate the height of the terrain. It is used for
collision detection and enabling height in an otherwise 2D movement system. This
data is stored alongside individual maps.
```

docs/maps/bdhc.md
```diff
@@ -0,0 +1,268 @@
...
+
+> [!NOTE]
+> Throughout this document, we consider that a 2D point is represented by a
+> pair of coordinates $(x, z)$. This allows more understandable conversions
```
lhearachel, line 17: **thought**: It may be worthwhile for us to explain the sourcing of the `X, Z, Y` coordinate system (being that it is right-handed and Y-up).

docs/maps/bdhc.md
```diff
@@ -0,0 +1,268 @@
...
+The animation below represents a plate, in blue, and Dawn moving on it. The height
+of Dawn is calculated as she is moving on the plate.
+
+https://github.com/user-attachments/assets/10251494-9bd6-41d4-b2b7-866d5b784109
+
+### How to calculate the height of a point?
+
+This section is targeted at people who are interested in the math necessary to
+calculate the height of a point on a plate. For instance, this can be useful for
+implementing a game engine.
```
lhearachel, line 63: **suggestion**: Rather than repeat this information for each of the following subsections, I would group them beneath a parent section that contains this disclaimer, so that users know that they can skip the entire section if they are not interested.

docs/maps/bdhc.md
```diff
@@ -0,0 +1,268 @@
...
+$$
+\begin{align*}
+d &= -\hat{n} \cdot A \\
+\Leftrightarrow \quad d &= -\hat{n}_x * A_x - \hat{n}_y * A_y - \hat{n}_z * A_z
+\end{align*}
+$$
+
+## Strips
+
+Instead of having to check every plate in the map to find the ones that contains
```
lhearachel, line 212: **typo**:

```suggestion
Instead of having to check every plate in the map to find the ones that contain
```

docs/maps/bdhc.md
```diff
@@ -0,0 +1,268 @@
...
+a map object (some complex map can reach over 100 plates!), the game engine uses
+strips to speed up the process.
+
+Strips are basically a list of plates that are all intersecting a given Z coordinate.
+This Z coordinate is stored in the strip, alongside its plates, and is called the
+"scanline".
```
lhearachel, line 218: **suggestion**: I would be careful with this terminology, as it can mean something *very* different in computer graphics: https://en.wikipedia.org/wiki/Scan_line

docs/maps/file_format_specifications.md
```diff
@@ -0,0 +1,376 @@
...
+The following sections describe the various file formats encountered in the files
+containing map data.
+
+Values and offsets described in this document are those found in the vanilla ROM
+of Pokémon Platinum.
+
+Here's a graph representing the different files and how they point to data in
+other files:
+
+```mermaid
```
lhearachel, line 12: **question**: This is cool! Is this a Github specific extension? Does it work with other renderers?

docs/maps/file_format_specifications.md
```diff
@@ -0,0 +1,376 @@
...
+- `VecFx32`: structure containing three `fx32` values, representing a 3D vector.
+
+Moreover, for any type `T`, `T[]` represents an array of `T`. Any other type
+should be documented in the context where it is used.
+
+## Map headers
+
+The map headers are not stored in a file on the ROM, but are instead hard-coded
+in the game's code. Because of it is the starting point for loading a map, it
+is nevertheless included in this list.
```
lhearachel, line 69:
```suggestion
The map headers are not stored in a file on the ROM, but are instead stored in
a table inside the code binary. Because it is the starting point for loading a
map, it is nevertheless included in this list.
```

docs/maps/file_format_specifications.md
```diff
@@ -0,0 +1,376 @@
...
+    u16 isFlyAllowed : 1;
+};
+```
+
+## Map matrix (`map_matrix.narc`)
+
+This NARC contains 289 files.
+
+Each file size is a multiple of 4 bytes, padding with `0xFF` at the end if
+necessary.
```
lhearachel, line 103: **question**: Is this the case, or potentially misleading? The NARC specification itself defines that member files are padded to word-boundaries with `0xFF`.

docs/maps/loading_maps.md
```diff
@@ -0,0 +1,253 @@
...
+the player is in.
+
+For instance, if the player is standing in the top-left quadrant of the current
+map, this means that they are standing in the bottom-right quadrant of the loaded
+maps quadrants.
+
+In other words, this means that the maps loaded will be the map above the player,
+on the left of the player, and the map on the top-left of the player.
+
+Here's an illustration of this:
```
lhearachel, line 35: **suggestion**: If possible, I would like for us to make this diagram more colorblind-accessible.

docs/maps/maps.md
```diff
@@ -0,0 +1,157 @@
...
+This document aims to provide an overview of how the playground is constructed,
+excluding BDHC, which is covered in [its own document](bdhc.md).
+
+## Introduction
+
+Maps are the main building block of the playground. They are defined by various
+properties, which are explored throughout this document.
+
+Maps in Pokémon Platinum are a bit of a hybrid between 2D and 3D: they retain a
+similar 2D grid system that were used in previous games, on which the terrain attributes
```
lhearachel, line 12: **polish**:

```suggestion
2D grid system that was used in previous games, on which the terrain attributes
```

docs/maps/maps.md
```diff
@@ -0,0 +1,157 @@
...
+dictates the collision and behavior of each tile, and on which map objects (such
```
lhearachel, line 13: **polish**:

```suggestion
dictate the collision and behavior of each tile, and on which map objects (such
```

docs/maps/maps.md
```diff
@@ -0,0 +1,157 @@
...
+as the player) are constrained, but they also have 3D elements, such as the map
```
lhearachel, line 14: **grammar**: This is a bit of a run-on sentence.

```suggestion
as the player) are constrained, but they also have 3D elements, such as the map
```

docs/maps/maps.md
```diff
@@ -0,0 +1,157 @@
...
+### Map data
+
+Stored in another archive inside the ROM, the map data contains the bulk of
+what's needed to render a map and have a player interact with it.
+
+The map data contains the following:
+
+- attributes for each tile of the map, which dictates whether a tile can be
+  walked on (collision data), and how the tile behaves when walked on: is it
+  tall grass, water, a trash can, etc.
```
lhearachel, line 63: **suggestion**: The trash-can example is interesting, because those typically cannot be walked on; I would generalize this statement to include player interactions.

docs/maps/README.md
Kuruyia: Sure. [Apicula's license](https://github.com/scurest/apicula/blob/master/LICENSE) is pretty permissive so I don't think there should be any problem in doing so.
cc @scurest

docs/maps/bdhc.md
Kuruyia: It was there mostly to be consistent with the other files. I don't mind removing it.

docs/maps/bdhc.md
Kuruyia: Could you expand on that?

lhearachel: The coordinate system used by the game, as far as I'm aware, is right-handed and defines Y+ as "up". This is aligned with the models being built in Maya, which I believe we know for certain that they were. It's convenient that it uses the same system as Godot.

Here are some comparisons that I found, for reference:

image

docs/maps/bdhc.md
Kuruyia: This is basically what I had mind when seeing how the strips were generated, basically "scanning" from top to bottom for all Z coordinate that start/end a plate.

Do we have a better name for this? We could simply call this something along the lines of "the strip's `z`", but I'm worried it could be mixed up with the `z` coordinate of BDHC points.

From [Mikelan's original post](https://projectpokemon.org/home/forums/topic/37816-gen-iv-bdhc-files-terrain-settings), here's the description of that field:

> The structure of each element of Part T is:
>
> 00 00 YY YY NN NN UU UU
>
> Where YY YY is the Y-axis line where the lower limit of the strip is.

My original interpretation of this description was that this field contained the "bottom-most" Z coordinate of all plates in the strip (in other words, the max value of the Z coord of all bottom-right points), which is definitely not the case. Although it could have been just me misunderstanding this description.

(in the code, we currently documented this field as [`lowerBound`](https://github.com/pret/pokeplatinum/blob/5c4943582b158c01cc2cb538a1acb09f00adf492/include/overlay005/bdhc.h#L23))

docs/maps/file_format_specifications.md
Kuruyia: Not sure how widely implemented this is, but it also works with at least GitLab and the Markdown extension I use in VSCode ^^

docs/maps/file_format_specifications.md
Kuruyia: If I'm not mistaken, 237 files out of the 289 in this NARC require padding.

Didn't know this was part of the NARC specification. Should we instead add a warning at the beginning of the file, or leave this out entirely?

lhearachel: I think leaving it out is fine, or specifying that the structure sizes stored in the filesystem are padded to word-length by the NARC specification.

docs/maps/loading_maps.md
Kuruyia: Sorry, I'm not really knowledgeable about this ^^

Would adding text work? Thinking about something like this:
image

Should the text be black?

lhearachel: I think the text works well enough as-is. 😁

docs/maps/maps.md
Kuruyia: Looks like your suggestion didn't register properly :)

lhearachel: My bad. 😅 The comment remains, though. It would be good to split it up a bit for readability.

docs/maps/bdhc.md
Kuruyia: Thanks for the reference!

Added this bit at the end of the note:
```
The game's coordinate system follows the right-hand rule, with the Y axis pointing upwards.
```

docs/maps/file_format_specifications.md
Kuruyia: I decided on leaving a note at the beginning of the document:

```markdown
> [!NOTE]
> As per the NARC specification, the size of each file contained inside one of
> those archives must be a multiple of 4 bytes. If this is not the case, the
> file is padded with `0xFF` bytes at the end to reach this.
```

docs/maps/bdhc.md
Kuruyia: Done that, and moved everything to the end of the doc.

lhearachel (approved):
We can leave the request to copy `apicula`'s documentation from scurest open for now and revisit it at a later time. I don't want that to block getting this merged in.

Thanks for such thorough detail, the animations are excellent for helping understand what's going on. :pray:

docs/maps/README.md
scurest: Please feel free, although that document is highly imperfect.

PR #437, opened 2025-03-17
Dump map data binaries, texture sets, and matrices

ravepossum:
# This PR:
- Only dumps files; it does not document them.
- Dumps map data files that contain 3D models, tile collisions, tile behavior, and BDHC data as binary files. The thought here is that these will likely stay as binaries since they'll be handled by a map editor.
- Dumps map texture sets as BTXO files.
- Generates map ID constants to be used in map matrices.
- Dumps map matrices as json files in the following format:
```json
{
    "name": "m_dun0602_",
    "headers": [
        ["MAP_HEADER_EVERYWHERE", "MAP_HEADER_EVERYWHERE", "MAP_HEADER_EVERYWHERE", "MAP_HEADER_EVERYWHERE"],
        ["MAP_HEADER_EVERYWHERE", "MAP_HEADER_GREAT_MARSH_1", "MAP_HEADER_GREAT_MARSH_2", "MAP_HEADER_EVERYWHERE"],
        ["MAP_HEADER_EVERYWHERE", "MAP_HEADER_GREAT_MARSH_3", "MAP_HEADER_GREAT_MARSH_4", "MAP_HEADER_EVERYWHERE"],
        ["MAP_HEADER_EVERYWHERE", "MAP_HEADER_GREAT_MARSH_5", "MAP_HEADER_GREAT_MARSH_6", "MAP_HEADER_EVERYWHERE"],
        ["MAP_HEADER_EVERYWHERE", "MAP_HEADER_EVERYWHERE", "MAP_HEADER_EVERYWHERE", "MAP_HEADER_EVERYWHERE"]
    ],
    "altitudes": [
        [2, 2, 2, 2],
        [2, 0, 0, 2],
        [2, 0, 0, 2],
        [2, 0, 0, 2],
        [2, 2, 2, 2]
    ],
    "maps": [
        ["MAP_176", "MAP_176", "MAP_176", "MAP_176"],
        ["MAP_176", "MAP_570", "MAP_571", "MAP_176"],
        ["MAP_176", "MAP_572", "MAP_573", "MAP_176"],
        ["MAP_176", "MAP_574", "MAP_575", "MAP_176"],
        ["MAP_176", "MAP_176", "MAP_176", "MAP_176"]
    ]
}

```

The binaries also contain matrix width, height, and name length fields but those are inferred from the contents of the map 2D lists.

# Still todo in follow-up efforts:
- Map prop files need to be dumped.
- I'm planning to write a brief guide to insert maps using PDSMS while we still don't have a dedicated map editor.


tools/json2bin/map_matrix.py
```diff
@@ -0,0 +1,35 @@
...
+import pathlib
+import sys
+
+from convert import from_map_header, from_map, u8, u16, utf8
+
+def pack_matrix_list(data, type_func, from_func) -> bytes:
+    ret = []
+    for row in data:
+        ret += map(type_func, map(from_func, row))
+    return b"".join(ret)
```
lhearachel, line 12: **suggestion**:

```suggestion
    return b"".join([
        map(type_func, map(from_func, row))
        for row in data
    ])
```

ravepossum:
> What's the rationale behind that? To me, map matrices would also be edited with a map editor, but those were converted to JSON files.
> 
> Is there an additional reason behind this conversion?

I thought it'd be easier for the hypothetical future editor to parse it that way (and it's easily representable in that manner, unlike the 3D models). 

Also part of the reason I went ahead and dumped this stuff now is to make modding easier - It allows a modder to edit them easily now without having to use something like DSPRE.

lhearachel:
> It allows a modder to edit them easily now without having to use something like DSPRE.

I would say that "easily" is doing some heavy lifting here. They are certainly editable in this fashion -- and moreso than in binary format -- but a graphical application will still obviously be superior.

ravepossum:
Right, I definitely don't intend to suggest editing the .jsons directly as an end goal, but some potentially tedious text editing certainly beats hex editing or unpacking, editing, and then repacking a binary file for now.

Kuruyia:
Thanks!

The reason I brought this up is that, while the 3D models are indeed fairly complex to represent as text data, I think the other stuff present in the map data could easily be represented in a JSON format, and may gain from some "abstraction":

- terrain attributes could be split in a boolean for collision and in a string constant for each existing behavior
- map prop instances could:
  - have their model ID referenced as a string constant
  - have their VecFx32 parameters (position/rotation/scale) be represented as plain numbers for easier editing (although we'd still like to constraint them to legal fixed-point values at some point in the pipeline)
- BDHC data can theoretically be represented only by their plates (although in practice, there are a lot of normals/constants duplicated for seemingly no reason in the game's vanilla maps, so we would still need to store those as well) - this one is the most complex apart from 3D models

3D models would live in their own binary files, referenced by the appropriate JSON file.

Anyway, I think this PR is fine as-is. We'll see if the need to convert the map data to text arise in the future or not.

lhearachel:
> The reason I brought this up is that, while the 3D models are indeed fairly complex to represent as text data, I think the other stuff present in the map data could easily be represented in a JSON format, and may gain from some "abstraction"

Just to clarify: I *do* think that there is value in thinking about this with regards to our own mapping application, but probably not much when considering the use of PDSMS as an intermittent editor until such an application is production-ready. Given the goals here, yeah, I agree that this is in a good state for now. 👍

tools/json2bin/map_matrix.py
ravepossum: Couldn't get the list comprehension approach to not throw an error - per discord discussion, keeping as-is.

tools/json2bin/map_matrix.py
```diff
@@ -0,0 +1,35 @@
...
+
+def parse_matrix(data) -> bytes:
+    return b"".join([
+        u8(0 if len(data["maps"]) == 0 else len(data["maps"][0])), # width
+        u8(len(data["maps"])), # height
+        u8(len(data["headers"]) != 0), # boolean indicating the matrix contains headers
+        u8(len(data["altitudes"]) != 0), # boolean indicating the matrix contains altitudes
+        u8(len(data["name"])),
+        utf8(data["name"]), # matrix name
```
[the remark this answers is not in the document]
ravepossum: Sounds good, changed the `utf8` function to `ascii`.

ravepossum:
Force push 1 to change `utf8()`->`ascii()`, force push 2 to rename map texture set extensions from `.BTX0` -> `.nsbtx`. Should be ready for re-review.

lhearachel (approved):

PR #394, opened 2025-02-21
Use a flag to forbid using ScriptEntry after ScriptEntryEnd

lhearachel:
This is a pretty simple change that alleviates a potential foot-gun scenario where a user could compile a field script that contained a `ScriptEntry` directive after terminating the entry-points table with `ScriptEntryEnd`. Such a field script previously would assemble totally fine and be perfectly usable in-game, but would result in entry-points within that file that cannot actually be used.

There is still a bit of a foot-gun scenario here where a user could totally just set this flag back to `TRUE` and add another entry anyways, but I feel that the documentation is appropriate enough to warn against them doing that.

### Add a new `ScriptEntry` to a file before `ScriptEntryEnd`

```diff
diff --git a/res/field/scripts/scripts_pokemon_center_daily_trainers.s b/res/field/scripts/scripts_pokemon_center_daily_trainers.s
index 7a33e5227..f7aa43662 100644
--- a/res/field/scripts/scripts_pokemon_center_daily_trainers.s
+++ b/res/field/scripts/scripts_pokemon_center_daily_trainers.s
@@ -24,6 +24,7 @@
     ScriptEntry PokemonCenterDailyTrainers_CheckUnlockedVSSeeker
     ScriptEntry PokemonCenterDailyTrainers_FirstNPC
     ScriptEntry PokemonCenterDailyTrainers_SecondNPC
+    ScriptEntry PokemonCenterDailyTrainers_SecondNPC
     ScriptEntryEnd
 
 PokemonCenterDailyTrainers_CheckUnlockedVSSeeker:
```

```console
$ make rom
ninja -C build data
ninja: Entering directory `build'
[100% 2/2] Generating res/field/scripts/scr_seq.narc with a custom command
ninja -C build pokeplatinum.us.nds
ninja: Entering directory `build'
[100% 4/4] Generating pokeplatinum.us.nds with a custom command
```

### Add a new `ScriptEntry` to a file after `ScriptEntryEnd

```diff
diff --git a/res/field/scripts/scripts_pokemon_center_daily_trainers.s b/res/field/scripts/scripts_pokemon_center_daily_trainers.s
index 7a33e5227..19072747c 100644
--- a/res/field/scripts/scripts_pokemon_center_daily_trainers.s
+++ b/res/field/scripts/scripts_pokemon_center_daily_trainers.s
@@ -26,6 +26,8 @@
     ScriptEntry PokemonCenterDailyTrainers_SecondNPC
     ScriptEntryEnd
 
[...]

PR #438, opened 2025-03-17
Document script commands using functions from `overlay005/field_menu`

Gudf:
This includes script commands for windows showing the different currencies (money, coins, BPs) and elevator floors-related commands, as well as commands used to build menus through scripts.

The menu builder commands could use some streamlining, with all the different variants for regular `Menu`s, `ListMenu`s, the different text banks (current script's text bank/`common_strings_2`), the various `Show(List)Menu` variants... so I'm opening this PR as a draft to discuss the implementation.

My current idea would be to have a single `InitMenuBuilder` with additional, optional, arguments for the menu type, the text bank, possibly the number of columns (for `Menu`s)/the width (for `ListMenu`s) and the horizontal and vertical alignments, which sets assembler variables used to determine which `Add(List)MenuEntry` and `Show(List)Menu` variants to use, and replace these respectively with single `AddEntry` and `ShowMenu` commands that pick the variant accordingly.

However, I can see three main issues with this:
- The commands need to be in the right order in script files, which is always the case in vanilla afaict, but is something users of the repository would have to pay attention to.
- Vanilla uses two versions of `AddMenuEntry`: the regular `AddMenuEntry` which supports variables and large (2-byte) string IDs/entry indices, and what I've called `AddMenuEntryImmediate`, which only supports 1-byte arguments. Unfortunately, I don't think there's a good way to pick which one to use while still matching.
- `AddListMenuEntry` can take three arguments: the string ID for the option, the entry index, and the string ID for the option's description (for example in the PC menu), whereas `AddMenuEntry` only takes the first two. This could be confusing if merged into a single command.


asm/macros/scrcmd.inc
```diff
@@ -406,66 +406,91 @@
...
+    .short \selectedOptionVar
     .endm
 
-    .macro ScrCmd_042 arg0, arg1
+    /*
+     * AddMenuEntry variant that only accepts single-byte values,
+     * and therefore can't use variables or string IDs larger than 0xff,
+     * unlike the regular AddMenuEntry.
+     */
+    .macro AddMenuEntryImmediate entryStringID, entryIndex
```
lhearachel, line 442: **suggestion**: There's an option here to abbreviate `Immediate` to `Imm`. Non-technical users probably won't know the difference (or even what `Immediate` really means), but technically-oriented users may appreciate the familiar abbreviation from ARM assembly.

asm/macros/scrcmd.inc
```diff
@@ -406,66 +406,91 @@
     .short 63
     .endm
 
-    .macro ScrCmd_040 arg0, arg1, arg2, arg3, arg4
+    /*
+     * selectedOptionVar is set to 0xeeee as part of initializing a FieldMenuManager.
+     * Once ShowMenu or one of its variants is used, selectedOptionVar is set to the
+     * entryIndex of the option the player selected.
+     */
+    .macro InitGlobalTextMenu anchorX, anchorY, initialCursorPos, canExitWithB, selectedOptionVar
```
lhearachel, line 414: **thought**: We could be cheeky, since menus like these have a sort of scope:

```
.macro InitTextMenu anchorX, anchorY, initialCursorPos, canExitWithB, selectedOptionVar
.short 64
.byte \anchorX
.byte \anchorY
.byte \initialCursorPos
.byte \canExitWithB
.short \selectedOptionVar
.set F_ACCEPT_MENU_ENTRIES, TRUE
.endm

.macro AddMenuEntryImmediate entryStringID, entryIndex
.if F_ACCEPT_MENU_ENTRIES == TRUE
    .short 66
    .byte \entryStringID
    .byte \entryIndex
.else
    .error "cannot specify MenuEntry outside a TextMenu scope
.endif
.endm

.macro ShowMenu
.short 67
.set F_ACCEPT_MENU_ENTRIES, FALSE
.endm
```

We implemented similar in #394; see the description there for an example of what the error message looks like.

lhearachel (commented):
For posterity, some usage-statistics of these commands (`rg` is `ripgrep`, `rg -c` prints the number of matches in a given file):

```console
> rg -c 'InitGlobalTextMenu' res/field/scripts | awk -F':' '{ sum += $2 } END { print sum }'
64

> rg -c 'InitLocalTextMenu' res/field/scripts | awk -F':' '{ sum += $2 } END { print sum }'
39
```

Since `Local` is the less common of the two, we might be able to get away with dropping the `Global` midfix.

> The commands need to be in the right order in script files, which is always the case in vanilla afaict, but is something users of the repository would have to pay attention to.

I have some thoughts on this in one of my comments below.

asm/macros/scrcmd.inc
```diff
@@ -677,31 +702,31 @@
     .long \arg1
     .endm
 
-    .macro ScrCmd_072 arg0, arg1
+    .macro ShowMoneyWindow tilemapLeft, tilemapTop
```
lhearachel, line 705: **suggestion**: I think just `ShowMoney` is sufficient. Similar for `ShowCoins` below.

asm/macros/scrcmd.inc
```diff
@@ -1610,17 +1635,20 @@
     .short \arg4
     .endm
 
-    .macro ScrCmd_11C arg0
+    // destVar will be set to the number of floors above the current floor for maps where there's an elevator, or 1 when there isn't
+    // The result is hardcoded per-map ID in overlay005/field_menu.c
```
lhearachel, line 1639: **polish**: List the function/table name instead of the filename; the latter is a bit more brittle.

asm/macros/scrcmd.inc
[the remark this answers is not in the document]
Gudf: The issue I have with `topleft` is that it doesn't always end up being the actual top-left corner (when `SetMenuHorizAnchor 1`/`SetMenuVertAnchor 1` is called). Maybe `origin`?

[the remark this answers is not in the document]
lhearachel: `origin` would also work. Let's go with that.

Gudf: Hmm, turns out I was wrong and sometimes they're out of order in vanilla too. See for example [`scripts_battle_factory.s`](https://github.com/pret/pokeplatinum/blob/0497e6082321efb52fc31ba36a5d114ef2b63db4/res/field/scripts/scripts_battle_factory.s). Line 62 is `ShowMenu` , but the corresponding `InitLocalTextMenu` is at line 376/384, via a call to label `_057A`/`_0592`... 

I still think we can do better than just the raw script commands for users, but this means we won't be able to use the improved system everywhere in vanilla.

lhearachel: That is so odd... How does that invocation with the chain of `_0059` -> `_0091` -> `_00B7` actually work? Is `ScrCmd_313` related somehow?

Gudf: `_0059` sets var `0x4004` to 0, which results in `_057A` being called in `_00B7` at line 58, and `_057A` has the init (`ScrCmd_041`), so when it returns to `_00B7` the menu is properly initialized. No idea what `ScrCmd_313` does exactly, but it's unrelated (it checks for space in some heaps, and honestly looks like a debug command to crash on memory leaks or smth?)

lhearachel: I see, alright. Well, maybe we just leave these macros without the flag-setting idea, then. C'est la vie. Users will just have to be careful. 😓

Gudf: Yeah... we could make a set of unchecked macros for these cases (there are like a dozen from what I can tell?) but also adding entries conditionally seems like a common enough use case that it would be more annoying that necessary? I propose we skip it for this PR, we can always come back to it later or leave handling that to a possible future higher-level scripting language.

asm/macros/scrcmd.inc
```diff
@@ -4560,14 +4590,14 @@
...
-    .byte \arg0
+    .byte \isRightSide
     .endm
 
-    .macro ScrCmd_33B arg0
+    .macro SetMenuOriginIsBottom isBottom
     .short 827
-    .byte \arg0
+    .byte \isBottom
     .endm
```
lhearachel, line 4601: **suggestion**: These are a bit awkward to read in-code. Could we do something like this instead?

```asm
    .macro SetMenuXOriginToLeft
    .short 826
    .byte FALSE
    .endm

    .macro SetMenuXOriginToRight
    .short 826
    .byte TRUE
    .endm

    .macro SetMenuYOriginToTop
    .short 827
    .byte FALSE
    .endm

    .macro SetMenuYOriginToBottom
    .short 827
    .byte TRUE
    .endm
```

Obviously with some comments to link them back to the actual script-command routines.

res/field/scripts/scripts_battle_arcade.s
```diff
@@ -167,12 +167,12 @@ _02A3:
 
 _02F1:
     Message 43
-    ScrCmd_040 30, 1, 0, 1, 0x800C
-    ScrCmd_33A 1
-    ScrCmd_042 13, 0
-    ScrCmd_042 14, 1
-    ScrCmd_042 5, 2
-    ScrCmd_043
+    InitGlobalTextMenu 30, 1, 0, TRUE, 0x800C
```
lhearachel, line 170: **thought**: I had to go back to the definition to remember what `TRUE` means here. Should we provide an alternate constant to signify `EXIT_ON_B`? We could always provide it as the final parameter and specify a default for `FALSE` when it is not given, which would improve the semantics a bit:

```asm
    InitGlobalTextMenu 30, 1, 0, 0x800C, EXIT_ON_B ; Exits when B is pressed
    InitGlobalTextMenu 30, 1, 0, 0x800C ; Does not exit when B is pressed
```

res/field/scripts/scripts_battle_frontier_gate_to_fight_area.s
```diff
@@ -53,15 +53,15 @@ _007C:
...
-    ScrCmd_042 11, 0
-    ScrCmd_042 16, 1
-    ScrCmd_042 12, 2
-    ScrCmd_042 13, 3
-    ScrCmd_042 14, 4
-    ScrCmd_042 15, 5
-    ScrCmd_043
+    InitLocalTextMenu 31, 5, 0, TRUE, 0x800C
+    SetMenuOriginIsRight TRUE
+    AddMenuEntryImm 11, 0
```
lhearachel, line 58: **thought**: Do these entries always increment by 1 from 0 in vanilla, and are they always immediately followed by `ShowMenu`? If so, we might be able to improve the semantics with a temporary counter-value that increments a value specified by `.set`, i.e.,

```asm
    .macro AddMenuEntryImm entryStringID
    .ifdef MENU_ENTRY_INDEX
        .set MENU_ENTRY_INDEX, MENU_ENTRY_INDEX + 1
    .else
        .set MENU_ENTRY_INDEX, 0
    .endif

    .short 66
    .byte \entryStringID
    .byte MENU_ENTRY_INDEX
    .endm

    .macro ShowMenu
    .set MENU_ENTRY_INDEX, -1 ; Set to -1 so that the next AddMenuEntryImm increments to 0
    .short 67
    .endm
```

Granted, I haven't tried this, so it might not work how we want it to; hopefully the idea is at least clear enough to follow.

asm/macros/scrcmd.inc
```diff
@@ -3650,17 +3680,17 @@
     .short \destVarID
     .endm
 
-    .macro ScrCmd_294 arg0, arg1
+    .macro ShowBP tilemapLeft, tilemapTop
```
lhearachel, line 3683: **suggestion**: Prefer to expand this abbreviation (`BattlePoints`).

src/overlay005/field_menu.c
```diff
@@ -731,11 +731,11 @@ void FieldMenu_PrintMoneyToWindow(FieldSystem *fieldSystem, Window *window)
     Window_ScheduleCopyToVRAM(window);
 }
 
-Window *FieldMenu_CreateCoinsWindow(FieldSystem *fieldSystem, u8 tilemapLeft, u8 tilemapTop)
+Window *FieldMenu_CreateCoinWindow(FieldSystem *fieldSystem, u8 tilemapLeft, u8 tilemapTop)
```
lhearachel, line 734: **polish**: Synonymize this name with `FieldMenu_DeleteCoinBPWindow` so that their association is clear.

As an additional thought, we could also name it `FieldMenu_CreateSpecialCurrencyWindow`.

I also noticed that this performs the draw and print routines, so I think that we should prefer to reflect that in this routine's name: `FieldMenu_Draw...`

lhearachel (commented):
Thanks for sticking this one out, I know that it's been a doozy. I have a few more thoughts below about ways that we can improve some of the macro semantics, but the rest of this looks great. :heart:

res/field/scripts/scripts_battle_frontier_gate_to_fight_area.s
Gudf: Unfortunately, no. First of all, we would essentially run into the same issue we do with the context idea, where sometimes they're out of order because GF calls subroutines that come much later in the file. If we go back to the same `scripts_battle_factory.s` example, the subroutines both add one or two entries which wouldn't be accounted for when adding entries at lines 60 and 61.
Also, there are just times where they aren't sequential, see [scripts_unk_406](https://github.com/pret/pokeplatinum/blob/d37cbc9b4bac6ab858122f1ebe4bb33ad8a25784/res/field/scripts/scripts_unk_0406.s#L41), for example.

Plus, this seems like a bit of a potential footgun? Since now you'd have to keep track of the order of your options to know what value to check for when branching afterwards instead of being able to use descriptive constants.

lhearachel: Unfortunate that vanilla isn't consistent enough to even permit the consideration, but your reasoning that it's a potential footgun makes sense. :heart:

res/field/scripts/scripts_battle_arcade.s
Gudf: Sounds reasonable. I'll have it default to `TRUE` instead, since that's way more common (114 times, versus 8 times only where it's `FALSE`), and make a NO_EXIT_ON_B constant.

src/overlay005/field_menu.c
lhearachel: That sounds good to me!

lhearachel (approved):

PR #441, opened 2025-03-19
Forms/Species-related functions and scripts

AsparagusEduardo:


AsparagusEduardo:
Stopping for now, ready for review :)

res/field/scripts/scripts_rotoms_room.s
```diff
@@ -173,316 +179,316 @@ _0325:
...
     WaitFadeScreen
-    GoToIfEq 0x8000, 0xFF, _03F3
-    ScrCmd_198 0x8000, 0x800C
-    GoToIfEq 0x800C, 0, _045D
-    GoToIfNe 0x800C, 0x1DF, _0468
-    ScrCmd_305 0x8000, 0x800C
+    GoToIfEq VAR_PARTY_SLOT, 0xFF, RotomsRoom_End
+    GetPartyMonSpecies VAR_PARTY_SLOT, 0x800C
+    GoToIfEq 0x800C, 0, RotomsRoom_AnEggCantEnterAMotor
+    GoToIfNe 0x800C, 0x1DF, RotomsRoom_ThatPokemonCantEnterAMotor
```
lhearachel, line 220: **polish**:

```suggestion
    GoToIfEq 0x800C, SPECIES_NONE, RotomsRoom_AnEggCantEnterAMotor
    GoToIfNe 0x800C, SPECIES_ROTOM, RotomsRoom_ThatPokemonCantEnterAMotor
```

Not sure why they return `0` for eggs, but...

include/constants/forms.h
```diff
@@ -18,35 +18,64 @@
...
+#define DEOXYS_FORM_DEFENSE 2
+#define DEOXYS_FORM_SPEED   3
+#define DEOXYS_FORM_COUNT   4
+
+#define BURMY_FORM_PLANT 0
+#define BURMY_FORM_SAND  1
+#define BURMY_FORM_TRASH 2
+#define BURMY_FORM_COUNT 3
+
+#define ROTOM_FORM_NORMAL 0
```
lhearachel, line 62: **suggestion**: I don't think that Rotom has a distinct name for this form, so maybe `BASE` is the right nomenclature.

include/constants/forms.h
```diff
@@ -18,35 +18,64 @@
...
+#define DEOXYS_FORM_NORMAL  0
+#define DEOXYS_FORM_ATTACK  1
+#define DEOXYS_FORM_DEFENSE 2
+#define DEOXYS_FORM_SPEED   3
+#define DEOXYS_FORM_COUNT   4
+
+#define BURMY_FORM_PLANT 0
+#define BURMY_FORM_SAND  1
+#define BURMY_FORM_TRASH 2
+#define BURMY_FORM_COUNT 3
```
lhearachel, line 60: **suggestion**: Denote that these apply to both Burmy and Wormadam.

include/constants/forms.h
```diff
@@ -18,35 +18,64 @@
...
-    UNOWN_FORM_U,
-    UNOWN_FORM_V,
-    UNOWN_FORM_W,
-    UNOWN_FORM_X,
-    UNOWN_FORM_Y,
-    UNOWN_FORM_Z,
-    UNOWN_FORM_EXC,
-    UNOWN_FORM_QUE
-};
+#define UNOWN_FORM_A     0
```
lhearachel, line 21: **thought**: Do we want to include forms for, e.g., Shellos/Gastrodon and Arceus?

res/field/scripts/scripts_rotoms_room.s
```diff
@@ -1,39 +1,45 @@
 #include "macros/scrcmd.inc"
 #include "generated/distribution_events.h"
+#include "constants/forms.h"
+#include "constants/moves.h"
```
lhearachel, line 4: **thought**: This one might be better included in `scrcmd.inc`, getting rid of `generated/moves.h`; thoughts?

src/battle/ov16_0223DF00.c
```diff
@@ -1675,22 +1672,21 @@ u32 BattleSystem_CalcMoneyPenalty(Party *party, TrainerInfo *trainerInfo)
...
-    v2 = BattleContext_Get(battleSystem, battleSystem->battleCtx, 2, param1);
-    v1 = BattleSystem_PartyPokemon(battleSystem, param1, v2);
+    int battlerType = Battler_Type(battleSystem->battlers[param1]);
+    int v2 = BattleContext_Get(battleSystem, battleSystem->battleCtx, 2, param1);
+    Pokemon *mon = BattleSystem_PartyPokemon(battleSystem, param1, v2);
 
     if ((battleSystem->battleType & (BATTLE_TYPE_LINK | BATTLE_TYPE_FRONTIER)) == FALSE) {
-        if ((v0 & 0x1) || (battleSystem->battleType == BATTLE_TYPE_AI_PARTNER) || (battleSystem->battleType == (BATTLE_TYPE_TRAINER_DOUBLES | BATTLE_TYPE_2vs2 | BATTLE_TYPE_AI))) {
-            Pokedex_Encounter(battleSystem->pokedex, v1);
+        if (battlerType & BATTLER_THEM
```
lhearachel, line 1680: **thought**: I'm surprised that this doesn't get bungled by operator precedence... :sweat_smile: Maybe we should keep the parens around bitwise-operations as a rule?

src/overlay005/ov5_021F6454.c
```diff
@@ -622,59 +621,52 @@ BOOL ScrCmd_2F1(ScriptContext *param0)
     return 0;
 }
 
-BOOL ScrCmd_303(ScriptContext *param0)
+BOOL ScrCmd_GetPartyRotomCountAndFirst(ScriptContext *ctx)
```
lhearachel, line 624: **thought**: How much do we need to communicate the `FirstSlot` aspect of this command? Is it something that we can effectively hide from users, or would that make for bad precedent? I'd ideally like for this to be named `CountRotomInParty` or something along similar lines, but grabbing the first-slot *is* relevant info.

I'm curious to hear other maintainers' thoughts on this. @ravepossum @cbt6

src/unk_0203D1B8.c
```diff
@@ -421,20 +421,19 @@ void *sub_0203D3E4(int param0, FieldSystem *fieldSystem)
     return v0;
 }
 
-int sub_0203D408(void *param0)
+int PartyManagementData_GetSelectedSlot(PartyManagementData *partyMan)
```
lhearachel, line 424: **thought**: This struct/file could really do with some cleanup. I think that I documented this ages ago before we had strong naming paradigms. :sweat_smile:

src/unk_0204CFFC.c
```diff
@@ -825,59 +829,59 @@ BOOL ScrCmd_0A0(ScriptContext *param0)
     return 0;
 }
 
-BOOL ScrCmd_31C(ScriptContext *param0)
+BOOL ScrCmd_CheckPartyHasSpeciesWithFatefulEncounter(ScriptContext *ctx)
```
lhearachel, line 832: **suggestion**: `CheckPartyHasFatefulEncounter` feels succinct.

src/unk_0204CFFC.c
```diff
@@ -610,109 +612,111 @@ BOOL ScrCmd_1F0(ScriptContext *param0)
...
             }
         }
     }
 
-    return 0;
+    return FALSE;
 }
 
-BOOL ScrCmd_2DD(ScriptContext *param0)
+BOOL ScrCmd_GetPartySlotWithSpecies(ScriptContext *ctx)
```
lhearachel, line 692: **suggestion**: `FindPartySlotWithSpecies`.

src/unk_0204CFFC.c
```diff
@@ -610,109 +612,111 @@ BOOL ScrCmd_1F0(ScriptContext *param0)
...
     }
 
-    return 0;
+    return FALSE;
 }
 
-BOOL ScrCmd_2CB(ScriptContext *param0)
+// If you pass SPECIES_NONE, it returns 1 to check if there's repeated species.
+// This is not used in vanilla, however.
+BOOL ScrCmd_GetPartyRepeatedSpeciesCount(ScriptContext *ctx)
```
lhearachel, line 647: **suggestion**: `CountRepeatedSpeciesInParty`

src/unk_0204CFFC.c
```diff
@@ -825,59 +829,59 @@ BOOL ScrCmd_0A0(ScriptContext *param0)
...
             }
         }
     }
 
-    return 0;
+    return FALSE;
 }
 
-BOOL ScrCmd_32F(ScriptContext *param0)
+BOOL ScrCmd_PartyHasHeldItem(ScriptContext *ctx)
```
lhearachel, line 864: **suggestion**: `CheckPartyHasHeldItem`

asm/macros/scrcmd.inc
```diff
@@ -4106,15 +4110,15 @@
...
+    .short \partySlot
+    .short \move
     .endm
 
-    .macro ScrCmd_2E8 arg0
+    .macro GetSummarySelectedMoveSlot destVar
     .short 744
-    .short \arg0
+    .short \destVar
     .endm
```
lhearachel, line 4122: **thought**: I see that these are only ever used together... maybe they should be bundled into a `TeachMove` convenience macro?

I also think that these names could be a bit tighter. My vote would be for `OpenTeachMoveScreen` and `GetSelectedMoveSlot`

lhearachel (commented):
Great work following through on this! No major flags here, but lots of thoughts and collaborative suggestions. :grin:

res/field/scripts/scripts_rotoms_room.s
```diff
@@ -1,39 +1,45 @@
...
 #include "res/text/bank/rotoms_room.h"
 
+#define VAR_PARTY_SLOT 0x8000
+#define VAR_MOVE_TO_LEARN 0x8001
+#define VAR_ROTOM_FORM 0x8004
```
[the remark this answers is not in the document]
AsparagusEduardo: Shouldn't these only apply within this file? In any case, pokeemerald does this:
https://github.com/pret/pokeemerald/blob/master/data/maps/DewfordTown/scripts.inc

res/field/scripts/scripts_rotoms_room.s
AsparagusEduardo: That's how it is in `ScrCmd_GetPartyMonSpecies` ¯\_(ツ)_/¯

include/constants/forms.h
AsparagusEduardo: I'll just add them too

include/constants/forms.h
AsparagusEduardo: I'll do it for completeness sake and for their COUNT to be used in the dex

res/field/scripts/scripts_rotoms_room.s
AsparagusEduardo: Hmm, I'm not sure if that makes sense, as I don't see that move defines will be used in every map script

res/field/scripts/scripts_rotoms_room.s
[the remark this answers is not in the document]
lhearachel: > Shouldn't these only apply within this file?

The specific names here, yeah. I'm more wondering if this is a precedent that we want to set, or if we want to keep the definitions for `0x8000`, etc. generic; e.g., `VAR_RESULT` for `0x800C`.

> In any case, pokeemerald does this:

Mhm, we use the `LOCALID` paradigm in some other scripts already that @Viperio19 documented. I'd like to do the same for all scripts. :grin:

res/field/scripts/scripts_rotoms_room.s
lhearachel: Yep, just weird.

res/field/scripts/scripts_rotoms_room.s
lhearachel: Fair; once we figure out proper assembler error messages, we should do some cleanup of what's included in `scrcmd.inc`, I think, and distribute them as needed by individual files.

asm/macros/scrcmd.inc
AsparagusEduardo: What should be this new macro's name?

src/unk_0203D1B8.c
AsparagusEduardo: Can you make a note for it, I don't think I want to do this myself ^^;

src/overlay005/ov5_021F6454.c
AsparagusEduardo: Any thoughts? @ravepossum @cbt6

ravepossum: Whoops, missed the first ping. In general I think it's a losing battle to try to cover *every* piece of relevant information in the command name, but in this case something like what Edu proposed seems fine. Though I would opt for `CountAndSlot` instead of `CountAndFirst`.

I'm personally less concerned about having concise names for these hyper-specific script commands that get used in like one place.

include/constants/savedata/vars_flags.h
```diff
@@ -21,6 +21,12 @@
 #define FLAG_AZELF_DISAPPEARED   0x01E0
 #define FLAG_UXIE_DISAPPEARED    0x01E1
 
+#define FLAG_ROTOM_ROOM_HIDE_MICROWAVE_OVEN  0x02BB
+#define FLAG_ROTOM_ROOM_HIDE_WASHING_MACHINE 0x02BC
+#define FLAG_ROTOM_ROOM_HIDE_REFRIGERATOR    0x02BD
+#define FLAG_ROTOM_ROOM_HIDE_ROTARY_FAN      0x02BE
+#define FLAG_ROTOM_ROOM_HIDE_LAWN_MOWER      0x02BF
```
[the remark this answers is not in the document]
AsparagusEduardo: To be handled in #449

src/overlay005/ov5_021F6454.c
lhearachel: Fair point, and I agree with that reasoning.

src/unk_0203D1B8.c
lhearachel: I'll probably take care of it as one of my next actual-documentation tasks along with field-items.

asm/macros/scrcmd.inc
lhearachel: I was thinking:

- `OpenSummaryScreenTeachMove` -> `OpenTeachMoveScreen` (less attached to the summary screen, in case a modder wants to split those up)
- `GetSummarySelectedMoveSlot` -> `GetSelectedMoveSlot` (same as above)
- New macro:

```arm
    .macro TeachMove partySlot, move, destVar
    OpenTeachMoveScreen \partySlot, \move
    GetSelectedMoveSlot \destVar
    .endm
```

ravepossum: One thing to keep in mind here - at a glance, `GetSummarySelectedMoveSlot` looks nearly functionally redundant with the already documented command `GetSelectedPartyMonMove` (the latter just uses a `void**` instead of a `PokemonSummary*` and uses the encapsulation function `PokemonSummary_GetSelectedMoveSlot` instead of accessing the member directly. It also sets the result to `MOVE_NOT_SELECTED`/`-1` if nothing is selected). 

So whichever way we go, we should probably be consistent and try to make the differences obvious if possible.

The existing `SelectPartyMonMove` is also quite similar to `OpenSummaryScreenTeachMove`.

lhearachel (approved):

PR #442, opened 2025-03-19
Some improvements on map-related code

Kuruyia:
This includes a few (what I think are) improvements to code I have documented in previous PRs.

See individual commits for more details


src/overlay005/area_light.c
```diff
@@ -63,21 +61,20 @@ void AreaLightManager_Free(AreaLightManager **areaLightMan)
...
     int currentTime = GetSecondsSinceMidnight() / 2;
 
     if (areaLightMan->templateCount > 1) {
+        int previousEndTime;
+
         if (areaLightMan->activeTemplateIndex - 1 >= 0) {
             previousEndTime = areaLightMan->templates[areaLightMan->activeTemplateIndex - 1].endTime;
         } else {
             previousEndTime = 0;
         }
```
lhearachel, line 75: **suggestion**: I would check if this matches as a ternary.

lhearachel (approved):
Looks great. Only one suggestion below, approving in advance in case it doesn't match.

src/overlay005/area_light.c
Kuruyia: It does!

PR #444, opened 2025-03-21
Document dynamic terrain height

Kuruyia:
This documents the `ov5_021EF250.c` file, which is in charge of managing terrain features whose height can change.

Examples include the water floor of Pastoria Gym, and the various lift platforms in the game (in Iron Island, the Pokémon League...).


lhearachel (approved):
Quick and easy. Looks great!

PR #445, opened 2025-03-22
Document terrain collision manager

Kuruyia:
This documents the `unk_02054D00.c` file, which is in charge of managing height, collision and tile behavior between the terrain and the player/map objects.

This also identifies more well-known map prop model IDs:

-  `34`: generic ship, found in Canalave City for instance.
- `123`: the healing machine found in Pokémon Centers.
- `239`: the blue button found in the Pastoria City Gym, used to set water to its highest level.
- `240`: the green button found in the Pastoria City Gym, used to the water to half its highest level.
- `241`: the orange button found in the Pastoria City Gym, used to the water to its lowest level.
- `303`: the muddy slope found in the overworld that needs a bike traverse.
- `304`: the interior variant of the muddy slope.
- `502`: the lift platform found only in the Pokémon League.
- `507`: the "healing machine"-type machine that registers the in the Hall of Fame.
- `538`: the S.S. Spiral (aka S.S. Sinnoh) ship used to travel Snowpoint City and the Fight Area.


include/terrain_collision_manager.h
```diff
@@ -0,0 +1,47 @@
...
+
+void TerrainCollisionHitbox_Init(const int tileX, const int tileY, const int offsetTileX, const int offsetTileY, const u32 sizeX, const u32 sizeY, TerrainCollisionHitbox *hitbox);
+BOOL TerrainCollisionHitbox_CollidesWithMapProp(const MapProp *mapProp, const TerrainCollisionHitbox *hitbox, const VecFx32 *mapAbsoluteOrigin);
+
+BOOL FieldSystem_FindCollidingLoadedMapPropByModelID(const FieldSystem *fieldSystem, const int targetMapPropModelID, const TerrainCollisionHitbox *hitbox, MapProp **mapPropOut);
+BOOL FieldSystem_FindCollidingLoadedMapPropByModelIDs(const FieldSystem *fieldSystem, const int *mapPropModelIDs, const u8 mapPropModelIDsCount, const TerrainCollisionHitbox *hitbox, MapProp **mapPropOut, int *mapPropModelIDOut);
+BOOL FieldSystem_FindLoadedMapPropByModelID(const FieldSystem *fieldSystem, const int targetMapPropModelID, MapProp **mapPropOut, int *mapMatrixIndex);
+BOOL FieldSystem_FindLoadedMapPropByModelIDs(const FieldSystem *fieldSystem, const int *mapPropModelIDs, const u8 mapPropModelIDsCount, MapProp **mapPropOut, int *mapPropModelIDOut);
```
Kuruyia, line 45: Those four functions, and `TerrainCollisionManager_GetMapAbsoluteOrigin`, are a bit annoying as they arguably seem out of place. I'm wondering if we should move them to their own file.

ravepossum: Personally, I'm not too concerned about it. Stuff like that is all over the repo, and sometimes functions just don't all neatly fit into the ostensible purpose of a module.

ravepossum (commented):
This looks really great! My one tiny nitpick is that we tend to use `Func` over `Fn` for function pointers. Since I got to this basically right as you submitted it, I'm going to leave it up for a bit in case anyone else wants to take a look, but that's all the feedback I have and I'll merge tomorrow or so if nobody else chimes in.

Kuruyia:
> My one tiny nitpick is that we tend to use Func over Fn for function pointers.

Oh, my bad! I should refactor [the function pointers](https://github.com/pret/pokeplatinum/blob/5306770fd40aabd750c7d49a24c2f9a131b80bb1/src/overlay005/land_data.c#L87-L88) in `land_data.c` then in the future.

ravepossum (approved):

PR #446, opened 2025-03-24
Document dynamic map features

Kuruyia:
This PR documents code related to dynamic map features management.

Dynamic map features are defined as features that can change the layout of a map according to player inputs. They may share the following traits:

- have runtime per-dynamic map features data, stored in the field system struct
- have persisted per-dynamic map features data, saved in the save data misc save block (mostly to avoid the possibility of having the player stuck/softlocked after reloading a save)
- affects the map collision

There can be at most one dynamic map features active at one point, if the current map requires one.

Dynamic map features may encompass multiple map headers. Examples include the Sunyshore Gym dynamic map features with its bridges in multiple rooms, or the Platform Lift dynamic map features that handles maps with platform lifts (some rooms in Iron Island and the Pokémon League).

Note: this _does not_ include obstacles such as movable boulders or breakable rocks.


asm/macros/scrcmd.inc
```diff
@@ -2013,31 +2013,31 @@
     .short \destVar
     .endm
 
-    .macro ScrCmd_16F
+    .macro InitPersistedMapFeaturesForPastoriaGym
```
lhearachel, line 2016: **thought**: Is `InitPastoriaGym`, etc. sufficient? The two that worry me are `InitGreatMarsh` (could be taken to mean... well, a lot of things) and `InitVilla` (confusing on the surface).

We could be a bit unprofessional here and also call these "Map Gimmicks," which may get the same concept across.

src/overlay008/ov8_02249960.c
```diff
@@ -380,28 +380,28 @@ void ov8_0224997C(FieldSystem *fieldSystem)
...
+BOOL PastoriaGym_DynamicMapFeaturesCheckCollision(FieldSystem *fieldSystem, const int tileX, const int tileY, const fx32 height, BOOL *isColliding)
 {
-    u8 v0 = TerrainCollisionManager_GetTileBehavior(fieldSystem, param1, param2);
+    u8 v0 = TerrainCollisionManager_GetTileBehavior(fieldSystem, tileX, tileY);
 
     if (TileBehavior_IsPastoriaGymHighGround(v0)) {
-        if (param3 != 0) {
-            (*param4) = 1;
+        if (height != 0) {
+            (*isColliding) = 1;
```
lhearachel, line 389: **polish**: Use `TRUE` / `FALSE`.

res/field/scripts/scripts_sunyshore_city_gym_room_1.s
```diff
@@ -11,7 +11,7 @@
 
 _0012:
     SetVar 0x4000, 0
-    ScrCmd_175 0
+    InitPersistedMapFeaturesForSunyshoreGym 0
```
lhearachel, line 14: **thought**: Seems like this is a floor ID...?

asm/macros/scrcmd.inc
Kuruyia: I'm worried having such short names would convey that they are doing more work than they actually do. Those script commands are exclusively related to the persisted map features.

I can think of two, less verbose middle grounds we could use:

- `InitPastoriaGymPersistedMapFeatures`: saves only three letters but has the same meaning
- `InitPastoriaGymMapFeatures`: saves more letters, but I'm worried it could also mean that it inits runtime data (which is handled by the `_DynamicMapFeaturesInit` functions)

What was your idea about calling those "Map Gimmicks" related to naming?

res/field/scripts/scripts_sunyshore_city_gym_room_1.s
Kuruyia: Looks like it, yeah. Should we have constants for those?

asm/macros/scrcmd.inc
lhearachel: My idea was that `Gymmick` is a fun portmanteau. :slightly_smiling_face: But hard to read, so I wouldn't really push for it.

Looking at usage, I think it's fine to keep long names here. These commands are only used once each.

res/field/scripts/scripts_sunyshore_city_gym_room_1.s
lhearachel: I think they're fine as just numbers, but we should document the argument name in the macro and C code.

asm/macros/scrcmd.inc
Kuruyia: Ahh I didn't pick it up ^^' We can always write a nod to this name in future Markdown doc :)

PR #447, opened 2025-03-26
Document terrain attributes loader

Kuruyia:
This documents the `unk_02054BD0.c` file, which is in charge of loading the terrain attributes from the land data NARC for some map load modes.

This also documents the following fields in the map load mode struct:

- `useTerrainAttributesLoader`: whether to load terrain attributes using the terrain attributes loader.
- `terrainAttributesLoaderBlockCount`: the number of terrain attributes block to load when using the terrain attributes loader. Each block corresponds to the terrain attributes contained in a land data file.

Note: when using the simple terrain collisions in a map load mode, it is mandatory to also use the terrain attributes loader, as this is where the terrain collision manager will get terrain attributes.


src/terrain_attributes_loader.c
```diff
@@ -0,0 +1,89 @@
...
+
+void TerrainAttributesLoader_Free(FieldSystem *fieldSystem)
+{
+    if (fieldSystem->terrainAttributesLoader != NULL) {
+        Heap_FreeToHeap(fieldSystem->terrainAttributesLoader);
+        fieldSystem->terrainAttributesLoader = NULL;
+    }
+}
+
+u16 const *TerrainAttributesLoader_GetTerrainAttributes(const u32 mapMatrixIndex, const TerrainAttributesLoader *terrainAttributesLoader)
```
lhearachel, line 34: **polish**: This should match as `const u16 *` on the return type of the signature, and is more aligned with the rest of the repository.

include/terrain_attributes_loader.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include "field/field_system_decl.h"
+
+#define TERRAIN_ATTRIBUTES_LOADER_MAP_MATRIX_COUNT 225
+#define TERRAIN_ATTRIBUTES_LOADER_MAX_BLOCK_COUNT  16
+
+typedef struct TerrainAttributesLoader {
+    u8 mapMatrixIndexToBlockIndex[TERRAIN_ATTRIBUTES_LOADER_MAP_MATRIX_COUNT];
+    u16 terrainAttributes[TERRAIN_ATTRIBUTES_COUNT * TERRAIN_ATTRIBUTES_LOADER_MAX_BLOCK_COUNT];
+} TerrainAttributesLoader;
```
lhearachel, line 16: **thought**: Is this really a `Loader`? The code suggests to me that it is the loaded (past-tense) data.

Kuruyia: I was calling it that way because the `TerrainAttributesLoader_Load` function is the meat of this module, but I do see your point.

Do you have any suggestion? I can quickly think of a few other names that would be correct:
- `TerrainAttributes`: too concise?
- `TerrainAttributesBuffer`: too generic?
- `SimpleTerrainAttributes`: refers to where it's used in the terrain collision manager - needs too much context to understand quickly? https://github.com/pret/pokeplatinum/blob/0497e6082321efb52fc31ba36a5d114ef2b63db4/src/terrain_collision_manager.c#L194-L210

Keeping in mind that for the majority of the maps, terrain attributes are normally loaded and handled through the land data manager, ideally we'd have a name that reflects the fact that this module is used exceptionally.

lhearachel: I feel like `TerrainAttributes` is perfectly fine. No need to get fancy with it, unless these are explicitly dynamic terrain attributes...?

Kuruyia: Perfect, let's go with that then.

> unless these are explicitly dynamic terrain attributes...?

If there are, I'm not aware of them.

lhearachel (approved):

PR #450, opened 2025-03-29
Document platform lift

Kuruyia:
This documents the `unk_0207160C.c` file, which is in charge of setting up and managing the platform lift dynamic map feature for maps that contains platform lifts (Iron Island and the Pokémon League).

Related functions, map script commands and map object status flag were also documented as part of this PR.

Check individual commit descriptions for more details.


include/platform_lift.h
```diff
@@ -0,0 +1,11 @@
+#ifndef POKEPLATINUM_PLATFORM_LIFT_H
```
lhearachel, line 1: **thought**: That this is embedded in the static binary is a little baffling.

src/platform_lift.c
```diff
@@ -0,0 +1,489 @@
...
+} PlatformLiftFieldTaskEnv;
+
+static BOOL FieldTask_PlatformLiftGoUp(FieldTask *taskMan);
+static BOOL FieldTask_PlatformLiftGoDown(FieldTask *taskMan);
+static void PlatformLift_PlaySoundEffect(const int platformLiftKind);
+static void PlatformLift_StopSoundEffect(const int platformLiftKind);
+
+static const PlatformLiftConfiguration sPerMapPlatformLiftConfiguration[PLATFORM_LIFT_COUNT] = {
+    [PLATFORM_LIFT_IRON_ISLAND_B1F_RIGHT_ROOM] = {
+        { (MAP_OBJECT_TILE_SIZE * 1), (MAP_OBJECT_TILE_SIZE * 9) },
```
lhearachel, line 91: **polish**: Use designated initializers for struct-fields.

src/platform_lift.c
```diff
@@ -0,0 +1,489 @@
...
+    PLATFORM_LIFT_STATE_START_MOVING = 0,
+    PLATFORM_LIFT_STATE_MOVE,
+    PLATFORM_LIFT_STATE_FINISH_MOVING,
+    PLATFORM_LIFT_STATE_END,
+};
+
+typedef struct PlatformLiftConfiguration {
+    fx32 floorHeights[PLATFORM_LIFT_FLOOR_COUNT];
+    int startTileX;
+    int startTileY;
```
lhearachel, line 65: **thought**: Should this be `startTileZ`?

src/platform_lift.c
```diff
@@ -0,0 +1,489 @@
...
+    const PlatformLiftConfiguration *platformLiftConfiguration = &sPerMapPlatformLiftConfiguration[data->id];
+    fx32 currentHeight = platformLiftConfiguration->floorHeights[data->floorID];
+
+    MapProp *platformLiftMapProp;
+
+    if (platformLiftConfiguration->isIronIslandB2FLeftRoom) {
+        VecFx32 platformLiftInitialPosition = {
+            (FX32_ONE * (16 * 19 + 8)),
+            0,
+            (FX32_ONE * (16 * (32 + 12) + 8))
```
lhearachel, line 168: **question**: Do these values have any special meanings that could be `const`-ified for better semantics?

src/platform_lift.c
```diff
@@ -0,0 +1,489 @@
...
+    data->notUsedWhenEnteredMap = TRUE;
+    data->kind = PLATFORM_LIFT_KIND_IRON_ISLAND;
+
+    switch (fieldSystem->location->mapId) {
+    case MAP_HEADER_IRON_ISLAND_B1F_RIGHT_ROOM:
+        if (fieldSystem->location->z == 26) {
+            data->floorID = PLATFORM_LIFT_BOTTOM_FLOOR;
+        } else {
+            data->floorID = PLATFORM_LIFT_TOP_FLOOR;
+        }
```
lhearachel, line 207: **suggestion**: Prefer ternary-initializers, if they match.

lhearachel (commented):
Looks great! A few comments, mostly for polish.

include/platform_lift.h
Kuruyia: Oh yeah. Don't worry, this will be a recurring theme regarding everything related to dynamic map features (that is, everything in overlay008 and the Villa in `ov5_021F8370.c`).

src/platform_lift.c
Kuruyia: Those are the map (3D) coordinates for the platform lift found in the Iron Island B2F Left room map.

Because they are only used here (afaik), and specific to a single map, I initially decided against having constants defined for those coordinates (and any other map-specific coordinates, such as the ones found throughout `PersistedMapFeatures_InitForPlatformLift`).

We could have constants such as `PLATFORM_LIFT_IRON_ISLAND_B2F_LEFT_ROOM_X`/`PLATFORM_LIFT_IRON_ISLAND_B2F_LEFT_ROOM_Z` if you find those clearer.

lhearachel: I think that we should not shy away from having semantic-constants for seemingly magic numbers, even if they are local to a single file. In this instance, I think that these are eligible for some `#define`s within the `.c` file, as no other module seems to want them.

src/platform_lift.c
Kuruyia: They do not seem to match, unfortunately.

src/platform_lift.c
Kuruyia: Those are map object (2D) coordinates, which I've documented as `tileX`/`tileY` in previous contributions.

This is a conversation I dreaded to have for quite a bit of time now, as `tileY` of course turns into the `z` component when transformed into the 3D map coordinate space (e.g. multiplying by `FX32_ONE * 16`). 

Do we want to use `x`/`z` for the 2D coordinate components? I know this is what's already being done in `struct Location`.
If so, I'll also need to do quite a bit of refactor to fix all `tileY` in the codebase in a future PR. Shouldn't take that long, as it's just replacing `tileY` with `tileZ` everywhere.

lhearachel: As we discussed on Discord, the plan is for all map-related coordinate systems (whether 2D or 3D) to follow the paradigm of "X is east-west, Z is north-south, Y is up-down."

lhearachel (approved):

PR #451, opened 2025-03-30
Update Game Corner Prizes

FrozenMetroid:
This updates all of the relevant files for the Game Corner prizes in Veilstone City to be in a more easily readable state. This is my first pull request, so please let me know if there are issues.


asm/macros/scrcmd.inc
```diff
@@ -3749,7 +3749,7 @@
     .short 677
     .endm
 
-    .macro ScrCmd_2A6 arg0, arg1, arg2
+    .macro ScrCmd_GetGameCornerPrizeData arg0, arg1, arg2
```
lhearachel, line 3752: **polish**: In `.inc` and `.s` files, we name commands without the `ScrCmd_` prefix; this makes for better semantics in the script files themselves.

include/game_corner_prize.h
```diff
@@ -0,0 +1,8 @@
+#ifndef POKEPLATINUM_GAME_CORNER_PRIZE_H
```
lhearachel, line 1: **suggestion**: Rename this file (and its corresponding source file) with a `scrcmd_` prefix to associate it with the other script-command modules.

src/game_corner_prize.c
```diff
@@ -0,0 +1,42 @@
...
+#include "generated/items.h"
+
+#include "field_script_context.h"
+#include "inlines.h"
+
+BOOL ScrCmd_GetGameCornerPrizeData(ScriptContext *param0)
+{
+    u16 index = ScriptContext_GetVar(param0);
+    u16 *item = ScriptContext_GetVarPointer(param0);
+    u16 *price = ScriptContext_GetVarPointer(param0);
```
lhearachel, line 15: **polish**: Propagate these parameter-names upwards into the `scrcmd.inc` file, so that they are also tied to the assembly macros.

src/game_corner_prize.c
```diff
@@ -0,0 +1,42 @@
...
+        { ITEM_TM13,            10000 },
+        { ITEM_TM29,            10000 },
+        { ITEM_TM74,            15000 },
+        { ITEM_TM68,            20000 }
+    };
+
+    *item = GameCornerPrizeData[index][0];
+    *price = GameCornerPrizeData[index][1];
+
+    return 0;
```
lhearachel, line 41: **polish**: Prefer `return FALSE` to match the `BOOL` on the function-signature.

src/game_corner_prize.c
```diff
@@ -0,0 +1,42 @@
...
+
+#include <nitro.h>
+#include <string.h>
+
+#include "generated/items.h"
+
+#include "field_script_context.h"
+#include "inlines.h"
+
+BOOL ScrCmd_GetGameCornerPrizeData(ScriptContext *param0)
```
lhearachel, line 11: **polish**: Rename `param0` here (and in the header) to `ctx`.

src/game_corner_prize.c
```diff
@@ -0,0 +1,42 @@
...
+{
+    u16 index = ScriptContext_GetVar(param0);
+    u16 *item = ScriptContext_GetVarPointer(param0);
+    u16 *price = ScriptContext_GetVarPointer(param0);
+    static const u16 GameCornerPrizeData[][2] = {
```
lhearachel, line 16: **suggestion**:

1. Prefix `static` data variables with `s`.
2. This double-table may match as a struct:

```c
typedef struct GameCornerPrize {
    u16 item;
    u16 price;
} GameCornerPrize;

static const GameCornerPrize sGameCornerPrizeData[] = {
    { ITEM_SILK_SCARF, 1000 },
    // snip...
};
```

This would then allow some slightly-better semantics in the function:

```c
*item = sGameCornerPrizeData[index].item;
*price = sGameCornerPrizeData[index].price;
```

lhearachel (commented):
Thanks for your contribution! I have some light feedback below. :grin:

lhearachel (approved):
I took the liberty of applying formatting rules as needed; the rest looks good here. Thanks again for your contribution! :pray:

PR #454, opened 2025-03-31
Start Markdown documentation for dynamic map features, add reference to NSBTP

Kuruyia:
This starts the Markdown documentation for the dynamic map features, explaining their purpose, how they are used, the data structures and interfaces that are defined.

The platform lift dynamic map feature is documented as the first one of the long list.

This also adds a reference to the NSBTP file format (where to find doc for it, and the fact that it is used in the map prop animations NARC).

### DIFF
--- a/docs/maps/dynamic_map_features.md
+++ b/docs/maps/dynamic_map_features.md
@@ -1,0 +1,170 @@
+# Dynamic map features
+
+This document aims to provide an overview of the dynamic map features system,
+as well as the ones that are implemented in the game.
+
+## Introduction
+
+Dynamic map features are defined as features that can change the layout of a
+map according to player inputs.
+
+We distinguish between two kinds of data that can be used to define those features:
+
+- **Dynamic map features data** (or "_runtime data_"): this is data that is specific to a
+  dynamic map feature, and is only available at runtime. For example, in the Canalave Gym,
+  this contains the instances of the map props that are used to create the movable platforms.
+- **Persisted map features data** (or "_persisted data_"): this is a small 32 bytes buffer
+  that is allocated in the save file for dynamic map features to store state that must persist
+  if the game is saved and loaded.
+
+    For example, in the Canalave Gym, this contains the position of the movable platforms,
+    so that they can be restored when the game is loaded. This avoids having the player
+    stuck in a part of the Gym because the platforms were reset to their original
+    position.
+
+    This saved data is considered valid only when the player has saved the game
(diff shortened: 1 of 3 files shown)


docs/maps/dynamic_map_features.md
```diff
@@ -0,0 +1,170 @@
...
+- `X_DynamicMapFeaturesFree` (optional): this frees up the resources used
+  by the dynamic map feature, when it is no longer needed.
+- `X_DynamicMapFeaturesCheckCollision` (optional): this is called by the
+  terrain collision manager, and allows dynamic map features to affect the
+  collision of the map. For example, this is used in the Villa to prevent the
+  player from walking through furniture that they have bought.
+
+As you may have guessed by now, all dynamic map features are defined in the
+game code directly, and adding new ones, modifying existing ones, or using an
+existing one on a custom map requires modifying the game code.
```
lhearachel, line 79: **suggestion**: Remove the personification here.

docs/maps/dynamic_map_features.md
```diff
@@ -0,0 +1,170 @@
...
+information specific to this/these map(s).
+
+## Dynamic map features in the game
+
+This section will quickly describe the dynamic map features that are implemented in the
+game.
+
+### Platform Lift
+
+<img src="https://github.com/user-attachments/assets/136f90ae-6276-445c-9b5b-c816bd4576b0" alt="Platform Lift" height="350">
```
lhearachel, line 93: **question**: Do we need to include this as a static asset in the repository?

Kuruyia: That's a question I had in mind when I first started those docs.

Basically my worries with using the GitHub CDN for storing those assets are those:
- We depend on GitHub for storing those, and not on Git only anymore
- If we want to migrate to another hosting service, such as GitLab, this complicates transition
- Their asset storage policy doesn't seem to be clearly defined, and could change (i.e. we don't)

We could store that in the repository, but I'm worried about the size impact this can have on Git's object store (especially if images were to change).

To alleviate that, we could use Git LFS, but it is limited to 1 GB (afaik) on GitHub for free repos.

lhearachel: Depending on the overall size, Git LFS could be viable here. We could also consider switching docs to a separate branch, maybe. I'll have to look into the LFS details for GitHub's free plan.

lhearachel (approved):

PR #464, opened 2025-04-06
Start documenting Distortion World and its camera

Kuruyia:
This begins the documentation of overlay009, which is in charge of managing stuff related to the Distortion World.

For now, the scope of this PR has been limited to the following:

- start documenting structures used throughout this overlay, more specifically the `DistortionWorldSystem` struct
- start documenting the structures found inside the files of the `tw_arc.narc` NARC
- document functions and structures related to camera management inside the Distortion Word, including the templates found inside the `tw_arc.narc` NARC

This also documents the `ScrCmd_31F` map script command, that is only used in the Giratina Room map of the Distortion World to reset the camera angles that are present in the persisted map features data.


src/overlay009/ov9_02249960.c
ravepossum, line 1: quip: Sweet jesus, this module is massive. 10k+ lines??

src/overlay009/ov9_02249960.c
```diff
@@ -1167,59 +1184,55 @@ const UnkStruct_ov9_02252EB4 Unk_ov9_02252EB4[];
 
 void DistortionWorld_DynamicMapFeaturesInit(FieldSystem *fieldSystem)
 {
-    PersistedMapFeatures *v0;
-    UnkStruct_02071C5C *v1;
-    UnkStruct_ov9_02249B04 *v2;
+    PersistedMapFeatures *persistedMapFeatures = MiscSaveBlock_GetPersistedMapFeatures(FieldSystem_GetSaveData(fieldSystem));
+    DistortionWorldPersistedData *data = PersistedMapFeatures_GetBuffer(persistedMapFeatures, DYNAMIC_MAP_FEATURES_DISTORTION_WORLD);
+    DistortionWorldSystem *distortionWorldSystem = Heap_AllocFromHeap(HEAP_ID_FIELD, sizeof(DistortionWorldSystem));
```
ravepossum, line 1189: thought: Do we need to use the full unabbreviated name for the local variable here?

src/overlay009/ov9_02249960.c
```diff
@@ -1462,42 +1475,42 @@ static BOOL ov9_02249E00(UnkStruct_ov9_02249B04 *param0, u32 param1)
...
-    u32 v1 = ov9_022510D0(v0);
+    DistortionWorldSystem *distortionWorldSystem = fieldSystem->unk_04->dynamicMapFeaturesData;
+    u32 mapHeaderID = DistortionWorld_GetMapHeaderID(distortionWorldSystem);
 
-    if (v1 == 582) {
-        if ((tileX == 15) && (tileZ == 26)) {
-            *isColliding = 1;
-            return 1;
+    if (mapHeaderID == MAP_HEADER_DISTORTION_WORLD_GIRATINA_ROOM) {
+        if (tileX == 15 && tileZ == 26) {
```
ravepossum, line 1498: question: Do we know what these refer to? Would it be worth making constants to make it more clear?

src/overlay009/ov9_02249960.c
```diff
@@ -1568,167 +1581,166 @@ static void ov9_02249F50(UnkStruct_ov9_02249B04 *param0)
...
-    }
+    CameraConfiguration cameraConfig = {
+        .distance = DISTORTION_WORLD_CAMERA_BASE_DISTANCE,
+        .cameraAngle = {
+            DISTORTION_WORLD_CAMERA_BASE_ANGLE_X,
+            DISTORTION_WORLD_CAMERA_BASE_ANGLE_Y,
+            DISTORTION_WORLD_CAMERA_BASE_ANGLE_Z },
+        .projectionMtx = DISTORTION_WORLD_CAMERA_BASE_PROJECTION_MTX,
+        .fovY = DISTORTION_WORLD_CAMERA_BASE_FOVY,
+        .padding_0F = 0
```
ravepossum, line 1639: question: Does this match without this line?

src/overlay009/ov9_02249960.c
```diff
@@ -1568,167 +1581,166 @@ static void ov9_02249F50(UnkStruct_ov9_02249B04 *param0)
...
-    v1->unk_04 = 0;
-    v1->unk_06 = 0;
-    v1->unk_08 = 0;
+    data->cameraAngleX = 0;
+    data->cameraAngleY = 0;
+    data->cameraAngleZ = 0;
 }
 
-static void ov9_02249FF4(UnkStruct_ov9_02249B04 *param0)
+static void DistortionWorld_CameraInit(DistortionWorldSystem *distortionWorldSystem)
```
ravepossum, line 1624: question: Do we need `DistortionWorld` preprended to the static functions here? It's a bit of a mouthful so it may make function names more concise if we drop it.

src/overlay009/ov9_02249960.c
```diff
@@ -245,46 +257,51 @@ typedef struct {
...
 } UnkStruct_ov9_02249B04_sub1;
 
 typedef struct {
-    int unk_00;
-    int unk_04;
-    VecFx32 unk_08;
-    VecFx32 unk_14;
-} UnkStruct_ov9_0224A0DC;
+    BOOL running;
+    int stepsLeft;
```
ravepossum, line 292: nitpick: Would something like `stepsRemaining` be more clear here?

src/overlay009/ov9_02249960.c
```diff
@@ -2207,27 +2219,27 @@ BOOL ov9_0224A800(FieldSystem *fieldSystem, int param1)
     return 0;
 }
 
-static BOOL ov9_0224A8A0(UnkStruct_ov9_02249B04 *param0, int param1, int param2, int param3, int param4)
+static BOOL DistortionWorld_ApplyCorrectCameraAngleForPlayer(DistortionWorldSystem *distortionWorldSystem, int playerX, int playerY, int playerZ, int playerDir)
```
ravepossum, line 2222: thought: I think it may be unclear what `Correct` means here. I don't immediately have a better suggestion for it, though.

src/overlay009/ov9_02249960.c
```diff
@@ -3708,204 +3720,203 @@ static void ov9_0224C050(const UnkStruct_ov9_0224BFE0 *param0, int param1, int *
...
 }
 
-static void ov9_0224C070(UnkStruct_ov9_02249B04 *param0, int param1, int *param2, int *param3, int *param4)
+static void ov9_0224C070(DistortionWorldSystem *param0, int param1, int *param2, int *param3, int *param4)
 {
     ov9_0224C050(&param0->unk_169C.unk_00, param1, param2, param3, param4);
 }
 
-static void ov9_0224C088(UnkStruct_ov9_02249B04 *param0, UnkStruct_ov9_0224C14C *param1, int param2)
+static void DistortionWorldFile_Load(DistortionWorldSystem *distortionWorldSystem, DistortionWorldFile *file, int mapHeaderID)
```
ravepossum, line 3728: polish: `enum MapHeader mapHeaderID` (same for the other functions here that accept a map header param)

asm/macros/scrcmd.inc
```diff
@@ -4424,7 +4424,7 @@
     .short \arg1
     .endm
 
-    .macro ScrCmd_31F
+    .macro ResetDistortionWorldPersistedCameraAngles
```
ravepossum, line 4427: question: Would this be nearly as clear as just `ResetDistortionWorldCameraAngles`?

ravepossum (commented):
This is a real behemoth of a module, nice first chunk! 

Overall thought I want to hear opinions on - would it be worth us picking some prefix to use for distortion world functionality e.g. `DW`? I'm usually not a fan of stuff like that since it reduces grep-ability, but in this case there's *so* much DW-specific functionality and preprending everything with `DistortionWorld` makes it pretty verbose. I'm kinda torn about it myself, but that just came to mind when looking at some of the function and struct names here.

include/overlay009/camera_configuration.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include "overlay115/camera_angle.h"
+
+typedef struct {
+    fx32 distance;
+    CameraAngle cameraAngle;
+    u8 projectionMtx;
+    u16 fovY;
+    u8 padding_0F;
+} CameraConfiguration;
```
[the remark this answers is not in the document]
Kuruyia: This is also used in overlays 77 and 93 which, afaik, aren't (strictly) related to the distortion world. This is why I did not add the prefix here.

src/overlay009/ov9_02249960.c
```diff
@@ -245,46 +257,51 @@ typedef struct {
...
     UnkStruct_ov9_0224C788 unk_74;
 } UnkStruct_ov9_02249B04_sub1;
 
 typedef struct {
-    int unk_00;
-    int unk_04;
-    VecFx32 unk_08;
-    VecFx32 unk_14;
-} UnkStruct_ov9_0224A0DC;
+    BOOL running;
```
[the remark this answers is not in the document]
Kuruyia: `isActive` sounds good!

src/overlay009/ov9_02249960.c
```diff
@@ -245,46 +257,51 @@ typedef struct {
...
-    CameraAngle unk_14;
-    UnkStruct_ov9_0224A0DC unk_1C;
-    SysTask *unk_3C;
-} UnkStruct_ov9_02249F9C;
+    CameraAngle baseAngle;
+    CameraAngle currentAngle;
+    CameraAngle targetAngle;
+    DistortionWorldCameraTransition transition;
+    SysTask *transitionTask;
+} DistortionWorldCameraManager;
```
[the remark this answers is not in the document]
Kuruyia: The thinking here was that this struct (and related functions) "manages" camera angle transition and its related systask. So it sounded appropriate to call that a manager.

src/overlay009/ov9_02249960.c
Kuruyia: Yeah... Although I think there might be potential for file-splitting here (as is the case with overlay 8). I'd like to attempt that once I have a better overall understanding of this overlay, tho'.

src/overlay009/ov9_02249960.c
Kuruyia: How would you abbreviate this name here? Just `system`?

Also, would we want to use such an abbreviation everywhere (mainly in function parameter names)?

src/overlay009/ov9_02249960.c
Kuruyia: Indeed!

src/overlay009/ov9_02249960.c
Kuruyia: I don't mind dropping this prefix from all static functions.

src/overlay009/ov9_02249960.c
Kuruyia: We could add "Position" to the end of those function names?

E.g. `ApplyCorrectCameraAngleForPlayerPosition` and `FindCorrectCameraAngleForPlayerPosition`.

asm/macros/scrcmd.inc
Kuruyia: This script command calls the following function:

https://github.com/pret/pokeplatinum/blob/c15cb4f31584eca0c8b80a8e5f1c1622649ba5fc/src/overlay009/ov9_02249960.c#L1612-L1622

Which only acts upon the camera angle in the persisted data. I'm worried calling it `ResetDistortionWorldCameraAngles` would imply there is an effect that is immediately visible (although it may have such an effect indirectly, I didn't test).

Also, this script command is only called in map script `scripts_distortion_world_giratina_room.s`, so I don't think having a longer name has a huge impact.

src/overlay009/ov9_02249960.c
Kuruyia: Well, now we do know what these refer to. Those are the tiles that are right after the tile that teleports the player between Giratina room and B7F.

| `MAP_HEADER_DISTORTION_WORLD_B7F` | `MAP_HEADER_DISTORTION_WORLD_GIRATINA_ROOM` |
|--------|--------|
| image | image |

In the above screenshots, Dawn stands on the teleporting tile. 
The red arrow points to the tiles that the code refers to. Those tiles are unreachable anyway as the player is teleported before being able to reach them.

How shall we name those? I can think of something like `DISTORTION_WORLD_GIRATINA_ROOM_TILE_AFTER_TELEPORT_X`, but that's a mouthful.

Kuruyia:
> Overall thought I want to hear opinions on - would it be worth us picking some prefix to use for distortion world functionality e.g. `DW`? I'm usually not a fan of stuff like that since it reduces grep-ability, but in this case there's _so_ much DW-specific functionality and preprending everything with `DistortionWorld` makes it pretty verbose. I'm kinda torn about it myself, but that just came to mind when looking at some of the function and struct names here.

I don't think it would reduce grep-ability that much. You could grep for `DW_` (with the trailing underscore) to get the results you want. Doing so currently only leads to two results:

```
❯ rg DW_
include/dw_warp/dw_warp.h
1:#ifndef POKEPLATINUM_DW_WARP_H
2:#define POKEPLATINUM_DW_WARP_H
```

I'm more worried about people not getting right away what "DW" stands for (Doctor Who?), but it's only a matter of learning once the acronym.

We could also use the `TornWorld_` prefix, which is a certified based name. Certainly a less understandable one too.

include/overlay009/ov9_02249960.h
```diff
@@ -9,11 +9,24 @@
...
+    u32 unk_00_1 : 24;
+    u32 unk_00_25 : 4;
+    u32 : 3;
+    u16 cameraAngleX;
+    u16 cameraAngleY;
+    u16 cameraAngleZ;
+    u16 unk_0A;
+    u32 unk_0C;
+    u8 reserved_10[16];
+} DistortionWorldPersistedData;
```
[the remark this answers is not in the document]
ravepossum: Sounds good. If I'm understanding what this does right, persistent definitely sounds more correct to me, but we could do a rename separate from this.

src/overlay009/ov9_02249960.c
ravepossum: Yeah I think if we could organize it in a way that makes sense and still matches, that would definitely be nice.

src/overlay009/ov9_02249960.c
ravepossum: Yeah honestly `system` seems fine here to me. It should be clear enough from the surrounding context. If we do decide to lean into `DW` and want more specificity we could go with `dwSystem` or something.

I feel like in general as long as it's sufficiently obvious what you're referring to, it should be fine to use that everywhere. E.g. I would use it like this:
```
DistortionWorld_Foo(DistortionWorldSystem *system)
```
But I would be more specific in a situation like this:
```
DistortionWorld_Bar(FieldSystem *fieldSystem, DistortionWorldSystem *dwSystem)
```
(just some sort of elaboration, I'm just using `dw` for the example here)

src/overlay009/ov9_02249960.c
ravepossum: I more meant that `Correct` itself doesn't have an obvious meaning to me. After reviewing the functions again, I'm not sure it really adds much actually. We could maybe add position and drop correct? `ApplyCameraAngleForPlayerPosition` and `FindCameraAngleForPlayerPosition` sound a bit better to me.

asm/macros/scrcmd.inc
ravepossum: Sounds good. Might be a good candidate to substitute in `DW` if we do want to abbreviate that across the board.

src/overlay009/ov9_02249960.c
ravepossum: Huh, interesting. I wonder if it would be more easily understood as something like `GIRATINA_ROOM_TELEPORT_TILE_X + 1` or something.

src/overlay009/ov9_02249960.c
ravepossum: Going along with what you've said here, I wonder if it makes sense to use `DistortionWorldSystem_` for `GetMapHeaderID` since it's an accessor function.

src/overlay009/ov9_02249960.c
Kuruyia: Well, on the function you commented on we do have a `FieldSystem` passed as a parameter ^^'

include/overlay009/ov9_02249960.h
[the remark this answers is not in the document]
Kuruyia: Let's rename it later then.

src/overlay009/ov9_02249960.c
Kuruyia: Really like the idea, I'll do that.

src/overlay009/ov9_02249960.c
lhearachel: The only element that matters is function layout, really. There may be some finagling to be done with rodata, though.

Kuruyia: Is there any methodology that can be followed or help to efficiently  split files ? I'm also interested in the answer if you have to do that on a huge asm file and not necessarily on a decompiled file as we have here.

I was thinking clustering the functions that crossref other, nearby ones, and functions that manipulate the same structs would be a good first approach.

ravepossum:
My bad, this slipped off my radar - re: the `DW`/`DistWorld`/etc. shorthand, one of the primary things I was concerned about the size of was the name of structs and their associated functionality - e.g. `DistortionWorldFileCameraAngleSection` and it looks like you've only used it for the module prefix. (as an aside, I think I still prefer something like `DW` over `DistWorld` to really commit to saving space but that is by no means a hill I would die on).

Beyond abbreviation concerns, I think I would rather us apply the shorthand universally rather than just when using it as a module prefix _or_ not at all - applying it some of the time feels like it adds some inconsistency that may make the Distortion World-related code slightly harder to traverse.

I don't want to hold this PR up much longer though - we can also revisit the naming in future PRs. I just wanted to avoid ground-retreading in future Distortion World PRs since it'd be nice to stay consistent.

Kuruyia:
> My bad, this slipped off my radar

No problem, life got (and still gets) on my way for me ^^ I'll need to take some time to properly get back to this PR and module.

> the DW/DistWorld/etc. shorthand, one of the primary things I was concerned about the size of was the name of structs and their associated functionality

Agreed, I just pushed an update to the name of the structs. Shortened the names to `DistWorld` for consistency with the prefix that was already chosen for the function names.

> Beyond abbreviation concerns, I think I would rather us apply the shorthand universally rather than just when using it as a module prefix or not at all - applying it some of the time feels like it adds some inconsistency that may make the Distortion World-related code slightly harder to traverse.

Agreed. We still have some `DistortionWorld` in the codebase, do we want to change all of them to `DistWorld`/`DW`?

Those appear mostly in the middle of function names (e.g. `LandDataManager_DistortionWorldX`, `SystemVars_XDistortionWorldY`, `SystemFlag_XDistortionWorldY`...). There's also the `DistortionWorldWarp` struct.

> I don't want to hold this PR up much longer though - we can also revisit the naming in future PRs. I just wanted to avoid ground-retreading in future Distortion World PRs since it'd be nice to stay consistent.

Yep, let's ask ourselves the question later on once the module is more documented. Having a better vision of what this module interacts with may help.

I'll rebase this PR and fix conflicts once it is greenlit.

lhearachel:
> Yep, let's ask ourselves the question later on once the module is more documented. Having a better vision of what this module interacts with may help.

Sorry for my lack of response here; I think that `DistWorld` is pretty reasonable here, given that it sensibly maps to `DistortionWorld` within the repository's context and is hard to mentally map onto some other concept once that link is made. Let's go with that one. :+1:

ravepossum:
Apologies again, still quite busy. The newest set of changes look good to me. Should just need to sort out the conflicts and we'll call this one good.

Kuruyia:
Should be good to merge :)

PR #466, opened 2025-04-08
Sound documentation followup

AsparagusEduardo:
Some stuff I found


include/sound.h
```diff
@@ -75,7 +139,7 @@ void Sound_SetSubScene(u8 scene);
 int Sound_LoadSoundEffectsForScene(u8 scene);
 BOOL Sound_SetSceneAndPlayBGM(u8 scene, u16 bgmID, int unused);
 void Sound_LoadSoundDataForFieldBGM(u16 seqID, u16 currentBankID);
-void sub_02004950(u16 param0);
+void Sound_PauseOrStopFieldBGMAndPlayNewBGM(u16 bgmID);
```
lhearachel, line 142: **suggestion**: This one feels a bit unwieldy. Does `Sound_SwapBGM` convey everything sufficiently?

src/unk_020553DC.c
```diff
@@ -20,101 +20,101 @@
...
-    { 0x35, 0x455 },
-    { 0x41, 0x459 },
-    { 0x42, 0x459 },
-    { 0x43, 0x459 },
-    { 0x44, 0x459 },
-    { 0x45, 0x45A },
-    { 0x62, 0x458 }
+// clang-format off
+// Class, Eyes meet theme
+const static u16 sTrainerEncounterBGMs[][2] = {
```
lhearachel, line 25: **thought**: This is probably a data-table that we will want to port to the JSON data files and generate at build-time.

src/unk_020553DC.c
```diff
@@ -137,71 +137,68 @@ void Sound_ClearSpecialBGM(FieldSystem *fieldSystem)
     *bgm = 0;
 }
 
-u16 sub_02055428(FieldSystem *fieldSystem, int param1)
+u16 sub_02055428(FieldSystem *fieldSystem, int mapID)
```
lhearachel, line 140: **suggestion**: `Sound_GetOverrideBGM`? `Sound_GetSpecialBGMByMapID`?

`int mapID` may also match as `enum MapHeader mapID` (from `generated/map_headers.h`).

src/unk_020553DC.c
```diff
@@ -281,52 +278,52 @@ static void sub_020555CC(FieldSystem *fieldSystem, int param1, int *param2, int
     }
 }
 
-u16 sub_0205560C(int param0)
+u16 Trainer_GetEncounterBGM(int trainerID)
```
lhearachel, line 281: **suggestion**: `int trainerID` should match as `enum Trainer trainerID`.

lhearachel (commented):
A few notes here, mostly decorating enums and suggesting alternate names.

AsparagusEduardo:
Review suggestions applied :)

generated/meson.build
```diff
@@ -84,7 +84,7 @@ metang_generators = {
     'species': { 'type': 'enum', 'tag': 'Species' },
     'species_data_params': { 'type': 'enum', 'tag': 'SpeciesDataParam' },
     'text_banks': { 'type': 'enum', 'tag': 'TextBank' },
-    'trainers': { 'type': 'enum', 'tag': 'Trainer' },
+    'trainers': { 'type': 'enum', 'tag': 'TrainerId' },
```
lhearachel, line 87: **quibble**: I personally feel that `Trainer` is more descriptive, since it refers to both the trainer's internal data and their party.

If we do keep the suffix, it should be `ID`.

AsparagusEduardo: > **quibble**: I personally feel that `Trainer` is more descriptive, since it refers to both the trainer's internal data and their party.

Right, but there's already a `Trainer` struct, so I wanted to avoid that duplicity.

> If we do keep the suffix, it should be `ID`.

Can do. Just let me know if it's ok to keep it.

lhearachel: > Right, but there's already a Trainer struct, so I wanted to avoid that duplicity.

Fair point. Let's go with the `ID` suffix, then. :grin:

AsparagusEduardo: Done 👍

PR #470, opened 2025-04-10
Text Bank message ID formatting

AsparagusEduardo:
From the discussions on #441, I'm finally back to standarize the text bank ID formatting to how I structured in the Rotom Room.


AsparagusEduardo:
Location names will be handled in a separate PR.

lhearachel (approved):
Thanks for taking the time to do this. :heart:

PR #471, opened 2025-04-11
Generating and documenting accessories constants

CharlesFolz4:
Adding `/generated/accessories.txt` to generate accessory constants

Replacing most instances of those constants in the code


generated/meson.build
```diff
@@ -12,6 +12,7 @@ metang_exe = find_program('metang', native: true)
 
 metang_generators = {
     'abilities': { 'type': 'enum', 'tag': 'Ability' },
+    'accessories': {'type': 'enum', 'tag': 'Accessories'},
```
ravepossum, line 15: suggestion: The tag here should be `Accessory` singular - that's what the enum ends up being named.

src/unk_0204CDDC.c
```diff
@@ -108,22 +109,22 @@ BOOL ScrCmd_217(ScriptContext *ctx)
 }
 
 static const u16 Unk_020EBF64[] = {
```
ravepossum, line 111: suggestion: If we know what this is well enough to confirm that those are item and accessory IDs, we should name this array (and the ScrCmd that uses it, if you're feeling froggy 😄 ). Looks like it's used to get a random item in Amity Square or something?

src/unk_0204CDDC.c
```diff
@@ -25,12 +26,12 @@ BOOL ScrCmd_GetAmitySquareStepCount(ScriptContext *ctx)
 }
 
 static const u16 Unk_020EBF84[6][10] = {
```
ravepossum, line 28: thought: We could probably go ahead and name this too (and constants for the dimensions would be nice as well).

src/scrcmd.c
```diff
@@ -6942,28 +6943,28 @@ static BOOL ScrCmd_26C(ScriptContext *ctx)
...
+    u16 *accessoryID = ScriptContext_GetVarPointer(ctx);
 
     v0 = sub_0202A750(ctx->fieldSystem->saveData);
     v1 = sub_02029D04(v0);
 
     v4 = 0;
 
-    for (v3 = 0; v3 < (49 - 34 + 1); v3++) {
-        if (sub_02029D50(v1, 34 + v3, 1) == 1) {
+    for (v3 = 0; v3 < (ACCESSORY_GLITTER_POWDER - ACCESSORY_PRETTY_DEWDROP + 1); v3++) {
```
ravepossum, line 6953: question: Do we know what the significance of these particular accessories is? If so, it'd be nice to make a descriptive constant or two to make it clearer.

ravepossum (commented):
Looks good! Just a couple thoughts.

src/scrcmd.c
CharlesFolz4: These are just the first and last accessories in the range available from the [Veilstone City Massage Girl.](https://bulbapedia.bulbagarden.net/wiki/Veilstone_City#Massage_Girl)  Rather than making more constants, it might be a better approach to clarify this by naming the function into something that gives it that relevant context?

ravepossum: Personally I think you can never have too many descriptive constants (and `NUM_MASSAGE_GIRL_ACCESSORIES` or something would fit well here IMO) but I do agree that if the rest of the function is documented it should be pretty clear.

src/unk_0204CDDC.c
```diff
@@ -102,36 +103,36 @@ BOOL ScrCmd_217(ScriptContext *ctx)
...
+    ACCESSORY_GLITTER_BOULDER,
+    ACCESSORY_BLACK_PEBBLE,
+    ACCESSORY_BIG_SCALE,
+    ACCESSORY_BIG_LEAF,
+    ACCESSORY_THICK_MUSHROOM,
+    ACCESSORY_STUMP
 };
 
-BOOL ScrCmd_2DF(ScriptContext *param0)
+BOOL ScrCmd_GetAmitySquareBerryAndAccessoryManOptionID(ScriptContext *param0)
```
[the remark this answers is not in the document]
CharlesFolz4: Not sure it's really possible to make them significantly more concise without losing important information, such as that there are two completely distinct accessory sources, and that the ID retrieved from `ScrCmd_GetAmitySquareBerryAndAccessoryManOptionID` is not an itemID for a berry or an accessoryID, but is instead is an ID for indexing the sBerryAndAccesoryManOptions array, but the other function for the berry/item man does give a berryID or an accessoryID. 

Using Calc for two of the functions here does seem more descriptive though.

src/unk_0204CDDC.c
```diff
@@ -102,36 +103,36 @@ BOOL ScrCmd_217(ScriptContext *ctx)
...
-    0xB6,
-    0xB7,
-    0xB,
-    0x7,
-    0xA,
-    0x10,
-    0x1B,
-    0x20,
-    0x21
+static const u16 sBerryAndAccesoryManOptions[] = {
```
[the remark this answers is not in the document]
CharlesFolz4: I'm not a big fan of `option` here either, but the problem is that accessories are not technically items in the same way that berries are.  Calling this `items` would imply that every ID in here should be indexed to be items, but we have an ID 11 in here, which is accessory ID 11 for mini pebble, not to be confused with item ID 11, which is the luxury ball.

ravepossum (approved):

PR #472, opened 2025-04-11
Document location data

AsparagusEduardo:


AsparagusEduardo:
Ready for review

include/unk_0203A944.h
```diff
@@ -3,6 +3,6 @@
 
 #include "field/field_system_decl.h"
 
-int FieldSystem_GetWeather(FieldSystem *fieldSystem, int param1);
+int FieldSystem_GetWeather(FieldSystem *fieldSystem, int mapHeaderID);
```
lhearachel, line 6: **suggestion**:

```suggestion
int FieldSystem_GetWeather(FieldSystem *fieldSystem, enum MapHeader mapHeaderID);
```

I think that we can also name this header, or merge it into the primary `field_system` header.

src/unk_0203A944.c
```diff
@@ -1846,31 +1846,28 @@ static const u8 Unk_020E98F4[] = {
...
+    GF_ASSERT(dayOfYear >= 0 && dayOfYear < 366);
 
-        if ((v2->date.month > 2) && !inline_0203A944(v2->date.year)) {
-            v1++;
-        }
-
-        if (FieldSystem_HasPenalty(fieldSystem)) {
-            v1 = 1;
-        }
+    if ((v2->date.month > 2) && !inline_0203A944(v2->date.year)) {
```
lhearachel, line 1862: **suggestion**: This is probably for leap years, no? Could we document that one while we're in here?

src/unk_0203A944.c
```diff
@@ -1846,31 +1846,28 @@ static const u8 Unk_020E98F4[] = {
...
+        dayOfYear++;
+    }
 
-        v0 = Unk_020E98F4[5 * v1 + (v0 - 32)];
+    if (FieldSystem_HasPenalty(fieldSystem)) {
+        dayOfYear = 1;
     }
 
-    return v0;
+    weather = Unk_020E98F4[5 * dayOfYear + (weather - OVERWORLD_WEATHER_32)];
```
lhearachel, line 1870: **question**: Is this simulating random weather for a given day?

src/unk_0203A944.c
Viperio19: can confirm that is this for leap years:

```c
static inline BOOL inline_0203A944(u32 param0)
{
    if (((param0 % 4 == 0) && (param0 % 100 != 0)) || (param0 % 400 == 0)) {
        return 1;
    } else {
        return 0;
    }
}
```

lhearachel: Looks like it's only used in this file, too.

Let's:

1. Move the function definition out of `inlines.h` and into this file, and
2. Document it. :grin:

AsparagusEduardo:
Applied changes

lhearachel (approved):

PR #473, opened 2025-04-14
Document Follower Partners

AsparagusEduardo:
Also added defines for in-between vars


res/field/scripts/scripts_follower_partners.s
```diff
@@ -0,0 +1,250 @@
...
+    .byte 0
+    .byte 8
+    .byte 49
+    .byte 0
+    .byte 52
+    .byte 0
+    .byte 97
+    .byte 0
+    .byte 2
+    .byte 0
```
Viperio19, line 52: **question:** any idea what these bytes could mean?

res/field/scripts/scripts_follower_partners.s
```diff
@@ -0,0 +1,250 @@
...
+    ScriptEntry FollowerPartners_Buck
+    ScriptEntry FollowerPartners_Mira
+    ScriptEntry FollowerPartners_Marley
+    ScriptEntryEnd
+
+FollowerPartners_Rival:
+    PlayFanfare SEQ_SE_CONFIRM
+    LockAll
+    FacePlayer
+    GoToIfGe VAR_FOLLOWER_RIVAL_STATE, 3, _003D
```
Viperio19, line 18: **suggestion:**
```diff
- GoToIfGe VAR_FOLLOWER_RIVAL_STATE, 3, _003D
+ GoToIfGe VAR_FOLLOWER_RIVAL_STATE, 3, FollowerPartners_Rival_PostStarterMessage
```

res/field/scripts/scripts_follower_partners.s
```diff
@@ -0,0 +1,250 @@
+#include "macros/scrcmd.inc"
+#include "res/text/bank/follower_partners.h"
+
+    .data
+
+    ScriptEntry FollowerPartners_Rival
+    ScriptEntry _00F9
```
Viperio19, line 7: **question:** is this script used anywhere?

Viperio19 (commented):
Looks great! Just a couple small things about the script file, and besides that it looks good to me.

res/field/scripts/scripts_follower_partners.s
```diff
@@ -0,0 +1,250 @@
...
+    ReleaseAll
+    End
+
+FollowerPartners_Mira_IncreaseTimesTalked:
+    AddVar VAR_FOLLOWER_MIRA_TIMES_TALKED, 1
+    WaitABXPadPress
+    CloseMessage
+    ReleaseAll
+    End
+    End
```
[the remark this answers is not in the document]
AsparagusEduardo: Dang, it doesn't

res/field/scripts/scripts_follower_partners.s
AsparagusEduardo: > **question:** any idea what these bytes could mean?

I don't, sorry

res/field/scripts/scripts_follower_partners.s
AsparagusEduardo: > **question:** is this script (`_00F9`) referenced anywhere?

`FollowerPartners_Rival` isn't referenced either, as this seems to be a table to store scripts.
This script is most likely Riley's, but the messages in the text bank are not actually referenced in this file, so I'm not sure if I should call it `FollowerPartners_RileyUnused` or not, as it may actually be used somehow.
There are duplicates for the text in Iron Island's banks, so I'm not sure how to proceed here.

Viperio19: The scripts in this table are referenced by scripts `970X`, for example:
```json
            "graphics_id": "OBJ_EVENT_GFX_BARRY",
            "movement_type": "MOVEMENT_TYPE_LOOK_SOUTH_AND_WEST",
            "flag": 370,
            "script": 9700,
```
9700 (rival), 9702 (cheryl), 9703 (buck) and 9705 (marley) are used, but 9701 (likely riley) and 9704 (mira) are not used, those seem to be completely handled in the scripts of their respective locations. It's interesting that `FollowerPartners_Mira` uses the same var `VAR_FOLLOWER_MIRA_TIMES_TALKED` as the script for Mira in `scripts_wayward_cave_1f.s`, while `_00F9` uses `0x40CB` instead of `0x40E3` which seems to be kinda like `VAR_FOLLOWER_RILEY_TIMES_TALKED` from `scripts_iron_island_b2f_left_room.s`. Although it's likely that's Riley, I'd say just leave it unnamed since we can't be 100% sure from what I can tell.

src/script_manager.c
```diff
@@ -228,7 +228,7 @@ static u16 ScriptContext_LoadAndOffsetID(FieldSystem *fieldSystem, ScriptContext
         ScriptContext_Load(fieldSystem, ctx, scripts_unk_0212, TEXT_BANK_UNK_0217);
         retScriptID -= 9800;
     } else if (retScriptID >= 9700) {
-        ScriptContext_Load(fieldSystem, ctx, scripts_unk_0422, TEXT_BANK_UNK_0429);
+        ScriptContext_Load(fieldSystem, ctx, scripts_follower_partners, TEXT_BANK_FOLLOWER_PARTNERS);
         retScriptID -= 9700;
```
Viperio19, line 232: **polish:** define a constant for this in `script_manager.h`:
`#define SCRIPT_ID_OFFSET_FOLLOWER_PARTNERS      9700`

res/field/scripts/scripts_follower_partners.s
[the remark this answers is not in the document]
lhearachel: That makes sense to me; generally, these byte-code scripts are interpreted in a literal manner, so there's nothing "smart" going on to clean up unnecessary duplicates.

PR #474, opened 2025-04-14
Define and use Var constants

AsparagusEduardo:
~~Requires #473~~


include/constants/savedata/vars_flags.h
```diff
@@ -4109,10 +4109,59 @@
 #define FLAG_UNK_0x0FFE                             0x0FFE
 #define FLAG_UNK_0x0FFF                             0x0FFF
 
-#define VARS_START 0x4000
+#define VARS_START        0x4000
```
lhearachel, line 4112: **note**: Some general feedback on these from knowledge from the binary-hacking folks:

1. Range `0x4000` to `0x401F` are general-purpose local variables that are zero-filled when the current map header changes.
2. Range `0x4020` to `0x402F` are persistent and related to NPC sprites. I'm not certain about their exact function, though.
3. Range `0x4030` to `0x406F` are persistent and control various game functions, as documented below.
4. Range `0x4070` to `0x411F` are persistent and used for tracking story progression.
5. Range `0x8000` to `0x800C` are general-purpose local variables that are zero-filled when the current script finishes execution.
6. `0x800D` always contains the value of the overworld object which initiated the script, if any (basically, the NPC that you interacted with or the trainer that saw you).

include/constants/savedata/vars_flags.h
```diff
@@ -4298,7 +4347,78 @@
...
+#define VAR_UNK_0x411A                        (SYSTEM_VARS_START + 234)
+#define VAR_UNK_0x411B                        (SYSTEM_VARS_START + 235)
+#define VAR_UNK_0x411C                        (SYSTEM_VARS_START + 236)
+#define VAR_UNK_0x411D                        (SYSTEM_VARS_START + 237)
+#define VAR_UNK_0x411E                        (SYSTEM_VARS_START + 238)
+#define VAR_UNK_0x411F                        (SYSTEM_VARS_START + 239)
+#define VARS_END                              (SYSTEM_VARS_START + 240)
+#define NUM_VARS                              (VARS_END - VARS_START)
 
 #define SPECIAL_VARS_START 0x8000
```
lhearachel, line 4408: **suggestion**: Based on the above, should we rename `SPECIAL_VARS_START` to `SCRIPT_LOCAL_VARS_START`, or is the familiarity of this name from `pokeemerald` valuable?

include/constants/savedata/vars_flags.h
```diff
@@ -4298,7 +4347,78 @@
...
+#define VAR_0x8003         (SPECIAL_VARS_START + 3)
+#define VAR_0x8004         (SPECIAL_VARS_START + 4)
+#define VAR_0x8005         (SPECIAL_VARS_START + 5)
+#define VAR_0x8006         (SPECIAL_VARS_START + 6)
+#define VAR_0x8007         (SPECIAL_VARS_START + 7)
+#define VAR_0x8008         (SPECIAL_VARS_START + 8)
+#define VAR_0x8009         (SPECIAL_VARS_START + 9)
+#define VAR_0x800A         (SPECIAL_VARS_START + 10)
+#define VAR_0x800B         (SPECIAL_VARS_START + 11)
+#define VAR_0x800C         (SPECIAL_VARS_START + 12)
```
lhearachel, line 4421: **suggestion**: I believe that variable `0x800C` is overwhelmingly used as a result-variable. Could we rename it to `VAR_RESULT`?

include/constants/savedata/vars_flags.h
```diff
@@ -4298,7 +4347,78 @@
...
+#define VAR_0x800D         (SPECIAL_VARS_START + 13)
```
lhearachel, line 4422: **suggestion**: Per the above, let's rename this to `VAR_LAST_TALKED` (borrowing from `pokeemerald`).

res/field/scripts/scripts_acuity_cavern.s
```diff
@@ -34,10 +34,10 @@ AcuityCavern_Uxie:
...
     StartLegendaryBattle SPECIES_UXIE, 50
     ClearFlag FLAG_UNK_0x008E
-    CheckWonBattle 0x800C
-    GoToIfEq 0x800C, FALSE, AcuityCavern_LostBattle
-    CheckDidNotCapture 0x800C
-    GoToIfEq 0x800C, TRUE, AcuityCavern_UxieDisappeared
+    CheckWonBattle VAR_0x800C
+    GoToIfEq VAR_0x800C, FALSE, AcuityCavern_LostBattle
+    CheckDidNotCapture VAR_0x800C
+    GoToIfEq VAR_0x800C, TRUE, AcuityCavern_UxieDisappeared
```
lhearachel, line 40: **thought**: I'm wondering if we want to undertake the effort of setting up `Switch` and `Case` macros like `pokeemerald` has: https://github.com/pret/pokeemerald/blob/f11bddf8548c2f47e2881f6211a24bc427a708b6/asm/macros/event.inc#L1871-L1878

lhearachel (commented):
This looks great, thanks for doing this! I have a few thoughts about the names of some symbols below.

include/constants/savedata/vars_flags.h
AsparagusEduardo: > **suggestion**: Based on the above, should we rename `SPECIAL_VARS_START` to `SCRIPT_LOCAL_VARS_START`, or is the familiarity of this name from `pokeemerald` valuable?

I believe that the new name is a better descriptor on what they do, so I'm good with renaming them :)

include/constants/savedata/vars_flags.h
AsparagusEduardo: What do you think of the following recommendations for the ranges:

1. Range `0x4000` to `0x401F`: `MAP_LOCAL_VARS_x`
2. Range `0x4020` to `0x402F`: I took a peek at it and I believe they're used for setting dynamic object event graphics. Would `NPC_GFX_VARS_x` be good?
> 3. Range `0x4030` to `0x406F`: Keeping these as `SYSTEM_VARS_x`
> 4. Range `0x4070` to `0x411F`: `STORY_VARS_x`

For the other 2 I'm using the recommendations given in other comments.

res/field/scripts/scripts_acuity_cavern.s
AsparagusEduardo: > **thought**: I'm wondering if we want to undertake the effort of setting up `Switch` and `Case` macros like `pokeemerald` has: [pret/pokeemerald@`f11bddf`/asm/macros/event.inc#L1871-L1878](https://github.com/pret/pokeemerald/blob/f11bddf8548c2f47e2881f6211a24bc427a708b6/asm/macros/event.inc#L1871-L1878)

Maybe on a separate PR

include/constants/savedata/vars_flags.h
AsparagusEduardo: Edit for point 2: pokeemerald has `VAR_OBJ_GFX_ID_0` and such. Maybe this is better?

lhearachel: I like that idea, let's go with that.

AsparagusEduardo:
Rebased and applied range renames. Ready for re-review.

lhearachel (approved):

PR #477, opened 2025-04-19
Tag previously untagged enums

Gudf:
This leaves out undocumented enums and a few enums I couldn't find a name for and feel would fit better as `#define`s instead.

I also replaced one instance of a `typedef`'d (and untagged) enum to use the new tag instead.

I'll probably open other PRs to replace types to use these newly tagged enums and potentially standardize variable names where appropriate, starting with the NARC IDs, but I felt that was out of scope for this PR.


lhearachel (approved):
Thanks for doing this!

PR #479, opened 2025-04-20
Switch to `enum NarcID`, standardize names

Gudf:
This replaces instances of NARC IDs with type `int` or `u32` to use `enum NarcID` instead, and replaces most instances of a NARC ID to be named `narcID` instead of e.g. `narc`, `narcIndex` or `narcIdx`.

Undocumented instances were left untouched for this PR.


lhearachel (approved):
God bless you for doing this. :pray:

PR #482, opened 2025-04-21
Integrate metroskrew as the default compiler

lhearachel:
Supersedes the previous PR #254.

This removes the use of `mwrap` as a wrapper around the metrowerks compiler and replaces it with `metroskrew`, a re-link of the proprietary tooling to function on modern Unix systems. The meat of this PR is in 9a2adc819adde9457207da1bd3aa1908742be731, which performs the actual integration, updates documentation, and provides some end-user tooling to make installation straight-forward. The rest is plumbing/compatibility or lingering elements from #478, on which this branch is based.

This PR also deprecates the Docker installation procedure, which is unmaintained and (as far as I am aware) unused, as it does not function. We can explore rebuilding it at a later time.

### Testing Procedure

For each Platform listed in the table below, in order:

1. ensure that the New Dependencies are installed using the system package manager;
2. run `make skrew` (potentially with necessary root privileges);
3. verify that the Expected Archive is downloaded; 
4. remove any existing build directory and run `make configure`; and
5. verify that `grep 'Detecting compiler via:' build/meson-logs/meson-log.txt` contains lines with the Expected Compiler.

| Platform                         | New Dependencies             | Expected Archive        | Expected Compiler |
| -------------------------------- | ---------------------------- | ----------------------- | ----------------- |
| Windows via MSYS2    | [p7zip][p7zip], [wget][wget] | metroskrew-windows.7z   | `mwccarm.exe`     |
| WSL1 (Windows mount) | [p7zip][p7zip], [wget][wget] | metroskrew-windows.7z   | `mwccarm.exe`     |
| WSL2 (Linux mount)   | [xz][xz], [wget][wget]       | metroskrew-linux.tar.xz | `mwccarm`         |
| Native x86 Linux     | [xz][xz], [wget][wget]       | metroskrew-linux.tar.xz | `mwccarm`         |
| macOS                | [xz][xz], [wget][wget]       | metroskrew-linux.tar.xz | `mwccarm`         |

[xz]: https://repology.org/project/xz/versions
[p7zip]: https://repology.org/project/p7zip/versions
[wget]: https://repology.org/project/wget/versions

### Verified Platform Support

- [x] Native x86 Linux - confirmed `OK` by @lhearachel 
- [x] MSYS2 - confirmed `OK` by @lhearachel 
- [x] WSL1 (Windows drive) - confirmed `OK` by @mlastes1987; ~~**bug on repeated builds** identified by @cbt6~~ fixed on the current revision
- [x] WSL2 (Linux drive) - confirmed `OK` by @mid-kid and @Fexty12573 
- [x] macOS - confirmed `OK` on both ARM and Intel Macs by @lhearachel

### Further Pre-Merge Action Items

- [x] Migrate the installation script to target a directory within the project
- [x] Add `skrew` as a dependency in the initial `make`
- [x] Update `INSTALL.md` with the new procedure for all platforms
- [x] Update the Nitro subprojects to support a new option `mwrap_sdk` to reflect the `-wrap:sdk` option passed to `metroskrew`


meson.build
```diff
@@ -85,27 +86,27 @@ link_args = [
 ############################################################
 nitrosdk_dep = dependency('NitroSDK',
     default_options: {
-        'mwrap_ver': '2.0/sp1p2',
+        'mwrap_ver': '3.0-126',
```
mid-kid, line 89: While I suppose this works, I'd rather update the submodules to support an `mwrap_sdk` option, reflecting the `-wrap:sdk` option. This doesn't just change the compiler, but also the header files used.

tools/devtools/get_metroskrew.sh
```diff
@@ -0,0 +1,34 @@
...
+*)
+    echo "Unrecognized platform $PLATFORM"
+    exit 1
+    ;;
+esac
+
+SKREW_GH="https://github.com/mid-kid/metroskrew"
+SKREW_ARC="metroskrew-$PLATFORM$SKREW_EXT"
+SKREW_URL="$SKREW_GH/releases/download/$SKREW_VER/$SKREW_ARC"
+DESTDIR="/usr/local"
```
mid-kid, line 27: This is a root-protected directory. We currently install both meson as well as any subprojects within the repo, so I'd keep it there as well. That said, maybe an install script like this should be part of metroskrew?

mid-kid (commented):
Preliminary comments, haven't ran any tests or anything yet.

tools/devtools/get_metroskrew.sh
mid-kid: Oh, now I get it. This is meant to be a separate step executed as `sudo make skrew`. That works, but I'll need a second to think about it.

lhearachel: > We currently install both meson as well as any subprojects within the repo, so I'd keep it there as well.

I would definitely prefer that, but my understanding of metroskrew's functionality is that everything needed to live within `/usr/local`. Or is that just to automatically pick them up via the user's `PATH` and `LD_LIBRARY_PATH`? What about the contents that go into `/usr/local/share`?

lhearachel:
Summary of force-push to 1b2aa37:

- Installations now target a directory within the project, and the Meson cross-files have been updated to use this path to avoid modifying the user's `PATH`.
- Installation instructions for Linux (and, by consequence, WSL2) now explicitly instruct users to install 32-bit `glibc`.
- The installation script drops the extracted archive to `/tmp` before inflating it into the project-local destination.
- The check in the `Makefile` for `WSL_ACCESSING_WINDOWS` wraps `$(CWD)` in `$(realpath)` to correct detection for users that access the project directory via a symlink.

tools/devtools/get_metroskrew.sh
lhearachel: Resolved by `ad648eb`.

lhearachel:
The current state of this PR `OK`s on an ARM M3 Macbook Pro with the following additional changes:

```diff
diff --git a/Makefile b/Makefile
index 09cff93be..c513cecaf 100644
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ else
 ifneq (,$(findstring Darwin,$(UNAME_S)))
 NATIVE := native_macos.ini
 CROSS := cross_unix.ini
-SKREW := linux
+SKREW := wine
 else
 ifneq (,$(findstring BSD, $(UNAME_S)))
 NATIVE := native.ini
```

Before committing this change, though, I will test with an older Intel Macbook, as well. Older Intel Macbooks may not require `wine`. If so, then this may necessitate a refactor of the `INSTALL.md` to account for different host-platform architectures.

lhearachel:
The latest force-push above adds a `wine` platform for MacOS systems and utilization of the `mwrap_sdk` option in the Nitro-libraries to set the specific compiler versioning to be used. Further, updates have been made to the installation instructions for each platform to clarify usage, and `metroskrew` is now automatically installed as a build dependency within the `Makefile`.

lhearachel:
Installation instructions have been updated for MacOS for clarity. Intel Mac is now OK, which means that we have no degradation of platform-support. :tada:

lhearachel:
Commit 5e3d4ed has been modified to:

1. Resolve a bug with the `metroskrew` dependency-resolution chain.
2. Tidy up the `get_metroskrew` script.
3. Reorganize changes to the `Makefile` into more logical sections.

PR #490, opened 2025-04-26
Document Particle System

Fexty12573:
This PR documents the particle system used by the game which is essentially just a thin wrapper around SPL with some added QoL features like automatic resource cleanup.

Relevant files are `particle_system.h/c`


include/particle_system.h
```diff
@@ -0,0 +1,100 @@
...
+    VRAM_AUTO_RELEASE_TEXTURE_FRM = (1 << 0), // Auto-release VRAM using the Frame Texture VRAM Manager
+    VRAM_AUTO_RELEASE_TEXTURE_LNK = (1 << 1), // Auto-release VRAM using the Linked-List Texture VRAM Manager
+    VRAM_AUTO_RELEASE_PALETTE_FRM = (1 << 2), // Auto-release VRAM using the Frame Palette VRAM Manager
+    VRAM_AUTO_RELEASE_PALETTE_LNK = (1 << 3), // Auto-release VRAM using the Linked-List Palette VRAM Manager
+};
+
+typedef struct ParticleSystem {
+    SPLManager *manager;
+    void *resource;
+    SPLEmitter *latestEmitter; // Last created emitter
```
lhearachel, line 28: **suggestion**: Is `tailEmitter` more succinct/clearer, since this seems to be a linked-list type structure?

include/particle_system.h
```diff
@@ -0,0 +1,100 @@
+#ifndef POKEPLATINUM_UNK_02014000_H
```
lhearachel, line 1: **polish**: Update the include guards and the comment at the bottom.

src/particle_system.c
```diff
@@ -0,0 +1,750 @@
...
+static void *sEmitterCallbackParam;
+
+// Particle system whose resources are currently being uploaded to VRAM.
+// Used so custom Texture/Palette allocators can save the resources.
+// See ParticleSystem_Register*Key
+static ParticleSystem *sUploadingParticleSystem;
+
+static const VecFx32 sParticleSystemDefaultCameraPos = { FX32_CONST(0), FX32_CONST(0), FX32_CONST(4) };
+static const VecFx32 sParticleSystemDefaultCameraUp = { FX32_CONST(0), FX32_CONST(1), FX32_CONST(0) };
+static const VecFx32 sParticleSystemDefaultCameraTarget = { FX32_CONST(0), FX32_CONST(0), FX32_CONST(0) };
```
lhearachel, line 42: **thought**: Would it be valuable to have our own macro for creating vectors of `Fx32`? e.g.,

```c
#define VEC_FX32(__x, __y, __z) (VecFx32){ FX32_CONST(__x), FX32_CONST(__y), FX32_CONST(__z) }
```

src/particle_system.c
```diff
@@ -0,0 +1,750 @@
...
+    ParticleSystem14_AllocMemory,
+    ParticleSystem15_AllocMemory,
+};
+
+void ParticleSystem_InitAll(void)
+{
+    for (int i = 0; i < MAX_PARTICLE_SYSTEMS; i++) {
+        sParticleSystems[i] = NULL;
+    }
+}
```
lhearachel, line 88: **suggestion**: This is more of a `ZeroAll`, IMO, or a `Clear`.

src/particle_system.c
```diff
@@ -0,0 +1,750 @@
...
+    particleSystem->heapStart = heap;
+    particleSystem->heap = heap;
+    particleSystem->heapEnd = (void *)((u32)heap + heapSize);
+    particleSystem->id = id;
+    sParticleSystems[id] = particleSystem;
+
+    if (hasCamera == TRUE) {
+        particleSystem->camera = Camera_Alloc(heapID);
+
+        {
```
lhearachel, line 129: **polish**: Unnecessary scope-block.

lhearachel (commented):
This looks great to my eyes. Just a few little suggestions about specific names.

include/particle_system.h
Fexty12573: It is, yes, but that is really an implementation detail of SPL. The intent seems like to provide easy access to the last emitter you created without having to pass it around everywhere. In that regard I think calling it "tail" is kind of unintuitive. Then again, that field isn't actually used anywhere except in the places it's assigned 😅

src/particle_system.c
Fexty12573: That seems like a good idea. Question is where to put it. `fx_util.h` or `math.h` maybe?

include/particle_system.h
lhearachel: That's valuable info, thanks! `lastAddedEmitter` would probably be the most accurate, then, but I `latestEmitter` also gets the idea across, IMO. 🙂

src/particle_system.c
lhearachel: I think that it fits better in `fx_util.h`, since it exposes some vector-functions, whereas `math_util.h` does not.

lhearachel (approved):
Pending rebase before merge.

PR #491, opened 2025-04-27
Pal Park map script documentation

AsparagusEduardo:


res/field/scripts/scripts_canalave_city.s
```diff
@@ -53,7 +53,7 @@ _00F8:
 
 _0100:
     SetObjectEventPos 16, 55, 0x2CC
-    ScrCmd_189 16, 3
+    SetObjectEventDir 16, 3
```
Viperio19, line 56: **suggestion:** put the `DIR_.....` constants in a generated file and use them here too

res/text/pal_park_lobby.gmm
```diff
@@ -1,246 +1,246 @@
...
 		<attribute name="window_context_name">used</attribute>
 		<language name="English">Would you like to take part in\na Catching Show now?</language>
 	</row>
-	<row id="pl_msg_00000515_00029" index="29">
+	<row id="PalParkLobby_Text_AwwThatsTooBad" index="29">
 		<attribute name="window_context_name">used</attribute>
 		<language name="English">Aww, that’s too bad.\rWell, just give me a shout if you’d\nlike to participate.</language>
 	</row>
-	<row id="pl_msg_00000515_00030" index="30">
+	<row id="PalParkLobby_Text_HiTherePlayerDidYouComeToParticipateInACatchingShow_PlayerMale" index="30">
```
Viperio19, line 130: **quibble:** I think it would be good to look critically at some of these text names and see if they can be more compact

src/scrcmd.c
```diff
@@ -6158,16 +6158,16 @@ static BOOL ScrCmd_1D5(ScriptContext *ctx)
...
-    u16 *v3 = ScriptContext_GetVarPointer(ctx);
+    u16 backdrop = ScriptContext_GetVar(ctx);
+    u16 *destVar = ScriptContext_GetVarPointer(ctx);
 
     v0 = sub_0202A750(ctx->fieldSystem->saveData);
     v1 = sub_02029D04(v0);
-    *v3 = sub_02029D80(v1, v2);
+    *destVar = sub_02029D80(v1, backdrop);
 
     return 0;
```
Viperio19, line 6172:
```suggestion
    return FALSE;
```

Viperio19 (commented):
I don't have time atm for a full thorough look through `res/field/scripts/scripts_pal_park.s` yet but here are just some small comments in general so far

res/text/pal_park_lobby.gmm
AsparagusEduardo: > **quibble:** I think it would be good to look critically at some of these text names and see if they can be more compact

We could set up a max char length for these if you want

res/field/scripts/scripts_canalave_city.s
AsparagusEduardo: > **suggestion:** put the `DIR_.....` constants in a generated file and use them here too

Done

res/text/pal_park_lobby.gmm
Viperio19: I don't love a set character limit because the header part (`PalParkLobby_Text_`) also varies in length, it's more that I think that names should not include almost the entire message, but I'm interested to hear other people's opinions about this

res/field/scripts/scripts_pal_park.s
```diff
@@ -59,46 +59,46 @@ _0020:
...
-    GoToIfEq VAR_RESULT, 0, _0114
-    SetVar VAR_UNK_0x40F3, 2
+    GoToIfEq VAR_RESULT, FALSE, _0114
+    SetVar VAR_PAL_PARK_STATE, 2
     ReleaseAll
-    Call _01C0
+    Call PalPark_ClearFlagAndWarpOut
     End
 
 _0114:
```
Viperio19, line 104: **suggestion:** for completion sake, it would be nice if this could be named as well (same for `_0159`)

res/field/scripts/scripts_pal_park_lobby.s
```diff
@@ -1,159 +1,161 @@
...
-    GoToIfNe VAR_UNK_0x40F3, 0, _004D
+    SetFlag FLAG_FIRST_ARRIVAL_POKE_PARK_FRONT_GATE
+    GoToIfNe VAR_PAL_PARK_STATE, 0, _004D
     End
 
 _004D:
     SetObjectEventPos 7, 8, 9
-    ScrCmd_189 7, 2
+    SetObjectEventDir 7, DIR_WEST
     End
```
Viperio19, line 33: **question:** the localID `7` shows up a lot through this file, do you know what it is? If so, I think it would be nice to use a constant like this:
`#define LOCALID_.... 7`

res/field/scripts/scripts_pal_park_lobby.s
```diff
@@ -288,116 +290,117 @@ _03F8:
...
 
-_04B4:
+PalParkLobby_Oak:
     LockAll
-    ClearFlag FLAG_UNK_0x0242
-    SetVar VAR_UNK_0x40E0, 0x7D0
+    ClearFlag FLAG_ETERNA_CITY_SOUTH_HOUSE_HIDE_PROF_OAK
+    SetVar VAR_CATCHING_SHOW_RECORD, 2000
     ApplyMovement 1, _0548
```
Viperio19, line 380: **suggestion:** the localID `1` seems to be for professor oak, so like the previous comment, I'd suggest:
`#define LOCALID_PROF_OAK 1`

res/text/pal_park_lobby.gmm
lhearachel: I think shortening is fine; for something like this, e.g., `PalParkLobby_Text_Greeting_PlayerMale`

Viperio19: yeah these shortened names look much nicer imo, nicely done!

res/field/scripts/scripts_pal_park_lobby.s
AsparagusEduardo: > **question:** the localID `7` shows up a lot through this file, do you know what it is? If so, I think it would be nice to use a constant like this: `#define LOCALID_.... 7`

Sorry for the delay, but I figured it out. It's the receptionist. I'll add his local ID alongside Prof Oak's

AsparagusEduardo:
Ready for re-review

generated/facing_directions.txt
lhearachel, line 1: **question**: Maybe a dumb question, but is there a good reason to `enum`-ify these for `c` / `python`? From what I can tell, they are just `#define`s at present, and there isn't a strong rationale to change them (namely, having them used as `struct` members or function parameters)...?

AsparagusEduardo: Reverted

lhearachel (approved):

PR #492, opened 2025-04-27
Add exception handling to json data parsing

CharlesFolz4:
Added exception handling to most instances of `json.load()` to improve the signal-to-noise ratio of error messages for incorrectly formatted json files.

json formatting errors now look like this:
Screenshot_20250427_143915

### DIFF
--- a/tools/datagen/datagen_trainer.cpp
+++ b/tools/datagen/datagen_trainer.cpp
@@ -215,7 +215,11 @@ int main(int argc, char **argv)
         try {
             fs::path trainerDataPath = dataRoot / (trainerStem + ".json");
             std::string json = ReadWholeFile(trainerDataPath);
-            doc.Parse(json.c_str(), json.length());
+            rapidjson::ParseResult ok = doc.Parse(json.c_str(), json.length());
+            if (!ok) {
+                ReportJsonError(ok, json, trainerDataPath);
+                std::exit(EXIT_FAILURE);
+            }
 
             TrainerHeader trdata = ParseTrainerData(doc);
             narc_pack_file_copy(trdataVFS, reinterpret_cast<unsigned char *>(&trdata), sizeof(trdata));
--- a/tools/datagen/datagen_trainer.cpp
+++ b/tools/datagen/datagen_trainer.cpp
@@ -212,20 +212,19 @@ int main(int argc, char **argv)
 
     rapidjson::Document doc;
     for (auto &trainerStem : trainerRegistry) {
-        try {
-            fs::path trainerDataPath = dataRoot / (trainerStem + ".json");
-            std::string json = ReadWholeFile(trainerDataPath);
-            rapidjson::ParseResult ok = doc.Parse(json.c_str(), json.length());
-            if (!ok) {
-                ReportJsonError(ok, json, trainerDataPath);
-                std::exit(EXIT_FAILURE);
-            }
+        fs::path trainerDataPath = dataRoot / (trainerStem + ".json");
+        std::string json = ReadWholeFile(trainerDataPath);
+        rapidjson::ParseResult ok = doc.Parse(json.c_str(), json.length());
+        if (!ok) {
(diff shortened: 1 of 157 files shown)


tools/json2bin/encdata_ex_elusive_rod.py
```diff
@@ -14,10 +14,12 @@ def as_species(s: str) -> bytes:
...
-
-data = {}
-with open(input_path, 'r', encoding='utf-8') as input_file:
-    data = json.load(input_file)
+try:
+    data = {}
+    with open(input_path, 'r', encoding='utf-8') as input_file:
+        data = json.load(input_file)
+except json.decoder.JSONDecodeError as e:
+    sys.exit(f"Error parsing {input_path}\n{e.msg} at line {e.lineno}\n")
```
lhearachel, line 22: **suggestion**: Incorporate [`colno`](https://docs.python.org/3.10/library/json.html#json.JSONDecodeError) into this error message.

---

**question**: What would `doc` and `pos` emit in these cases?

---

**question**: How difficult would it be to also emit the precise line of the output that produced the error? We have the file handle, so I imagine that it should not be difficult.

lhearachel (commented):
Thank you so much for doing this! Could we also implement some similar functionality in the `datagen` cases? I'm not sure how familiar you are with C++, but `rapidjson` has some good [example code](https://rapidjson.org/group___r_a_p_i_d_j_s_o_n___e_r_r_o_r_s.html) for handling potentially invalid JSON inputs.

CharlesFolz4:
Updated error messages to match gcc error style standard.
image

c++ datagen json parsing has not been addressed.

lhearachel (approved):
Pre-approving this ahead of my own changes (which will need their own approval).

tools/datagen/datagen_trainer.cpp
```diff
@@ -215,7 +215,11 @@ int main(int argc, char **argv)
         try {
```
CharlesFolz4, line 215: **suggestion:**  The old try/catch message should probably be updated as well.  The new parsing detection handles parsing errors, not this try/catch, so the displayed error `"exception parsing data file"` is misleading.  The only error that I can see this try/catch actually catching is if the file that is specified does not exist.

lhearachel: Worth noting that the parse-detection handles only verification that the loaded file is a valid JSON document; I did a short test for what happens if I remove necessary-fields from a species document, and the error is still quite unintuitive:

```
[100% 2/2] Generating res/pokemon/datagen_speci... a custom command (wrapped by meson to set env)
FAILED: res/pokemon/pl_personal.narc res/pokemon/evo.narc res/pokemon/wotbl.narc res/pokemon/ppark.narc res/pokemon/height.narc res/pokemon/pl_poke_data.narc res/pokemon/tutorable_moves.h res/pokemon/species_learnsets_by_tutor.h res/pokemon/species_footprints.h
/opt/homebrew/opt/python@3.13/bin/python3.13 /Users/rachel/code/github.com/lhearachel/pokeplatinum/subprojects/meson-1.7.0/meson.py --internal exe --unpickle /Users/rachel/code/github.com/lhearachel/pokeplatinum/build/meson-private/meson_exe_datagen-species_0994334cbba4386f827e8bb65e34827663679770.dat
while executing ['/Users/rachel/code/github.com/lhearachel/pokeplatinum/build/tools/datagen/datagen-species', '/Users/rachel/code/github.com/lhearachel/pokeplatinum/build/res/pokemon', '/Users/rachel/code/github.com/lhearachel/pokeplatinum/res/pokemon', 'res/pokemon/form_data.order', '../res/pokemon/move_tutors.json']
--- stdout ---

--- stderr ---
Assertion failed: (false), function operator[], file document.h, line 1053.

ninja: build stopped: subcommand failed.
make: *** [target] Error 1
```

But I will handle this as part of a separate PR.

PR #502, opened 2025-05-07
Generate pms.narc

h2o-DS:
Added offspring field to species data files
Added pms.narc to datagen_species


lhearachel (commented):
Code looks good. Please also update the header- comments in each of `res/pokemon/meson.build` and `tools/datagen/datagen_species.cpp` to list the new output file (purely for documentation purposes).

lhearachel (approved):

h2o-DS:
Added the comments.

tools/datagen/datagen_species.cpp
```diff
@@ -289,18 +290,18 @@ static std::optional<u16> TryParseOffspring(rapidjson::Document &root)
...
+    offspringData.emplace_back(498); // deoxys speed
+    offspringData.emplace_back(499); // wormadam sandy
+    offspringData.emplace_back(500); // wormadam trash
+    offspringData.emplace_back(501); // shaymin sky
+    offspringData.emplace_back(502); // giratina origin
+    offspringData.emplace_back(503); // rotom heat
+    offspringData.emplace_back(504); // rotom wash
+    offspringData.emplace_back(505); // rotom frost
+    offspringData.emplace_back(506); // rotom fan
+    offspringData.emplace_back(507); // rotom mow
```
lhearachel, line 304: **thought**: We probably want to emplace the actual `SPECIES_` constants here in the event that a user performs a dex expansion, no? Up to you if you want to add that here, or I can handle it alongside the datagen refactoring.

lhearachel (changes requested):
Flipping from `Approve` to `Request Changes` per my above comment.

tools/datagen/datagen_species.cpp
h2o-DS: It would be trivial to add an offspring field to the forms' data files. The reason I didn't was that those forms do not map to a species value. I can look into alternative solutions.

lhearachel: I think that just handling the form-registry in-order is a perfectly reasonable solution, and it keeps the data-sets relatively clean.

h2o-DS:
I was overthinking it. Now if there is no `offspring` field, it just points to itself. This also has the bonus of making the field optional without breaking the table.

tools/datagen/datagen_species.cpp
```diff
@@ -539,11 +524,12 @@ int main(int argc, char **argv)
         }
 
         try {
+            u16 personalValue = find(speciesRegistry.begin(), speciesRegistry.end(), species) - speciesRegistry.begin();
```
lhearachel, line 527: **suggestion**: This method will invoke a `find` on each iteration, which is costly; at worst, you'll transform the compute here from *O(n)* to *O(n^2)*. Instead, keep a rolling `u16` starting at `SPECIES_NONE` and increment it on each iteration.

lhearachel (commented):
Good thought! I have one comment below about a specific of the approach.

lhearachel (approved):

PR #503, opened 2025-05-07
Document ByteFlagSet

Morglom:
Just a small section that stood out. Still a bit unsure about what's going on with the direction in the struct.


include/byte_flag_set.h
```diff
@@ -0,0 +1,27 @@
...
+    u8 rightIndex;
+} ByteFlagSet;
+
+enum ByteFlagSetDirection {
+    BYTE_FLAG_SET_DIRECTION_UP = 0,
+    BYTE_FLAG_SET_DIRECTION_DOWN,
+    BYTE_FLAG_SET_DIRECTION_LEFT,
+    BYTE_FLAG_SET_DIRECTION_RIGHT,
+    BYTE_FLAG_SET_DIRECTION_NO_DIRECTION,
+};
```
lhearachel, line 21: **thought**: This structure makes me think that this is not as general as a `byte_flag_set` and that it is a cursor-tracker. I think that we should analyze how this module is used, as that will give us some clues into how we should document it.

include/byte_flag_set.h
```diff
@@ -0,0 +1,27 @@
+#ifndef POKEPLATINUM_BYTE_FLAG_SET_H
+#define POKEPLATINUM_BYTE_FLAG_SET_H
+
+typedef struct {
```
lhearachel, line 4: **polish**: Tag the `struct` with the same name as the `typedef`, e.g., `struct ByteFlagSet`.

src/overlay013/ov13_02228A38.c
```diff
@@ -96,39 +94,39 @@ static u8 ov13_02228ACC(UnkStruct_ov13_02228A50 *param0)
...
-        ov16_0226DD7C(param0->unk_00, param0->unk_04[param0->unk_09].unk_00, param0->unk_04[param0->unk_09].unk_02, param0->unk_04[param0->unk_09].unk_01, param0->unk_04[param0->unk_09].unk_03);
+        ov16_0226DD7C(param0->unk_00, param0->unk_04[param0->unk_09].flagA, param0->unk_04[param0->unk_09].flagC, param0->unk_04[param0->unk_09].flagB, param0->unk_04[param0->unk_09].flagD);
         Sound_PlayEffect(SEQ_SE_CONFIRM);
     }
 
     return 0;
 }
 
-static u8 ov13_02228B18(const ByteFlagSet *param0, u8 param1)
+static BOOL ov13_02228B18(const ByteFlagSet *byteFlagSet, u8 direction)
```
lhearachel, line 104: **thought**: Continuing from thoughts above, we know that this overlay is used for the in-battle Bag and Party interfaces on the touch screen. My thought, then, is that this `ByteFlagSet` represents some data needed for the cursors on these interfaces to function.

I noticed earlier that what are currently called `flagA` and `flagB` are tied to one another. Maybe these are actually the X-Y coordinates of the cursor?

src/unk_0207E0B8.c
```diff
@@ -8,7 +8,6 @@
 #include "struct_defs/struct_0207C690.h"
```
lhearachel, line 8: **thought**: Going off of the earlier comments about overlay 13 and looking through the code within this file, I would conclude that this is the in-field party list interface (shown on the top screen).

include/byte_flag_set.h
```diff
@@ -0,0 +1,27 @@
...
+    u8 leftIndex;
+    u8 rightIndex;
+} ByteFlagSet;
+
+enum ByteFlagSetDirection {
+    BYTE_FLAG_SET_DIRECTION_UP = 0,
+    BYTE_FLAG_SET_DIRECTION_DOWN,
+    BYTE_FLAG_SET_DIRECTION_LEFT,
+    BYTE_FLAG_SET_DIRECTION_RIGHT,
+    BYTE_FLAG_SET_DIRECTION_NO_DIRECTION,
```
lhearachel, line 20: **polish**: Tighten the final entry's name to `NONE`, e.g. `BYTE_FLAG_SET_DIRECTION_NONE`.

lhearachel (changes requested):
Some preliminary thoughts. I think that this module needs to be renamed, judging on how it's used.

Morglom:
I've followed up on the points of interest you listed and you're definitely right that A and B are x and y coordinates and so are C and D. But I'm not sure what C and D are for exactly. They're only being used in a section that has some item and (I think) bag functionality to set coordinates for a couple of sprites. I'm happy to rename the flags and give C and D something placeholder like "secondX" or "subX" and leave it things are a bit clearer.

I've also kind of gotten side tracked trying to document sections of ov13 in order to get said clearer picture and I think that including that work would kind of bloat this pull request. Would you prefer I just update the struct members with more accurate names or would it be better to mothball this pull request until I've done enough to better know what's happening with that struct?

lhearachel:
> I've followed up on the points of interest you listed and you're definitely right that A and B are x and y coordinates and so are C and D. But I'm not sure what C and D are for exactly. They're only being used in a section that has some item and (I think) bag functionality to set coordinates for a couple of sprites. I'm happy to rename the flags and give C and D something placeholder like "secondX" or "subX" and leave it things are a bit clearer.

I'm totally fine with renaming `C` and `D` to some pair of values that clearly relates them to the `x` and `y` coordinate values also in the structure. That makes perfect sense to me. :grin: 

> I've also kind of gotten side tracked trying to document sections of ov13 in order to get said clearer picture and I think that including that work would kind of bloat this pull request. Would you prefer I just update the struct members with more accurate names or would it be better to mothball this pull request until I've done enough to better know what's happening with that struct?

Let's not balloon this PR and keep it tight. I think the next place to start is less overlay 13 and more the party-list, since the latter may give us some insight into the former.

include/byte_flag_set.h
```diff
@@ -2,10 +2,10 @@
 #define POKEPLATINUM_BYTE_FLAG_SET_H
 
 typedef struct ByteFlagSet {
```
lhearachel, line 4: **suggestion**: Just in case it slipped through how prior feedback was communicated, we should also rename the `struct` to reflect that it is some sort of cursor and rename the module's header and source files to match.

Morglom: Sorry got distracted and forgot to commit the rename. Also took the opportunity to rename again to be "PartyMenuCursor" instead of "CursorInfo".

lhearachel (approved):

PR #508, opened 2025-05-12
Unpack item_icon.narc and refactor pl_item_data

memory5ty7:
New Enum to document the items icons and palettes from item_icon.narc.
This will be useful later for file naming conventions of PNG and PAL files.

How I envision the new structure:
-An icons/palette order file that specifies the order of the files in the final NARC
-This enum generated automatically from the icons/palettes order file
-Every palette or icon named according to the enum eg. "plate.png" or "flame_plate.pal"
-Every item has its own json file (properties from pl_item_data.csv AND content of sItemArchiveIDs in item.c)

Every NCGR / NCLR is then built in the correct order according to the icons/palettes order file and the item data array is built using the items.txt order (There are some unused items that appear in only one of the two arrays)


lhearachel (commented):
> How I envision the new structure:
> - An icons/palette order file that specifies the order of the files in the final NARC
> - This enum generated automatically from the icons/palettes order file
> - Every palette or icon named according to the enum eg. "plate.png" or "flame_plate.pal"
> - Every item has its own json file (properties from pl_item_data.csv AND content of sItemArchiveIDs in item.c)

Is there any reason not to go ahead and implement this structure along with decompiling the graphics files themselves? The `enum` is a bit superfluous, in that regard, as we will just replace it with the generated constants built when packing the NARC.

There is a good bit of implicit feedback here in that these generated constants are currently bound to the names of the output files emitted by `nitrogfx`. Ideally, those would be decoupled... or we could at least output an arbitrary filename and tell `nitrogfx` what conversion routine to use as an override.

memory5ty7:
Alright, I've updated the structure. We now have two json files (one for the cell, one for the anim), png files and pal files.
This generates a naix header which is used in item.c.
I read on the guidelines that enum values should be uppercased, however I couldn't find a way to make it work and I find it impractical to have filenames in lowercase and then the enum in uppercase

res/items/item_icon/icon_347.png
lhearachel, line 1: **suggestion**:

1. Can we apply palettes to these icons so that they are easily moddable in a graphics editor that supports indexed PNGs? You can apply the palette using the `-palette` option when converting `NCGR` to `png` and providing a `NCLR` file as an argument for the option. This will probably require a custom packing-script, but that's okay with me, if the scope of work doesn't feel too bad.
2. I think it would be valuable to reorganize the folders a bit... I had started working on this in the past, and my aim was to structure `items` similarly to `pokemon`, where a unique folder exists for each item and contains a (potentially empty) data file and its icon.

res/items/item_icon/icon_709.png
lhearachel, line 1: **suggestion**: `icon_bag_cancel.png` (or whatever the menu option is called in vanilla)

memory5ty7:
I've restructured the icons to be similar to pokemon : there is now a folder for every item with a data.json file obtained from pl_item_data.csv (this file is not used for the moment), and an icon.png for items that use their respective palette and icon. For the other icons, their icons and palettes stay in the root folder

memory5ty7:
Things left to do are :
-actually using every item's data.json to build the narc
-potentially adding nuts_data.csv's informations to the berries' data.json

PR #509, opened 2025-05-13
Clean up and finish documenting `unk_02096420` -> `party_use_item`

Gudf:
Use constants/enum values in place of literals, rename the public functions to be more in line with the repo's naming scheme, finish documenting variables and parameters, and finally rename the module.


src/party_use_item.c
```diff
@@ -0,0 +1,818 @@
...
+    }
+
+    vars[CHECK_EFFECTS_STATUS] = Pokemon_GetValue(mon, MON_DATA_STATUS_CONDITION, NULL);
+
+    if (Item_Get(item, ITEM_PARAM_HEAL_SLEEP) != FALSE) {
+        if ((vars[CHECK_EFFECTS_STATUS] & MON_CONDITION_SLEEP) != 0) {
+            Heap_FreeToHeap(item);
+            return TRUE;
+        }
+    }
```
lhearachel, line 97: **suggestion**: Lots of repetitive code here. Maybe worth a macro:

```c
#define CHECK_STATUS(__itemParam, __condition) do { \
    if (Item_Get(item, __itemParam) != FALSE        \
        && (vCheckStatus & (__condition)) != 0) {   \
        Heap_FreeToHeap(item);                      \
        return TRUE;                                \
    }                                               \
} while (0)

CHECK_STATUS(ITEM_PARAM_HEAL_SLEEP, MON_CONDITION_SLEEP);
// etc.
```

src/party_use_item.c
```diff
@@ -0,0 +1,818 @@
+#include "party_use_item.h"
+
+#include <nitro.h>
+#include <string.h>
+
+#include "constants/battle/condition.h"
```
lhearachel, line 6: **suggestion**: I know that this is for the status conditions, but those should probably be pulled out of the battle folder, at this stage.

src/party_use_item.c
```diff
@@ -0,0 +1,818 @@
...
+    if (Item_Get(item, ITEM_PARAM_HEAL_PARALYSIS) != FALSE) {
+        if ((vars[CHECK_EFFECTS_STATUS] & MON_CONDITION_PARALYSIS) != 0) {
+            Heap_FreeToHeap(item);
+            return TRUE;
+        }
+    }
+
+    vars[CHECK_EFFECTS_CURRENT_HP] = Pokemon_GetValue(mon, MON_DATA_CURRENT_HP, NULL);
+
+    if (((Item_Get(item, ITEM_PARAM_REVIVE) != FALSE) || (Item_Get(item, ITEM_PARAM_REVIVE_ALL) != FALSE)) && (Item_Get(item, ITEM_PARAM_LEVEL_UP) == FALSE)) {
```
lhearachel, line 129: **polish**: Extraneous parens + move the operand and right-hand-side of the `&&` clause to their own line.

src/party_use_item.c
```diff
@@ -0,0 +1,818 @@
...
+            return TRUE;
+        }
+    }
+
+    if (Item_Get(item, ITEM_PARAM_LEVEL_UP) != FALSE) {
+        if (Pokemon_GetValue(mon, MON_DATA_LEVEL, NULL) < MAX_POKEMON_LEVEL) {
+            Heap_FreeToHeap(item);
+            return TRUE;
+        }
+    }
```
lhearachel, line 146: **thought**: Maybe could generalize the macro that I suggested earlier to work with any condition, but that's probably not worth the maintenance mess.

src/party_use_item.c
```diff
@@ -0,0 +1,818 @@
...
+                    Heap_FreeToHeap(item);
+                    return TRUE;
+                }
+
+                if (CheckFriendshipItemEffect(mon, item) == TRUE) {
+                    Heap_FreeToHeap(item);
+                    return TRUE;
+                }
+            }
+        }
```
lhearachel, line 205: **suggestion**: Definitely macro-ify these.

src/party_use_item.c
```diff
@@ -0,0 +1,818 @@
...
+        Pokemon_SetValue(mon, MON_DATA_MOVE1_CUR_PP + moveSlot, &ppCurr);
+        return TRUE;
+    }
+
+    return FALSE;
+}
+
+static u8 IncreaseMovePPUps(Pokemon *mon, u32 moveSlot, u32 amount)
+{
+    u8 currPp; // forward declaration required to match
```
lhearachel, line 637: **suggestion**: `currPP` (since it is an acronym)

lhearachel (commented):
Looks great overall. I think that we should macro-ify the repetitive sections of code so that they are easier to modify in unison and generally make the semantics of the code easier to read.

src/party_use_item.c
```diff
@@ -0,0 +1,818 @@
...
+#define HEAL_FULL_HP    255
+#define HEAL_HALF_HP    254
+#define HEAL_QUARTER_HP 253
+
+#define EV_UNCHANGED -1
+
+#define HEAL_FULL_PP 127
+
+enum CheckEffectsVarsIndex {
+    CHECK_EFFECTS_STATUS = 0,
```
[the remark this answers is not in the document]
Gudf: Hmm, that would probably make sense yeah. While obviously not required, would it make sense to have these defines in the function body, or would you rather keep them at the top of the file, with a comment in the function body pointing to them?

src/party_use_item.c
Gudf: Not worth the extra complexity considering realistically it could only be reused for two, maybe three of these checks, imo.

src/party_use_item.c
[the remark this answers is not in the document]
lhearachel: If they are unique to the function, then you could define them at the start of this scope and `undef` them at the end. That aside, I think just defining them at the top of the module and explaining what they are used for is also fine.

src/party_use_item.c
lhearachel: I agree. 💯

Gudf:
Discussed on Discord: rename to `item_use_party`/`item_use_pokemon` so the file sorts near `item.c`; replace the module prefixes with `Pokemon`/`Party` to make them easier to find for auto-complete suggestions and clarify the leading operand.

src/party_use_item.c
Gudf: While I agree, I'll consider this out of scope for this PR. I'll be happy to open another one to move it out of this folder though! Do we simply want to move it to `include/constants/condition.h`?

src/party_use_item.c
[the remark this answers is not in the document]
Gudf: Ended up deciding to have them at the top, I didn't like how it looked with the defines inside the function body.

lhearachel (approved):

PR #510, opened 2025-05-14
refactor: Cleanup scrcmd.c

tillgeissler:
This PR includes some refactorings for the `scrcmd.c` file. Those include:

- Combining variable declaration and definition, in case it matches
- Return `TRUE`/`FALSE` for BOOL functions
- Usage of constants, where it is obvious
- Remove unused variable, when possible

### DIFF
--- a/src/scrcmd.c
+++ b/src/scrcmd.c
@@ -7,6 +7,7 @@
 #include "constants/battle.h"
 #include "constants/daycare.h"
 #include "constants/field/dynamic_map_features.h"
+#include "constants/field/window.h"
 #include "constants/heap.h"
 #include "constants/items.h"
 #include "constants/overworld_weather.h"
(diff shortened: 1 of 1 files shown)


src/scrcmd.c
```diff
@@ -756,7 +756,7 @@ static const u8 sConditionTable[6][3] = {
     { TRUE, FALSE, TRUE }, //  !=
 };
 
-static const WindowTemplate Unk_020EAB84 = {
+static const WindowTemplate YesNoWindowTemplate = {
```
lhearachel, line 759: **suggestion**:

1. Rename this to `sYesNoWindowTemplate`, per our conventions for naming `static` variables.
2. Use designated initializers for all fields of this definition (prefix each value with `.<fieldname> = `). This makes the definition easier to grok at a glance and makes it harder to mess up when modifying.
3. The first member (`bgLayer`) should use the constant `BG_LAYER_MAIN_3` from `bg_window.h`.
4. Convert the fields `tilemapLeft`, `tilemapTop`, `width`, `height`, and `palette` to decimal values.

lhearachel (changes requested):
Thanks for this! I'm not being too much of a stickler with this PR, save for one request on the `WindowTemplate` that is documented here, since much of the other code will be cleaned up by us in the future for documentation. :slightly_smiling_face:

lhearachel (approved):

PR #511, opened 2025-05-14
Document sub menu cursor and rename position struct.

Morglom:
Been working through ov_13 and was able to fully figure out the cursor and as a pleasant bonus also what was going on with that cursor struct in the previous pull request.

Some sections feel like they could do with comments but I couldn't come up with anything that didn't seem too long.


src/overlay013/battle_sub_menu_cursor.c
```diff
@@ -0,0 +1,207 @@
...
+    cursor->enabledPositionsMask = mask;
+}
+
+static u8 IsCursorVisible(BattleSubMenuCursor *cursor)
+{
+    if (cursor->isVisible == TRUE) {
+        return TRUE;
+    }
+
+    if (gSystem.pressedKeys & (PAD_KEY | PAD_BUTTON_B | PAD_BUTTON_A)) {
```
lhearachel, line 84: **question**: Would this match with `JOY_NEW` from `system.h`?

src/meson.build
```diff
@@ -73,7 +73,7 @@ pokeplatinum_c = files(
     'vram_transfer.c',
     'cell_transfer.c',
     'unk_0201DD00.c',
-    'party_menu_cursor.c',
+    'battle_sub_menu_cursor_position.c',
```
lhearachel, line 76: **suggestion**: I don't really like this rename, because we know that this cursor is also used by the field party menu, which is rendered on the top screen and also not battle-specific.

lhearachel (changes requested):

src/meson.build
Morglom: Would "sub_menu_cursor_position" work? If not did you have any preferences for potential names?

lhearachel: I'm just not that opposed to keeping `party_menu_cursor`, or maybe going with something like `grid_menu_cursor` if we're concerned about staying generic to use it in the battle bag.

lhearachel (commented):
Looks good; just a couple more suggested changes.

lhearachel (approved):

PR #512, opened 2025-05-16
Document battle bag utility

Morglom:
Documenting some battle bag functionality and renaming some related struct members.


src/overlay013/battle_bag_utils.c
```diff
@@ -0,0 +1,108 @@
...
+#define BATTLE_BAG_POCKET_SIZE 36
+#define BATTLE_BAG_POCKET_NUM  5
+
+enum BattleBagPocketMask {
+    BATTLE_BAG_POCKET_MASK_POKE_BALLS = 0x0,
+    BATTLE_BAG_POCKET_MASK_BATTLE_ITEMS,
+    BATTLE_BAG_POCKET_MASK_RECOVER_HP,
+    BATTLE_BAG_POCKET_MASK_RECOVER_STATUS,
+    BATTLE_BAG_POCKET_MASK_UNUSUED = 0x0,
+};
```
lhearachel, line 17: **suggestion**: Move these constants to `include/constants/items.h` and make use of them where possible in the structs.

I would also suggest:

```c
#define BATTLE_POCKET_SIZE 36
#define BATTLE_POCKET_MAX  5
```

etc., which aligns with the existing `POCKET_...` for the field bag.

lhearachel (changes requested):
Looks good to me, just one request to move the constants to `include/constants/items.h`. :grin:

src/overlay013/battle_bag_utils.c
```diff
@@ -5,23 +5,20 @@
...
-    BATTLE_BAG_POCKET_MASK_POKE_BALLS,
-    BATTLE_BAG_POCKET_MASK_BATTLE_ITEMS,
-    BATTLE_BAG_POCKET_MASK_UNUSUED
+static const u8 battlePocketMasks[] = {
+    BATTLE_POCKET_MASK_RECOVER_HP,
+    BATTLE_POCKET_MASK_RECOVER_STATUS,
+    BATTLE_POCKET_MASK_POKE_BALLS,
+    BATTLE_POCKET_MASK_BATTLE_ITEMS,
+    BATTLE_POCKET_MASK_UNUSUED
 };
```
lhearachel, line 22: **thought**: I recall that in the generation for `pl_item_data`, there are currently raw integers for the battle pocket for each item. We will probably want to pull these constants upwards into `include/constants/items.h` as `#define`s so that the data generation can make use of them. It's up to you if you want to do that yourself, or I can take care of it as part of moving the current CSV file to a set of JSON files.

lhearachel (approved):
This looks mergeable to me. I have one thought below that I want to get your input on before we merge.

src/overlay013/battle_bag_utils.c
Morglom: I'll look into it.

Morglom: This enum is pretty badly named now that I think about it and should be called something else like "BattlePocketIndex" and it's use in items.h would probably be more indirect. They're not really a bit mask and more represent how many bits to shift in order to get the actual bit mask.

We could still use them in items.h but it would probably be something like.
`#define BATTLE_POCKET_RECOVER_HP (1 << BATTLE_POCKET_INDEX_RECOVER_HP)`

But I wonder if it would be better to do
`#define BATTLE_POCKET_RECOVER_HP (1 << 2)`
or
`#define BATTLE_POCKET_RECOVER_HP 0x4`

I still the suggested change is good and I'm  happy to setup the defines and make the appropriate changes in the CSV. Just wanted to clear up my naming mistake and get input on what the best approach should be.

lhearachel: I would document them like so:

```c
enum BattlePocketShift {
    BATTLE_POCKET_SHIFT_POKE_BALLS = 0x0,
    BATTLE_POCKET_SHIFT_BATTLE_ITEMS,
    BATTLE_POCKET_SHIFT_RECOVER_HP,
    BATTLE_POCKET_SHIFT_RECOVER_STATUS,
    BATTLE_POCKET_SHIFT_UNUSUED = 0x0, 
};

#define BATTLE_POCKET_POKE_BALLS     (1 << BATTLE_POCKET_SHIFT_POKE_BALLS)
#define BATTLE_POCKET_BATTLE_ITEMS   (1 << BATTLE_POCKET_SHIFT_BATTLE_ITEMS)
#define BATTLE_POCKET_RECOVER_HP     (1 << BATTLE_POCKET_SHIFT_RECOVER_HP)
#define BATTLE_POCKET_RECOVER_STATUS (1 << BATTLE_POCKET_SHIFT_RECOVER_STATUS)
```

Note that the definitions themselves must be `#define`s to function with the CSV processor.

src/overlay013/battle_bag_utils.c
```diff
@@ -0,0 +1,105 @@
...
+
+enum BattlePocketMask {
+    BATTLE_POCKET_MASK_POKE_BALLS = 0x0,
+    BATTLE_POCKET_MASK_BATTLE_ITEMS,
+    BATTLE_POCKET_MASK_RECOVER_HP,
+    BATTLE_POCKET_MASK_RECOVER_STATUS,
+    BATTLE_POCKET_MASK_UNUSUED = 0x0,
+};
+
+static const u8 battlePocketMasks[] = {
```
lhearachel, line 16: **polish**:

```suggestion
static const u8 sBattlePocketMasks[] = {
```

src/overlay013/battle_bag_utils.c
```diff
@@ -0,0 +1,105 @@
...
+
+#include "bag.h"
+#include "item.h"
+
+enum BattlePocketMask {
+    BATTLE_POCKET_MASK_POKE_BALLS = 0x0,
+    BATTLE_POCKET_MASK_BATTLE_ITEMS,
+    BATTLE_POCKET_MASK_RECOVER_HP,
+    BATTLE_POCKET_MASK_RECOVER_STATUS,
+    BATTLE_POCKET_MASK_UNUSUED = 0x0,
```
lhearachel, line 13: **typo**:

```suggestion
    BATTLE_POCKET_MASK_UNUSED = 0x0,
```

lhearachel (commented):
Two more small things that I noticed.

src/overlay013/battle_bag_utils.c
Morglom: In working on the csv I discovered what the fifth bit shift was for and why points to the same index as the HP restore. It's not unused it's for PP restoring items. I decided trying to get the bit shifts to correspond in some clean way with the battle pocket indexes wasn't working so I just made them separate and had the bit shifts and masks be their own thing. Not sure if we should keep the shift defines though as they aren't being used for the masks due to the csv issue.

PR #514, opened 2025-05-23
Unpack b_pl_btdpm.narc

tillgeissler:
This PR unpacks the `res/prebuilt/battle/b_pl_tower/pl_btdpm.narc` which contains the Pokémon data for the Battle Tower opponents to `res/battle/battle_tower/pokemon` as individual JSON files.


tillgeissler:
Since this is my first attempt at unpacking a .narc, I am happy about any feedback about file names, directory structures, JSON keys, the build script etc. 😄 
Are there any recommendations how to handle bit flags as used for the EVs of the Pokémon data? I am not too keen about representing those as int values. 
I am also considering to documenting some of the structures used to read the data from the .narc, if this doesn't balloon the PR.

generated/natures.txt
```diff
@@ -0,0 +1,25 @@
...
+NATURE_MODEST
+NATURE_MILD
+NATURE_QUIET
+NATURE_BASHFUL
+NATURE_RASH
+NATURE_CALM
+NATURE_GENTLY
+NATURE_SASSY
+NATURE_CAREFUL
+NATURE_QUIRCKY
```
lhearachel, line 25: **typo**:

```suggestion
NATURE_QUIRKY
```

res/battle/battle_tower/pokemon/meson.build
```diff
@@ -0,0 +1,26 @@
+pl_btdpm_bin_gen = generator(
+    battletower_data_py,
+    arguments: [ '@INPUT@', '@OUTPUT@', ],
+    output: '@BASENAME@.bin'
+)
+
+bt_pokemon_data_files = []
+
+foreach i : range(951)
```
lhearachel, line 9: **suggestion**: We definitely want this to be less static and more dynamic. How are individual Pokemon picked from the data files for use in the tower? Is there a constant value used somewhere?

tools/json2bin/battletower_data.py
```diff
@@ -0,0 +1,33 @@
...
+    with open(input_path, 'r', encoding='utf-8') as input_file:
+        data = json.load(input_file)
+except json.decoder.JSONDecodeError as e:
+    print(e)
+    sys.exit(1)
+
+packables = bytearray([])
+packables.extend(u16(from_species(data['species'])))
+packables.extend(parse_moves(data['moves']))
+packables.extend(u8(data['evFlag']))  # evFlag, it it enough to convert the int to a u8 byte?
```
lhearachel, line 26: **question**: Do we have a sense of what `evFlag` entails? That might help us better-decompose the structure.

lhearachel (commented):
I think that this looks fine without context, but it would be good to get a sense of how the Tower Pokemon are actually chosen and how some of the fields like `evFlag` are used. If you want to do that as a precursor PR and leave this one dangling for now, I'd be fine with that.

tools/json2bin/battletower_data.py
tillgeissler: It is a 6 bit flag (+ 2 bits padding) that represents which stats should be invested. Each of them will be set to 510 divided by the number of stats that should be set. The order seems to be Sp.Def., Sp.Atk., Speed, Defense, Attack, HP. So `010010` would represent investments of 255 into Sp.Atk. and Attack. 
I was thinking about storing a list of stats that should be invested to make it more user friendly.

tillgeissler: No, the 510 total is still adhered to. E.g. if three flags are set, each of those stats will obtain 170 EVs. (`100101` results in 170 in Sp.Def., Defense and HP). There are no sets with more than three active flags, but the EVs would be divided equally among the stats as well.

res/battle/battle_tower/pokemon/meson.build
tillgeissler: I took a deeper look into how they are loaded now. At first, a random trainer ID is chosen based on the current progress in the battle facility. (I am not sure about the differences in the different facilities yet, though.) Each trainer has an associated list of potential Pokémon sets. (This is part of the `pl_btdtr.narc`.) Based on this list, the IDs of the data files are determined. As far as I could see, there is no constant value for this in the code.

tools/json2bin/battletower_data.py
lhearachel: Cool, let's just have this be an array of booleans for stats. I would mimic the names used by the `data.json` files for each species. :slightly_smiling_face:

res/battle/battle_tower/pokemon/meson.build
lhearachel: Interesting. With that in mind, would these files better-belong to the `trainers` subfolder, since there is an inherent relation? Maybe `res/trainers/battle_tower/...`?

tillgeissler:
> I think that this looks fine without context, but it would be good to get a sense of how the Tower Pokemon are actually chosen and how some of the fields like `evFlag` are used. If you want to do that as a precursor PR and leave this one dangling for now, I'd be fine with that.

I have created #531 to document the structures used for loading frontier trainer and pokemon data to give a bit more context how they are used. Will come back to this PR when the other is finalized.

res/trainers/frontier/data/ace_trainer_charlie.json
```diff
@@ -0,0 +1,73 @@
+{
```
lhearachel, line 1: **thought**: I'm a little concerned about the efficacy of how a user might modify this. A few questions spring to mind:

1. Are the `setIds` here referring to individual Pokemon from `b_pl_btdpm.narc`?
2. If so, is it possible to list them by an identifier and do a lookup at build-time? Should these NARCs be packed in tandem, perhaps? If so, can we think of a way to do that?

res/trainers/frontier/pokemon/zapdos_4.json
```diff
@@ -0,0 +1,20 @@
+{
```
lhearachel, line 1: **praise**: I like this structure much better, and I also like the identifiers to separate them from one another. Kudos. :pray:

res/trainers/frontier/data/ace_trainer_charlie.json
tillgeissler: Finally got around to this PR again. With 1cdecc74af581285313979d060fbb048d6c6446c I move the packing script to a c++ implementation (`datagen_frontier.cpp`). Now the available sets are identified by their names that are the same for the respective file.

lhearachel (approved):
Fantastic work, thank you! :pray:

PR #526, opened 2025-05-24
Remove some redundant struct declaration headers

AsparagusEduardo:


lhearachel:
What's the purpose of this PR? It seems to be moving a lot of struct definitions to headers that they shouldn't or don't need to be in...?

PR #527, opened 2025-05-25
Unpack Trainer Graphics

h2o-DS:
This PR unpacks the following NARCs:
`/res/prebuilt/poketool/trgra/trfgra.narc`
`/res/prebuilt/poketool/trgra/trbgra.narc`
`/res/prebuilt/graphics/field_encounteffect.narc`

They are now created in the directory `/res/trainers/classes`
Each trainer class has a folder within this directory.

This PR cannot be merged until nitrogfx implements TACU chunks.


src/overlay005/encounter_effect_core.c
```diff
@@ -2726,28 +2728,28 @@ typedef struct EliterFourChampionEncounterParam {
 } EliterFourChampionEncounterParam;
 
 static const EliterFourChampionEncounterParam sEliteFourChampionEncounterParams[5] = {
-    { .mugshotPlttIdx = 87, // Narc indices of the Elite Four and Champion
-        .bannerPlttIdx = 39,
+    { .mugshotPlttIdx = mugshot_elite_four_aaron_NCLR, // Narc indices of the Elite Four and Champion
```
lhearachel, line 2731: **suggestion**: We can remove this comment now.

tools/scripts/make_field_encounteffect.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python3
```
lhearachel, line 1: **thought**: We have a lot of specialized scripts that don't really do different things. Maybe we can generalize them into a program that takes a configuration file or somesuch...?

lhearachel (commented):
Some preliminary comments.

src/overlay005/encounter_effect_core.c
```diff
@@ -680,7 +682,7 @@ void EncounterEffect_Trainer_Grass_LowerLevel(SysTask *task, void *param)
         ov5_021DE47C(&trainerEffect->unk_48, 2, 1);
 
         ov5_021DE4CC(
-            encEffect->narc, &trainerEffect->unk_48, &trainerEffect->unk_1E8, 0, 1, 5, 7, 6, 600000);
+            encEffect->narc, &trainerEffect->unk_48, &trainerEffect->unk_1E8, enc_trainer_NCLR, 1, enc_trainer_high_NCGR, enc_trainer_high_NCER, enc_trainer_high_NANR, 600000);
```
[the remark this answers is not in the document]
h2o-DS: The `1` appears to be `paletteIdx` for `PaletteResourceData`

tools/scripts/make_field_encounteffect.py
h2o-DS: Front and Back could certainly use a shared script. encounter effects might take a little work; would a series of generators be preferable to a script?

lhearachel (commented):
Mmh, the `cell` change is interesting, and I do like how it makes things look. How does it work?

h2o-DS:
I use the NCER to place the tiles instead of going sequentially. The downside to this is that you then need to pass the NCER when going back to NCGR to undo the shuffle.

h2o-DS:
Will close this PR until `TACU` chunks are supported by `nitrogfx`.

PR #529, opened 2025-05-26
Document screen fades and hardware windows

lhearachel:
This is a small wrapper for controlling screen-fades, typically to and from white or black. Both brightness-backed fades and window-backed fades are supported.

I'll leave the `screen_fade_funcs` module for another time, as it is quite large and requires a bit of elbow grease to iron out. At the very least, the external interface surfaced by `screen_fade.h` should help document other sections of the repository more easily.

Some relevant reading for the associated hardware components:

- https://problemkaputt.de/gbatek.htm#lcdiowindowfeature
- https://www.coranac.com/tonc/text/gfx.htm#sec-win

These articles are centered on the GBA, but the architecture here is fundamentally identical.


ravepossum:
Adding all comments here because GitHub sucks and isn't letting me add comments to lines for some reason. Not too much for this one fortunately. I like the inline documentation in `hardware_window.h`, I think it's a helpful touch.

`screen_fade.h Fade enums`
It seems like you've made a relatively consistent decision to prepend `FADE_` but not necessarily specify the whole enum name in the constants here, which is concise but might impair clarity on ones like `FADE_WINDOW` and `FADE_BRIGHTNESS` that are a bit more ambiguous. I also wonder if `FADE_` would make more sense as the prefix for `FadeMode` than `MODE_` (`FADE_BOTH_SCREENS` is more clear to me than `MODE_BOTH_SCREENS)`.

`screen_fade.h Color constants`
Would it make more sense as something like `FADE_SAVED_COLOR` to make it more clear that it signals to use the saved color specifically? Since e.g. `WHITE` is implicitly a color, but `SAVED` is not.

`screen_fade.c GetFadeColor()`
Missing a `FADE_SAVED` constant. (exact change might depend on above though)

PR #534, opened 2025-06-01
Update metroskrew to 0.1.3

lhearachel:
Small PR to:

1. Add some light infrastructure to update `metroskrew` alongside Meson subprojects
2. Swallow errors when Meson throws an error during subproject update (which is often just a warning)
3. Do some light cleanup of the Makefile
4. Bump the `metroskrew` version

Changelog for `metroskrew-0.1.3`:

- [`9a25f02`](https://github.com/mid-kid/metroskrew/commit/9a25f02877d31418b8447c6201884f6f0ffa6efd) - Fix relative path generation on Linux
- [`6693130`](https://github.com/mid-kid/metroskrew/commit/6693130c796ba2b6056e6d5b8c179ce7e6548171) - Fix depfiles only outputting half of the dependencies

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,24 @@
-.PHONY: all release debug check rom data target format clean distclean purge update setup_release setup_debug configure skrew meson
+# Keep this list in alphabetical order for ease of reference.
+.PHONY:           \
+	all           \
+	check         \
+	clean         \
+	configure     \
+	data          \
+	distclean     \
+	debug         \
+	format        \
+	meson         \
+	purge         \
+	release       \
+	rom           \
+	setup_debug   \
+	setup_release \
+	skrew         \
+	skrewrm       \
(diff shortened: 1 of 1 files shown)


mid-kid:
Suggestion: If you add the version number to SKREW_DIR, just like we do with the meson version, it will be downloaded automatically whenever the newer version is missing. That way you don't need to redownload it for every `make update`, and it aligns the behavior with the meson download.

lhearachel:
> Suggestion: If you add the version number to SKREW_DIR, just like we do with the meson version, it will be downloaded automatically whenever the newer version is missing. That way you don't need to redownload it for every `make update`, and it aligns the behavior with the meson download.

I considered as much, but opted not to rock the boat on this one too much since it was a crucial depfile fix.

I'll incorporate this into the next update.

PR #537, opened 2025-06-03
Document some things related to type rendering

YakosWG:


src/applications/pokedex/ov21_021DE668.c
```diff
@@ -650,66 +652,67 @@ static void ov21_021DF098(UnkStruct_ov21_021DF374 *param0, PokedexGraphicData **
     }
 }
 
-int ov21_021DF180(int param0)
+// This maps the type to an animation ID used in the Pokedex graphics. See zukan.narc -> files 88 (NCER), 89 (NANR), 90 (NCGR)
+int PokedexGraphics_GetAnimIDfromType(int monType)
```
lhearachel, line 656: @h2o-DS This one is probably of interest to you whenever unpacking Pokedex graphics.

src/battle/ov16_0223DF00.c
```diff
@@ -297,12 +297,12 @@ UnkStruct_ov12_0221FCDC *ov16_0223E008(BattleSystem *battleSystem)
     return battleSystem->unk_8C;
 }
 
-SpriteSystem *ov16_0223E010(BattleSystem *battleSystem)
+SpriteSystem *BattleSystem_GetSpriteSystem(BattleSystem *battleSystem)
 {
     return battleSystem->unk_90;
```
lhearachel: **polish**: If it's easy to do so, go ahead and mass-rename these struct-fields to match the fronting accessor-function.

src/battle/ov16_0226871C.c
```diff
@@ -2812,48 +2812,48 @@ static void ov16_0226AEA0(UnkStruct_ov16_02268A14 *param0, const Strbuf *param1,
     }
 
     Window_Init(&param3->unk_00);
-    Window_AddToTopLeftCorner(BattleSystem_BGL(param0->unk_00), &param3->unk_00, v1, 16 / 8, 0, 0);
+    Window_AddToTopLeftCorner(BattleSystem_BGL(param0->battleSys), &param3->unk_00, v1, 16 / 8, 0, 0);
     Text_AddPrinterWithParamsColorAndSpacing(&param3->unk_00, param2, param1, 0, 0, TEXT_SPEED_NO_TRANSFER, param4, 0, 0, NULL);
 }
 
-static void ov16_0226AF0C(UnkStruct_ov16_02268A14 *param0)
+static void DisplayMoveTypeIcons(UnkStruct_ov16_02268A14 *param0)
```
lhearachel: **suggestion**: `DrawMoveTypeIcons`

src/battle/ov16_0226871C.c
```diff
@@ -2894,21 +2894,21 @@ static void ov16_0226B028(UnkStruct_ov16_02268A14 *param0)
     }
 }
 
-static void ov16_0226B06C(UnkStruct_ov16_02268A14 *param0, int param1, int param2)
+static void LoadPaletteForMoveInSlot(UnkStruct_ov16_02268A14 *param0, int moveType, int moveSlot)
```
lhearachel: **suggestion**: `LoadMoveTypePalette`

src/overlay011/ov11_0221F800.c
```diff
@@ -415,21 +415,21 @@ void include_unk_ov11_0221FBAC()
...
-        v0 = Unk_ov11_0221F96C;
+        paletteSourceAdr = Unk_ov11_0221F96C;
     }
 
-    return v0;
+    return paletteSourceAdr;
 }
 
-void ov11_0221F81C(PaletteData *param0, int param1, int param2, int param3, int param4)
+void LoadMoveSelectionPalette(PaletteData *paletteSys, int moveType, int unused, int bufferID, int paletteDest)
```
lhearachel: **suggestion**: `LoadMoveSelectPalette`.

**polish**:

```suggestion
void LoadMoveSelectionPalette(PaletteData *paletteSys, int moveType, int unused, enum PaletteBufferID bufferID, enum Palette paletteDest)
```

The former is from `palette.h`, the latter from `constants/palette.h`. I would argue that the latter header should be merged with the former. Up to you if you want to do that; I won't block the PR on that change.

src/overlay011/ov11_0221F800.c
```diff
@@ -415,21 +415,21 @@ void include_unk_ov11_0221FBAC()
...
 {
-    const u16 *v0 = ov11_0221F800(param1);
-    PaletteData_LoadBuffer(param0, v0, param3, param4 * 16, 0x20);
+    const u16 *paletteSource = GetPaletteSourceFromMoveType(moveType);
+    PaletteData_LoadBuffer(paletteSys, paletteSource, bufferID, paletteDest * 16, 0x20);
```
lhearachel: **polish**:

```suggestion
    PaletteData_LoadBuffer(paletteSys, paletteSource, bufferID, PLTT_DEST(paletteDest), PALETTE_SIZE_BYTES);
```

The former is a new addition; you could/should add it to `palette.h`, if you're so-inclined:

```c
#define PLTT_DEST(i) ((i) * PALETTE_SIZE)
```

src/overlay011/ov11_0221F800.c
```diff
@@ -415,21 +415,21 @@ void include_unk_ov11_0221FBAC()
     Unk_ov11_0221FBAC[0];
 }
 
-const u16 *ov11_0221F800(int param0)
+const u16 *GetPaletteSourceFromMoveType(int moveType)
```
lhearachel: **suggestion**: `GetMoveTypePalette`. You're returning an array containing the raw palette data, which is really what any client of the function cares about.

src/overlay011/ov11_0221F800.c
```diff
@@ -5,7 +5,7 @@
 
 #include "palette.h"
 
-__attribute__((aligned(4))) const u16 Unk_ov11_0221F94C[] = {
+__attribute__((aligned(4))) const u16 MovePaletteNormal[] = {
```
lhearachel: **polish**: For each of these arrays:

1. Use `static const` and prefix names with `s` (project convention)
2. Use `ALIGN_4` instead of `__attribute__((aligned(4)))` (equivalent, but the former is more flexible and concise)
3. Ideally, surface the `RGB` macro from `palette.c` into `palette.h` and replace the raw hexadecimal values in each array with that macro, separating out the individual color components (I would just write a script to do this translation)

src/battle/ov16_0226871C.c
```diff
@@ -2894,21 +2894,21 @@ static void ov16_0226B028(UnkStruct_ov16_02268A14 *param0)
...
 {
-    ov11_0221F81C(BattleSystem_PaletteSys(param0->unk_00), param1, 5, 1, 8 + param2);
+    LoadMoveSelectionPalette(BattleSystem_PaletteSys(param0->battleSys), moveType, 5, 1, 8 + moveSlot);
```
lhearachel: **polish**:

```suggestion
    LoadMoveSelectionPalette(BattleSystem_PaletteSys(param0->battleSys), moveType, 5, PLTTBUF_SUB_BG, PLTT_8 + moveSlot);
```

src/overlay019/ov19_021DA92C.c
```diff
@@ -33,12 +33,12 @@ typedef struct {
...
 u32 sub_0207C920(void);
-u32 sub_0207C924(void);
-u32 sub_0207C928(void);
-u8 sub_0207C92C(int param0);
-enum NarcID sub_0207C944(void);
+u32 GetMoveTypeIconCellBank(void);
+u32 GetMoveTypeIconAnimBank(void);
+u8 GetPaletteFromMoveType(int moveType);
+enum NarcID GetBattleGraphicNarcIndex(void);
 void Window_PutRectToTilemap(Window *param0, u32 param1, u32 param2);
```
lhearachel: **polish**: These declarations should be removed and instead retrieved from their respective `.h` files.

src/unk_0207C908.c
lhearachel, line 1: **thought**: I personally feel like we have enough information to rename this module to `type_icon` and decorate the exposed functions appropriately according to project standards. What do you think?

Assuming that we do:

- `GetIconFromMoveType` -> `TypeIcon_GetChar`
- `GetMoveTypeIconCellBank` -> `TypeIcon_GetCell`
- `GetMoveTypeIconAnimBank` -> `TypeIcon_GetAnim`
- `GetPaletteFromMoveType` -> `TypeIcon_GetPltt`
- `GetBattleGraphicNarcIndex` -> `TypeIcon_GetNARC`
- `GetIconFromMoveCategory` -> `CategoryIcon_GetChar`
- `GetPaletteFromMoveCategory` -> `CategoryIcon_GetPltt`

etc.

I think we could also just proceed to document the rest of the file, and maybe even unpack the graphics. It doesn't look too bad. If you'd rather not take care of that now, then we can add it to #325 as a bite-sized introduction for new contributors.

src/unk_0207C908.c
```diff
@@ -8,8 +8,8 @@
 #include "palette.h"
 #include "sprite_system.h"
 
-__attribute__((aligned(4))) static const u32 Unk_020F0AF0[] = {
-    0xEA,
+__attribute__((aligned(4))) static const u32 sMoveTypeIconIndex[] = {
```
lhearachel: **polish**: The same `ALIGN_4` comment applies to these arrays, as well.

YakoSWG:
This isn't all of it but it's as far as I got today. I feel like both ov11_0221F800.c and unk_0207C908.c (as well as their respective headers) could be renamed. I am not too familiar with the build system so I might need a bit of a hand with that.

YakoSWG:
I would like to discuss the last commit as well. Do you agree with the name and the naming itself?

include/battle/ov16_0223DF00.h
```diff
@@ -227,8 +228,8 @@ void ov16_0223F24C(BattleSystem *battleSys, int param1);
 void ov16_0223F268(BattleSystem *battleSys);
 void BattleSystem_SetCommandSelectionFlags(BattleSystem *battleSys, int flags);
 void ov16_0223F290(BattleSystem *battleSys, int param1);
-void *Battle_WaitDial(BattleSystem *battleSys);
-void Battle_SetWaitDial(BattleSystem *battleSys, void *param1);
+WaitDial *Battle_WaitDial(BattleSystem *battleSys);
```
lhearachel: **polish**: `Battle_GetWaitDial`.

include/battle/struct_ov16_02268A14_decl.h
```diff
@@ -1,6 +1,6 @@
 #ifndef POKEPLATINUM_STRUCT_OV16_02268A14_DECL_H
 #define POKEPLATINUM_STRUCT_OV16_02268A14_DECL_H
 
-typedef struct UnkStruct_ov16_02268A14_t UnkStruct_ov16_02268A14;
+typedef struct BattleDisplayData UnkStruct_ov16_02268A14;
```
lhearachel: **polish**: Rename the `typedef` as well to match the `struct` tag.

```suggestion
typedef struct BattleDisplayData BattleDisplayData;
```

src/overlay019/ov19_021DA92C.c
```diff
@@ -33,12 +33,6 @@ typedef struct {
     UnkStruct_ov19_021DA9E0 *unk_08;
 } UnkStruct_ov19_021DAE2C;
 
-u32 GetIconFromMoveType(int moveType);
-u32 sub_0207C920(void);
-u32 GetMoveTypeIconCellBank(void);
-u32 GetMoveTypeIconAnimBank(void);
-u8 GetPaletteFromMoveType(int moveType);
-enum NarcID GetBattleGraphicNarcIndex(void);
 void Window_PutRectToTilemap(Window *param0, u32 param1, u32 param2);
```
lhearachel, line 36: **polish**: Delete this declaration.

src/battle/ov16_0226871C.c
```diff
@@ -192,7 +192,7 @@ typedef struct {
     UnkStruct_ov16_0226AEA0 unk_C8[4];
 } UnkStruct_ov16_0226ABD4;
 
-typedef struct UnkStruct_ov16_02268A14_t {
+typedef struct BattleDisplayData {
```
lhearachel: **polish**: Remove this `typedef`, which is non-standard (as it is a duplicate of a previous `typedef`).

lhearachel:
> This isn't all of it but it's as far as I got today. I feel like both ov11_0221F800.c and unk_0207C908.c (as well as their respective headers) could be renamed. I am not too familiar with the build system so I might need a bit of a hand with that.

For this bit, I recommend using a global find+replace utility (like what's built into most modern editors / IDEs) to just search for the module name minus its extension, then replace it with the target name. Then, you can just manually rename the files, and it should Just Work.

If you have any issues with getting that to work, feel free to nudge me on Discord. 😁

YakoSWG:
I've thought about it a bit and maybe it's a bit early to name the stuct after all. There are some parts I don't think I fully understand so it's tough to say if the name would be entirely correct. I think it would be better to leave it at that point.

lhearachel (approved):

PR #543, opened 2025-06-08
Document battle bag buttons

Morglom:
Documented "battle bag buttons". Also made some minor changes to the other "battle bag" files.


include/overlay013/battle_bag_text.h
```diff
@@ -3,6 +3,50 @@
 
 #include "overlay013/battle_bag.h"
 
+enum BattleBagMenuScreenWindow {
```
lhearachel: **suggestion**: We recently had a short discussion about `Menu` vs `Screen` as terminology, and it was judged that `Menu` is a preferable term due to the implication of interactivity. I would say we can thus name these as just `BattleBagMenuWindow`, etc.

src/overlay013/battle_bag.c
```diff
@@ -79,16 +80,16 @@ static u8 BattleBagTask_Exit(BattleBag *battleBag);
...
 static void CleanupMessageLoader(BattleBag *battleBag);
 static u8 TryUseItem(BattleBag *battleBag);
 static void SetupBackgroundScroll(BattleBag *battleBag, enum BattleBagScreen screen);
 static void ChangeBattleBagScreen(BattleBag *battleBag, u8 screen);
-static BOOL CheckTouchRectIsPressed(BattleBag *battleBag, const TouchScreenRect *rect);
+static int CheckTouchRectIsPressed(BattleBag *battleBag, const TouchScreenRect *rect);
 static void UseBagItem(BattleSystem *battleSys, u16 item, u16 category, u32 heapID);
 
-static const TouchScreenRect menuScreenTouchRects[] = {
+static const TouchScreenRect sMenuScreenTouchRects[] = {
```
lhearachel: **suggestion**: In line with the above, `sMenuTouchRects`, etc.

src/overlay013/battle_bag_buttons.c
```diff
@@ -0,0 +1,559 @@
...
+typedef struct ButtonDimensions {
+    u8 xCoord;
+    u8 yCoord;
+    u8 width;
+    u8 height;
+} ButtonDimensions;
+
+static void LoadButtonData(u16 *buttonData, u16 *screenData, u8 xOffset, u8 yOffset, u8 width, u8 height);
+
+static const ButtonDimensions buttonDimensions[] = {
```
lhearachel: **polish**: `sButtonDimensions` (and applies to other similar variable names)

include/overlay013/battle_bag.h
```diff
@@ -65,22 +72,22 @@ typedef struct BattleBag {
...
+    u16 menuPocketButtonData[3][144];
+    u16 useItemButtonData[4][130];
+    u16 cancelButtonData[3][25];
+    u16 pocketItemButtonData[4][96];
+    u16 pocketPrevPageButtonData[4][25];
+    u16 pocketNextPageButtonData[4][25];
+    u16 menuScreenRecoverHPPocketIconData[3][16];
+    u16 menuScreenRecoverStatusPocketIconData[3][16];
+    u16 menuScreenPokeBallsPocketIconData[3][16];
+    u16 menuScreenBattleItemsIconData[3][16];
```
lhearachel: **question**: If I understand correctly:

- The first dimension of each of these arrays represents how many states that button can be in.
- The second dimension of each of these arrays represents the maximum area of that particular button, in pixels.

If those are true, should we pull the constants for the button dimensions into this file so that we can properly scale the sizes of these arrays? I'd also say that we want a semantic way to say how many states that a button can be in, but it'd require two different custom values inside the `ButtonState` enum above.

src/overlay013/battle_bag_buttons.c
```diff
@@ -0,0 +1,559 @@
...
+    LoadButtonData(battleBag->menuScreenBattleItemsIconData[BUTTON_STATE_PRESSED], screenData, MENU_POCKET_ICON_DATA_X_OFFSET + MENU_POCKET_ICON_WIDTH * 2, MENU_POCKET_ICON_DATA_Y_OFFSET + MENU_POCKET_ICON_HEIGHT * 3, MENU_POCKET_ICON_WIDTH, MENU_POCKET_ICON_HEIGHT);
+}
+
+static void LoadButtonData(u16 *buttonData, u16 *screenData, u8 xOffset, u8 yOffset, u8 width, u8 height)
+{
+    u16 i, l;
+
+    for (i = 0; i < height; i++) {
+        for (l = 0; l < width; l++) {
+            buttonData[i * width + l] = screenData[(yOffset + i) * PALETTE_SIZE_BYTES + xOffset + l];
```
lhearachel: **question**: Is this for sure palette data? If so, that's unclear from both the function name and the parameter `screenData`.

lhearachel (commented):
Good work! I have some questions and further thoughts below.

include/overlay013/battle_bag.h
Morglom: Yep, that is what those dimensions are. You're also right we should probably pull the dimension constants into this header. With regards to the number of states, we could possible have something like
``#define BATTLE_BAG_BUTTON_STATE_NUM 3
#define BATTLE_BAG_BUTTON_STATE_WITH_DISABLE_NUM 4``
to handle that, otherwise if we wanted to handle it in the enum we could double up on 3, but that has it's downsides.
``enum ButtonState {
    BUTTON_STATE_UNPRESSED = 0,
    BUTTON_STATE_PRESSING,
    BUTTON_STATE_PRESSED,
    BUTTON_STATE_DISABLED,
    BUTTON_STATE_MAX,
    BUTTON_STATE_MAX_WITHOUT_DISABLED = 3,
};``

src/overlay013/battle_bag_buttons.c
Morglom: On second glance I think
`buttonData[i * width + l] = screenData[(yOffset + i) * TILE_SIZE_4BPP + xOffset + l];`
would be more appropriate as it's a tilemap.
I'm not too familiar with the functionality around tilemaps, palettes, etc though, so I'd happy to defer to your judgement on what would be most appropriate.

include/overlay013/battle_bag.h
Morglom: Also just a thought. Seeing as how both the previous and next page buttons have the same dimensions (and are closely related) would it be better to combine the constants like this:
`#define POCKET_MENU_PAGE_BUTTON_WIDTH 5`
`#define POCKET_MENU_PAGE_BUTTON_HEIGHT 5`

lhearachel: Regarding the states, my thought is that the best that we can do here is realistically something like this:

```c
enum ButtonState {
    BUTTON_STATE_UNPRESSED = 0,
    BUTTON_STATE_PRESSING,
    BUTTON_STATE_PRESSED,
    BUTTON_STATE_DISABLED, // WARN: This state MUST be the last in the sequence!

    BUTTON_STATE_MAX,
    BUTTON_STATE_MAX_CANNOT_DISABLE = BUTTON_STATE_DISABLED - 1,
};
```

It's potentially brittle, but it at least communicates some amount of intent and permits a user to add more states, so long as they add them above `DISABLED`. This will ultimately be a game of trade-offs no matter what we choose to do, so I am happy to defer to whatever you believe the best solution is. :slightly_smiling_face: 

> Seeing as how both the previous and next page buttons have the same dimensions (and are closely related) would it be better to combine the constants...?

IMO, yes; we don't need to clutter the namespace with excess constants here. If a user wants to modify the UI and split the widths of each button, then that's on them.

src/overlay013/battle_bag_buttons.c
lhearachel: If you want to read up more on the structure of tilemaps, then I would recommend [this article](https://www.coranac.com/tonc/text/regbg.htm). GBA tilemaps work very similarly to DS tilemaps, and they follow the same structure:

```c
struct Tile {
    u16 index : 10; // Index of the tile to render from VRAM.
    u16 hFlip : 1; // If set to 1, flip the tile horizontally.
    u16 vFlip : 1; // If set to 1, flip the tile vertically.
    u16 palette : 4; // Index of the palette slot to be used when in 16-color mode.
};
```

include/overlay013/battle_bag.h
```diff
@@ -72,16 +93,16 @@ typedef struct BattleBag {
...
+    u16 menuPocketButtonData[BUTTON_STATE_MAX_CANNOT_DISABLE][MENU_POCKET_BUTTON_WIDTH * MENU_POCKET_BUTTON_HEIGHT];
+    u16 useItemButtonData[BUTTON_STATE_MAX][USE_ITEM_BUTTON_WIDTH * USE_ITEM_BUTTON_HEIGHT];
+    u16 cancelButtonData[BUTTON_STATE_MAX_CANNOT_DISABLE][CANCEL_BUTTON_WIDTH * CANCEL_BUTTON_HEIGHT];
+    u16 pocketItemButtonData[BUTTON_STATE_MAX][POCKET_ITEM_BUTTON_WIDTH * POCKET_ITEM_BUTTON_HEIGHT];
+    u16 pocketPrevPageButtonData[BUTTON_STATE_MAX][POCKET_MENU_PAGE_BUTTON_WIDTH * POCKET_MENU_PAGE_BUTTON_HEIGHT];
+    u16 pocketNextPageButtonData[BUTTON_STATE_MAX][POCKET_MENU_PAGE_BUTTON_WIDTH * POCKET_MENU_PAGE_BUTTON_HEIGHT];
+    u16 menuScreenRecoverHPPocketIconData[BUTTON_STATE_MAX_CANNOT_DISABLE][MENU_POCKET_ICON_WIDTH * MENU_POCKET_ICON_HEIGHT];
+    u16 menuScreenRecoverStatusPocketIconData[BUTTON_STATE_MAX_CANNOT_DISABLE][MENU_POCKET_ICON_WIDTH * MENU_POCKET_ICON_HEIGHT];
+    u16 menuScreenPokeBallsPocketIconData[BUTTON_STATE_MAX_CANNOT_DISABLE][MENU_POCKET_ICON_WIDTH * MENU_POCKET_ICON_HEIGHT];
+    u16 menuScreenBattleItemsIconData[BUTTON_STATE_MAX_CANNOT_DISABLE][MENU_POCKET_ICON_WIDTH * MENU_POCKET_ICON_HEIGHT];
```
lhearachel: **quibble**: I think this would be more readable by adding, e.g., `#define MENU_POCKET_BUTTON_SIZE (MENU_POCKET_BUTTON_WIDTH * MENU_POCKET_BUTTON_HEIGHT)`.

lhearachel (approved):
Just one quibble, non-blocking. Great work!

PR #544, opened 2025-06-09
Document Poketch move tester (Overlay 43)

YakosWG:
This should be the entire overlay done
We'll probably need to discuss some of the function names since the poketch apps all have the same structure and ideally they should use the same conventions.


platinum.us/main.lsf
```diff
@@ -1000,8 +1000,8 @@ Overlay overlay42
 Overlay overlay43
```
lhearachel: **suggestion**: We don't have a fully-defined structure for poketch apps yet, but I would say that this slots into belonging in the `applications` folder. To that end, I propose:

1. Create `applications/poketch/move_tester` in both `src` and `include` folders.
2. Move the source and header files accordingly. This will also necessitate changing the folder in `src/meson.build` and the path listed in `main.lsf` (`src_applications_poketch_move_tester_<>.c.o`).
3. It may be reasonable to rename the source and header modules themselves, since much of the info is communicated in the path. You can checkout how `pokemon_summary_screen` is structured for some ideas.
4. Finally, rename the overlay to `poketch_move_tester`, and replace all the other occurrences of `overlay43` in the code with `poketch_move_tester`.

include/overlay043/poketch_move_tester_graphics.h
```diff
@@ -0,0 +1,52 @@
...
+
+typedef struct PoketchMoveTesterGraphics PoketchMoveTesterGraphics;
+
+typedef struct {
+    u32 lastButtonPressed;
+    u16 exclamCount;
+    u16 attackType;
+    u16 defenderType1;
+    u16 defenderType2;
+} MoveTesterData;
```
lhearachel: **polish**:

```suggestion
typedef struct MoveTesterData {
    u32 lastButtonPressed;
    u16 exclamCount;
    u16 attackType;
    u16 defenderType1;
    u16 defenderType2;
} MoveTesterData;
```

src/overlay043/poketch_move_tester.c
```diff
@@ -0,0 +1,366 @@
...
+    if (moveTester->sysTaskState < NELEMS(stateFuncs)) {
+        PoketechSystem_UpdateButtonManager(moveTester->poketchSys, moveTester->buttonManager);
+
+        if (stateFuncs[moveTester->sysTaskState](moveTester)) {
+            PoketchMoveTester_Free(moveTester);
+            SysTask_Done(task);
+            PoketchSystem_NotifyAppUnloaded(moveTester->poketchSys);
+        }
+    } else {
+    }
```
lhearachel: **polish**:

```suggestion
    }
```

src/overlay043/poketch_move_tester.c
```diff
@@ -0,0 +1,366 @@
...
+{
+    PoketchMoveTester *moveTester = (PoketchMoveTester *)appData;
+
+    moveTester->moveTesterData.lastButtonPressed = buttonID;
+    moveTester->buttonState = buttonState;
+}
+
+static void PoketchMoveTester_Shutdown(void *appData)
+{
+    ((PoketchMoveTester *)appData)->shouldExit = TRUE;
```
lhearachel: **quibble**: I personally think that this is more clear as:

```suggestion
    PoketchMoveTester *moveTester = appData;
    moveTester->shouldExit = TRUE;
```

src/overlay043/poketch_move_tester.c
```diff
@@ -0,0 +1,366 @@
...
+            SysTask_Done(task);
+            PoketchSystem_NotifyAppUnloaded(moveTester->poketchSys);
+        }
+    } else {
+    }
+}
+
+static void PoketchMoveTester_ButtonChanged(u32 buttonID, u32 buttonState, u32 touchState, void *appData)
+{
+    PoketchMoveTester *moveTester = (PoketchMoveTester *)appData;
```
lhearachel: **polish**: Casting from `void *` is implicit.

```suggestion
    PoketchMoveTester *moveTester = appData;
```

src/overlay043/poketch_move_tester.c
```diff
@@ -0,0 +1,366 @@
...
+        if (PoketchMoveTesterGraphics_NoActiveTasks(appData->graphics)) {
+            PoketchMoveTester_SetTaskState(appData, TASK_SHUTDOWN);
+        }
+
+        return FALSE;
+    }
+
+    switch (appData->taskFuncState) {
+    case 0:
+        if (appData->buttonState == 1) {
```
lhearachel: **question**: Is there a suitable enum for these button states?

src/overlay043/poketch_move_tester.c
```diff
@@ -0,0 +1,366 @@
...
+        break;
+    }
+
+    return FALSE;
+}
+
+static u32 PoketchMoveTester_GetExclamation(u32 attackType, u32 defenderType1, u32 defenderType2)
+{
+    static const s8 sMoveTesterTypeChart[][MOVE_TESTER_NONE_SELECTED] = {
+        { MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NOT_VERY_EFFECTIVE, MOVE_TESTER_NEUTRAL, MOVE_TESTER_IMMUNE, MOVE_TESTER_NOT_VERY_EFFECTIVE, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL, MOVE_TESTER_NEUTRAL },
```
lhearachel: :skull:

src/overlay043/poketch_move_tester.c
```diff
@@ -0,0 +1,366 @@
...
+        }
+        break;
+    }
+
+    return FALSE;
+}
+
+static u32 PoketchMoveTester_GetExclamation(u32 attackType, u32 defenderType1, u32 defenderType2)
+{
+    static const s8 sMoveTesterTypeChart[][MOVE_TESTER_NONE_SELECTED] = {
```
lhearachel: **suggestion**:

1. Pull this table out into file-level scope and format it with array-initializers.
2. I think that it would be useful to also add a comment above the first line that denotes which column belongs to which type.
3. This is a table that I think is best formatted with aligned columns; you can wrap it in `// clang-format off` and `// clang-format on` to turn off auto-formatting.
4. Use `NUM_POKEMON_TYPES` instead of `MOVE_TESTER_NONE_SELECTED` for the size.

src/overlay043/poketch_move_tester.c
```diff
@@ -0,0 +1,366 @@
...
+        TYPE_GROUND,
+        TYPE_FLYING,
+        TYPE_PSYCHIC,
+        TYPE_BUG,
+        TYPE_ROCK,
+        TYPE_GHOST,
+        TYPE_DRAGON,
+        TYPE_DARK,
+        TYPE_STEEL
+    };
```
lhearachel: **thought**: This implies to me that we want a separate local enum for the move tester... sigh.

src/overlay043/poketch_move_tester_graphics.c
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    return FALSE;
+}
+
+static void SetupSprites(PoketchMoveTesterGraphics *graphics)
+{
+    static const PoketchAnimation_AnimationData sPoketchMoveTester_AnimDataMain[] = {
+        {
+            { (28 << FX32_SHIFT), (128 << FX32_SHIFT) },
```
lhearachel: **suggestion**: Use designated initializers for clarifying fields.

src/overlay043/poketch_move_tester_graphics.c
```diff
@@ -0,0 +1,355 @@
...
+            { (196 << FX32_SHIFT), (72 << FX32_SHIFT) },
+            2,
+            0,
+            2,
+            0,
+            0,
+        },
+    };
+    static const PoketchAnimation_AnimationData sPoketchMoveTester_AnimDataExclamation = {
+        { (44 << FX32_SHIFT), (48 << FX32_SHIFT) },
```
lhearachel: **thought**: I wonder if we want a separate `FX32` macro to encase this.

src/overlay043/poketch_move_tester_graphics.c
```diff
@@ -0,0 +1,355 @@
...
+    }
+}
+
+static const PoketchTask sMoveTesterTasks[] = {
+    { TASK_DRAW_APP_SCREEN, Task_DrawAppScreen, 0x0 },
+    { TASK_FREE_WINDOWS_AND_BG, Task_FreeWindowsAndBG, 0x0 },
+    { TASK_BUTTON_PRESSED, Task_ButtonPressed, 0x0 },
+    { TASK_BUTTON_RELEASED, Task_ButtonReleased, 0x0 },
+    { TASK_UPDATE_GRAPHICS, Task_UpdateGraphics, 0x0 },
+    { 0x0, NULL, 0x0 }
```
lhearachel: **suggestion**: The last line here can just be `{ 0 }`, which is idiomatic C for a zero-filled struct.

src/overlay043/poketch_move_tester.c
lucie-sharpe: There is relevant enums in `include/poketch/poketch_button.h`

src/overlay043/poketch_move_tester.c
```diff
@@ -0,0 +1,366 @@
...
+    TASK_SHUTDOWN,
+};
+
+static void NitroStaticInit(void);
+
+static BOOL PoketchMoveTester_New(void **appData, PoketchSystem *poketchSys, BgConfig *bgConfig, u32 appID);
+static BOOL PoketchMoveTester_Init(PoketchMoveTester *appData, PoketchSystem *poketchSys, BgConfig *bgConfig, u32 appID);
+static void PoketchMoveTester_InitData(MoveTesterData *moveTesterData);
+static void PoketchMoveTester_Free(PoketchMoveTester *appData);
+static void Task_MoveTesterMain(SysTask *task, void *appData);
```
lucie-sharpe: Some functions are common to all poketch apps. Do we want common naming for these? I did this with my poketch apps functions:

**note**: Feel free to drop the module-level prefix on `static` functions, if it makes the code semantics nicer.

_Originally posted by @lhearachel in https://github.com/pret/pokeplatinum/pull/388#discussion_r1958758608_

src/overlay043/poketch_move_tester_graphics.c
Fexty12573: From my experience you should be able to replace these with `FX32_CONST(44)`, etc. Don't think a separate macro should be necessary.

src/overlay043/poketch_move_tester.c
YakoSWG: I think dropping the module-level prefix is probably the way to go altough I do like the Task_NAME for the tasks. Maybe that's just preferance?
I also think we want a common naming for the functions every poketch app has. I had a look at the digital watch app and I'd say the names are fine too altough I would prefer "New" over "Init" for the function allocating heap space since that fits more with what it does.

src/overlay043/poketch_move_tester_graphics.c
YakoSWG: Had a look and I believe the macro `FX32_CONST` does something else.
We'd need a new one here...
Maybe this is a bit out of scope for this PR though.

Gudf: It's functionally equivalent (as would doing `* FX32_ONE`), and the compiler should constant-fold everything away so it ends up being the same once compiled.

src/overlay043/poketch_move_tester.c
lhearachel: I agree with `Task_NAME`, and it's consistent across the entire repository.

`New` over `Init` is also within our standard naming conventions; `New` is a routine that performs both an `Alloc` and an `Init` on the result of that `Alloc`.

src/applications/poketch/move_tester/graphics.c
```diff
@@ -24,6 +25,66 @@
 #include "sys_task_manager.h"
 #include "text.h"
 
+static const PoketchAnimation_AnimationData sPoketchMoveTester_AnimDataButtons[6] = {
```
lhearachel: **polish**: If there isn't a dedicated constant that this array should use for its dimension, then remove it and let the compiler determine the size.

src/applications/poketch/move_tester/main.c
```diff
@@ -0,0 +1,381 @@
...
+    u8 sysTaskState;
+    u8 taskFuncState;
+    u8 shouldExit;
+    u32 appID;
+    MoveTesterData moveTesterData;
+    PoketchMoveTesterGraphics *graphics;
+    PoketchSystem *poketchSys;
+    PoketchButtonManager *buttonManager;
+    u32 buttonState;
+} PoketchMoveTester;
```
lhearachel, line 29: **quibble**: `typedef struct PoketchMoveTester`

src/applications/poketch/move_tester/main.c
```diff
@@ -0,0 +1,381 @@
...
+    [TYPE_GRASS]    = { ______, ______, X(0_5), X(0_5), X(2_0), X(2_0), X(0_5), ______, X(0_5), ______, X(0_5), X(2_0), X(0_5), ______, ______, ______, X(0_5), ______ },
+    [TYPE_ELECTRIC] = { ______, ______, X(2_0), ______, X(0_0), ______, ______, ______, ______, ______, ______, X(2_0), X(0_5), X(0_5), ______, ______, X(0_5), ______ },
+    [TYPE_PSYCHIC]  = { ______, X(2_0), ______, X(2_0), ______, ______, ______, ______, X(0_5), ______, ______, ______, ______, ______, X(0_5), ______, ______, X(0_0) },
+    [TYPE_ICE]      = { ______, ______, X(2_0), ______, X(2_0), ______, ______, ______, X(0_5), ______, X(0_5), X(0_5), X(2_0), ______, ______, X(0_5), X(2_0), ______ },
+    [TYPE_DRAGON]   = { ______, ______, ______, ______, ______, ______, ______, ______, X(0_5), ______, ______, ______, ______, ______, ______, ______, X(2_0), ______ },
+    [TYPE_DARK]     = { ______, X(0_5), ______, ______, ______, ______, ______, X(2_0), X(0_5), ______, ______, ______, ______, ______, X(2_0), ______, ______, X(0_5) }
+};
+// clang-format on
```
lhearachel, line 71: **praise**: Thanks for making this look so much nicer! 🙏

platinum.us/main.lsf
lhearachel: Bump here to do a global rename across the whole project for `overlay43` -> `poketch_move_tester`; this should affect the following files, according to my search:

```console
> rg 'overlay43'
platinum.us/main.lsf
1000:Overlay overlay43

platinum.us/sbins.sha1
45:261b64d5fd09a304a5d7e78a18b44a6ecfe7ec93 *overlay43.sbin

src/poketch/poketch_system.c
46:FS_EXTERN_OVERLAY(overlay43);
101:    { POKETCH_APPID_MOVETESTER, FS_OVERLAY_ID(overlay43) },
```

This will give the overlay a clearer semantic association between `FS_OVERLAY_ID` and the associated TUs.

lhearachel (commented):
Just a few more requested changes; thanks for cleaning things up! 🙏

YakoSWG:
Unless I forgot something I believe that'll be all of it.
Pulled out one more const array that I initially forgot to the top of the graphics.c as well.

lhearachel (approved):

PR #547, opened 2025-06-18
Document the field system camera creation/deletion

Gudf:
Documents `ov5_021D5B40`.
This module contains the functions used to create and delete the field system's camera, as well as the camera presets (or types) used by the different maps (see `include/data/map_headers.h`).

Also moves `CameraAngle` from `overlay115/camera_angle.h` (now deleted) to `camera.h`.


lhearachel (approved):
Looks great! No comments. Thanks for doing this!

PR #548, opened 2025-06-19
Added debug logging functionality

LelouBil:
It works by writing to special debug addresses recognized by no$gba and MelonDS.
When logging is disabled, defines are present that nullify logging calls and so the rom check passes

Added 2 meson options : 
- logging_enabled (true/false)
- logging_subsystems (list of strings)

The EmulatorLog function takes a subsystem parameter as the first argument, it is simply a string that will get compared to the allowed subsystems. if it starts with any of the allowed subsystems it will get printed out.

The subsystem list can be specified as a make option like so
`make debug LOG_SUBSYSTEMS=scripts,messages`

Due to difficulties with meson, currently you can only specify 6 strings max because passing an array as command line defines is hard, so I splitted it into up to 6 different defines.

I could also generate a .h and .c file containing the array to fix this issue.


meson.build
```diff
@@ -40,7 +40,7 @@ c_args = [
 if get_option('gdb_debugging')
     c_args += [
         '-O1,p',
-        '-inline', 'off'
+#        '-inline', 'off'
```
lhearachel: **question**: Should we just remove this line?

src/debug.c
```diff
@@ -0,0 +1,142 @@
...
+void EmulatorPrintf(const char *text, ...)
+{
+    va_list args;
+    va_start(args, text);
+    EmulatorVPrintf(text, args);
+    va_end(args);
+}
+
+#define Q(X)     #X
+#define QUOTE(X) Q(X)
```
lhearachel: **quibble**: I would call these `STRINGIFY` and `STRINGIFY_`, personally:

```c
#define STRINGIFY_(__s) #__s
#define STRINGIFY(__s)  STRINGIFY_(__s)
```

The `__` prefix on the variable-name is a personal habit of mine when writing functional macros, just to ensure that I'm not hitting anything that may be in the namespace. Feel free to drop it. :slightly_smiling_face:

src/overlay005/ov5_021DD42C.c
```diff
@@ -145,6 +146,8 @@ static void ov5_021DD648(UnkStruct_ov5_021DD648 *param0, const MessageLoader *pa
 {
     MessageLoader_GetStrbuf(param1, param2, param0->unk_04);
     StringTemplate_Format(param0->unk_08, param0->unk_00, param0->unk_04);
+    EmulatorLog("message", "message id %d text is :", param2);
+    EmulatorPrintStrBuf(param0->unk_00);
```
lhearachel: **suggestion**: I, personally, don't think that we should be providing these values ourselves; we can provide the tools for users to implement them, certainly, but I feel that this creates expectations of long-term maintenance that is not within the scope of *this* repository.

I'm curious what other maintainers and contributors feel about this.

src/debug.c
```diff
@@ -0,0 +1,142 @@
...
+
+#ifndef LOGGING_SUBSYSTEM_COUNT
+
+const int nb_subsystems = 0;
+const char *allowedSubsystems[] = {};
+
+#else
+const char *allowedSubsystems[LOGGING_SUBSYSTEM_COUNT] = {
```
lhearachel: **thought**: The idea behind this is interesting; if we decide to remove the bundled logging statements from this PR, then we may want to remove it, which would also ease the burden of configuration quite a bit.

tools/devtools/gen_compile_commands.py
```diff
@@ -52,6 +56,24 @@
     "-DGAME_LANGUAGE=ENGLISH",
 ]
 
+with open(meson_options_file, 'r') as optf:
+    meson_options = json.load(optf)
+
+def find_option(opt: str):
+    return [o["value"] for o in meson_options if o["name"] == opt][0]
+gdb_debugging = find_option("gdb_debugging")
```
lhearachel: **polish**: Insert an empty line between this line and the preceding block, for clarity.

meson.build
LelouBil: I mean we should try to figure out why this makes the rom buggy, because this option is useful for debugging.

src/overlay005/ov5_021DD42C.c
LelouBil: So do I drop everything related to the StrBuf printing ? Including the embedding of the charmap then.

src/debug.c
LelouBil: Yeah, the logging statement in the PR are meant as examples really.

 I added this because si figured out if logging would be added everywhere the output would become unreadable pretty quickly without filtering. 

But a system like this may be overkill too. We could just ask the user to grep, or maybe just a single string for filtering, or something else.

src/overlay005/ov5_021DD42C.c
lhearachel: I think that keeping the functionality in-place for modders to make use of is good. I'm moreso worried about sprinkling log statements throughout the vanilla code. Users should be able to log exactly what they want and no more; while the subsystem idea accomplishes that, I worry that it introduces some scope creep for the repository.

src/debug.c
lhearachel: I think `grep` is a perfectly fine idea, or we don't provide any invocations of logging ourselves and the user is expected to implement them where they want them.

docs/logging.md
```diff
@@ -0,0 +1,29 @@
+# Logging
+
+The repository includes a basic logger, that works with emulators such as MelonDS and No$GBA
```
lhearachel: **polish**: Tack periods to the ends of these sentences.

tools/debug/charmap_gen.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
+
+import pathlib
+import sys
+import codecs
+
+homedir = pathlib.Path(__file__).resolve().parent.parent.parent
```
lhearachel: **suggestion**: Rather than do this, why not pass the target directory as an argument to the script?

tools/debug/charmap_gen.py
```diff
@@ -0,0 +1,58 @@
...
+input_file = sys.argv[1]
+output_file_c = sys.argv[2]
+output_file_h = sys.argv[3]
+
+with open(input_file,'r') as inp:
+    mapping = dict()
+    for l in inp.readlines()[:3000]:
```
lhearachel: **question**: Why a maximum number of lines?

tools/debug/charmap_gen.py
```diff
@@ -0,0 +1,58 @@
...
+# language=c
+c_file = f"""
+#ifdef LOGGING_ENABLED
+#include "charmap.h"
+const char* resolve_charmap(const charcode_t id) {{
+    switch(id) {{
+        {"\n".join(array_inner)}
+        default: return NULL;
+    }}
+}};
```
lhearachel: **thought**: I think that it would be better to instead output a table to `charmap.h` which contains this table as a `static` declaration, then the logging module can include it in the source code and reference it as-needed, rather than having to go through a function-branch and a jump-table. What do you think?

tools/debug/charmap_gen.py
LelouBil: Forgot to remove that, sorry

tools/debug/charmap_gen.py
LelouBil: I was having issues with a big table, so that's why I limited the number of lines and tried to do a function.
I'll try a static array and see if that works

tools/debug/charmap_gen.py
LelouBil: It's actually done, this line is unnecessary

PR #550, opened 2025-06-20
Renamed option struct elements

AsparagusEduardo:


lhearachel:
From the build logs, it seems that one was missed in the rename:

```
..\..\..\home\runner\work\pokeplatinum\pokeplatinum\src\overlay017\ov17_022476F8.c:204: 'unk_196C' is not a member of class 'struct '
```

AsparagusEduardo:
> From the build logs, it seems that one was missed in the rename:

yep, my bad, fixed

lhearachel (approved):

PR #554, opened 2025-06-20
Slight Ball Capsule documentation

AsparagusEduardo:


generated/seals.txt
```diff
@@ -0,0 +1,82 @@
+SEAL_DUMMY
```
lhearachel: **question**: I have a similar ask here as on #491; is there a strong justification to move these to a generated `enum` rather than keep them in a plain `enum`? Are these going to be used in any game-scripts or build-scripts...?

AsparagusEduardo: > **question**: I have a similar ask here as on #491; is there a strong justification to move these to a generated `enum` rather than keep them in a plain `enum`? Are these going to be used in any game-scripts or build-scripts...?

Reverted

PR #557, opened 2025-06-21
Poketch standardization

lucie-sharpe:
This patch standardizes some function and variable names for poketch apps. It also moves them into the applications folder. I have moved the OV25 files into applications/poketch/base but am not sure if this is the best name?


include/applications/poketch/digital_watch/graphics.h
```diff
@@ -1,16 +1,10 @@
-#ifndef POKEPLATINUM_POKETCH_DIGITAL_WATCH_DISPLAY_H
-#define POKEPLATINUM_POKETCH_DIGITAL_WATCH_DISPLAY_H
+#ifndef POKEPLATINUM_POKETCH_digital_watch_graphics_H
+#define POKEPLATINUM_POKETCH_digital_watch_graphics_H
```
lhearachel: **polish**: Use upper-case for include-guards.

src/meson.build
```diff
@@ -659,15 +659,15 @@ pokeplatinum_c = files(
...
-    'pre_poketch_subscreen/pre_poketch_subscreen.c',
-    'poketch/poketch_system.c',
-    'poketch/poketch_graphics.c',
-    'poketch/poketch_task.c',
-    'poketch/poketch_animation.c',
-    'poketch/poketch_button.c',
-    'poketch/ov25_02255DBC.c',
-    'poketch_digital_watch/main.c',
-    'poketch_digital_watch/display.c',
+    'applications/poketch/pre_poketch_subscreen/pre_poketch_subscreen.c',
```
lhearachel: **thought**: Maybe this would be better-named as `poketch/empty/poketch_empty.c`...? Or something to that effect...?

lhearachel (commented):
Thanks for taking charge on this! :heart: 

> I have moved the OV25 files into applications/poketch/base but am not sure if this is the best name?

I think `base` is fine, but there's also the option to store all of the core `poketch` system files directly under `applications/poketch`, which I think is more intuitive.

src/meson.build
lucie-sharpe: renamed to poketch_unavailable and moved poketch/base directly into poketch folder

lhearachel (approved):

PR #558, opened 2025-06-21
Document the size contest (module `ov5_021EE7D4`)

Gudf:
Not much to say here.

There's a bit of weirdness going on with the use of `u64`s and `vu16`s where they should not be required for the code to work properly, but it's needed for matching so here we go I guess.


generated/size_contest_results.txt
```diff
@@ -0,0 +1,3 @@
+SIZE_CONTEST_SMALLER
+SIZE_CONTEST_SAME_SIZE
+SIZE_CONTEST_LARGER
```
lhearachel, line 3: **thought**: This is suggesting to me that `metang` could do with some enhancements down the line to support building multiple `enum`s from a single input file... I mostly worry about the total number of files being presented to users here that are ultimately quite small. I'll think about that a bit.

lhearachel (approved):

PR #559, opened 2025-06-22
Document some underground record data and functions

CharlesFolz4:
Document some underground record data and functions
Renamed `SecretBaseRecord` to `UndergroundRecord`
Updated struct files for `UndergroundRecord` and `UndergroundData` to match struct names
Removed unnecessary struct decls


include/struct_defs/underground_record.h
```diff
@@ -0,0 +1,22 @@
...
+    u32 unk_18_0 : 20;
+    u32 numTrapsHit : 20;
+    u32 unk_20_0 : 20;
+    u32 unk_24_0 : 20;
+    u32 unk_28_0 : 20;
+    u32 unk_2C_0 : 20;
+    u32 unk_30_0 : 20;
+    u32 unk_34_0 : 20;
+    u32 capturedFlagCount : 20;
+} UndergroundRecord;
```
lhearachel, line 20: **suggestion**: The bitfield syntax here seems superfluous; it shouldn't affect the layout of the struct in memory, but maybe it affects codegen when assigning values? Does the codegen match without the bitfield defintions? i.e.,

```suggestion
typedef struct UndergroundRecord {
    u32 trainerScore;
    u32 peopleMet;
    u32 numGiftsGiven;
    u32 unk_0C_0;
    u32 numSpheresDug;
    u32 numFossilsDug;
    u32 unk_18_0;
    u32 numTrapsHit;
    u32 unk_20_0;
    u32 unk_24_0;
    u32 unk_28_0;
    u32 unk_2C_0;
    u32 unk_30_0;
    u32 unk_34_0;
    u32 capturedFlagCount;
} UndergroundRecord;
```

include/struct_defs/underground_record.h
```diff
@@ -0,0 +1,22 @@
+#ifndef POKEPLATINUM_STRUCT_UNDERGROUND_RECORD_H
+#define POKEPLATINUM_STRUCT_UNDERGROUND_RECORD_H
+
+typedef struct UndergroundRecord {
+    u32 trainerScore : 20;
+    u32 peopleMet : 20;
+    u32 numGiftsGiven : 20;
+    u32 unk_0C_0 : 20;
+    u32 numSpheresDug : 20;
+    u32 numFossilsDug : 20;
```
lhearachel, line 10: **suggestion**: Maybe `num[Sphere,Fossils]Mined`?

src/unk_0202854C.c
```diff
@@ -16,58 +18,64 @@
...
-    0x139
+#define FLAG_CAPTURED_COUNT_PLATINUM 50
+#define FLAG_CAPTURED_COUNT_GOLD     10
+#define FLAG_CAPTURED_COUNT_SILVER   3
+#define FLAG_CAPTURED_COUNT_BRONZE   1
+#define FLAG_CAPTURED_COUNT_NORMAL   0
+
+static void sub_02028B48(UndergroundData *undergroundData, int param1, int param2);
+
+u16 MiningItems[] = {
```
lhearachel: **polish**:

```suggestion
static const u16 sMiningItems[] = {
```

src/unk_0202854C.c
```diff
@@ -80,30 +88,30 @@ int sub_02028554(void)
...
 void Underground_Init(UndergroundData *param0)
 {
-    u32 v0 = 0;
+    u32 timestamp = 0;
     RTCDate v1;
     RTCTime v2;
 
     GetCurrentDateTime(&v1, &v2);
-    v0 = (((((((u32)v1.year * 32ULL + v1.month) * 32ULL) + v1.day) * 32ULL + v2.hour) * 32ULL + v2.minute) * 32ULL + (v2.second + gSystem.vblankCounter));
+    timestamp = (((((((u32)v1.year * 32ULL + v1.month) * 32ULL) + v1.day) * 32ULL + v2.hour) * 32ULL + v2.minute) * 32ULL + (v2.second + gSystem.vblankCounter));
```
lhearachel: **thought**: There is some similar code in `CommSys_Seed`, though it's not 100% identical:

```c
    GetCurrentDateTime(&date, &time);
    seed = (((((((u64)date.year * 16ULL + date.month) * 32ULL) + date.day) * 32ULL + time.hour) * 64ULL + time.minute) * 64ULL + (time.second + gSystem.vblankCounter));
```

I wonder if there is a way that we can parameterize this into an inline function that is shared between them...? We may need to inspect the code-gen of each instance.

src/unk_0202854C.c
```diff
@@ -80,30 +88,30 @@ int sub_02028554(void)
...
-    return v0;
+    UndergroundRecord *undergroundRecord = Heap_AllocFromHeap(heapID, sizeof(UndergroundRecord));
+    MI_CpuFill8(undergroundRecord, 0, sizeof(UndergroundRecord));
+    return undergroundRecord;
 }
 
 void Underground_Init(UndergroundData *param0)
 {
-    u32 v0 = 0;
+    u32 timestamp = 0;
```
lhearachel: **suggestion**: I would call this `seed`, since it *is* being used as a seed-value for the RNG functions.

src/unk_0202854C.c
```diff
@@ -80,30 +88,30 @@ int sub_02028554(void)
...
     RTCDate v1;
     RTCTime v2;
```
lhearachel: **polish**: Name these `date` and `time`.

src/unk_0202854C.c
```diff
@@ -980,18 +988,18 @@ BOOL sub_02029274(UndergroundData *param0, int param1)
     return 1;
 }
 
-void sub_0202929C(UndergroundData *param0)
+void UndergroundData_IncrementStepCount(UndergroundData *undergroundData)
 {
-    if (param0->unk_9AB >= (100 - 1)) {
-        param0->unk_9AB = 0;
+    if (undergroundData->stepCount >= (100 - 1)) {
+        undergroundData->stepCount = 0;
```
lhearachel, line 994: **polish**: Use `> 100` rather than `>= 99`; the former creates a clear semantic relationship to the associated game record.

src/unk_0202854C.c
```diff
@@ -1040,17 +1048,18 @@ void sub_02029364(UnkStruct_02029894 *param0, int param1, int param2, int param3
     sub_02029308(param0, param1, v0, param2, param3);
 }
 
+// this is off-by-one compared to the actual flag type
 static int sub_02029384(const UnkStruct_02029894 *param0)
```
lhearachel, line 1052: **question**: Any reason not to name this function / enumify its returned values? The off-by-one is frustrating, but maybe this function's return value isn't used in places that expect `enum BaseFlagType`...?

src/unk_0202854C.c
```diff
@@ -1098,144 +1107,144 @@ int sub_0202942C(const UnkStruct_02029894 *param0, int param1)
...
 {
-    if (4 == sub_0202958C(param0)) {
-        return 1;
+    if (BASE_FLAG_PLATINUM == UndergroundRecord_GetFlagRank(undergroundRecord)) {
+        return TRUE;
     }
 
-    return 0;
+    return FALSE;
 }
```
lhearachel: **question**: Does this function-body match as `return BASE_FLAG_PLATINUM == UndergroundRecord_GetFlagRank(undergroundRecord);`?

lhearachel (commented):
Exciting! Great to see us starting to look at Underground code. :grin:

src/unk_0202854C.c
CharlesFolz4: Yes on both counts. 
I don't really understand why or what it is doing so I can't accurately name it.  
Then the only place it is used, `sub_02029384`, directly returns this function's value, but only under certain conditions, and returns either a 0, 6, or a different undocumented variable under different conditions, which makes me suspect that these values are part of a different unknown enum.

src/unk_0202854C.c
CharlesFolz4: That somehow doesn't match.

src/unk_0202854C.c
CharlesFolz4: Doesn't match with `const`

include/struct_defs/underground_record.h
CharlesFolz4: This name was chosen to maintain consistency with how it is displayed in-game:
image

include/struct_defs/underground_record.h
CharlesFolz4: It does not match without those fields.

lhearachel: I figured that might happen. It must be generating bit-shifts on value-access. :disappointed:

include/struct_defs/underground_record.h
lhearachel: I can vibe with that.

src/unk_0202854C.c
lhearachel: That is *extremely* surprising, but alright.

PR #562, opened 2025-06-22
save_data: playTimestamp -> firstCompletionTimestamp

ellieplayswow:
hello! the `playTimestamp` in save data is actually recorded as the first time the HOF is entered (see: https://github.com/kwsch/PKHeX/blob/master/PKHeX.Core/Saves/SAV4.cs#L353 ). i've updated some of the surrounding methods to reflect this too


ellieplayswow:
drafting this for a bit as i have a few more commits inc labeling a few more  parts of save data

ellieplayswow:
for reference, with `dwcProfileId` i am _kinda_ educated guessing here. i have no idea what it actually is, but it seems to be some kind of networked ID (https://twlsdk.randommeaninglesscharacters.com/docs/nitro/NitroDWC/dwc/base/DWC_GetGsProfileId.html ). indeed, a fresh save has this as 0, and so it looks like the associated functions all manage getting/setting that.

i tried to look for references for `unk_49` and `unk_50` but i couldn't spot any? i'd like to try and figure out what they are too, from my own tests:
- `unk_49`: is `0` across all reference saves i have
- `unk_50`: is _only_ partially set on a reference save with a `dwcProfileId`..

asm/macros/scrcmd.inc
```diff
@@ -3902,7 +3902,7 @@
     .short \arg0
     .endm
 
-    .macro ScrCmd_2AC
+    .macro ScrCmd_UnlockMysteryGift
```
lhearachel: **polish**: Documented script-commands in `.inc` and `.s` files should omit the `ScrCmd_` prefix. We still want to keep it in the associated `.h` and `.c` functions, but removing it in the assembler-macros makes the semantics of the field scripts more elegant.

src/system_data.c
```diff
@@ -67,25 +67,25 @@ u8 SystemData_GetOwnerBirthDayOfMonth(const SystemData *systemData)
     return systemData->ownerBirthDayOfMonth;
 }
 
-BOOL sub_02025D64(const SystemData *param0)
+BOOL SystemData_IsMysteryGiftUnlocked(const SystemData *param0)
```
lhearachel: **polish**: Update the names of the arguments for these functions in both declaration (`.h`) and definition (`.c`). This applies to all renamed functions. No need to rename parameters in all clients that call these functions, but it's a nice addition if it's little effort. :slightly_smiling_face:

lhearachel (changes requested):

ellieplayswow:
i would like to try and figure out what the remaining 2 vars in systemdata are, but i can't find any references to either of those being set in the codebase. any ideas on where to check?

lhearachel:
> i would like to try and figure out what the remaining 2 vars in systemdata are, but i can't find any references to either of those being set in the codebase. any ideas on where to check?

If you cannot find references to them being set, then they very well could be unused. That's not uncommon. :grin:

If there are functions that act as accessors for these fields, then you can grep the codebase to see if there are any extern calls from other translation units. If there aren't, then:

1. Remove the function declaration from the header file
2. Mark the function inside of the source file as `static` and give it a name that signals that it is never used. Maybe we want some kind of macro that can tag such functions with a compiler attribute, or is a functional preproc no-op? (`#define UNUSED`, replacement content intentionally omitted)
3. Name the struct members in a way that signals that they are unused.

ellieplayswow:
> If you cannot find references to them being set, then they very well could be unused. That's not uncommon. 

actually my childhood save (which is from actual cart & hardware, no AR) has `unk_50` partially set, but indeed on an emulated new save & a reference save i found online it's unset, this is why i'm wondering if it's just an artifact from godmode or perhaps just an artifact of how saving works (looks like it just dumps the memory, so if it's not zeroed it might save?)

interestingly, if i just rename the fields in the struct, the build passes, which would suggest neither are actually used? my concern is if they're used in some pointer hell...

i'm fine to rename these fields to something like `unusedByte` (`unk_49`) and `unusedArrayOfUints` (`unk_50`) unless there's a convention elsewhere?

lhearachel:
> interestingly, if i just rename the fields in the struct, the build passes, which would suggest neither are actually used? my concern is if they're used in some pointer hell...

I would not worry about that right now; if everything still matches, then just rename the struct members and call them good for the scope of this PR. :grin: 

> i'm fine to rename these fields to something like unusedByte (unk_49) and unusedArrayOfUints (unk_50) unless there's a convention elsewhere?

Our typical style here is just `unused`, with an optional byte-offset (prefixed by `_`) from the vanilla struct layout if there are multiple such members in the same struct. e.g., for these, `unused_49` and `unused_50`.

ellieplayswow:
ty! updated ^^

include/system_data.h
```diff
@@ -22,9 +22,9 @@ typedef struct SystemData {
     u8 ownerBirthDayOfMonth;
     GameTime gameTime;
     u8 isMysteryGiftUnlocked;
-    u8 unk_49;
+    u8 unused_49;
     s32 dwcProfileId;
-    u32 unk_50[3];
+    u32 unused_4E[3];
```
lhearachel: **quibble**: Shouldn't this be at `50` due to padding bytes inserted by the compiler between `unused_49` and `dwcProfileId`?

ellieplayswow: i was confused about this as well - is an `s32` not 4 bytes stored? afaik it's just a rust `i32`?

ellieplayswow: for ref (highlighting at `0x4A`, where `dwcProfileId` starts)

image

Gudf: It is, but the GBA [doesn't support misaligned accesses](https://problemkaputt.de/gbatek.htm#armcpumemoryalignments), so the compiler adds two padding bytes in between `unk_48` and `dwcProfileId`.

lhearachel: It is, but the target ABI enforces that values are aligned to boundaries according to their type. e.g., an `s32` must be aligned to a word boundary (address divisible by 4), a `u16` must be aligned to a half-word boundary (address divisible by 2), etc. Compilers implement this for structures by inserting padding-bytes between struct-members, and some will also support a way to [tag a struct as "packed"](https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Packed-Structures.html) to suppress such insertions.

My suspicion in the highlighted image is that `0x4A` and `0x4B` are both these sorts of padding bytes, and that the value of `dwcProfileId` spans from `0x4C..0x4F`. You could verify this by adding the following to the struct definition and checking if the code-gen still matches:

```diff
typedef struct SystemData {
    s64 rtcOffset;
    u8 macAddress[6];
    u8 ownerBirthMonth;
    u8 ownerBirthDayOfMonth;
    GameTime gameTime;
    u8 isMysteryGiftUnlocked;
    u8 unk_49;
    u8 unused_49;
+   u8 padding_4A[2];
    s32 dwcProfileId;
    u32 unused_4E[3];
} SystemData;
```

ellieplayswow: aha! looking at diamond it has 3 unused bytes at the same location: https://github.com/pret/pokediamond/blob/master/include/save_system_info.h#L27

will update :D thanks a lot for that

PR #564, opened 2025-06-22
docs(install): on debian/ubuntu, install nasm

ellieplayswow:
i needed to install `nasm` on pop!_os 24.04 (based off ubuntu 24.04) in order to be able to run `make`.

in addition, add a note to install meson via pip3 (pop has an outdated version of meson)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -208,7 +208,7 @@ the repository](#2-downloading-the-repository).
 2. Install the following packages via `apt`:
 
     ```bash
-    sudo apt install bison flex g++ gcc-arm-none-eabi git make ninja-build pkg-config wget python3 xz-utils libc6:i386
+    sudo apt install bison flex g++ gcc-arm-none-eabi git make ninja-build pkg-config wget python3 xz-utils nasm libc6:i386
     ```
 
 #### Arch Linux (and derivatives, e.g., Manjaro, Endeavour)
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -208,7 +208,13 @@ the repository](#2-downloading-the-repository).
...
     ```bash
-    sudo apt install bison flex g++ gcc-arm-none-eabi git make ninja-build pkg-config wget python3 xz-utils libc6:i386
+    sudo apt install bison flex g++ gcc-arm-none-eabi git make ninja-build pkg-config wget python3 xz-utils nasm libc6:i386
+    ```
+
+3. Install `meson` via `pip3`:
+
+    ```bash
+    pip3 install --user meson --break-system-packages
     ```
```
lhearachel: **note**: There's actually no need to do this; we "install" `meson` for a user by cloning the repository into the `subprojects` folder and checking out a version-tag. The Makefile is wired to track this as a build-dependency and perform this pull as-needed.

The bigger issue with that system is that it makes it difficult to interact with the build configuration via the typical `meson` commands that one might issue on the command line.

ellieplayswow: ah, i see, sorry! will update this, i think i was trying to get meson working via an ide to help track references :) good to know for the future

lhearachel (approved):

PR #566, opened 2025-06-22
save_data: adding explicit padding to PlayerSave

ellieplayswow:
via #562 : `Options` is a u16 so there's 2 bytes of padding after for alignment

### DIFF
--- a/include/save_player.h
+++ b/include/save_player.h
@@ -7,7 +7,8 @@
 #include "trainer_info.h"
 
 typedef struct PlayerSave {
-    Options options;
+    Options options; // u16 bitfield
+    u8 padding_02[2];
     TrainerInfo info;
     u16 coins;
     PlayTime playTime;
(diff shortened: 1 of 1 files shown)


lhearachel:
I'm wary of including this one as an explicit member, because it's likely that modders will introduce more elements to the `Options` struct, and that may introduce an additional padding-word by mistake.

I think that keeping the documentation is valuable, though; what if we had this as an inline comment?

```c

typedef struct PlayerSave {
    Options options; // u16 bitfield
    // u8 padding_02[2]; // implicit padding in vanilla
    TrainerInfo info;
    u16 coins;
    PlayTime playTime;
} PlayerSave;
```

ellieplayswow:
i think it's a reasonable alternative, i tried playing around with a few variants like adding some extra padding bits to `Options` but couldn't get any to play nicely. i don't really have any argument for or against so i'll change it as you mentioned ^^

lhearachel (approved):

PR #567, opened 2025-06-23
Rename object event movement type symbols

AsparagusEduardo:


src/scrcmd.c
```diff
@@ -5443,12 +5443,12 @@ static BOOL ScrCmd_187(ScriptContext *ctx)
     return FALSE;
 }
 
-static BOOL ScrCmd_188(ScriptContext *ctx)
+static BOOL ScrCmd_SetObjectEventMovType(ScriptContext *ctx)
```
```suggestion
static BOOL ScrCmd_SetObjectEventMovementType(ScriptContext *ctx)
```

AsparagusEduardo: Should this also apply to the script macro too?

cbt6: Yes obviously, it makes no sense for them to have different names.

AsparagusEduardo: > Yes obviously, it makes no sense for them to have different names.

done 👍

PR #571, opened 2025-06-23
Document the GBA Pokémon game header struct

Gudf:
Emerald, Fire Red and Leaf Green have information on the ROM and save data structures to be read from other games (GC games, gen 4 games...) held in bytes 0x100 to 0x203 in their ROM (although Platinum only reads up to byte 0x1FB).

This was pretty much pulled straight from pokeemerald (`src/rom_header_gf.c`), since most of these fields are unread in Platinum.


src/overlay097/ov97_02235D18.c
```diff
@@ -62,12 +162,12 @@ int ov97_02235D2C(void *param0)
...
 
     if (sLoadedGBACartInfo->gameSet == AGB_SET_RUBYSAPP) {
-        Unk_ov97_0223F450.unk_88 = 2192;
-        Unk_ov97_0223F450.unk_8C = 15040;
-        Unk_ov97_0223F450.unk_50 = 4640;
-        Unk_ov97_0223F450.unk_98 = 9;
+        Unk_ov97_0223F450.saveBlock2Size = 2192;
+        Unk_ov97_0223F450.saveBlock1Size = 15040;
+        Unk_ov97_0223F450.flagsOffset = 4640;
+        Unk_ov97_0223F450.warpFlagsOffset = 9;
```
lhearachel, line 168: **question**: Are there interesting constants that we can pull from `pokeruby`, etc. for these values? Or are those computed using `sizeof` on their particular structs?

lhearachel (approved):
Approved with one curious question.

src/overlay097/ov97_02235D18.c
Gudf: So, in pokeemerald, the [saveBlockXSize are computed w/ `sizeof`](https://github.com/pret/pokeemerald/blob/e1b83d605698f22bee4ebcb1717378a1a10f3717/src/rom_header_gf.c#L140-L141) and the offsets w/ `offsetof` ([flags](https://github.com/pret/pokeemerald/blob/e1b83d605698f22bee4ebcb1717378a1a10f3717/src/rom_header_gf.c#L113), [warp flags](https://github.com/pret/pokeemerald/blob/e1b83d605698f22bee4ebcb1717378a1a10f3717/src/rom_header_gf.c#L144)), and I couldn't find any reference to the specific values in pokeruby. (And Ruby/Sapphire don't seem to include the header afaict, so I can't check there.)

We could copy the struct definitions, I guess?

lhearachel: > We could copy the struct definitions, I guess?

That seems ill-advised. :sweat_smile: We can merge this as-is.

PR #578, opened 2025-06-25
save_data: document OverworldMapHistory

ellieplayswow:
not totally sure what to do with the files that are now 100% documented, is it just to rename them (eg: `include/struct_defs/struct_020556C4_sub1.h` -> `include/struct_defs/struct_overworld_map_history_item.h`)?
should the headers for maphistory & maphistoryitem be combined?

the logic for renaming one of the params `unused_03` is that it never seems to have any meaningful data, it's always set to 1. i'm not sure if this matches any real definition for a param being marked as "unused" but i'm happy to change it if needed


include/struct_defs/struct_020556C4.h
```diff
@@ -4,10 +4,10 @@
...
-    int unk_00;
-    u8 unk_04;
-    u8 unk_05;
-    UnkStruct_020556C4_sub1 unk_06[6];
-} UnkStruct_020556C4;
+    int historyPointer;
+    u8 currentMapX;
+    u8 currentMapZ;
+    OverworldMapHistoryItem items[6];
+} OverworldMapHistory;
```
lhearachel: **style**: Our style in this project is to both tag and typedef a struct with the same name, e.g., here:

```c
typedef struct OverworldMapHistory {
   // snip
} OverworldMapHistory;
```

include/struct_defs/struct_020556C4_sub1.h
```diff
@@ -2,10 +2,10 @@
...
 typedef struct {
-    u8 unk_00;
-    u8 unk_01;
-    u8 unk_02;
-    u8 unk_03;
-} UnkStruct_020556C4_sub1;
+    u8 mapX;
+    u8 mapZ;
+    u8 faceDirection;
+    u8 unused_03; // this is "unused" because it's always set to 1
```
lhearachel: **suggestion**: Our typical style for variables that are set but never meaningfully used is `dummy`; that's not 100% consistent, but it tracks with a good portion of the repo.

ellieplayswow:
thanks! i've left it at `overworld_map_history` just because i think it's more descriptive from my understanding. i was considering `field_overworld_map_history` so in your IDE it shows up near the `field_overworld_state` but unless there's a reason for the state to be `field_overworld_state` i would actually suggest maybe renaming that just to `overworld_state`?

ellieplayswow:
as a note, for whatever reason i can't properly install clang-format with my repos, so to track this down i had to end up using this docker container: https://hub.docker.com/r/xianpengshen/clang-tools

it would be nice if the workflow dropped a comment or diff with what's wrong, it looks like the linter stage supports it but needs some flags to enable? unsure if that's desired or what works best

lhearachel:
> as a note, for whatever reason i can't properly install clang-format with my repos, so to track this down i had to end up using this docker container

There are instructions in the `CONTRIBUTING.md` for setting up `clang-format` as a pre-commit action that use a Python venv as a wrapper for compatibility. Those might work for you.

ellieplayswow:
oh, awesome, that works :)

src/overworld_map_history.c
```diff
@@ -0,0 +1,51 @@
+#include "overworld_map_history.h"
+
+#include <nitro.h>
+#include <string.h>
+
+void OverworldMapHistory_Clear(OverworldMapHistory *mapHistory)
+{
+    u8 v0;
+
+    for (v0 = 0; v0 < 6; v0++) {
```
lhearachel: **polish**: Use inline variable declaration + definition style where possible, e.g., here:

```suggestion
    for (u8 i = 0; i < 6; i++) {
```

(And also update simple variable names).

src/overworld_map_history.c
```diff
@@ -0,0 +1,51 @@
...
+        mapHistory->historyPointer = (mapHistory->historyPointer + 1) % 6;
+    }
+}
+
+void OverworldMapHistory_Push(OverworldMapHistory *mapHistory, int x, int z, int faceDirection)
+{
+    u8 mapX = x, mapZ;
+    mapZ = z;
+
+    if ((mapX != mapHistory->currentMapX) || (mapZ != mapHistory->currentMapZ)) {
```
lhearachel: **polish**: Remove redundant parens; I usually don't mention this outright, but I'm tossing it in since I'm requesting a change as-is. :slightly_smiling_face:

PR #580, opened 2025-06-26
document PalPad

ellieplayswow:
still need to clean a few things

### DIFF
--- a/include/pal_pad.h
+++ b/include/pal_pad.h
@@ -53,7 +53,7 @@ u8 PalPad_GetTrainerRegionCode(const PalPad *palPad, int trainerIndex);
  * @param second
  * @return TRUE if trainer name & ID matches, FALSE otherwise
  */
-BOOL PalPad_TrainersEqual(const PalPad *first, const PalPad *second);
+BOOL PalPad_AreTrainersEqual(const PalPad *first, const PalPad *second);
 
 /**
  * @brief Push {numberToCopy} entries from the beginning of {source} to the beginning of {destination}
--- a/src/pal_pad.c
+++ b/src/pal_pad.c
@@ -22,9 +22,7 @@ PalPad *SaveData_GetPalPad(SaveData *saveData)
 
 void PalPad_Init(PalPad *palPad)
 {
-    int i;
-
-    for (i = 0; i < PAL_PAD_ENTRIES; i++) {
+    for (int i = 0; i < PAL_PAD_ENTRIES; i++) {
         memset(&palPad[i], 0, sizeof(PalPad));
         palPad[i].trainerName[0] = CHAR_EOS;
     }
@@ -40,7 +38,7 @@ u8 PalPad_GetTrainerRegionCode(const PalPad *palPad, int trainerIndex)
     return palPad->associatedTrainerRegionCodes[trainerIndex];
 }
 
-BOOL PalPad_TrainersEqual(const PalPad *first, const PalPad *second)
+BOOL PalPad_AreTrainersEqual(const PalPad *first, const PalPad *second)
 {
     if (0 == CharCode_Compare(first->trainerName, second->trainerName)) {
         if (first->trainerId == second->trainerId) {
@@ -53,7 +51,7 @@ BOOL PalPad_TrainersEqual(const PalPad *first, const PalPad *second)
 
 void PalPad_PushEntries(PalPad *destination, PalPad *source, int numberToCopy, int heapID)
 {
-    int i, j, newIndex;
+    int i;
     int diffs[5];
     PalPad *newPad = Heap_AllocFromHeap(heapID, sizeof(PalPad) * PAL_PAD_ENTRIES);
     PalPad_Init(newPad);
@@ -61,14 +59,14 @@ void PalPad_PushEntries(PalPad *destination, PalPad *source, int numberToCopy, i
     for (i = 0; i < numberToCopy; i++) {
         diffs[i] = -1;
 
-        for (j = 0; j < PAL_PAD_ENTRIES; j++) {
-            if (PalPad_TrainersEqual(&destination[j], &source[i])) {
+        for (int j = 0; j < PAL_PAD_ENTRIES; j++) {
+            if (PalPad_AreTrainersEqual(&destination[j], &source[i])) {
                 diffs[i] = j;
             }
         }
     }
 
-    newIndex = 0;
+    int newIndex = 0;
 
     for (i = 0; i < numberToCopy; i++) {
         newPad[newIndex] = source[i];


ellieplayswow:
for reference, i think `PalPad_GetTrainerRegionCode` is basically an unused function; it's only used in a call to `StringTemplate_SetStrbuf` as the 6th arg (which is marked as `unused5`)

on top of that, i've tried to make as much sense of this as possible, it took me a bit  longer than i'd like to admit to get my head around some of this :p

src/pokedex.c
```diff
@@ -52,6 +52,7 @@ typedef struct Pokedex {
     u32 rotomFormsSeen;
     u8 shayminFormsSeen;
     u8 giratinaFormsSeen;
+    u8 padding[2];
```
lhearachel: **polish**: Comment out this padding to suggest that it is implicit.

src/unk_0207A6DC.c
```diff
@@ -651,22 +650,22 @@ static void sub_0207ADB4(int param0, int param1, void *param2, void *param3)
...
-    param2->unk_10 = TrainerInfo_ID(param0);
-    param2->unk_14 = TrainerInfo_RegionCode(param0);
-    param2->unk_15 = TrainerInfo_GameCode(param0);
-    param2->unk_16 = TrainerInfo_Gender(param0);
+    destination->trainerId = TrainerInfo_ID(trainerInfo);
+    destination->regionCode = TrainerInfo_RegionCode(trainerInfo);
+    destination->gameCode = TrainerInfo_GameCode(trainerInfo);
+    destination->gender = TrainerInfo_Gender(trainerInfo);
 
     for (v0 = 0; v0 < 16; v0++) {
```
lhearachel: **polish**: Document `v0` and merge the declaration with this definition, if possible.

src/pal_pad.c
```diff
@@ -0,0 +1,117 @@
...
+                break;
+            }
+        }
+    }
+
+    memcpy(destination, newPad, sizeof(PalPad) * PAL_PAD_ENTRIES);
+    Heap_FreeToHeap(newPad);
+}
+
+int PalPad_TrainerIsFriend(PalPad *palPad, u32 trainerId)
```
lhearachel, line 98: **suggestion**: This name feels a bit dubious to me, but I don't have a strong alternative to suggest at the moment. Namely, the `trainerIdHistory` bit is throwing me a bit for a loop; is that not just a record of trainers the player has met in the past?

ellieplayswow: the trainer id history is effectively your friends' friends lists. its not explicitly just trainers your friends have met because in order to get a pal pad entry you actually have to manually add someone

Agree that `*History` kinda sucks as variable names but i couldn't think of anything else that wasn't a silly level of verbose, happy to take in any ideas there

src/unk_0207A6DC.c
ellieplayswow: sorry what do you mean by the latter half here? move this function into pal_pad.c?

ellieplayswow:
will implement the feedback in the morning

src/pokedex.c
ellieplayswow: i'm happy to  do this, but is there some logic i can consider when adding these padding vars in?

is the logic something along the lines of marking it implicit in places where the padding may have some value to modders? (eg: in pokedex a modder may use it to add some extra flags)

as an example, i know there's an undocumented padding byte in `BerryPatch` (i'll get to it, promise!):

```c
typedef struct BerryPatch {
    u8 berryID;
    u8 growthStage;
    u16 stageMinutesRemaining;
    u16 moistureMinutesRemaining;
    u8 replantCount;
    //u8 padding; <--
    u16 yield;
    u8 moistureRating;
    u8 yieldRating;
    u8 mulchType;
    u8 isGrowing;
} BerryPatch;
```

would this be considered a place for it to be noted as implicit?

ellieplayswow:
all done barring the last point, but need some guidance on what would be best here, sorry!

src/pokedex.c
lhearachel: > is the logic something along the lines of marking it implicit in places where the padding may have some value to modders? (eg: in pokedex a modder may use it to add some extra flags)

Yes, but we prefer to mark it as implicit via comments, because adding the bytes to the structure *directly* is potentially dangerous and can have cascading effects on overall alignment.

**thought**: This might be a good thing for us to add to our documentation, both for contributors and for potential users.

---

This particular structure is interesting. I would mark it like this (save for the actual `word-boundary here` comments, which are just for illustration):

```c
typedef struct BerryPatch {
    u8 berryID;
    u8 growthStage;
    u16 stageMinutesRemaining;     // <-- word-boundary here
    u16 moistureMinutesRemaining;
    u8 replantCount;
    // u8 padding_07;              // <-- word-boundary here
    u16 yield;
    u8 moistureRating;
    u8 yieldRating;                // <-- word-boundary here
    u8 mulchType;
    u8 isGrowing;
    // u8 padding_0E[2];           // <-- word-boundary here
} BerryPatch;
```

src/pal_pad.c
lhearachel: How do `friends` and `friendsOfFriends` feel?

ellieplayswow: im not sure if that works? afaik the only thing iffy here are the arrays at the end, and those are for friends of friends so you'd have to denote them as `friendsOfFriendsTrainerIds`, `friendsOfFriendsGenders` or am i misunderstanding?

src/pokedex.c
ellieplayswow: makes sense, moving forwards ill only add them in as comments ^^

ellieplayswow:
btw: what is the projects' stance on new constants? eg here i define `PAL_PARK_ENTRIES` and try to make good use of it, but in #578 i could define a similar constant for [here](https://github.com/pret/pokeplatinum/pull/578/files#diff-787ce0ab7f5bd125a397b94a55d848ee35d3989691635b067908341f4a421aa4R15) - the number 6 is actually used a lot in those systems because of the history pointer, otherwise i would agree it probably isn't super useful.

another constant that would be genuinely useful (imo) is something like `STRING_TERMINATOR` (defined as `0xffff`). there are other constants defined as `0xffff`, but none of which thematically make sense to use eg [here](https://github.com/pret/pokeplatinum/pull/580/files#diff-476e34c09ec5eba6492c375a7ae8afb5f83b73070761ee0e742b348d298ffb94R27) (even though, strictly speaking, it would be closer to something like `STRING_NULL` or `STRING_EMPTY` in this case... guh)

just some food for thought, i personally would like to minimise the use of any magic numbers, imo it's a big help in reading code

Fexty12573:
I say define constants as needed. Of course check if there is already a constant defined that correlates to your number thematically. I tend to define constants very liberally.

Tho for something as generic and wide-reaching as a string terminator it might be worth briefly discussing it in the discord before throwing it into a PR.

lhearachel:
> btw: what is the projects' stance on new constants? eg here i define PAL_PARK_ENTRIES and try to make good use of it, but in https://github.com/pret/pokeplatinum/pull/578 i could define a similar constant for [here](https://github.com/pret/pokeplatinum/pull/578/files#diff-787ce0ab7f5bd125a397b94a55d848ee35d3989691635b067908341f4a421aa4R15) - the number 6 is actually used a lot in those systems because of the history pointer, otherwise i would agree it probably isn't super useful.

Generally, define constants if they are useful, well-scoped, and represent a value that needs to be synchronized across multiple locations. One of my guiding principles here is to ask myself: "Does this make it harder for a user to make a mistake?"

> another constant that would be genuinely useful (imo) is something like STRING_TERMINATOR (defined as 0xffff). there are other constants defined as 0xffff, but none of which thematically make sense to use eg [here](https://github.com/pret/pokeplatinum/pull/580/files#diff-476e34c09ec5eba6492c375a7ae8afb5f83b73070761ee0e742b348d298ffb94R27) (even though, strictly speaking, it would be closer to something like STRING_NULL or STRING_EMPTY in this case... guh)

There is one, actually: [`CHARCODE_EOS`](https://github.com/pret/pokeplatinum/blob/b96901ce5c591316c952d86200ebaba8dc9dd087/include/constants/charcode.h#L360). This `enum` is just a mapping of Game Freak's internal fixed-width encoding to semantic names. If you want to see what each character renders to on-screen, you can reference [the font glyphs used in-game](https://github.com/pret/pokeplatinum/blob/main/res/fonts/font_message.png), which is organized as a 16-column row-wise table.

ellieplayswow:
aha! i was looking in `includes/constants/string.h` for one and then blanket searching for  `0xffff` so i missed that :D

src/pal_pad.c
lhearachel: Fair point! You could go with `acquaintance`, but that doesn't fully convey the meaning that we want.

> a person whom one knows but who is not a particularly close friend 

But, I don't think the `History` names really convey it particularly well, either. This may just be a judgment call.

ellieplayswow: what about `associatedTrainer`? its fully your call though ^^

lhearachel: That seems fine to me! I would also include a small comment next to the field explaining the usage.

ellieplayswow: hopefully this is looking a bit better now? :D 

it's a confusing system in general, maybe the original devs knew we'd be having this discussion and made it look funky in the memory view :thinking:

src/pal_pad.c
```diff
@@ -0,0 +1,119 @@
...
+}
+
+PalPad *SaveData_GetPalPad(SaveData *saveData)
+{
+    return SaveData_SaveTable(saveData, SAVE_TABLE_ENTRY_PAL_PAD);
+}
+
+void PalPad_Init(PalPad *palPad)
+{
+    int i;
```
Fexty12573: Turn this into a loop variable

include/pal_pad.h
```diff
@@ -0,0 +1,77 @@
...
+u8 PalPad_GetTrainerRegionCode(const PalPad *palPad, int trainerIndex);
+
+/**
+ * @brief Compare two Pal Pads by their trainer name & ID
+ *
+ * @param first
+ * @param second
+ * @return TRUE if trainer name & ID matches, FALSE otherwise
+ */
+BOOL PalPad_TrainersEqual(const PalPad *first, const PalPad *second);
```
Fexty12573: `PalPad_AreTrainersEqual` (verbs for function names)

include/pal_pad.h
```diff
@@ -0,0 +1,77 @@
...
+    u8 padding_17;
+
+    // these are the 16 trainers inside the owners' pal pad
+    u32 associatedTrainerIds[PAL_PAD_ENTRIES];
+    u8 associatedTrainerGameCodes[PAL_PAD_ENTRIES];
+    u8 associatedTrainerRegionCodes[PAL_PAD_ENTRIES];
+    u8 associatedTrainerGenders[PAL_PAD_ENTRIES];
+} PalPad;
+
+int PalPad_SaveSize(void);
```
Fexty12573, line 27: `PalPad_GetSaveSize`

src/pal_pad.c
```diff
@@ -0,0 +1,119 @@
...
+            return TRUE;
+        }
+    }
+
+    return FALSE;
+}
+
+void PalPad_PushEntries(PalPad *destination, PalPad *source, int numberToCopy, int heapID)
+{
+    int i, j, newIndex;
```
Fexty12573: Can any of these be declared in the loops/where they're first used?

include/pal_pad.h
ellieplayswow: i don't necessarily disagree, but all the others are `*_SaveSize` - so should we rename them all? https://github.com/pret/pokeplatinum/blob/main/src/savedata/save_table.c#L58

src/pal_pad.c
ellieplayswow: `i` can't, the others can, have done what i can for t his

include/pal_pad.h
lhearachel: Yep, but out-of-scope for this PR

src/pal_pad.c
lhearachel: That happens sometimes; it's very likely a case where the compiler picked a specific register for a C89-style declaration (separate from its definition, at the beginning of scope) that does not match as a C99-style declaration (inline with its definition).

lhearachel (approved):

PR #584, opened 2025-06-29
Sync Heap functions with pokeheartgold

AsparagusEduardo:


AsparagusEduardo:
Updated description with the renames applied

PR #585, opened 2025-06-29
Sync Pokemon get/set data names with pokeheartgold

AsparagusEduardo:
Matching PR: https://github.com/pret/pokeheartgold/pull/385


AsparagusEduardo:
Closing, as its contents are also contained in #589 and since both affect a lot of files, might as well handle it in a single PR than 2.

PR #586, opened 2025-06-29
[WIP] document GTS & associated stuff

ellieplayswow:
gts: begin documentation
scrcmd: document PartyHasBadEgg function, CommonScript_HasBadEgg (0x2338)
text: unk_0671 -> gts
text: unk_0214 -> network_errors
text: general cleanup of some Message calls in scripts

this is very much a wip atm, so a lot of things are placeholders / bad names / etc. feel free to shout if anything is particularly egregious, otherwise i'll mark this as ready when i'm happy with it :)

main priority rn is to figure things out before naming them, so you'll see a lot of temp comments / `pl_*` / `chungus_silly_method_name_doesthis` as i chunk through things in a way that makes it more readable for me


Gudf:
Text bank unk_0695 also has a bunch of network-related strings, including errors, so `network_errors` might not be the best name for unk_0214?

It looks to me like unk_0695 is more widely used throughout the codebase: in overlays 97 (main menu, etc.), 98 (Wii Messages/email), 61 (VS Recorder Wi-Fi stuff), 96 (Battle Tower Wi-Fi stuff) and some Wi-Fi plaza overlays (overlays 65, 67 and 70), so I would suggest a less general name for unk_0214, maybe `gts_network`?

ellieplayswow:
unk_0214 is also used outside of the gts, it looks like actually it's a catch-all for any uncaught network errors based off of the networked area you're in , so i think maybe a better name is `uncaught_network_errors`?

Gudf:
Oh, you're right, I missed `error_message_reset.c`.
It looks like unk_0214 has strings for more severe errors ("You will be returned to the title screen", "You will be returned to wherever you last saved"...). unk_0695 also seems more specifically WFC related, whereas unk_0214 could also be local wireless? `network_errors` seems fine for unk_0214 then, and unk_0695 could be `nintendo_wfc` or `wifi_connection` or smth.

include/constants/map_object.h
```diff
@@ -53,4 +53,6 @@
 #define MAP_OBJ_UNK_A0_09 9
 #define MAP_OBJ_UNK_A0_10 10
 
+#define MAP_OBJ_PLAYER_AVATAR 0xff
```
lhearachel, line 56: **polish** / **quibble**: Prefer capitalized hex-digits, e.g., `0xFF`. (note the lowercase `x` delimiter

include/overlay094/const_ov94_02245FD8.h
```diff
@@ -1,6 +1,6 @@
 #ifndef POKEPLATINUM_CONST_OV94_02245FD8_H
 #define POKEPLATINUM_CONST_OV94_02245FD8_H
 
-extern const int Unk_ov94_02245FD8[];
+extern const int GTS_GENDER_PREFERENCE_MESSAGES[];
```
lhearachel, line 4: **suggestion**: The style for these sorts of constant data-tables in the repository is, e.g., `gGTSGenderPreferenceMessages`. The `g` prefix is used to denote that it is globally-visible (`extern`), and the rest of the name follows our typical `camelCase` convention from there (though it looks more like `PascalCase` due to the prefixed `g`).

Ideally, this declaration should also find a home in another header, like struct declarations / definitions. If there is no such header at present, then rename this header and update the include-guards as appropriate.

include/overlay094/gts_application_state.h
```diff
@@ -113,7 +113,9 @@ typedef struct UnkStruct_ov94_0223FD4C_t {
     Sprite *unk_F30;
     Sprite *unk_F34[8];
     Sprite *unk_F54[2];
-    Window unk_F5C;
+    // this window provides a running instruction at the bottom of the screen
+    // ex. https://www.youtube.com/watch?v=vH8U16wV7-E&t=94
+    Window bottomInstructionWindow;
```
lhearachel, line 118: **suggestion**: While nice, we typically don't leave comments like this in the codebase. This should just be the bottom message window, right? Would `messageWindow` be more appropriate?

include/overlay094/struct_ov94_0223FD4C_sub1.h
```diff
@@ -4,6 +4,6 @@
 typedef struct {
     int unk_00;
     int unk_04;
-} UnkStruct_ov94_0223FD4C_sub1;
+} GTSApplicationState_sub1;
```
lhearachel, line 7: **question**: Can these substructs be zipped into the same header as their parent struct?

include/struct_defs/gts_player_data.h
```diff
@@ -28,12 +28,12 @@ typedef struct {
...
     GameRecords *records;
-    JournalEntry *unk_2C;
+    JournalEntry *journalEntry;
     Bag *unk_30;
     int unk_34;
     int unk_38;
     int unk_3C;
     int unk_40;
-} UnkStruct_0203E0FC;
+} GTSPlayerData;
```
lhearachel, line 37: **style**: Apply this as a `struct` tag, as well. (`typedef struct GTSPlayerData ...`)

include/unk_020393C8.h
```diff
@@ -5,7 +5,7 @@
 #include "struct_defs/struct_020E5EB4.h"
 
 UnkStruct_020393C8 *sub_020393C8(u32 param0, u32 heapID, int param2, int param3, BOOL param4, const UnkStruct_020E5EB4 *param5[], int param6);
-void sub_02039434(UnkStruct_020393C8 *param0, int param1);
+void GTS_UnsureYet_SetNetworkStrength(UnkStruct_020393C8 *param0, int networkStrength);
```
lhearachel, line 8: **question**: Is `UnsureYet` meant to be a note for yourself?

res/field/scripts/scripts_unk_0213.s
```diff
@@ -92,14 +94,14 @@ _0108:
...
-    AddListMenuEntry 10, 4
-    AddListMenuEntry 11, 5
+    AddListMenuEntry pl_msg_00000361_00163, LIST_MENU_BUILDER_HEADER
+    AddListMenuEntry pl_msg_00000361_00006, 0
+    AddListMenuEntry pl_msg_00000361_00007, 1
+    AddListMenuEntry pl_msg_00000361_00008, 2
+    AddListMenuEntry pl_msg_00000361_00164, LIST_MENU_BUILDER_HEADER
+    AddListMenuEntry pl_msg_00000361_00009, 3
+    AddListMenuEntry pl_msg_00000361_00010, 4
+    AddListMenuEntry pl_msg_00000361_00011, 5
```
lhearachel, line 104: **suggestion**: Let's start documenting this text bank's IDs. Feel free to limit it to just these elements for now. :slightly_smiling_face:

src/overlay094/ov94_0223D0C4.c
```diff
@@ -280,10 +282,10 @@ static const int Unk_ov94_02245B94[][4] = {
     { 0x7, 0x5, 0x7, 0x2 }
 };
 
-static void ov94_0223D438(UnkStruct_ov94_0223FD4C *param0)
+static void ov94_0223D438(GTSApplicationState *param0)
```
lhearachel, line 285: **polish**: Document `param` names for functions where we have a sufficiently clear idea of their purpose.

lhearachel (commented):
Some preliminary feedback; I've yet to dig into the changes to files in `overlay094`.

include/overlay094/gts_application_state.h
ellieplayswow: yeah this is really more just a comment for myself that i'll remove & itll get squashed out of history after the fact, otherwise i have a habit of keeping a running text editor with all my notes that i forget to save somewhere when i shutdown. compared to some of the work with save data otoh where im prodding and probing with hex editors and a separate ide to confirm data & offsets its just a bit messier

i'm fine to rename it, i think actually another window occupies the same space with a different objective so i wanted to differentiate the 2 but i don't think i renamed that one yet

include/unk_020393C8.h
ellieplayswow: yep, its also why you'll see a few `ov94_gts_x` functions where i renamed but hadn't settled on a name yet :laughing:  (it'll get picked up when i revisit the function next, i just didnt have as much context as i did when i renamed it originally)

res/field/scripts/scripts_unk_0213.s
ellieplayswow: agreed, i tried to start just replacing various `Message` / `AddListMenuEntryX` calls with their appropriate ids so i could just grep and replace once ^^

ellieplayswow:
going to close this PR & reopen it when it's a little more cooked as in hindsight i think i jumped the gun a bit on making this

include/unk_020393C8.h
Gudf: It is kind of the network strength (or `link level` as it's called in the SDK), though it seems its (second) parameter is usually `WM_LINK_LEVEL_3 - (WM|DWC)_GetLinkLevel()`?
It's typically called via `sub_020397B0`, which is responsible for passing it some `static` state struct (that's the first parameter). 

For context, `unk_020393C8` is responsible for handling the wireless strength icons in general, both the one used for local wireless and the one used for Wi-Fi. (`res/prebuilt/data/pl_(wm|wifi).NCGR` and the associated palette.) It sets up the systask, loads the sprite, renders it, updates it, etc...

PR #587, opened 2025-06-29
document scrcmd_09a

scbroede:


src/scrcmd.c
```diff
@@ -922,7 +922,7 @@ const ScrCmdFunc Unk_020EAC58[] = {
     ScrCmd_GiveEgg,
     ScrCmd_098,
     ScrCmd_CheckPartyMonHasMove,
-    ScrCmd_09A,
+    ScrCmd_GetSlotOfPartyMonWithMove,
```
```suggestion
    ScrCmd_FindPartySlotWithMove,
```
This is to be more in line with other commands like `ScrCmd_FindPartySlotWithSpecies`. 
Remember to change the script macros accordingly too.

res/field/scripts/scripts_unk_0409.s
lhearachel, line 1: **suggestion**: This file looks like it's ready to finally be identified: It is the scripts used for execution of field moves. I can see, for example, a bunch of badge checks and prompts for messages.

Looking at `ScriptContext_LoadAndOffsetID` in `src/script_manager.c`, it maps to text bank `0383`. Let's go ahead and name these files while we're at it; feel free to give the scripts and text bank entries some nicer labels and such, too. 😁

scbroede: seems to be specifically hms (and flash); field moves like sweet scent, teleport, etc. seem to be elsewhere. would calling it `scripts_hm_field_moves` be okay despite the fact that flash isn't an hm anymore in this game? not sure how to work in flash without making the name look ridiculous.

lhearachel: I would just call them `scripts_field_moves`; `hms` would be misleading.

cbt6: But then as mentioned above there are also other field moves elsewhere - I suggest holding off renaming this for now and coming back to this when that "elsewhere" has been identified (besides this is out of scope for this PR)

lhearachel: We know where the other field moves are defined; they're in `field_move_task.c`.

But, fair.

scbroede: the scripts/text for sweet scent are in common and the others might not have dedicated scripts/text banks either, so with that in mind i guess calling these files `field_moves` works.

i have the changes ready locally and can push them here or in a separate pull request.

lhearachel: A separate PR sounds great! Let's close this one out.

PR #591, opened 2025-07-02
document field moves scripts and text

scbroede:


res/field/scripts/scripts_field_moves.s
```diff
@@ -1,43 +1,43 @@
...
-    ScriptEntry _0551
-    ScriptEntry _0704
-    ScriptEntry _060C
-    ScriptEntry _064C
+    ScriptEntry FieldMoves_CutTree
+    ScriptEntry FieldMoves_Rock
+    ScriptEntry FieldMoves_Boulder
+    ScriptEntry FieldMoves_RockyWall
+    ScriptEntry FieldMoves_Water
+    ScriptEntry FieldMoves_Fog_Unused
```
cbt6, line 11:
```suggestion
    ScriptEntry FieldMoves_Fog
```
And all other entries appended with `_Unused`

src/script_manager.c
```diff
@@ -216,7 +216,7 @@ static u16 ScriptContext_LoadAndOffsetID(FieldSystem *fieldSystem, ScriptContext
         ScriptContext_Load(fieldSystem, ctx, scripts_unk_1115, TEXT_BANK_UNK_0622);
         retScriptID -= 10100;
     } else if (retScriptID >= 10000) {
```
cbt6:
```suggestion
    } else if (retScriptID >= SCRIPT_ID_OFFSET_FIELD_MOVES) {
```

src/script_manager.c
```diff
@@ -216,7 +216,7 @@ static u16 ScriptContext_LoadAndOffsetID(FieldSystem *fieldSystem, ScriptContext
...
-        ScriptContext_Load(fieldSystem, ctx, scripts_unk_0409, TEXT_BANK_UNK_0381);
+        ScriptContext_Load(fieldSystem, ctx, scripts_field_moves, TEXT_BANK_FIELD_MOVES);
         retScriptID -= 10000;
```
cbt6:
```suggestion
        retScriptID -= SCRIPT_ID_OFFSET_FIELD_MOVES;
```

res/field/scripts/scripts_field_moves.s
scbroede: i mean, being able to interact with fog isn't actually a thing, is it? searching for `10005` also brings up nothing that suggests this script is used anywhere. then again i guess it might not be necessary to clarify that with the name

YakoSWG: Since I happen to have previously looked into this:
The script it runs also isn't functional (never calls the _DefogAnimation_ command). I suspect this just went unused.

lhearachel: It almost certainly did; you can't interact with fog via a script-activation in these games, it must be done from the party menu.

PR #592, opened 2025-07-02
Extract shareable graphics-related constants to a single header

lhearachel:
For full details, refer to related conversation [in the Discord channel](https://discord.com/channels/442462691542695948/1044260641382346793/1389700209524604938).

The essence here is to make these constants easy to share in a single uniform place. Fexty and I each need to share some of these values for our own PRs, so having them in this one header will be nice.


Fexty12573:
Looks good to me 👍

PR #593, opened 2025-07-02
Rename `ov97_2236380` and standardize using GBA as a prefix

Gudf:
Looking through undocumented overlay 97 files, I realized this one was already documented but not renamed.
Also decided to changes names to have GBA as a prefix to align with the rest of the GBA-related names in the overlay.

### DIFF
--- a/include/overlay097/gba_save.h
+++ b/include/overlay097/gba_save.h
@@ -77,12 +77,18 @@ typedef struct GBAPokemonSubstruct3 {
     u32 modernFatefulEncounter : 1;
 } GBAPokemonSubstruct3;
 
+#define NUM_SUBSTRUCT_BYTES (max(      \
+    max(sizeof(GBAPokemonSubstruct0),  \
+        sizeof(GBAPokemonSubstruct1)), \
+    max(sizeof(GBAPokemonSubstruct2),  \
+        sizeof(GBAPokemonSubstruct3))))
+
 union GBAPokemonSubstruct {
     GBAPokemonSubstruct0 type0;
     GBAPokemonSubstruct1 type1;
     GBAPokemonSubstruct2 type2;
     GBAPokemonSubstruct3 type3;
-    u16 raw[6];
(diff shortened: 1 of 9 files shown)


include/overlay097/gba_save.h
```diff
@@ -75,17 +75,17 @@ typedef struct {
...
-    PokemonGBASubstruct1 type1;
-    PokemonGBASubstruct2 type2;
-    PokemonGBASubstruct3 type3;
+union GBAPokemonSubstruct {
+    GBAPokemonSubstruct0 type0;
+    GBAPokemonSubstruct1 type1;
+    GBAPokemonSubstruct2 type2;
+    GBAPokemonSubstruct3 type3;
     u16 raw[6];
 };
```
lhearachel: **note**: Just leaving a remark here that these match [the names from `pokeemerald`](https://github.com/pret/pokeemerald/blob/aac44d0f915eb6366e99aeffe416e89e31785c8a/include/pokemon.h#L187-L194).

As a follow-up thought, why not compute the `6` value according to the `sizeof` these structs?

lhearachel (approved):
One thought below for final polish; the rest looks great. 😁

include/overlay097/gba_save.h
Gudf: Ah, good idea. Do we simply use the `sizeof` any of these, write some macro to find the max, or used this cursed thing since we can have anonymous unions?
```
union GBAPokemonSubstruct {
    union _GBAPokemonSubstruct {
        GBAPokemonSubstruct0 type0;
        GBAPokemonSubstruct1 type1;
        GBAPokemonSubstruct2 type2;
        GBAPokemonSubstruct3 type3;
    };
    u16 raw[sizeof(union _GBAPokemonSubstruct) / sizeof(u16)];
};
```

lhearachel: I would rather we have a `max` macro. We'll surely find a use-case for it elsewhere. 😁

lhearachel (approved):

PR #595, opened 2025-07-03
further document mining minigame

scbroede:
- not sure if this was a good way of handling the indices within the ug_parts narc, are there any examples i should follow?
- there were a fair amount of variables that are unused and ignored by the compiler that i removed here
- should the remaining values in `sMiningObjects` be converted to decimal? they're just probabilities and dimensions, so they might be more straightforward that way


src/overlay023/ov23_0223E140.c
```diff
@@ -1107,26 +1103,24 @@ int ov23_0223EBFC(void)
     return 2;
 }
 
-void ov23_0223EC00(int param0, int param1, void *param2, void *param3)
+void Mining_ReceiveTouchScreenInputFromConnectedPlayer(int param0, int param1, void *src, void *param3)
```
lhearachel: **suggestion**: `Mining_TakeLinkInput`

`Link` is the standard term we've been using for any sort of connected peer.

lhearachel (changes requested):
> not sure if this was a good way of handling the indices within the ug_parts narc, are there any examples i should follow?

These indices should not be hard-coded *nor* explicitly defined as constants within the repository. Existing indices are generated when we build the asset packs so that filenames in the asset tree have an implicit association to an index.

As you've built it now, it *works*, but it's not very flexible. I would rather that particular change-set wait until we unpack the corresponding NARC and rebuild it from source-files.

> there were a fair amount of variables that are unused and ignored by the compiler that i removed here

Yep, totally expected. Thanks for doing that! :heart: 

> should the remaining values in `sMiningObjects` be converted to decimal? they're just probabilities and dimensions, so they might be more straightforward that way

We typically just document / normalize these as we go. The exact representation typically depends on the use-case. Looking at this table, it seems like the `struct` definition implies that things should just all be in decimal, since they are just values. If you do modify the table to tidy it up, make sure to use C99-style designated initializers to make it a bit more readable and easier to modify: https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html

scbroede:
> If you do modify the table to tidy it up, make sure to use C99-style designated initializers to make it a bit more readable and easier to modify

done. also took the liberty of switching the order of `shape` and `itemID` because `itemID` seems like the most useful thing to list first, and listing `shape` next to the dimensions also makes sense imo

src/overlay023/ov23_0223E140.c
```diff
@@ -114,18 +113,18 @@ typedef struct {
     u8 width;
     u8 height;
     u8 itemID;
-    u16 unk_10;
-    u16 unk_12;
+    u16 spriteNARCIndex;
+    u16 paletteNARCIndex;
 } MiningObject;
 
 typedef struct {
```
lhearachel: **polish**: Tag structs with the same name as their `typedef`, e.g., for this one:

```c
typedef struct BuriedObject {
    // fields
} BuriedObject;

File: src/overlay023/ov23_0223E140.c [hunk 3]
```diff
@@ -484,91 +483,91 @@ static u8 sDampRockShape[3][4] = {
...
+    { .itemID = MINING_ROCK_5, .oddTIDWeight = 1, .evenTIDWeight = 1, .oddTIDNatDexWeight = 1, .evenTIDNatDexWeight = 1, .width = 3 * 2, .height = 2 * 2, .shape = sRockSShape, .spriteNARCIndex = 36, .paletteNARCIndex = 26 },
+    { .itemID = MINING_ROCK_5, .oddTIDWeight = 1, .evenTIDWeight = 1, .oddTIDNatDexWeight = 1, .evenTIDNatDexWeight = 1, .width = 2 * 2, .height = 3 * 2, .shape = sRockSShape90deg, .spriteNARCIndex = 37, .paletteNARCIndex = 26 },
+    { .itemID = MINING_ROCK_6, .oddTIDWeight = 1, .evenTIDWeight = 1, .oddTIDNatDexWeight = 1, .evenTIDNatDexWeight = 1, .width = 3 * 2, .height = 3 * 2, .shape = NULL, .spriteNARCIndex = 38, .paletteNARCIndex = 26 },
+    { .itemID = MINING_ROCK_7, .oddTIDWeight = 1, .evenTIDWeight = 1, .oddTIDNatDexWeight = 1, .evenTIDNatDexWeight = 1, .width = 2 * 2, .height = 4 * 2, .shape = NULL, .spriteNARCIndex = 39, .paletteNARCIndex = 26 },
+    { .itemID = MINING_ROCK_7, .oddTIDWeight = 1, .evenTIDWeight = 1, .oddTIDNatDexWeight = 1, .evenTIDNatDexWeight = 1, .width = 4 * 2, .height = 2 * 2, .shape = NULL, .spriteNARCIndex = 40, .paletteNARCIndex = 26 }
```
lhearachel, line 570: **question**: Are these multiplications valuable to keep? What information do they convey over pure constants?

src/overlay023/ov23_0223E140.c
```diff
@@ -581,27 +580,27 @@ static const WindowTemplate Unk_ov23_0225630E = {
...
-static void ov23_0223E140(void)
+static void Mining_InitGameState(void)
 {
     int i;
 
-    for (i = 0; i < (7 + 1); i++) {
-        Unk_ov23_02257740->unk_910[i] = 0xff;
-        Unk_ov23_02257740->unk_918[i] = 0xff;
+    for (i = 0; i < MAX_CONNECTED_PLAYERS; i++) {
+        Unk_ov23_02257740->linkTouchX[i] = TOUCHSCREEN_INPUT_NONE;
```
lhearachel, line 588: **question**: Is the static constant `Unk_ov23_02257740` here able to be documented?

src/overlay023/ov23_0223E140.c
scbroede: i believe so. these values are for display-related code and they get halved whenever used for actual logic. the halved versions also correspond with what a user would expect the dimensions to be both visually and from looking at the defined shapes. for example, sThunderstoneShape is 3x3, but the dimensions for a thunderstone in this table are 6x6

src/overlay023/ov23_0223E140.c
scbroede: kind of? the problem is that i can't think of any name for the UnkStruct_ov23_02257740 struct that wouldn't also fit for the UnkStruct_ov23_0223EE80 struct. otherwise i'd call it (and its associated constant) something along the lines of MiningEnv

lhearachel: That's fine. We can leave this for a future PR. 😁

PR #599, opened 2025-07-06
decompile ov110_021D1808

scbroede:


src/overlay110/ov110_021D0D80.c
```diff
@@ -621,205 +621,66 @@ static u32 ov110_021D17AC (UnkStruct_ov110_021D0F78 * param0, u8 param1)
...
+    }
+
+    ov110_021D1324(param0, &param0->unk_10[v0], v1, v2, 0, 1, 2, 0, FONT_SYSTEM, 0, 2);
+    Window_ScheduleCopyToVRAM(&param0->unk_10[v0]);
+    v0 = 1;
+    ov110_021D19B0(param0);
+    ov110_021D1324(param0, &param0->unk_10[v0], 39, 0, 0, 1, 2, 0, FONT_SYSTEM, 0, 0);
+    Window_ScheduleCopyToVRAM(&param0->unk_10[v0]);
+    v0 = 2;
+    ov110_021D13CC(param0, &param0->unk_10[v0], 29, 1);
```
cbt6, line 655: Is there a particular reason for doing this instead of simply `&param0->unk_10[2]`? 
As a side note I think it might be useful to name the windows in `UnkStruct_ov110_021D0F78` individually instead of having them in an array and then referencing them by hard-coded indexes.

scbroede: >Is there a particular reason for doing this instead of simply &param0->unk_10[2]?

no particular reason on my part; i'm just copying the way it's done in the other functions in the same file.

>As a side note I think it might be useful to name the windows in UnkStruct_ov110_021D0F78 individually instead of having them in an array and then referencing them by hard-coded indexes.

while they're mostly referenced that way, the array does get iterated through in the call to `ov110_021D2124` in line 133 of this file, for example

PR #601, opened 2025-07-06
decompile remaining functions in ov110_021D0D80.c

scbroede:


src/overlay110/ov110_021D0D80.c
```diff
@@ -697,235 +698,49 @@ static void ov110_021D19B0 (UnkStruct_ov110_021D0F78 * param0)
...
-    add r0, r5, #0
-    bl Window_ScheduleCopyToVRAM
-    add sp, #0x1c
-    pop {r3, r4, r5, r6, pc}
+    u8 v0;
+    int v1;
+    u32 v2;
+
+    v0 = 0;
+    Window_FillTilemap(&param0->unk_10[v0], 0);
```
cbt6:
```suggestion
    Window_FillTilemap(&param0->unk_10[0], 0);
```
Let's do away with the intermediate variables and use the indexes directly. Makes it easier to replace them with constants via search-and-replace down the road.

cbt6 (approved):

PR #610, opened 2025-07-10
Document PokeRadar_ClearIfAllOutOfView

CybernetCatgirl:
This one seems to build and I keep borking the prior one. HOPEFULLY this one works?

### DIFF
--- a/src/pokeradar.c
+++ b/src/pokeradar.c
@@ -262,7 +262,7 @@ const BOOL sub_02069798(const RadarChain *chain)
     return chain->unk_18;
 }
 
-void sub_0206979C(FieldSystem *fieldSystem)
+void PokeRadar_ClearIfAllOutOfView(FieldSystem *fieldSystem)
 {
     BOOL patchInView;
     GrassPatch *patch;
--- a/include/pokeradar.h
+++ b/include/pokeradar.h
@@ -26,7 +26,7 @@ int PokeRadar_ShouldDoRadarEncounter(const int param0, const int param1, FieldSy
 void SetRadarMon(RadarChain *chain, const int species, const int level);
 void GetRadarMon(RadarChain *chain, int *species, int *level);
 const BOOL sub_02069798(const RadarChain *chain);
-void sub_0206979C(FieldSystem *fieldSystem);
+void PokeRadar_ClearIfAllOutOfView(FieldSystem *fieldSystem);
 BOOL GetRadarChainActive(const RadarChain *chain);
 BOOL RefreshRadarChain(FieldTask *param0);
 void RadarChain_Increment(FieldSystem *fieldSystem);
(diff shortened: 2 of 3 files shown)


src/pokeradar.c
```diff
@@ -274,7 +274,7 @@ void sub_0206979C(FieldSystem *fieldSystem)
 
     for (patchRing = 0; patchRing < NUM_GRASS_PATCHES; patchRing++) {
         patch = &(fieldSystem->chain->patch[patchRing]);
-        v0 = GFXBoxTest_IsBoxAtPositionInView(&(patch->position), &(fieldSystem->chain->unk_BC));
+        v0 = GFXBoxTest_IsBoxAtPositionInView(&(patch->position), &(fieldSystem->chain->grassPatchVolume));
```
lhearachel: **polish**: Remove the unnecessary parentheses around either variable passed to the `&` operator:


```suggestion
        v0 = GFXBoxTest_IsBoxAtPositionInView(&patch->position, &fieldSystem->chain->grassPatchVolume);
```

---

**suggestion**: Name `v0`, if we have a good name in mind for it. Maybe `patchInView`?

src/pokeradar.c
```diff
@@ -49,7 +49,7 @@ typedef struct RadarChain {
     BOOL unk_14;
     BOOL unk_18;
     GrassPatch patch[NUM_GRASS_PATCHES];
-    GFXTestBox unk_BC;
+    GFXTestBox grassPatchVolume;
```
lhearachel, line 52: **question**: What is the `volume` portion of this name meant to convey?

lhearachel (commented):
Thanks for your contribution! I have one open question and two pieces of mini-feedback below. 😁

src/pokeradar.c
Gudf: Since I'm the one who suggested the name ([on Discord](https://discord.com/channels/442462691542695948/1044260641382346793/1392918451458412554)), I'm gonna chime in. The idea was that this is the volume occupied by a grass patch in world-space, that's then positioned at a grass patch's position to check for visibility.

lhearachel: Oh, it's the volume of the 3D object...? Interesting.

Gudf: Fwiw, I don't know whether it's the volume(/AABB) of an actual 3D object? But that's the idea yeah, it's a box capable of enclosing an entire grass patch/enough of a grass patch that if this box were visible, the grass patch would be as well.

CybernetCatgirl: I accepted the provided names. I also applied another suggestion I put in the Discord. However, I am still stuck on the "this looks like a function meant to verify that at least one patch is actually active". I [posted my thought process](https://discord.com/channels/442462691542695948/1044260641382346793/1393707369074262177) but am still a TOUCH unsure on it; I was hoping for feedback from people better at this than me.

PR #611, opened 2025-07-10
Partially document battle anims (overlay12)

Fexty12573:
So this documents the core battle anim system which is used for any and all pokemon move visuals.

The main file of this PR is `overlay12/battle_anim_system`. Various other files in overlay12 have been slightly documented to simplify documenting the main battle anim file. Additionally, I have added `constants/battle/battle_anim.h` which includes various constants related to the system. This is also the main file planned to be included in the battle anim scripts.

`overlay12/battle_anim_system` is still not 100% documented, but most of the script VM, script commands and auxillary functions are.

I will continue work on overlay12 in subsequent PRs but I wanted to make one now so it doesn't get out of hand. The main file I did here is already more than 4000 lines of code so arguably this one is already too big. My next PR will probably include unpacking most of the archives in `res/prebuilt/wazaeffect/`.


include/overlay012/funcptr_ov12_02239E64_1.h
```diff
@@ -1,11 +1,11 @@
...
 #define POKEPLATINUM_FUNCPTR_OV12_02239E64_1_H
 
-#include "overlay012/struct_ov12_0221FCDC_decl.h"
+#include "overlay012/battle_anim_system.h"
 
 #include "sprite.h"
 #include "sprite_system.h"
 
-typedef void (*UnkFuncPtr_ov12_02239E64_1)(UnkStruct_ov12_0221FCDC *, SpriteSystem *, SpriteManager *, Sprite *);
+typedef void (*UnkFuncPtr_ov12_02239E64_1)(BattleAnimSystem *, SpriteSystem *, SpriteManager *, Sprite *);
```
lhearachel: **question**: Should this be part of `battle_anim_system.h`?

src/overlay012/battle_anim_system.c
```diff
@@ -0,0 +1,4119 @@
...
+static void BattleAnimScriptCmd_CreateEmitterForMove(BattleAnimSystem *system)
+{
+    u32 resourceTable[6];
+
+    BattleAnimScript_Next(system);
+
+    u32 psIndex = (u32)BattleAnimScript_ReadWord(system->scriptPtr);
+    BattleAnimScript_Next(system);
+
+    for (int i = 0; i < 6; i++) {
```
lhearachel: **question**: Should this refer to the resource maximum in `sprite_system.h`? Or is this related to `spl_resource.h`?

include/overlay012/funcptr_ov12_02239E64_1.h
Fexty12573: It should realistically be in `ov12_02226998.c` because that is the only place it's "used" lol
image

src/overlay012/battle_anim_system.c
Fexty12573: Neither, if I tied it to a constant it would be `BATTLER_TYPE_ENEMY_SIDE_SLOT_2 + 1` Because each entry in that resource table corresponds to one of the `BATTLER_TYPE_*` constants. Could add a `BATTLER_TYPE_COUNT` to `constants/battle.h` maybe?
Actually. It's not even exactly that, it's more of a "which direction is the move going in". For example
Index 0 is any of:
- 1v1: Player attacking Enemy
- 2v2: Player1 attacking Enemy2
- 2v2: Player2 attacking Enemy1

Index 1 is:
- 2v2: Player2 attacking Enemy2
etc. 

(See the table in `BattleAnimSystem_GetParticleResIdxForMove`)

Fexty12573:
> Are you intending to write documentation similar to the 3D system for this once it feels complete?

Absolutely. I also plan to have one or two examples for adding new move animations.

include/overlay012/funcptr_ov12_02239E64_1.h
lhearachel: Does that match with the other function signature declared in `battle_anim_system.h`? The one that references `ManagedSprite` instead of `Sprite`.

src/overlay012/battle_anim_system.c
lhearachel: I think that we should create some kind of semantic relationship between these, if possible. Even if we don't directly use the constants in that table for the sake of alignment to keep the table tidy, having them for use in the accessing code would be beneficial.

Fexty12573: Wait are we talking about the `resourceTable` in `BattleAnimScriptCmd_CreateEmitterForMove` or the `indexTable` in `BattleAnimSystem_GetParticleResIdxForMove`?

Because for the former, I could imagine something along the lines of
```c
enum BattleAnimParticleResource {
    PARTICLE_RESOURCE_PL_PARALLEL_ATTACK = 0, // P>E, P1>E2, P2>E1
    PARTICLE_RESOURCE_PL_DIAGONAL_ATTACK_2, // P2>E2
    PARTICLE_RESOURCE_PL_DIAGONAL_ATTACK_1, // P1>E1
    PARTICLE_RESOURCE_EM_PARALLEL_ATTACK, // E>P, E1>P2, E2>P1
    PARTICLE_RESOURCE_EM_DIAGONAL_ATTACK_2, // E2>P2
    PARTICLE_RESOURCE_EM_DIAGONAL_ATTACK_1, // E1>P1
    PARTICLE_RESOURCE_COUNT,
};
```
These would serve as indices into `resourceTable`.

include/overlay012/funcptr_ov12_02239E64_1.h
Fexty12573: Yeah, could even be a `void*` for that matter. I'll turn it into an array of `BattleAnimScriptSpriteFunc` for now.

lhearachel (approved):
Rebase to resolve conflicts and this will be good to merge.

Also feel free to squash some history while rebasing, if you're so inclined.

PR #612, opened 2025-07-10
initial documentation of trainer card code

scbroede:
- ~~the stuff around the signature field is pretty half-baked at the moment, i only really touched it at this point because it was using an incorrect function name that i needed (edit: i think the other field in `UnkStruct_0202C834` is the cleanliness of the player's badges, so `TrainerCardSignature` is a bad name. maybe naming the functions `TrainerCard_SaveSize` and `TrainerCard_InitBadgeCleanliness` would be better?)~~ edit2: cleaned this up
- maybe the functions in the file for displaying trainer card text should be prefixed with `TrainerCardScreen` instead of `TrainerCard`?


include/overlay071/struct_ov71_0223B620.h
```diff
@@ -3,19 +3,33 @@
...
+    TRAINER_CARD_WINDOW_TIME,
+    TRAINER_CARD_WINDOW_ADVENTURE_STARTED,
+    TRAINER_CARD_WINDOW_HOF_DEBUT,
+    TRAINER_CARD_WINDOW_TIMES_LINKED,
+    TRAINER_CARD_WINDOW_LINK_BATTLES,
+    TRAINER_CARD_WINDOW_LINK_TRADES,
+    TRAINER_CARD_WINDOW_COUNT
+};
 
 typedef struct {
```
lhearachel, line 29: **question**: I assume that this is the application struct for the trainer card scene?

include/play_time.h
```diff
@@ -1,6 +1,11 @@
 #ifndef POKEPLATINUM_PLAYTIME_H
 #define POKEPLATINUM_PLAYTIME_H
 
+#define SECONDS_PER_MINUTE   60
+#define MINUTES_PER_HOUR     60
+#define PLAYTIME_MAX_HOURS   999
+#define PLAYTIME_MAX_MINUTES (MINUTES_PER_HOUR - 1)
+#define PLAYTIME_MAX_SECONDS (SECONDS_PER_MINUTE - 1)
 typedef struct PlayTime {
```
lhearachel: **polish**: Insert a blank line between this `define` and the `struct` declaration.

include/constants/heap.h
```diff
@@ -143,6 +143,7 @@ enum HeapSize {
 
     HEAP_SIZE_CHOOSE_STARTER_APP = 0x40000,
     HEAP_SIZE_LIBRARY_TV = 0x40000,
+    HEAP_SIZE_TRAINER_CARD_SCREEN = 0x28000,
```
lhearachel, line 146: **question**: Do we want to take this as an opportunity to bikeshed `SCREEN` vs `APP` for these heap IDs? I can see some potential confusion, maybe, with `POKETCH_APP`.

src/trainer_card.c
```diff
@@ -0,0 +1,281 @@
...
+} UnkStruct_02072204;
+
+static void sub_02072014(const u8 param0, const u8 gameVersion, const u8 stars, const u8 param3, const u8 regionCode, TrainerCard *trainerCard);
+static void TrainerCard_SetTrainerInfo(const u16 id, const u8 gender, const u16 *name, const u32 money, const u32 seenPokemon, const BOOL pokedexObtained, const u32 score, TrainerCard *trainerCard);
+static void TrainerCard_SetDates(const u8 gameCompleted, const PlayTime *playTime, const RTCDate *adventureStartedDate, const RTCDate *hofDebutDate, const RTCTime *hofDebutTime, const u8 param5, TrainerCard *trainerCard);
+static void TrainerCard_SetLinkDataAndSignature(const u32 param0, const u32 param1, const u32 param2, const u32 param3, const u8 *param4, TrainerCard *trainerCard);
+static void sub_0207216C(TrainerInfo *param0, FieldSystem *fieldSystem, TrainerCard *trainerCard);
+static BOOL sub_02072230(FieldTask *param0);
+
+void TrainerCard_Init(const u8 param0, const u8 param1, const u8 param2, const u8 param3, FieldSystem *fieldSystem, TrainerCard *trainerCard)
```
lhearachel: **polish**: Remove `const` from these fields; they're primitives, so they're necessarily copied for the call, and thus the caller has no need to care what happens to them internally.

src/trainer_card.c
```diff
@@ -0,0 +1,281 @@
...
+    } else {
+        trainerCard->playTime = NULL;
+    }
+}
+
+static void TrainerCard_SetLinkDataAndSignature(const u32 timesLinked, const u32 linkBattleWins, const u32 linkBattleLosses, const u32 linkTrades, const u8 *signature, TrainerCard *trainerCard)
+{
+    trainerCard->timesLinked = timesLinked;
+
+    if (trainerCard->timesLinked > 999999) {
```
lhearachel, line 191: **thought**: Do these deserve source-level constants?

src/overlay071/trainer_card_display_text.c
```diff
@@ -0,0 +1,284 @@
...
+
+#define TRAINER_CARD_WINDOW_PARTIAL_WIDTH 17
+#define TRAINER_CARD_WINDOW_FULL_WIDTH    28
+#define TRAINER_CARD_WINDOW_HEIGHT        2
+
+static void TrainerCard_DrawNumber(Window *window, const u32 windowWidth, const u32 endXOffset, const u32 yOffset, Strbuf *strbuf, const u32 num, const u8 maxDigits, enum PaddingMode paddingMode);
+static void TrainerCard_DrawString(Window *window, const u32 windowWidth, const u32 endXOffset, const u32 yOffset, const Strbuf *strbuf);
+
+static const WindowTemplate sTrainerCardWindowTemplates[] = {
+    [TRAINER_CARD_WINDOW_ID] = { .bgLayer = BG_LAYER_SUB_3, .tilemapLeft = 2, .tilemapTop = 4, .width = TRAINER_CARD_WINDOW_PARTIAL_WIDTH, .height = TRAINER_CARD_WINDOW_HEIGHT, .palette = 0xF, .baseTile = 0x6F },
```
lhearachel: **suggestion**: Lay these out vertically; you can force `clang-format` to respect that layout by tacking a `,` after the final struct-member (which is preferable anyways, as it makes for nicer diffs for users):


```suggestion
    [TRAINER_CARD_WINDOW_ID] = {
        .bgLayer = BG_LAYER_SUB_3,
        .tilemapLeft = 2,
        .tilemapTop = 4,
        .width = TRAINER_CARD_WINDOW_PARTIAL_WIDTH,
        .height = TRAINER_CARD_WINDOW_HEIGHT,
        .palette = 0xF,
        .baseTile = 0x6F,
    },
```

I would also set the `palette` fields to decimal, for clarity.

lhearachel (commented):
Nice work! I have a few quibbly comments below for polish.

include/overlay071/struct_ov71_0223B620.h
scbroede: i'm fairly sure it is. was going to wait until `ov71_0223B140.c` is further documented before going ahead with a rename, though

include/constants/heap.h
scbroede: my vote goes to screen, personally. in the context of the poketch i would think dowsing machine, calculator, etc. when i see "app"

src/trainer_card.c
scbroede: i believe these are only used here, for displaying these values to the user. the underlying records seem to have their own limits that don't necessarily match these. i personally don't think constants are necessary, but i don't feel all that strongly about it and am willing to add them if others feel differently

lhearachel: I think even just some constants local to this source file would be fine. No need to worry about putting them in global namespace.

include/constants/heap.h
lhearachel: Let's raise this as a discussion topic in the channel; I don't want to block this PR over it.

lhearachel (approved):

PR #617, opened 2025-07-12
Generate Dex Sizes and Polish Species Related Scripts

h2o-DS:
The equivalent of `include/constants/species.h` is now generated by `tools/scripts/make_shinzukan.py`
It is `res/pokemon/pokedex_sizes.h`

Several scripts were cleaned up to remove hard-coded dex sizes.

`pl_font.narc`'s ignore file was updated to reflect the new naix filetype.


res/fonts/pl_font.ignore
```diff
@@ -1,3 +1,3 @@
 font_special_chars.NCGR
 *.narc
-*.naix
```
lhearachel, line 3: **suggestion**: Let's keep this line just so this doesn't happen again.

res/pokemon/meson.build
```diff
@@ -283,6 +286,7 @@ shinzukan_narc = custom_target('shinzukan.narc',
         sinnoh_pokedex
     ]
 )
+pokedex_sizes = shinzukan_narc[1]
```
lhearachel: **issue**: Add this to the list of sources to the `main` target in the top-level `meson.build`.

---

**thought**: Would it be good to create a master array of these types of headers that can be appended to within each subdirectory as needed? e.g.,

```diff
main = executable('main',
    sources: [
        pokeplatinum_c,
        pokeplatinum_asm,
        c_consts_generators,
        naix_headers,
-       tutorable_moves_h,
-       species_learnsets_by_tutor_h,
-       species_footprints_h,
+       data_headers,
    ],

    # the rest...
)
```

include/constants/species.h
```diff
@@ -1,19 +0,0 @@
-#ifndef POKEPLATINUM_CONSTANTS_SPECIES_H
```
lhearachel: **thought**: Does it make sense to wholly delete this file? Maybe we could keep it as a meta-header that includes the generated files? It would also contract the size of this change-set quite a bit.

lhearachel (changes requested):
This looks good! A few comments below, one blocking-issue.

res/pokemon/meson.build
h2o-DS: That is done with naix_headers. Is there any reason not to?

lhearachel: I don't think so, and I think that it will be good to set the precedent now.

h2o-DS:
Addressed comments. Sorry for all the commits, I was failing to link here but not locally. The build dependencies should be settled now.

lhearachel (approved):

PR #619, opened 2025-07-12
Document Title Screen

Fexty12573:
Documents the title screen, which includes everything from the sky pillar portal animation up until the user presses start/A. to transition into the main menu.

In addition to the code, this PR also unpacks `res/prebuilt/demo/title/titledemo.narc` into `res/graphics/title_screen`.

The main file here is `applications/title_screen.c`. It is very self-contained (i.e. no headers).  The reason for the 100+ other changed files is because I renamed one of the `FADE_TYPE_` enum members as well as a few other constants.


res/graphics/meson.build
```diff
@@ -3,3 +3,4 @@ subdir('pokemon_summary_screen')
 subdir('shop_menu')
 subdir('signposts')
 subdir('windows')
+subdir('title_screen')
```
lhearachel: **polish**: Keep these directories sorted lexicographically.

res/graphics/title_screen/meson.build
```diff
@@ -0,0 +1,108 @@
+pngs_4bpp = files([
+    'bottom_screen_border.png',
+])
```
lhearachel: **polish**: Remove internal array-braces from `files` calls. `files` automatically coerces variadic arguments to an array.

res/graphics/title_screen/meson.build
```diff
@@ -0,0 +1,108 @@
...
+pals_explicit = files([
+    'gf_presents.pal',
+    'top_screen_border.pal',
+    'copyright.pal',
+])
+
+ncgrs_4bpp = ncgr_gen.process(pngs_4bpp,
+    extra_args: [ '-sopc', '-version101', ],
+)
```
lhearachel, line 24: **thought**: It might make sense to create a set of common generators with various typical argument-combinations. But maybe that's over-engineering...

res/graphics/title_screen/meson.build
```diff
@@ -0,0 +1,108 @@
...
+
+tilemaps = copy_gen.process(files([
+    'bottom_screen_border.NSCR',
+    'bottom_screen_border_2.NSCR',
+    'top_screen_border.NSCR',
+    'top_screen_border_2.NSCR',
+    'copyright.NSCR',
+    'gf_presents.NSCR',
+    'logo.NSCR',
+    'logo_jp.NSCR'
```
lhearachel: **polish**: Trailing commas are idiomatic in `meson`'s grammar, and they make diffs slightly-nicer for end-users.

res/graphics/title_screen/meson.build
```diff
@@ -0,0 +1,108 @@
...
+]))
+
+models = copy_gen.process(files([
+    'giratina.nsbmd',
+    'giratina_face.nsbmd',
+    'giratina_portal.nsbmd'
+]))
```
lhearachel: **thought**: I've been wondering for a while if the `copy_gen` thing makes sense to keep around. [`fs.copyfile`](https://mesonbuild.com/Fs-module.html#copyfile) will produce a `custom_target`, which should function basically identically. 

Would you be up to try out using it in this PR? If it goes well and we can make the semantics feel nice, then we can open a separate issue to update all of the existing instances and remove `copy_gen` (which would also remove a `PATH`-scan for `cp`, which is quite slow).

src/unk_0209A2C4.c
```diff
@@ -259,7 +259,7 @@ static BOOL sub_0209A544(UnkStruct_0209A3D0 *param0)
     case 2:
         Bg_MaskPalette(BG_LAYER_MAIN_0, 0x6c21);
         Bg_MaskPalette(BG_LAYER_SUB_0, 0x6c21);
-        StartScreenFade(FADE_BOTH_SCREENS, FADE_TYPE_UNK_1, FADE_TYPE_UNK_1, FADE_TO_BLACK, 6, 1, param0->heapID);
+        StartScreenFade(FADE_BOTH_SCREENS, FADE_TYPE_BRIGHTNESS_IN, FADE_TYPE_BRIGHTNESS_IN, FADE_TO_BLACK, 6, 1, param0->heapID);
```
lhearachel: **question**: Just tacking onto this line: Do we want to replace `FADE_TO_BLACK` / `FADE_TO_WHITE` with `COLOR_BLACK` and `COLOR_WHITE`?

We would also need to decide on a new name for `FADE_TO_SAVED_COLOR`. Maybe just `COLOR_SAVED`?

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
+#include "nitro/gx/g3.h"
+#include "nitro/types.h"
+#include <nitro.h>
+#include <string.h>
```
lhearachel: **polish**: Use angle-brackets for SDK and compiler headers, and remove the unused ones (probably `nitro.h` and `string.h` here?).

platinum.us/main.lsf
```diff
@@ -1289,7 +1289,7 @@ Overlay overlay76
 Overlay overlay77
```
lhearachel: **question**: Can we name this overlay? I assume that the other files are related to the opening cutscene. `game_opening` comes to mind, but doesn't necessarily imply the title screen.

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+
+#include "res/graphics/title_screen/titledemo.naix.h"
+
+FS_EXTERN_OVERLAY(overlay77);
+FS_EXTERN_OVERLAY(overlay89);
+FS_EXTERN_OVERLAY(overlay97);
+FS_EXTERN_OVERLAY(d_startmenu);
+
+#define KEY_PRESSED(key)     ((JOY_NEW(key)) == (key))
+#define KEY_HELD(key)        ((JOY_HELD(key)) == (key))
```
lhearachel: **suggestion**: Move these to `system.h` and update names to reflect direct equality:


```suggestion
#define JOY_NEW_ONLY(key)     (JOY_NEW(key) == (key))
#define JOY_HELD_ONLY(key)    (JOY_HELD(key) == (key))
```

The parens around `JOY_NEW` and `JOY_HELD` should be unnecessary, as `JOY_NEW` and `JOY_HELD` already encase themselves in parens.

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+#define LIGHT_COLOR(r, g, b) ((((r) << 0) & GX_RGB_R_MASK) | (((g) << 5) & GX_RGB_G_MASK) | (((b) << 10) & GX_RGB_B_MASK))
+#define ANGLE(angle)         FX_DEG_TO_IDX(FX32_CONST(angle))
+
+#define TITLE_SCREEN_LAYER_GIRATINA    BG_LAYER_MAIN_0
+#define TITLE_SCREEN_LAYER_COPYRIGHT   BG_LAYER_MAIN_1
+#define TITLE_SCREEN_LAYER_GIRATINA_BG BG_LAYER_MAIN_3
+#define TITLE_SCREEN_LAYER_PRESS_START BG_LAYER_SUB_0
+#define TITLE_SCREEN_LAYER_LOGO_BG_2   BG_LAYER_SUB_1
+#define TITLE_SCREEN_LAYER_LOGO        BG_LAYER_SUB_2
+#define TITLE_SCREEN_LAYER_LOGO_BG     BG_LAYER_SUB_3
```
lhearachel, line 55: **thought**: Interesting! Maybe we want to take this approach for other layers. I've definitely used similar approaches when building features against, e.g., `pokeemerald`, just because the semantics are much nicer, IMO.

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+#include "res/graphics/title_screen/titledemo.naix.h"
+
+FS_EXTERN_OVERLAY(overlay77);
+FS_EXTERN_OVERLAY(overlay89);
+FS_EXTERN_OVERLAY(overlay97);
+FS_EXTERN_OVERLAY(d_startmenu);
+
+#define KEY_PRESSED(key)     ((JOY_NEW(key)) == (key))
+#define KEY_HELD(key)        ((JOY_HELD(key)) == (key))
+#define LIGHT_COLOR(r, g, b) ((((r) << 0) & GX_RGB_R_MASK) | (((g) << 5) & GX_RGB_G_MASK) | (((b) << 10) & GX_RGB_B_MASK))
```
lhearachel, line 46: **thought**: This one feels a bit awkward, as it's not really *that* much different from this macro in `palette.h`:

```c
#define RGB(r, g, b)   (((b) << 10) | ((g) << 5) | (r))
```

But it's different enough that I assume that it affects codegen. 🙁

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+    int unused2;
+    int titleScreenTimer; // How long the title screen has been shown for
+    int exitFadeTimer;
+} TitleScreenAppData;
+
+extern const ApplicationManagerTemplate Unk_020F8A48;
+extern const ApplicationManagerTemplate gStartMenuAppTemplate;
+extern const ApplicationManagerTemplate gOpeningCutsceneAppTemplate;
+
+void EnqueueApplication(FSOverlayID overlayID, const ApplicationManagerTemplate *template);
```
lhearachel: **polish**: Remove; this is covered by `#include "main.h"`.

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+static BOOL TitleScreen_Main(ApplicationManager *appMan, int *state)
+{
+    TitleScreenAppData *appData = ApplicationManager_Data(appMan);
+
+    switch (*state) {
+    case TITLE_SCREEN_APP_STATE_INIT_RESOURCES:
+        if (TitleScreen_LoadGfx(&appData->titleScreen, appData->bgConfig, appData->heapID) == TRUE) {
+            appData->titleScreen.state = INTRO_STATE_FADE_FROM_BLACK;
+
+            if (!gSystem.unk_6C) {
```
lhearachel: **question**: Could we name this from the surrounding context? Feel free to make that a separate PR, if it's a big task.

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+
+        if (KEY_HELD(PAD_BUTTON_B | PAD_KEY_UP | PAD_BUTTON_SELECT)) {
+            appData->nextApp = 2;
+            StartScreenFade(FADE_BOTH_SCREENS, FADE_TYPE_BRIGHTNESS_OUT, FADE_TYPE_BRIGHTNESS_OUT, FADE_TO_BLACK, 6, 1, appData->heapID);
+            *state = TITLE_SCREEN_APP_STATE_CLEANUP;
+            break;
+        }
+
+        if (appData->titleScreenTimer > TITLE_SCREEN_REPLAY_OPENING_FRAMES) {
+            appData->nextApp = 3;
```
lhearachel: **polish**: `NEXT_APP_REPLAY_OPENING`.

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+            appData->nextApp = NEXT_APP_START_MENU;
+            Sound_FadeOutBGM(0, 60);
+            Sound_PlayPokemonCry(SPECIES_GIRATINA, 1);
+            TitleScreen_ShowBlurEffect(appData->bgConfig, appData->heapID);
+            *state = TITLE_SCREEN_APP_STATE_EXIT_NORMAL;
+            break;
+        }
+
+        if (KEY_HELD(PAD_BUTTON_B | PAD_KEY_UP | PAD_BUTTON_SELECT)) {
+            appData->nextApp = 2;
```
lhearachel: **suggestion**: This would be `NEXT_APP_2`, but based on the `KEY_HELD` combination above, this should be `NEXT_APP_CLEAR_SAVE_FILE` (or similar).

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+
+static void TitleScreen_VBlank(void *param)
+{
+    TitleScreenAppData *appData = param;
+    Bg_RunScheduledUpdates(appData->bgConfig);
+}
+
+static void TitleScreen_SetVRAMBanks(void)
+{
+    UnkStruct_02099F80 banks = {
```
lhearachel, line 491: **thought**: We should probably name this struct, at this point. I feel like we have a solid grasp on what it does, and it seems like it's just an SDK wrapper...?

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+    GXLayers_SetBanks(&banks);
+}
+
+static BOOL TitleScreen_ShouldSkipIntro(void)
+{
+    if (KEY_PRESSED(PAD_BUTTON_A) || KEY_PRESSED(PAD_BUTTON_START) || KEY_PRESSED(PAD_BUTTON_SELECT)) {
+        return TRUE;
+    }
+
+    return FALSE;
```
lhearachel: **question**: Does this match as:


```suggestion
    return KEY_PRESSED(PAD_BUTTON_A) || KEY_PRESSED(PAD_BUTTON_START) || KEY_PRESSED(PAD_BUTTON_SELECT);
```

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+        Camera_ComputeProjectionMatrix(CAMERA_PROJECTION_PERSPECTIVE, gfx->titleCamera);
+        Camera_SetAsActive(gfx->titleCamera);
+    }
+
+    titleScreen->giratinaHoverAngle += 2;
+    titleScreen->giratinaHoverAngle %= 360;
+
+    // BUG: This should pass the value in directly, since the function takes degrees as input, not an index.
+    // But in practice this just slightly reduces the range of the hover so it isn't very noticeable.
+    fx32 offset = CalcSineDegrees_Wraparound((titleScreen->giratinaHoverAngle * 0xFFFF) / 360);
```
lhearachel, line 656: **docs**: Potentially worth documenting in `bugs_and_glitches`?

src/applications/title_screen.c
```diff
@@ -0,0 +1,1443 @@
...
+    G2_BlendNone();
+    G3X_EdgeMarking(FALSE);
+
+    gSystem.whichScreenIs3D = DS_SCREEN_MAIN;
+    GXLayers_SwapDisplay();
+
+    return TRUE;
+}
+
+static void TitleScreen_Load2DGfx(BgConfig *bgConfig, enum HeapId heapID, TitleScreen *param2)
```
lhearachel: **polish**: Label `param2`.

lhearachel (commented):
Great work! Most of my feedback below are questions and polish, though there are some leading-thoughts that may create additional PRs. Up to you how much of it you want to address within this PR or leave for later cleanup to limit scope.

src/unk_0209A2C4.c
```diff
@@ -58,7 +58,7 @@ static const WindowTemplate Unk_020F8A58 = {
     0x16D
 };
 
-const ApplicationManagerTemplate Unk_020F8AB4 = {
+const ApplicationManagerTemplate gStartMenuAppTemplate = {
```
Gudf: **thought**: this feels like it could be confused for the overworld menu, since we decided to name that the start menu. I'm also not quite sure I understand what this is referring to?

res/graphics/title_screen/meson.build
Fexty12573: I think I will leave that to someone else more familiar with nitrogfx and 2D graphics files. 😅

res/graphics/title_screen/meson.build
Fexty12573: How would I do that exactly? I'm not very familiar with meson. Simply replacing the `copy_gen.process` calls with `fs.copyfile` calls doesn't work. And wouldn't switching to that result in having to specify the target directory everywhere as well?

src/unk_0209A2C4.c
Fexty12573: YES. PLEASE. I mentioned this on discord earlier today but I don't think anyone saw it lol. It would be much better to use `COLOR_BLACK`/`COLOR_WHITE` etc. My reasoning is in [this discord message](https://discord.com/channels/442462691542695948/1044260641382346793/1393560309842837564).

And yeah, I think `COLOR_SAVED` works just fine.

res/graphics/title_screen/meson.build
lhearachel: My thought here is that we could do it like this:

```
files_to_copy = files(
    'giratina.nsbmd',
    'giratina_face.nsbmd',
    'giratina_portal.nsbmd',
    # plus other files that are directly copied using `copy_gen`
)

copied_files = []
foreach source_file : files_to_copy
    # `fs.copyfile` takes a single source file and produces a single
    # `custom_target`. So we must call it iteratively.
    #
    # `dest` is an optional argument; when unspecified, it uses the basename of
    # the input file.
    copied_files += fs.copyfile(source_file)
endforeach
```

Then, you would replace the `models`, etc. variables in the final target with `copied_files`.

platinum.us/main.lsf
Fexty12573: I'm not 100% sure if all files in ov77 are actually title screen but it seems pretty likely so I'm not opposed.

lhearachel: At least some of them are related to the opening cutscene. One is probably related to the save-wipe screen.

src/unk_0209A2C4.c
lhearachel: Feel free to do that in this PR, then; I am in-favor. :grin:

src/applications/title_screen.c
Fexty12573: Neither `RGB` nor `GX_RGB` match unfortunately due to the masks.

src/applications/title_screen.c
Fexty12573: Yes it is. I would suggest something like `VRAMBankSpec` or `VRAMBankConfig`. It's used exclusively as a parameter to `GXLayers_SetBanks` so I don't think it's too far fetched.

src/applications/title_screen.c
Fexty12573: I didn't know if this was significant enough to warrant putting it in there but I'm down

src/applications/title_screen.c
Fexty12573: Weirdly enough it doesn't

src/unk_0209A2C4.c
Fexty12573: Whoops I was kind of on autopilot when labelling this. I was under the impression it went straight to the main menu after this but it looks like there's another application before that which does some savedata checks. I'll just revert it to the `Unk_*` name.

res/graphics/title_screen/meson.build
Fexty12573: This copies the files into `build/res/graphics/title_screen`, how can I make it copy them into `build/res/graphics/title_screen/titledemo.narc.p`?

src/applications/title_screen.c
Fexty12573: I find it much easier to reason about what is happening if the layers are given a name. Especially in a situation like this where main and sub are also flipped. Can recommend 👍

src/applications/title_screen.c
Fexty12573: It's ONLY use seems to be to skip the title screen intro when set to `FALSE`. But I find it so weird that it is part of `gSystem`. So I've been hesitant to name it. But if I had to I'd say `showTitleScreenIntro` would be a reasonable name.

PR #624, opened 2025-07-13
Document G3D Pipeline State

Fexty12573:
Documents the G3D Pipeline State (called `GenericPointerData` before).

I also "documented" all the calls to `G3DPipelineState_New`, i.e. replaced numbers with enum values.


src/g3d_pipeline_state.c
```diff
@@ -0,0 +1,71 @@
...
+
+    G2_SetBG0Priority(0);
+    G3X_SetShading(GX_SHADING_TOON);
+    G3X_AntiAlias(FALSE);
+    G3X_AlphaTest(FALSE, 0);
+    G3X_AlphaBlend(FALSE);
+    G3X_EdgeMarking(FALSE);
+    G3X_SetFog(FALSE, GX_FOGBLEND_COLOR_ALPHA, GX_FOGSLOPE_0x8000, 0);
+    G3X_SetClearColor(COLOR_BLACK, 0, 0x7FFF, 63, FALSE);
+    G3_ViewPort(0, 0, 255, 191);
```
Gudf: **suggestion**: use the screen size constants
```suggestion
    G3_ViewPort(0, 0, HW_LCD_WIDTH - 1, HW_LCD_HEIGHT - 1); // Inclusive bounds
```

src/g3d_pipeline_state.c
```diff
@@ -0,0 +1,71 @@
...
+    GXLayers_EngineAToggleLayers(GX_PLANEMASK_BG0, TRUE);
+
+    G2_SetBG0Priority(0);
+    G3X_SetShading(GX_SHADING_TOON);
+    G3X_AntiAlias(FALSE);
+    G3X_AlphaTest(FALSE, 0);
+    G3X_AlphaBlend(FALSE);
+    G3X_EdgeMarking(FALSE);
+    G3X_SetFog(FALSE, GX_FOGBLEND_COLOR_ALPHA, GX_FOGSLOPE_0x8000, 0);
+    G3X_SetClearColor(COLOR_BLACK, 0, 0x7FFF, 63, FALSE);
```
Gudf: **question**: I noticed `easy3d.c` has a `G3X_DEPTH_MAX` constant whose value is also 0x7FFF, should it maybe be moved to some header and used here as well?

include/g3d_pipeline_state.h
```diff
@@ -0,0 +1,43 @@
+#ifndef POKEPLATINUM_G3D_PIPELINE_STATE_H
+#define POKEPLATINUM_G3D_PIPELINE_STATE_H
+
+#include "constants/heap.h"
+
+typedef struct G3DPipelineState {
```
Gudf: **note**: I'm not sure about the `State` suffix, as this struct doesn't really seem to store much state at all.

IIRC, the 3D pipeline is quite OGL-like, with all the actual state stored on the GPU equivalent and commands you can use to affect this state, but you don't ever really access it directly.

Gudf (commented):
Not much to say here for me, just a note regarding the name and a few suggestions.

include/g3d_pipeline_state.h
Fexty12573: Disagree. While yes, the struct itself doesn't actually store any state, it represents it. As that is the only way it is used. Calling it what it represents makes much more sense than something like `VramManagerBuffers`, which is entirely meaningless besides describing what is actually inside of the struct.

And yes, the 3D engine is very similar to legacy opengl.

src/g3d_pipeline_state.c
Fexty12573: Right, I forgot I had that. I'll move it to `constants/graphics.h`

include/g3d_pipeline_state.h
Gudf: If it *did* represent the pipeline state, surely you'd have to pass it to every state-modifying function, which you obviously don't; this struct is only ever constructed and freed.

What it *is* is the buffers used by the 3D pipeline, so you could name it `G3DPipelineBuffers`, and if you think it represents more than that, why not simply call it `G3DPipeline`? It is the entire pipeline as much as it is only the pipeline's state, after all.

include/g3d_pipeline.h
```diff
@@ -0,0 +1,43 @@
+#ifndef POKEPLATINUM_G3D_PIPELINE_STATE_H
```
lhearachel: **polish**: Update the include guard.

PR #635, opened 2025-07-16
Move Egg Moves and Icon Palettes to Species Data

h2o-DS:
Moves egg moves and icon palette offsets from header files to species data files. The two files are now generated by `tools/datagen/datagen_species.cpp`

`res/pokemon/species_egg_moves.h`
`res/pokemon/species_icon_palettes.h`

`datagen_species.cpp` is now provided all forms and parses form types.


res/pokemon/shellos/forms/east_sea/data.json
```diff
@@ -0,0 +1,3 @@
+{
+    "icon_palette": 0
+}
```
lhearachel: **thought**: How much do we expect to expand this kind of data file? If it's not very much, then would it make more sense to make this field an `int | array[int]`, where each index of the array is a palette for that form? I think that a `dict` could make sense to account for mapping to some key? Maybe to the key in `form_registry.json`?

tools/datagen/datagen_species.cpp
```diff
@@ -363,6 +367,64 @@ static void TryEmitTutorableLearnset(rapidjson::Document &root, std::ofstream &o
...
 
+static void TryEmitEggMoves(rapidjson::Document &root, std::ofstream &ofs, std::string species)
+{
+    const rapidjson::Value &learnsets = root["learnset"];
+    if (!learnsets.HasMember("egg_moves")) {
+        return;
+    }
+
+    std::transform(species.begin(), species.end(), species.begin(), ::toupper);
+    ofs << "    EGG_MOVES(SPECIES_"
```
lhearachel: **thought**: Is it worth keeping this macro if we are just generating this header?

tools/datagen/datagen_species.cpp
```diff
@@ -512,7 +601,27 @@ int main(int argc, char **argv)
...
+                 << " * @brief This table defines which index of the shared palette file is used\n"
+                 << " * for the icons used by Pokemon in PC boxes, the party list, etc. Alternate\n"
+                 << " * forms are listed after the full National Dex using a custom ordering.\n"
+                 << " */\n"
+                 << "const u8 sPokemonIconPaletteIndex[] = {\n";
+
     u16 personalValue = 0;
+    BOOL unownDupe = FALSE;
     rapidjson::Document doc;
     for (auto &species : speciesRegistry) {
```
lhearachel: **question**: For iterating over the forms, could we just read `form_registry.json` directly and manipulate it into the various orderings in-memory?

h2o-DS: I see that line of thought and it would help with Unown. If we want to process that data here anyway would we want to generate the order files in species_datagen as well?

lhearachel: Maybe...? I don't have an issue with that, I think. It shouldn't be too hard to adapt. But I do wonder if we might be putting too much weight onto this particular driver.

res/pokemon/shellos/forms/east_sea/data.json
h2o-DS: I decided to leave `icon_palette` in the "data" forms' `data.json` but implement the array for "icon" forms. This eliminates the data files with only a single field and minimizes bloat on the base forms.

lhearachel (approved):

PR #638, opened 2025-07-16
Remove dummy from BgTemplate

ljtpetersen:
And update **all** references.

note: I could not have done this without neovim. I wrote a lua script that automatically converted the bare struct declaration to have designated initializers and removed the dummy and converted the screen size to the enum value if necessary.


lhearachel (approved):
LGTM; thanks for doing this! :pray:

PR #641, opened 2025-07-17
document trainer messages

scbroede:


include/constants/trainer.h
```diff
@@ -114,10 +114,27 @@ enum TrainerClass {
 };
 
 enum TrainerMessageType {
```
lhearachel: **issue**: Remove this direct enum definition in favor of `generated/trainer_message_types.h`.

res/text/npc_trainer_messages.gmm
lhearachel, line 1: **praise**: I was not familiar with your game. Thanks for identifying all of these!

src/battle/ov16_0223DF00.c
```diff
@@ -1719,17 +1719,17 @@ void Battle_SetDefaultBlend(void)
     G2_BlendNone();
 }
 
-u8 ov16_0223F9FC(BattleSystem *battleSystem, int param1, int param2, int param3, int param4)
+u8 ov16_0223F9FC(BattleSystem *battleSystem, int trainerID, int param2, int msgType, int param4)
```
lhearachel: **question**: Should `int msgType` reference the generated enum type?

src/unk_02048614.c
```diff
@@ -29,14 +30,14 @@ BOOL ScrCmd_0B6(ScriptContext *param0);
...
-BOOL ScrCmd_0E7(ScriptContext *param0);
-BOOL ScrCmd_0E8(ScriptContext *param0);
-BOOL ScrCmd_0E9(ScriptContext *param0);
-BOOL ScrCmd_BlackOutFromBattle(ScriptContext *param0);
-BOOL ScrCmd_CheckHasTwoAliveMons(ScriptContext *param0);
+BOOL ScrCmd_GetTrainerMessageTypes(ScriptContext *ctx);
+BOOL ScrCmd_GetTrainerRematchMessageTypes(ScriptContext *ctx);
+BOOL ScrCmd_CheckIsTrainerDoubleBattle(ScriptContext *ctx);
+BOOL ScrCmd_BlackOutFromBattle(ScriptContext *ctx);
+BOOL ScrCmd_CheckHasTwoAliveMons(ScriptContext *ctx);
```
lhearachel: **question**: Do these come from the header file?

src/unk_02048614.c
```diff
@@ -176,83 +177,83 @@ BOOL ScrCmd_StartTagBattle(ScriptContext *ctx)
...
+            notEnoughPokemonMessageType = TRMSG_DOUBLE_BATTLE_NOT_ENOUGH_POKEMON_2;
         }
     }
 
-    *v6 = v2;
-    *v7 = v3;
-    *v8 = v4;
+    *destVar0 = preBattleMsgType;
+    *destVar1 = postBattleMsgType;
+    *destVar2 = notEnoughPokemonMessageType;
```
lhearachel: **suggestion**: Can we rename the arguments here and in the macro to reflect these assignments?

lhearachel (changes requested):

include/constants/trainer.h
scbroede: should the trainer class enum in this file also be removed in favour of its generated file then?

res/text/npc_trainer_messages.gmm
scbroede: thanks, i wrote a script to copy all this info from `trtbl_0.bin` because doing it manually would have taken forever

src/unk_02048614.c
scbroede: yes, i'm not sure why they're redeclared here. should i remove them?

lhearachel: Yes, please. :heart:

include/constants/trainer.h
lhearachel: Yes.

res/text/npc_trainer_messages.gmm
cbt6, line 1: We could probably store the text strings for each trainer individually in `res/trainers/data` and then have this text bank generated from there. But this could be explored in a separate PR.

cbt6 (approved):
Do you intend to unpack `poketool/trmsg/trtbl.narc` / `poketool/trmsg/trtblofs.narc` next? Now that the message types are documented.

scbroede:
>Do you intend to unpack poketool/trmsg/trtbl.narc / poketool/trmsg/trtblofs.narc next?

not immediately, but i do plan on doing that at some point

res/text/npc_trainer_messages.gmm
scbroede: i think this could be a good thing to do along with unpacking those two narcs (the data from trtbl can be put in `res/trainers/data` along with the actual corresponding strings)

PR #645, opened 2025-07-21
address some todos in pokemon.c

scbroede:


src/pokemon.c
```diff
@@ -59,34 +61,182 @@
...
-    { 0x0, 0x1, 0x0, 0x0, -0x1 },
-    { 0x0, 0x1, -0x1, 0x0, 0x0 },
-    { 0x0, 0x0, 0x0, 0x0, 0x0 },
-    { 0x0, 0x1, 0x0, -0x1, 0x0 },
-    { -0x1, 0x0, 0x0, 0x1, 0x0 },
-    { 0x0, 0x0, 0x0, 0x1, -0x1 },
-    { 0x0, 0x0, -0x1, 0x1, 0x0 },
-    { 0x0, -0x1, 0x0, 0x1, 0x0 },
-    { 0x0, 0x0, 0x0, 0x0, 0x0 },
+    [NATURE_HARDY] = {
```
cbt6, line 65: Format the table similarly to that in `BattleAnimSystem_GetParticleResIdxForMove`

scbroede: is there a particular reason you want that? imo the added clarity from filling this with designated initializers is worth the verbosity

cbt6: Not really, no. I thought it might look better when compact but it doesn't really matter either way.

PR #646, opened 2025-07-22
Set brace wrapping after function to false

AsparagusEduardo:
Potential PR in case functions are decided to have the following format across the gen 4 repos:
```
ret Function(args) {
    // code
}
```
Should this format not be accepted, the opposite PR should be accepted instead: https://github.com/pret/pokeheartgold/pull/394, which sets their braces to pokeplatinum's format.


cbt6:
Closing this, as the Allman style is the already established convention to adopt for opening braces of functions in pokeplatinum.

PR #647, opened 2025-07-24
more trainer card documentation

scbroede:


scbroede:
so clang-format doesn't seem to like this, but imo it's full of shit here

instead of 
`template.position.x = FX32_ONE * sBadgeCoordinates[badgeID].x;`
it seems to want
`template.position.x = FX32_ONE *sBadgeCoordinates[badgeID].x;`
for some reason?

lhearachel:
> so clang-format doesn't seem to like this, but imo it's full of shit here

This sounds like a case where `clang-format` isn't sure what to do with `FX32_ONE`; it might think that it's a type, for some reason? Or a cast. In either case, you can use the [`Macros`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#macros) property of the `.clang-format` file to explain how it should be treated.

include/trainer_card.h
```diff
@@ -5,25 +5,34 @@
...
+#define SIGNATURE_WIDTH  24
+#define SIGNATURE_HEIGHT 8
+
+#define BADGE_POLISH_THRESHOLD_DIRTY      100
+#define BADGE_POLISH_THRESHOLD_NORMAL     140
+#define BADGE_POLISH_THRESHOLD_2_SPARKLES 170
+#define BADGE_POLISH_THRESHOLD_4_SPARKLES 190
+#define MAX_BADGE_POLISH                  199
+
 typedef struct {
```
lhearachel: **polish**: Tag the struct with the name of its `typedef`.

include/sound.h
```diff
@@ -29,6 +29,11 @@
 #define WAVE_OUT_PAN_CENTER 64
 #define WAVE_OUT_PAN_RIGHT  127
 
+#define SOUND_PLAYBACK_TRACK(TRACK) (1 << (TRACK))
+#define SOUND_PLAYBACK_TRACK_ALL    0xFFFF
```
lhearachel, line 33: **question**: Why do these need to be in `sound.h` instead of `sound_playback.h`?

include/trainer_card_screen/struct_trainer_card_screen.h
```diff
@@ -0,0 +1,73 @@
+#ifndef POKEPLATINUM_STRUCT_TRAINERCARD_SCREEN_H
+#define POKEPLATINUM_STRUCT_TRAINERCARD_SCREEN_H
+
+#include <nnsys.h>
+
+#include "trainer_card_screen/struct_trainer_card_badge_chime_state.h"
+#include "trainer_card_screen/struct_trainer_card_sprite_data.h"
```
lhearachel: **question**: Can these files not be merged into this one?

include/trainer_card_screen/trainer_card_screen.h
```diff
@@ -0,0 +1,19 @@
...
+#include "overlay_manager.h"
+
+enum BadgePolishLevels {
+    BADGE_POLISH_LEVEL_FILTHY,
+    BADGE_POLISH_LEVEL_DIRTY,
+    BADGE_POLISH_LEVEL_NORMAL,
+    BADGE_POLISH_LEVEL_2_SPARKLES,
+    BADGE_POLISH_LEVEL_4_SPARKLES,
+    BADGE_POLISH_LEVEL_COUNT
+};
```
lhearachel: **thought**: Does this `enum` better belong alongside the `struct` definitions?

include/trainer_card_screen/trainer_card_sprites.h
```diff
@@ -0,0 +1,17 @@
+#ifndef POKEPLATINUM_TRAINERCARD_SPRITES_H
+#define POKEPLATINUM_TRAINERCARD_SPRITES_H
+
+#include "trainer_card_screen/struct_trainer_card_sprite_data.h"
```
lhearachel: **question**: Leading from the previous question regarding merging-files, should all of the `struct` and `enum` definitions be in this file? If not, then I think the definitions file should be `trainer_card_defs.h` rather than `struct_trainer_card_screen.h`.

src/trainer_card_screen/trainer_card_sprites.c
```diff
@@ -0,0 +1,232 @@
...
+        [BADGE_ID_BEACON] = 23
+    };
+
+    for (badgeID = 0; badgeID < MAX_BADGES; badgeID++) {
+        spriteData->badgePaletteBuffers[badgeID] = NARC_AllocAndReadWholeMember(narc, badgePaletteNarcIndices[badgeID], HEAP_ID_TRAINER_CARD_SCREEN);
+
+        if (spriteData->badgePaletteBuffers[badgeID] != NULL) {
+            if (NNS_G2dGetUnpackedPaletteData(spriteData->badgePaletteBuffers[badgeID], &spriteData->badgePalettes[badgeID]) == FALSE) {
+                Heap_Free(spriteData->badgePalettes[badgeID]);
+                GF_ASSERT(0);
```
lhearachel: **polish**: `GF_ASSERT(FALSE)`

src/trainer_card_screen/trainer_card_screen.c
```diff
@@ -0,0 +1,1323 @@
...
+static const u8 sPolishThresholds[BADGE_POLISH_LEVEL_COUNT] = {
+    1,
+    3,
+    4,
+    15,
+    15
+};
+
+static const TouchScreenRect sTouchRectangles_BadgesEnabled[] = {
+    [TOUCH_RECTANGLE_OPEN_CLOSE_BADGE_CASE] = { .rect.top = 152, .rect.bottom = 183, .rect.left = 120, .rect.right = 151 },
```
lhearachel: **question**: Do these lines match as, e.g.:

```c
[TOUCH_RECTANGLE_OPEN_CLOSE_BADGE_CASE] = { .rect = { .top = 152, .bottom = 183, .left = 120, .right = 151 } },
```

src/trainer_card_screen/trainer_card_screen.c
```diff
@@ -0,0 +1,1323 @@
...
+};
+
+enum BadgeCaseButtonSpriteIndices {
+    BUTTON_NOT_PRESSED,
+    BUTTON_HALF_PRESSED,
+    BUTTON_FULLY_PRESSED
+};
+
+typedef struct TouchRectangleSet {
+    const TouchScreenRect *touchRectangles[2];
```
lhearachel: **suggestion**: Semantically tie this to `BadgeCaseOpenCloseStates` in some manner.

include/sound.h
scbroede: `sound.h` also includes function that have `tracks` as a parameter and therefore would want these defined (it has more of them, actually; 3 vs 1), and i thought it made more sense for files using `sound_playback.h` to also have to import `sound.h` than it did for files using `sound.h` to also have to import `sound_playback.h`.

i don't feel too strongly about it, though. willing to revert if you disagree

lhearachel: I don't mind the change; I was more asking for clarity. 😁

lhearachel (approved):

PR #649, opened 2025-07-27
Document Battle Font engine

whoispoosto:
`unk_0200C440` from [good first documentation efforts list](https://github.com/pret/pokeplatinum/issues/325)

* Wasn't too sure on some of the names for functions and parameters.
* Still not sure on a few parts of the source file. Namely, I have no idea what the different cases in the switch statement inside `BattleFont_Init` represent, or what the `Unk_020E4FEC` table values are.


src/battle_font.c
```diff
@@ -0,0 +1,149 @@
...
+
+BattleFontContext *BattleFont_Init(u32 color, u32 outline, u32 transparent, u32 heapID)
+{
+    BattleFontContext *context = Heap_AllocFromHeap(heapID, sizeof(BattleFontContext));
+
+    if (context) {
+        u32 idx;
+        u8 *rawData;
+
+        context->ncgrBuffer = Graphics_GetCharData(NARC_INDEX_GRAPHIC__PL_FONT, 4, 1, &context->charData, heapID);
```
lhearachel: **thought**: Based on this line and the context of the overall usage. I'm hesitant to call this `battle_font` given that it is used in a lot of non-battle places:

- `shop_menu`
- `overlay019`, which is the PC box interface
- `overlay084`, which is the Bag interface in the overworld
- `pokemon_summary_screen`

This line infers that this font pulls from `font_special_chars.png` in `res/fonts`. I think that we should rename this, then, as `font_special_chars` to create a semantic mapping between the code and the asset.

---

**suggestion**: Add `#include "res/fonts/pl_font.naix.h"` to this file's includes, then:

```suggestion
        context->ncgrBuffer = Graphics_GetCharData(NARC_INDEX_GRAPHIC__PL_FONT, font_special_chars_NCGR_lz, TRUE, &context->charData, heapID);
```

src/battle_font.c
```diff
@@ -0,0 +1,149 @@
...
+#include <nitro.h>
+#include <string.h>
+
+#include "constants/charcode.h"
+
+#include "bg_window.h"
+#include "graphics.h"
+#include "heap.h"
+
+#define BTIMAP_SIZE 0x20
```
lhearachel: **polish**: Remove this constant and instead use `TILE_SIZE_4BPP` from `constants/graphics.h`.

src/battle_font.c
```diff
@@ -0,0 +1,149 @@
...
+
+typedef struct BattleFontContext_t {
+    void *ncgrBuffer;
+    NNSG2dCharacterData *charData;
+    charcode_t charcodes[16];
+    u32 transparent;
+} BattleFontContext;
```
lhearachel: **polish**: Extract this to `battle_font.h`, delete the corresponding `struct_decls/battle_font_decl.h` file, and replace all occurrences of `#include "struct_decls/battle_font_decl.h"` with `#include "battle_font.h"`.

```suggestion
typedef struct BattleFontContext {
    void *ncgrBuffer;
    NNSG2dCharacterData *charData;
    charcode_t charcodes[16];
    u32 transparent;
} BattleFontContext;
```

src/battle_font.c
```diff
@@ -0,0 +1,149 @@
...
+    u16 unk_02;
+} Unk_020E4FEC[] = {
+    { 0x140, 0x8 },
+    { 0x160, 0x10 },
+    { 0x1A0, 0x10 },
+    { 0x1E0, 0x10 },
+    { 0x220, 0x10 },
+    { 0x260, 0x10 },
+    { 0x2A0, 0x10 }
+};
```
lhearachel: **thought**: Looking at the asset file, my hunch is that this `struct` is used to determine the width of a non-numeric "character" in pixels. To that end, I think that we want to document it like this:

```c
enum NonNumericSpecialChar {
    SPECIAL_CHAR_SLASH = 0,
    SPECIAL_CHAR_LEVEL,
    SPECIAL_CHAR_NUMBER,
    SPECIAL_CHAR_ID,
    SPECIAL_CHAR_LEVEL_FEMALE_UNUSED,
    SPECIAL_CHAR_LEVEL_MALE_UNUSED,
    SPECIAL_CHAR_LEVEL_UNUSED,
};

static const struct {
    u16 offset;
    u16 width;
} sNonNumericWidths[] = {
    [SPECIAL_CHAR_SLASH] = { .offset = 10 * TILE_SIZE_4BPP, .width = TILES_TO_PIXELS(1) },
    /* etc. */
};
```

The name of the enum isn't set-in-stone, but the enum should probably also be exposed in the header file and used as the type of `int tableIdx` for `BattleFont_DrawPartyScreenLevelText` (and any other similar parameters).

poosto:
There's a lot of use of the magic number `8` throughout the source file. Should those be replaced with `TILE_WIDTH_PIXELS` and `TILE_HEIGHT_PIXELS` or is there a better constant to use?

And any idea what to do for the switch statement in `FontSpecialChars_Init`? I was able to figure out the parameters either fill a font sprite with color, outline it, or make it transparent, but beyond that I couldn't deduce what each case was or what the difference was between the upper and lower 4 bits (I'm assuming it's something to do with the 4bpp format).

src/font_special_chars.c
```diff
@@ -0,0 +1,142 @@
...
+    [SPECIAL_CHAR_SLASH] = { .offset = 10 * TILE_SIZE_4BPP, .width = TILES_TO_PIXELS(1) },
+    [SPECIAL_CHAR_LEVEL] = { .offset = 11 * TILE_SIZE_4BPP, .width = TILES_TO_PIXELS(2) },
+    [SPECIAL_CHAR_NUMBER] = { .offset = 13 * TILE_SIZE_4BPP, .width = TILES_TO_PIXELS(2) },
+    [SPECIAL_CHAR_ID] = { .offset = 15 * TILE_SIZE_4BPP, .width = TILES_TO_PIXELS(2) },
+    [SPECIAL_CHAR_LEVEL_FEMALE_UNUSED] = { .offset = 17 * TILE_SIZE_4BPP, .width = TILES_TO_PIXELS(2) },
+    [SPECIAL_CHAR_LEVEL_MALE_UNUSED] = { .offset = 19 * TILE_SIZE_4BPP, .width = TILES_TO_PIXELS(2) },
+    [SPECIAL_CHAR_LEVEL_UNUSED] = { .offset = 21 * TILE_SIZE_4BPP, .width = TILES_TO_PIXELS(2) }
+};
+
+FontSpecialCharsContext *FontSpecialChars_Init(u32 color, u32 outline, u32 transparent, u32 heapID)
```
lhearachel: **note**: I see what you mean about these values now. We typically call these `fgColor`, `shadowColor`, and `bgColor`, as they relate to font glyphs. You can reference `TextPrinterTemplate` in `render_text.h` to see what I mean.

src/font_special_chars.c
```diff
@@ -0,0 +1,142 @@
...
+                break;
+            case 32:
+                rawData[idx] = (outline << 4) | (transparent);
+                break;
+            case 33:
+                rawData[idx] = (outline << 4) | (color);
+                break;
+            case 34:
+                rawData[idx] = (outline << 4) | (outline);
+                break;
```
lhearachel: **note**: These case values, like you intuited, are related to the 4bpp format. Essentially, the upper "nibble" of a byte (the top four bits) is used for a pixel, and the lower nibble is used for the neighbor to its right. In bits, these cases are:

- `0000 0000` - pixel pair renders as `bgColor` and `bgColor`
- `0000 0001` - pixel pair renders as `bgColor` and `fgColor`
- `0000 0010` - pixel pair renders as `bgColor` and `shadowColor`
- `0001 0000` - pixel pair renders as `fgColor` and `bgColor`
- `0001 0001` - etc.
- `0001 0010` - etc.
- `0010 0000` - pixel pair renders as `shadowColor` and `bgColor`
- `0010 0001` - etc.
- `0010 0010` - etc.

Personally, I'm not certain how I would represent these semantically. I would probably just leave a comment explaining what's happening here: this is just altering the palette indices applied to each special-char.

lhearachel (commented):
> There's a lot of use of the magic number `8` throughout the source file. Should those be replaced with `TILE_WIDTH_PIXELS` and `TILE_HEIGHT_PIXELS` or is there a better constant to use?

Presuming that the values of `8` are referring to a tile dimension, then those would be the constants to use. If you aren't certain if that's the case for a particular instance, then don't worry about it.

> And any idea what to do for the switch statement in `FontSpecialChars_Init`? I was able to figure out the parameters either fill a font sprite with color, outline it, or make it transparent, but beyond that I couldn't deduce what each case was or what the difference was between the upper and lower 4 bits (I'm assuming it's something to do with the 4bpp format).

See my thoughts in the comments below.

The rest looks great! Good work. 😁

include/font_special_chars.h
```diff
@@ -0,0 +1,31 @@
...
+    SPECIAL_CHAR_ID,
+    SPECIAL_CHAR_LEVEL_FEMALE_UNUSED,
+    SPECIAL_CHAR_LEVEL_MALE_UNUSED,
+    SPECIAL_CHAR_LEVEL_UNUSED,
+};
+
+typedef struct FontSpecialCharsContext {
+    void *ncgrBuffer;
+    NNSG2dCharacterData *charData;
+    charcode_t charcodes[16];
```
lhearachel, line 20: **thought**: We will probably want to revisit some way to tie this to the underlying graphics file, but I do not think we currently have a way to do that. Non-blocking for now, just musing.

lhearachel (approved):

PR #650, opened 2025-07-27
Document script_manager

metallicity:
Documented nearly all the remaining auto-named functions and variables in `script_manager.c`, also going into portions of `overlay034`, which as best I can guess deals solely with the Dowsing Machine app for the Poketch. Analyzing `FieldSystem_GetNearbyHiddenItems` (`sub_0203F478`) naturally brought with it understanding of several functions higher up in the callstack, so it made sense to include them here too. The distinction between `HiddenItemOffsets` and `HiddenItemScreenPosition` (relative number of tiles away from the player vs. absolute pixel coordinates on screen) might need some clearer naming, but I wasn't quite sure what might work best there.

Otherwise, the main thing left on the table here is identifying the exact ways the different types of init scripts are called, to give them more descriptive labels. Nearly all maps use solely `INIT_SCRIPT_MODE_DYNAMIC` and `INIT_SCRIPT_UNK_STATIC_MODE_2`, with `3` and `4` appearing only a handful of times in all. Though `scripts_valley_windworks_outside.s` is notable for being the only script to use every type of "static" init script (as well as a "dynamic" script).

I don't know if anyone else has already looked into how init scripts are declared, but here's the understanding I came to:

```bash
#include "macros/scrcmd.inc"


    .byte 4 # When init script mode = 4
    .short 2, 0 # The second map script (1-based index, not 0-based)
    .byte 2 # When init script mode = 2
    .short 1, 0 # The first map script
    .byte 3 # When init script mode = 3
    .short 7, 0 # The seventh map script
    .byte 1 # When init script mode = 1 (dynamic)
    ScriptEntry _0015 # Start of dynamic scripts (always last, with a single ScriptEntry)
    .byte 0 # No more init script modes covered (though this example is exhaustive)

_0015:
    .short 0x411E, 1, 8 # The eighth map script, when TryGetVar(0x411E) = TryGetVar(1), though the latter is just 1.
    .short 0 # No more dynamic init scripts 

    .balign 4, 0
```

### DIFF
--- a/src/script_manager.c
+++ b/src/script_manager.c
@@ -746,9 +746,9 @@ void FieldSystem_RunScript(FieldSystem *fieldSystem, u16 scriptID)
 
 BOOL FieldSystem_RunInitScript(FieldSystem *fieldSystem, u8 initScriptType)
 {
-    const u8 *initScripts = MapHeaderData_GetInitScripts(fieldSystem);
+    const u8 *initScriptBytes = MapHeaderData_GetInitScriptBytes(fieldSystem);
 
-    if (initScripts == NULL) {
+    if (initScriptBytes == NULL) {
         return FALSE;
     }
 
@@ -772,19 +772,23 @@ BOOL FieldSystem_RunInitScript(FieldSystem *fieldSystem, u8 initScriptType)
     return TRUE;
 }
 
+#define GET8(bytes)  *bytes
+#define GET16(bytes) bytes[0] + (bytes[1] << 8)
+#define GET32(bytes) bytes[0] + (bytes[1] << 8) + (bytes[2] << 16) + (bytes[3] << 24)
+
 static u16 FieldSystem_GetFixedInitScriptID(const u8 *initScriptBytes, u8 initScriptType)
 {
     while (TRUE) {
-        if (*initScripts == 0) {
+        if (GET8(initScriptBytes) == 0) {
             return 0xffff;
         }
 
-        if (*initScripts == initScriptType) {
-            initScripts++;
-            return *initScripts + (*(initScripts + 1) << 8);
+        if (GET8(initScriptBytes) == initScriptType) {
+            initScriptBytes++;
+            return GET16(initScriptBytes);
         }
 
-        initScripts += (1 + 2 + 2);
+        initScriptBytes += (1 + 2 + 2);
     }
 
     return 0xffff;
@@ -796,46 +800,46 @@ static u16 FieldSystem_GetFirstMatchInitScriptID(FieldSystem *fieldSystem, const
     u32 offset = 0;
 
     while (TRUE) {
-        if (*initScripts == 0) {
+        if (GET8(initScriptBytes) == 0) {
             return 0xffff;
         }
 
-        if ((*initScripts) == initScriptType) {
-            initScripts++;
-            offset = (*initScripts + (*(initScripts + 1) << 8) + (*(initScripts + 2) << 16) + (*(initScripts + 3) << 24));
-            initScripts += 4;
+        if (GET8(initScriptBytes) == initScriptType) {
+            initScriptBytes++;
+            offset = GET32(initScriptBytes);
+            initScriptBytes += 4;
             break;
         }
 
-        initScripts += (1 + 4);
+        initScriptBytes += (1 + 4);
     }
 
     if (offset == 0) {
         return 0xffff;
     }
 
-    initScripts = (initScripts + offset);
+    initScriptBytes = (initScriptBytes + offset);
 
     while (TRUE) {
-        if (*initScripts == 0) {
+        if (GET8(initScriptBytes) == 0) {
             return 0xffff;
         }
 
-        currentVar = (*initScripts + (*(initScripts + 1) << 8));
+        currentVar = GET16(initScriptBytes);
 
         if (currentVar == 0) {
             return 0xffff;
         }
 
-        initScripts += 2;
-        compareVar = (*initScripts + (*(initScripts + 1) << 8));
-        initScripts += 2;
+        initScriptBytes += 2;
+        compareVar = GET16(initScriptBytes);
+        initScriptBytes += 2;
 
         if (FieldSystem_TryGetVar(fieldSystem, currentVar) == FieldSystem_TryGetVar(fieldSystem, compareVar)) {
-            return *initScripts + (*(initScripts + 1) << 8);
+            return GET16(initScriptBytes);
         }
 
-        initScripts += 2;
+        initScriptBytes += 2;
     }
 
     return 0xffff;
--- a/src/script_manager.c
+++ b/src/script_manager.c
@@ -39,8 +39,8 @@ static void ScriptContext_JumpToOffsetID(ScriptContext *ctx, u16 param1);
 static void *ScriptContext_LoadScripts(int headerID);
 static u32 MapHeaderToMsgArchive(int headerID);
 static BOOL ScriptManager_SetHiddenItem(ScriptManager *scriptManager, u16 scriptID);
-static u16 FieldSystem_GetStaticInitScriptID(const u8 *param0, u8 param1);
-static u16 FieldSystem_GetDynamicInitScriptID(FieldSystem *fieldSystem, const u8 *param1, u8 param2);
+static u16 FieldSystem_GetFixedInitScriptID(const u8 *param0, u8 param1);
+static u16 FieldSystem_GetFirstMatchInitScriptID(FieldSystem *fieldSystem, const u8 *param1, u8 param2);
 
 void ScriptManager_Set(FieldSystem *fieldSystem, u16 scriptID, MapObject *object)
 {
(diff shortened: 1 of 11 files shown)


include/overlay034/struct_ov34_02256540_1.h
```diff
@@ -1,14 +1,14 @@
 #ifndef POKEPLATINUM_STRUCT_OV34_02256540_1_H
 #define POKEPLATINUM_STRUCT_OV34_02256540_1_H
 
-#include "overlay034/struct_ov34_02256540_1_sub1.h"
+#include "overlay034/hidden_item_screen_position.h"
```
lhearachel: **suggestion**: Merge this struct definition into this header; we aim to have as few standalone-headers as possible, as a stylistic choice.

include/struct_defs/hidden_item_offsets.h
lhearachel, line 1: **question**: Do we know of a header that would be the best owner for this struct definition?

src/script_manager.c
```diff
@@ -492,10 +492,10 @@ u16 FieldSystem_TryGetVar(FieldSystem *fieldSystem, u16 varID)
...
 }
 
-u16 sub_0203F164(FieldSystem *fieldSystem, u16 varID)
+u16 FieldSystem_GetGraphicsID(FieldSystem *fieldSystem, u16 graphicsID)
 {
-    GF_ASSERT(varID < 16);
-    return FieldSystem_TryGetVar(fieldSystem, (((0 + VARS_START) + 32) + varID));
+    GF_ASSERT(graphicsID < 16);
+    return FieldSystem_TryGetVar(fieldSystem, (((0 + VARS_START) + 32) + graphicsID));
 }
```
lhearachel: **thought**: It might be worth clarifying that this is the overworld graphics ID.

src/script_manager.c
```diff
@@ -666,67 +666,67 @@ static BOOL ScriptManager_SetHiddenItem(ScriptManager *scriptManager, u16 script
...
-        if ((v1[v4].type == 2) && (FieldSystem_CheckFlag(fieldSystem, Script_GetHiddenItemFlag(v1[v4].script)) == 0)) {
-            if ((v1[v4].x >= v8) && (v1[v4].x <= v9) && (v1[v4].z >= v10) && (v1[v4].z <= v11)) {
-                v0[v6].unk_04 = Script_GetHiddenItemRange(v1[v4].script);
-                v7 = (v2 - v1[v4].x);
-                v0[v6].unk_00 = abs(7 - v7);
-                v7 = (v3 - v1[v4].z);
-                v0[v6].unk_02 = abs(7 - v7);
-                v6++;
+    for (eventIndex = 0; eventIndex < numBgEvents; eventIndex++) {
+        if ((bgEvents[eventIndex].type == 2) && (FieldSystem_CheckFlag(fieldSystem, Script_GetHiddenItemFlag(bgEvents[eventIndex].script)) == 0)) {
```
lhearachel: **question**: Is there a semantic name that we can give `type == 2` here?

lhearachel (commented):
> The distinction between `HiddenItemOffsets` and `HiddenItemScreenPosition` (relative number of tiles away from the player vs. absolute pixel coordinates on screen) might need some clearer naming, but I wasn't quite sure what might work best there.

Perhaps `HiddenItemDistance`? The latter of these names feels appropriately clear to my eye.

> I don't know if anyone else has already looked into how init scripts are declared

To my knowledge, the only contributor that might have done any of this investigation thus-far is @cbt6.

include/struct_defs/hidden_item_offsets.h
metallicity: I ended up putting it with the other DowsingMachine structs, since the two item position structs are shaped very similarly and used by the same chunk of code (more or less). I also went a bit further in documenting these structs, and ended up landing on a more clearly differentiated `HiddenItemTilePosition` vs. `HiddenItemPixelPosition`. Their fields then specify that they are relative to the screen, to avoid having to include all that information in the struct name.

src/script_manager.c
metallicity: The behavior of the different `type` values was undocumented, so I ended up doing a bit of work on `unk_0203C954.c` while figuring it out. Basically `type = 2` is for hidden items, while `type = 0` is for normal interactable static objects. There's one other option `type = 1`, for signs that are on northern walls that otherwise wouldn't show their text when you ran into those walls. This was only used twice, once in the player's room, and once in the underground man's house in Eterna City, but in the latter case it was placed on the wrong object (a computer that can't be walked into from the south) and so the sign there still needs to be interacted with manually.

cbt6:
I don't have the time to look through this PR, but one thing that stands out to me is `INIT_SCRIPT_MODE_DYNAMIC` / `INIT_SCRIPT_UNK_STATIC_MODE_X`. Not a fan of these name because they doesn't convey what dynamic / static mean in the context of init scripts.

Also let's try to keep PRs focused on one thing each. Imo this PR is covering too many things at once:
- init scripts
- dowsing machine subroutines
- bg events

> > I don't know if anyone else has already looked into how init scripts are declared
> 
> To my knowledge, the only contributor that might have done any of this investigation thus-far is @cbt6.

The init scripts get called at different points depending on how a map gets loaded (initialized). But I'd rather we hold off giving names to these different types until we have the map loading functions documented well enough.

metallicity:
> Also let's try to keep PRs focused on one thing each. Imo this PR is covering too many things at once:
> - init scripts
> - dowsing machine subroutines
> - bg events

I've gone ahead and split off the latter two areas in their own PRs (#652 for DowsingMachine and #653 for BgEvent). I did ask on the other PR I opened (#651) if it was preferable to break things out into smaller PRs or not, and wasn't asked to, but admittedly those changes all did have more of a cohesive focus.

> I'd rather we hold off giving names to these different types until we have the map loading functions documented well enough.

I felt it was important to give a name to `1`, since that was actually being used in the function logic, rather than just being generically checked against the value read from each init script entry. But I also didn't want the extra scope creep of figuring out what all the rest of them did before documenting this unique behavior. I did change the name to `INIT_SCRIPT_TYPE_FIRST_MATCH`, and the others to `INIT_SCRIPT_TYPE_FIXED_UNK_[N]`, which hopefully expresses that each value other than `1` is associated with at most a single script on each map, while `1` finds the first script among potentially several that matches some condition (namely, that the two given values for that script are equal). But I'm fine leaving it as `1` if necessary.

src/script_manager.c
```diff
@@ -753,17 +753,17 @@ BOOL sub_0203F5C0(FieldSystem *fieldSystem, u8 param1)
     }
 
     u16 scriptID;
-    if (param1 == 1) {
-        scriptID = sub_0203F638(fieldSystem, initScripts, param1);
+    if (initScriptType == INIT_SCRIPT_TYPE_FIRST_MATCH) {
+        scriptID = FieldSystem_GetDynamicInitScriptID(fieldSystem, initScripts, initScriptType);
```
lhearachel: **suggestion**: Change the name of the function to mirror `initScriptType`.

src/script_manager.c
```diff
@@ -772,70 +772,70 @@ BOOL sub_0203F5C0(FieldSystem *fieldSystem, u8 param1)
...
             return 0xffff;
         }
 
-        if ((*param1) == param2) {
-            param1++;
-            v2 = (*param1 + (*(param1 + 1) << 8) + (*(param1 + 2) << 16) + (*(param1 + 3) << 24));
-            param1 += 4;
+        if ((*initScripts) == initScriptType) {
+            initScripts++;
+            offset = (*initScripts + (*(initScripts + 1) << 8) + (*(initScripts + 2) << 16) + (*(initScripts + 3) << 24));
```
lhearachel: **suggestion**:

```suggestion
            offset = initScripts[0] + (initScripts[1] << 8) + (initScripts[2] << 16) + (initScripts[3] << 24);
```

I would also see if this matches with `|` instead of `+`, which is more typical for this kind of construction of a `u32` from a sequence of `u8`s.

This suggestion applies to other similar constructs elsewhere in the function where a `u16` or `u32` is computed from a sequence of `u8`s.

src/script_manager.c
```diff
@@ -772,70 +772,70 @@ BOOL sub_0203F5C0(FieldSystem *fieldSystem, u8 param1)
...
 
-        param0 += (1 + 2 + 2);
+        initScripts += (1 + 2 + 2);
     }
 
     return 0xffff;
 }
 
-static u16 sub_0203F638(FieldSystem *fieldSystem, const u8 *param1, u8 param2)
+static u16 FieldSystem_GetDynamicInitScriptID(FieldSystem *fieldSystem, const u8 *initScripts, u8 initScriptType)
```
lhearachel: **thought**: Is `initScripts` the right name to convey what is actually here? Should this be `initScriptBytes`, since the code performs an iteration over the byte-content?

lhearachel (commented):
I generally agree that it would be best to leave things with `unk` names if we aren't certain, but I think that the behavior here is clear enough, unless we want to mirror any name for similar behavior from `pokeemerald`.

src/script_manager.c
metallicity: I tried `|` and the checksum failed, but in looking through the codebase with any other such constructs, I noticed these `SET8`/`SET16`/`SET32` macros in [overlay097/gba_pokemon.c](https://github.com/pret/pokeplatinum/blob/e6b859f580f358be0f3059eafd906522e54fdb6f/src/overlay097/gba_pokemon.c#L663) doing a similar thing. What do you think of something like this?

```c
#define GET8(bytes)  *bytes
#define GET16(bytes) bytes[0] + (bytes[1] << 8)
#define GET32(bytes) bytes[0] + (bytes[1] << 8) + (bytes[2] << 16) + (bytes[3] << 24)
```

(not sure if the `GET8` is too trivial to bother with, but I can see wanting to have symmetry with the other operators here)

src/script_manager.c
metallicity: I just went with the name of the function that returned it in the first place, but agree it felt a little misleading. The better option is probably to rename that function to `MapHeaderData_GetInitScriptBytes`.

lhearachel (approved):

PR #651, opened 2025-07-28
Document misc script commands and identify all unused script commands

metallicity:
Not sure if this would be preferred as one PR or multiple, but I went through and documented a handful of script commands, and in the process also decided to identify all the unused commands to distinguish them from simply undocumented commands. I didn't remove them from `scrcmd.inc` though, just in case someone wanted to call one of these commands in their own project or something. But I did verify the ROM check passed when they were removed, so between that and the rename there shouldn't be any references I missed.

I also noticed while in `scripts_common.s` that `OpenPokemonStorage` couldn't used named enums from `BoxMode`, since it was only defined for C code. I did give names to all the menu text it referenced, so it could be understood which items where which, but felt like things were getting a little out of scope when the goal was just to have `OpenPokemonStorage` clearly documented.

## Documented Commands

- `CloseMessageWithoutErasing`: Not the most creative name, but that's what it does. Seems mostly used by shop menus.
- `LockObject`/`ReleaseObject`: Targeted equivalents of `LockAll` and `ReleaseAll`, taking a single object id
- `AddFreeCamera`/`RestoreCamera`: Swaps out the normal player-tracking camera for one that responds to `ApplyMovement`
- `ApplyFreeCameraMovement`: Not a native command, just a helper macro for `ApplyMovement 241`, which is the camera object id. Shouldn't do anything when the "normal" camera object is active, hence specifying the "free" camera.
- `AddCameraOverrideObject`/`RemoveCameraOverrideObject`: This one looks a bit weird, but it puts an invisible object in place of the camera object while keeping the camera locked in place. Only used for the Poke Radar tutorial, to "walk" through an empty grass patch and have it wiggle.
- `OpenPokemonStorage`: Takes a `BoxMode` and opens the Pokemon Storage System under that mode, exiting when closed

## Other Documented Functionality

- `MapObject_RecalculateObjectHeight`: Checks what Y position an object should be at on the map based on its current X and Z positions, then sets it to that height. Objects can configure via status flags whether they recognize "dynamic" sources of height or if they can change height at all.
- `ScriptContext_WaitForFinishedPrinting`/`ScriptContext_WaitForMovement`: Functions passed into `ScriptContext_Pause` to wait for certain actions to finish
- `PokemonStorageSession`: Struct that `OpenPokemonStorage` (and elsewhere) use to maintain state during box operations


src/scrcmd.c
```diff
@@ -3216,64 +3216,64 @@ static BOOL ScrCmd_RemoveObject(ScriptContext *ctx)
...
-static BOOL ScrCmd_067(ScriptContext *ctx)
+static BOOL SrcCmd_RestoreCamera(ScriptContext *ctx)
 {
-    MapObject **v0 = FieldSystem_GetScriptMemberPtr(ctx->fieldSystem, SCRIPT_MANAGER_CAMERA_OBJECT);
+    MapObject **freeCameraObject = FieldSystem_GetScriptMemberPtr(ctx->fieldSystem, SCRIPT_MANAGER_CAMERA_OBJECT);
 
-    MapObject_Delete(*v0);
+    MapObject_Delete(*freeCameraObject);
 
     {
```
lhearachel: **polish**: Remove unnecessary scoping

include/struct_defs/pokemon_storage_session.h
lhearachel, line 1: **question**: Is there a header that would be a suitable owner for this struct?

lhearachel (commented):
This looks great! I just have one leading question and a polish-comment.

include/struct_defs/pokemon_storage_session.h
metallicity: I don't have a good sense of the larger repo, but it's only used in a few places and I don't see any obvious header that those places already share. Looking more closely it's really only used for communicating between `scrcmd.c` and `overlay019` (which handles the pokemon storage interface), so I could see moving the file under `overlay019`. I've gone ahead and done so, but its still its own individual file. I also decided to document the bridge between these files a bit better, since for some reason I missed that.

PR #653, opened 2025-07-30
Document BgEvent types

metallicity:
Splitting this off from #650, focusing only on documenting the different types of BgEvents and a few functions related to their behavior.

Taking the summary from [this comment](https://github.com/pret/pokeplatinum/pull/650#discussion_r2238585726):

> Basically `type = 2` is for hidden items, while `type = 0` is for normal interactable static objects. There's one other option `type = 1`, for signs that are on northern walls that otherwise wouldn't show their text when you ran into those walls. This was only used twice, once in the player's room, and once in the underground man's house in Eterna City, but in the latter case it was placed on the wrong object (a computer that can't be walked into from the south) and so the sign there still needs to be interacted with manually.


generated/bg_event_types.txt
```diff
@@ -0,0 +1,3 @@
+BG_EVENT_TYPE_FACING
```
lhearachel, line 1: **question**: Are you intending to update the `.json` files with these values as part of this PR, or is that being saved for another time?

lhearachel (approved):
One leading question. These changes look good as-is.

generated/bg_event_types.txt
metallicity: I looked into it briefly, and it didn't look too bad (giant diff aside), but figured that could wait until I closed out some of the PRs I already had open. If you'd like, I can add it here, otherwise I'm fine with just getting this merged in first.

generated/bg_event_types.txt
```diff
@@ -0,0 +1,3 @@
...
+BG_EVENT_TYPE_WALL
```
cbt6:
```suggestion
BG_EVENT_TYPE_WALL_SIGN
```

src/unk_0203C954.c
```diff
@@ -96,23 +99,23 @@ u8 sub_0203CA40(FieldSystem *fieldSystem, MapObject **param1)
...
 
-    for (v3 = 0; v3 < param2; v3++) {
-        if ((v1 == v0[v3].x) && (v2 == v0[v3].z)) {
-            if (v0[v3].type == 2) {
-                if (sub_0203CAEC(fieldSystem, &v0[v3]) == 1) {
-                    return v0[v3].script;
+    for (eventIndex = 0; eventIndex < numBgEvents; eventIndex++) {
+        if ((facingX == events[eventIndex].x) && (facingZ == events[eventIndex].z)) {
+            if (events[eventIndex].type == BG_EVENT_TYPE_HIDDEN_ITEM) {
+                if (BgEvent_IsHiddenItemAvailable(fieldSystem, &events[eventIndex]) == TRUE) {
```
cbt6, line 113:
```suggestion
            if (events[eventIndex].type == BG_EVENT_TYPE_HIDDEN_ITEM && BgEvent_IsHiddenItemAvailable(fieldSystem, &events[eventIndex]) == TRUE) {
```

src/unk_0203C954.c
```diff
@@ -121,66 +124,66 @@ u16 sub_0203CA6C(FieldSystem *fieldSystem, const BgEvent *param1, int param2)
     return 0xffff;
 }
 
-static u8 sub_0203CAEC(FieldSystem *fieldSystem, const BgEvent *param1)
+static u8 BgEvent_IsHiddenItemAvailable(FieldSystem *fieldSystem, const BgEvent *param1)
 {
     if (param1->type != 2) {
```
cbt6:
```suggestion
    if (param1->type != BG_EVENT_TYPE_HIDDEN_ITEM) {
```

include/unk_0203C954.h
```diff
@@ -9,8 +9,8 @@
 
 void sub_0203C9D4(FieldSystem *fieldSystem, MapObject **param1);
 u8 sub_0203CA40(FieldSystem *fieldSystem, MapObject **param1);
-u16 sub_0203CA6C(FieldSystem *fieldSystem, const BgEvent *param1, int param2);
-u16 sub_0203CB80(FieldSystem *fieldSystem, const BgEvent *param1, int param2);
+u16 BgEvent_GetFacingPositionScript(FieldSystem *fieldSystem, const BgEvent *bgEvents, int bgEventCount);
+u16 BgEvent_GetFacingWallScript(FieldSystem *fieldSystem, const BgEvent *param1, int param2);
```
cbt6:
```suggestion
u16 FieldEvent_GetInteractedBgEventScript(FieldSystem *fieldSystem, const BgEvent *bgEvents, int numBgEvents);
u16 FieldEvent_GetInteractedWallSignScript(FieldSystem *fieldSystem, const BgEvent *bgEvents, int numBgEvents);
```

generated/bg_event_types.txt
metallicity: I actually made this exact change in the follow-up branch I worked on to update all the bg events in `.json` files with these constants. Agree that "wall sign" makes it more clear that it isn't necessary for every interactable object that happens to be along a wall.

src/unk_0203C954.c
metallicity: This change is a syntax error on its own, and even just trying to adjust the surrounding code somehow along these lines, I can't see any way to maintain the same logic structure, let alone compiling down to identical machine code. In the case of a collected hidden item, it looks like this would end up returning that hidden item script, rather than skipping it.

The full logic here is:

```
for event in events:
  if event is on the position the player faces:
    if event is a hidden item:
      if event is an uncollected hidden item:
        return event script
    else: // event is not a hidden item, never checking if it's specifically an uncollected hidden item
      if player facing direction is accepted by event:
        return event script
return no event script
```

The only minor possible change I can see is turning the `else` + `if` in the bottom part of the loop into a single `else if`, which does seem to compile down just fine. However, I'm unsure this would be an actual improvement to the clarity of the logic here. Having equal nesting levels for "is hidden item" and "is not hidden item", and then placing secondary conditions within those blocks, seems more sensible to me.

lhearachel: The conjunction-merge here is definitely incorrect, as there is a valid case where the event type is a hidden item, but it is not available, and that case should *not* fall into the `else` block below this.

PR #656, opened 2025-07-30
unpack trtbl.narc and trtblofs.narc

scbroede:
the next step is storing the actual message strings themselves in the json files along with the message types and replacing the existing npc_trainer_messages.gmm with one that's generated from them (an idea cbt had). this would make actually adding/removing messages for a given trainer a lot easier. parking this here for initial feedback before going ahead with that (and also to make sure that's something we want)

this was really annoying because the trainers are ordered in a way that seems mostly arbitrary, so to get it to match i needed a giant list to order them correctly. this is currently dumped in the middle of the cpp file, which probably isn't ideal. ideas for alternate locations would be appreciated

also sorry in advance, my c++ is rusty as hell


scbroede:
this works fine locally, not sure why the build check is failing here

lhearachel:
Uh-oh, 245 is a segfault; can you reproduce this from clean?

scbroede:
ran `make clean` and then `make` again, still works fine for me

tools/datagen/datagen_trainer.cpp
```diff
@@ -228,9 +247,1019 @@ int main(int argc, char **argv)
             std::cerr << e.what() << std::endl;
             std::exit(EXIT_FAILURE);
         }
+        trMsgs[trainerID] = ParseMessages(doc);
+        trainerID++;
     }
 
+    int order[] = {
```
lhearachel: **suggestion**: I would encode this as a `static constexpr std::array`, especially if you know the size at compile-time (which you should...?).

tools/datagen/datagen_trainer.cpp
```diff
@@ -228,9 +247,1019 @@ int main(int argc, char **argv)
...
+        872,
+        873,
+        916,
+        914,
+        915,
+        917,
+        918,
+        919,
+        920,
+        // entries without messages (left out of trtbl)
```
lhearachel: **thought**: This feels like it's a potential foot-gun for users adding messages to trainers. I wonder if there's a way that we can alleviate that.

res/trainers/data/ace_trainer_alyssa.json
```diff
@@ -50,5 +50,11 @@
             "power": 50,
             "ball_seal": 0
         }
+    ],
+    "messages": [
+        "TRMSG_PRE_BATTLE",
```
lhearachel: **thought**: Is there a way that we can map these to the message-archive themselves? I feel like that's what a user would *expect* to be able to edit here. Maybe, for example, this should instead be an object with a pointer to the message text, and then we generate the text archive itself in the same procedure?

tools/datagen/datagen_trainer.cpp
scbroede: i should probably make this comment more clear; including these in the array is an attempt at alleviating that. they're currently left out of trtbl due to being empty, but should be included if a user adds some messages (will have to test that though)

res/trainers/data/ace_trainer_abel.json
```diff
@@ -37,5 +37,10 @@
             "power": 50,
             "ball_seal": 0
         }
-    ]
+    ],
+    "messages": {
+        "TRMSG_PRE_BATTLE": "The Trainers and Pokémon that gather\nhere aren’t pushovers.\fThey all have tricks up their sleeves!\r",
```
cbt6:
```suggestion
        "TRMSG_PRE_BATTLE": [
            "The Trainers and Pokémon that gather\n",
            "here aren’t pushovers.\f",
            "They all have tricks up their sleeves!\r"
        ],
```

tools/datagen/datagen_trainer.cpp
```diff
@@ -195,6 +200,982 @@ static void ParseAndPackParty(const rapidjson::Document &doc, TrainerDataType mo
...
+            message.AddMember("garbage", member["garbage"].GetInt(), out->GetAllocator());
+        }
+
+        trainerMessages.PushBack(message, out->GetAllocator());
+    }
+    *outMessages = trainerMessages;
+    return ret;
+}
+
+static constexpr std::array<int, VANILLA_TRAINER_COUNT> trtblIndices = {
```
lhearachel: **suggestion**: Throw this into a header file just to reduce clutter in the source file. Alternatively, we could store these values in the JSON payloads, but maybe that's a detail that could potentially confuse users.

tools/datagen/datagen_trainer.cpp
```diff
@@ -195,6 +200,982 @@ static void ParseAndPackParty(const rapidjson::Document &doc, TrainerDataType mo
...
+static std::string ParseMessages(const rapidjson::Document &doc, int trainerID, std::string stem, rapidjson::Value *outMessages, rapidjson::Document *out)
+{
+    rapidjson::Value trainerMessages(rapidjson::kArrayType);
+    std::string ret = "";
+    for (const auto &member : doc["messages"].GetArray()) {
+        std::string type = member["type"].GetString();
+        ret += (char)(trainerID & 0xFF);
+        ret += (char)(trainerID >> 8);
+        ret += (char)LookupConst(type, TrainerMessageType);
+        ret += (char)0;
```
lhearachel, line 209: **thought**: There is probably a better way to do this, but I'm drawing a blank at the moment. Maybe you could serialize this using a `struct`, but I'm not sweating that right now.

PR #657, opened 2025-07-30
Attempt at documenting HM Cut-In

YakoSWG:
This is incomplete. I am mostly making this PR now so people are aware I am doing this and don't get blindsided by it.
The main goal of this PR is to document all the tasks and subtask in the hm cut-in and also get the structs reasonably documented.


lhearachel:
Closing in favor of #684

PR #659, opened 2025-07-31
replace remaining script bytes with macros

scbroede:


res/field/scripts/scripts_route_227.s
cbt6, line 1: Should we place `.balign 4, 0` at the end of all script files, even when they are already aligned (like this file for example)?

res/field/scripts/scripts_galactic_hq_3f.s
```diff
@@ -76,6 +76,4 @@ _00B8:
     ReleaseAll
     End
 
-    .byte 0
-    .byte 0
-    .byte 0
+    .balign 4, 0
```
cbt6: Retain newline at end of file (and for all other files too)

res/field/scripts/scripts_route_227.s
scbroede: might be nice in terms of consistency, but i'm not sure it's worth the effort

cbt6:
Thank you for doing this!

PR #667, opened 2025-08-04
Document character sprite retrieval functions

VicSevenT:
Hello,

In this PR I've tried to document `unk_020131EC.c`. The file's purpose looks to be retrieving the sprites for Pokémon and trainers, so I settled on the name `character_sprite.c` for the name based on other uses of the term `character` in the related files.

The file has two main chains of functions. The first is a series only for pokemon that accepts certain pieces of metadata about the sprite being loaded to determine whether Spinda dots need to be drawn. The second is more generic and can be used for both pokemon and trainers.

Besides just loading the sprite data from files, the function then also do some reordering of the pixels. First, the pixels are reordered from a untiled, row-major order to an arrangement of 8x8pixel tiles. Those tiles are then rearranged based on the pattern defined in `SUB_REGION_ORDER`. It is not clear to me why the tiles need to be reordered/why they would have to be out of order in the first place, any insight here or indication that I misinterpreted this section is welcome.

I did my best to rename everything per the contributing guide, but I struggled at times to come up with appropriate names, especially for some of the functions because they are so similar. I'm happy to rename/refactor anything as requested.


src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+    u8 *untiledSpriteData = NULL;
+    u8 *tiledData = NULL;
+    int startIndex;
+    int tiledDataOffset;
+    int spriteWidth;
+
+    GF_ASSERT(monSpriteData->W >= (startCol + width));
+    GF_ASSERT(monSpriteData->H >= (startRow + height));
+
+    untiledSpriteData = monSpriteData->pRawData;
```
h2o-DS, line 52: This data is not untitled, it is raw.

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+static BOOL CharacterSprite_SpeciesIsSpinda(enum Species species)
+{
+    int i;
+    enum Species listItem;
+    const int speciesList[] = {
+        SPECIES_SPINDA,
+        SPECIES_BAD_EGG,
+    };
+
+    for (i = 0; i < 495; i++) {
```
h2o-DS:
```suggestion
    for (i = 0; i < MAX_SPECIES; i++) {
```

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
+#include "character_sprite.h"
+
+#include <nitro.h>
+#include <string.h>
+
+#include "constants/graphics.h"
+#include "constants/heap.h"
+#include "constants/narc.h"
+#include "generated/species.h"
```
h2o-DS: We generally prefer to import `constants/species.h`

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+        }
+    }
+}
+
+static void CharacterSprite_CopyTilesRegion(int sourceWidth, int sourceHeight, const TileRegion *subRegion, int *destOffset, const void *sourceSprite, void *dest)
+{
+    CharacterSprite_CopyTilesRect(sourceWidth, sourceHeight, subRegion->leftMostCol, subRegion->topMostRow, subRegion->width, subRegion->height, destOffset, sourceSprite, dest);
+}
+
+static void CharacterSprite_GetSpriteRect(enum NarcID narcID, int characterIdx, int heapID, int startTileCol, int startTileRow, int width, int height, void *dest)
```
h2o-DS: It feels weird to me to name a function `Get..` without returning anything. Maybe load or something like that?
This applies to similarly named functions.

h2o-DS (commented):
Thank you for your contribution! This looks very good. I have a few small comments to get you started and some insight. What I believe that second rearranging you noticed is the placing of pixels into something called OAMs. They are like groups of tiles. I find it funny that they appear to hard code them instead of use the `NCER` files.

src/character_sprite.c
VicSevenT: That makes sense to me. I actually had used `Load` for some of the void functions, but I guess I wasn't very careful at keeping it consistent through the whole file

src/character_sprite.c
VicSevenT: The variable name is actually _untiled_ (no second "t"). The intent was to distinguish this from the subsequent `tiledData` which is the result of the data reorganization later in the function.

h2o-DS: That is reasonable. My bad for reading too fast.

include/character_sprite.h
```diff
@@ -0,0 +1,28 @@
...
+
+#define FRAME_0_REGION { 0, 0, 10, 10 }
+#define FRAME_1_REGION { 10, 0, 10, 10 }
+
+typedef struct {
+    int leftMostCol;
+    int topMostRow;
+    int width;
+    int height;
+} TileRegion;
```
lhearachel: **suggestion**: `x` and `y` should be sufficient for `leftMostCol` and `topMostRow`; `x` and `y` are generally presumed to apply to the top-left corner of any given image when working with computer graphics.

---

**polish**: Structs should have identical tags and typedefs:

```c
typedef struct TileRegion {
    // ...
} TileRegion;
```

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+#include "narc.h"
+#include "pokemon.h"
+#include "pokemon_sprite.h"
+
+#define SUB_REGION_ORDER { { 0x0, 0x0, 0x8, 0x8 }, { 0x8, 0x0, 0x2, 0x4 }, { 0x8, 0x4, 0x2, 0x4 }, { 0x0, 0x8, 0x4, 0x2 }, { 0x4, 0x8, 0x4, 0x2 }, { 0x8, 0x8, 0x2, 0x2 } }
+
+static void CharacterSprite_CopyTileFromUntiledArray(const u8 *src, u8 *dest, int *srcOffset, int *destOffset, int tileWidth, int rowWidth)
+{
+    int i;
+    for (i = 0; i < TILE_HEIGHT_PIXELS; i++) {
```
lhearachel: **polish**: Declare `i` inline with the loop:

```suggestion
    for (int i = 0; i < TILE_HEIGHT_PIXELS; i++) {
```

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+        *destOffset += tileWidth;
+    }
+}
+
+static void CharacterSprite_LoadSpriteData(enum NarcID narcID, int narcIdx, enum HeapId heapID, NNSG2dCharacterData **charData)
+{
+    BOOL success;
+    void *ngcrFile = NULL;
+
+    ngcrFile = NARC_AllocAndReadWholeMemberByIndexPair(narcID, narcIdx, heapID);
```
lhearachel: **question**: Can these assignments be collapsed and still match?

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+    GF_ASSERT(ngcrFile != NULL);
+
+    success = NNS_G2dGetUnpackedCharacterData(ngcrFile, &(*charData));
+    GF_ASSERT(charData != NULL);
+    GF_ASSERT(success != FALSE);
+
+    Heap_Free(ngcrFile);
+}
+
+static void CharacterSprite_TileUntiledPokemonSprite(int startCol, int startRow, int width, int height, NNSG2dCharacterData *monSpriteData, void *dest)
```
lhearachel: **suggestion**: (applies throughout the file and in any associated declarations in the header)

- `startCol` -> `x`
- `startRow` -> `y`

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+    u8 *tiledData = NULL;
+    int startIndex;
+    int tiledDataOffset;
+    int spriteWidth;
+
+    GF_ASSERT(monSpriteData->W >= (startCol + width));
+    GF_ASSERT(monSpriteData->H >= (startRow + height));
+
+    untiledSpriteData = monSpriteData->pRawData;
+    tiledData = (u8 *)dest;
```
lhearachel: **polish**: Remove unnecessary casts from `void *`; the compiler infers this cast, as per the C standard.

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+    {
+        int row, col;
+        for (row = startRow; row < (startRow + height); row++) {
+            for (col = startCol; col < (startCol + width); col++) {
+                startIndex = (col * TILE_WIDTH_4BPP) + (row * spriteWidth * 8);
+                CharacterSprite_CopyTileFromUntiledArray(untiledSpriteData, tiledData, &startIndex, &tiledDataOffset, TILE_WIDTH_4BPP, spriteWidth);
+            }
+        }
+    }
+}
```
lhearachel: **polish**: Remove unnecessary scope-blocks, and declare `row` and `col` inline, if it still matches.

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+
+static BOOL CharacterSprite_SpeciesIsSpinda(enum Species species)
+{
+    int i;
+    enum Species listItem;
+    const int speciesList[] = {
```
lhearachel, line 69: **polish**: This technically can be `const enum Species`.

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+    void *buffer;
+    int dataSize = ((width * height) * TILE_SIZE_4BPP);
+    buffer = Heap_AllocFromHeap(heapID, dataSize);
+
+    CharacterSprite_LoadTiledPokemonSprite(narcID, narcIdx, heapID, startCol, startRow, width, height, buffer, personality, isAnimated, face, species);
+
+    return buffer;
+}
+
+void CharacterSprite_LoadDataInTiles(enum NarcID narcID, int narcIdx, enum HeapId heapID, int startTileCol, int startTileRow, int width, int height, u8 *dest)
```
lhearachel: **suggestion**: `CharacterSprite_LoadTiledData`, for consistency with the rest of the file.

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+void CharacterSprite_LoadSpriteFrame0(enum NarcID narcID, int narcIdx, enum HeapId heapID, void *dest)
+{
+    const TileRegion region = FRAME_0_REGION;
+    CharacterSprite_LoadSpriteRegion(narcID, narcIdx, heapID, &region, dest);
+}
+
+void *CharacterSprite_GetSprite(enum NarcID narcID, int narcIdx, enum HeapId heapID)
+{
+    void *spriteData = NULL;
+    int size = (10 * 10 * TILE_SIZE_4BPP);
```
lhearachel: **suggestion**: `10` can be pulled from `constants/graphics` as `MON_SPRITE_FRAME_WIDTH_TILES` and `MON_SPRITE_FRAME_HEIGHT_TILES`.

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+    CharacterSprite_LoadSpriteRect(narcID, narcIdx, heapID, region->leftMostCol, region->topMostRow, region->width, region->height, dest);
+}
+
+void CharacterSprite_LoadSpriteFrame0(enum NarcID narcID, int narcIdx, enum HeapId heapID, void *dest)
+{
+    const TileRegion region = FRAME_0_REGION;
+    CharacterSprite_LoadSpriteRegion(narcID, narcIdx, heapID, &region, dest);
+}
+
+void *CharacterSprite_GetSprite(enum NarcID narcID, int narcIdx, enum HeapId heapID)
```
lhearachel: **suggestion**: `CharacterSprite_LoadTiles`

src/character_sprite.c
```diff
@@ -0,0 +1,296 @@
...
+    void *spriteData = NULL;
+    int size = (10 * 10 * TILE_SIZE_4BPP);
+    spriteData = Heap_AllocFromHeap(heapID, size);
+
+    CharacterSprite_LoadSpriteFrame0(narcID, narcIdx, heapID, spriteData);
+
+    return spriteData;
+}
+
+void *CharacterSprite_GetPaletteData(enum NarcID narcID, int paletteIndex, enum HeapId heapID)
```
lhearachel: **suggestion**: `CharacterSprite_LoadPalette`

lhearachel (commented):
This looks good, and thanks for your contribution! I have some suggestions and polishing notes below. :slightly_smiling_face:

src/character_sprite.c
VicSevenT: Adding a const here cause an error
```
..\src\character_sprite.c:75: illegal assignment to constant
```

src/character_sprite.c
VicSevenT: Yup. Made the change and also collapsed a few similar assignments.

VicSevenT:
Conflicts have been fixed.

PR #668, opened 2025-08-08
Document battle anim script funcs

Fexty12573:
This PR documents the script functions callable from battle anim scripts. Specifically the ones callable via the `CallFunc` script command.

A complete list of all of them is in `src/battle_anim/script_func_tables.c`.

Relevant files are:
- `battle_anim/script_funcs_0`
- `battle_anim/script_funcs_1`
- `battle_anim/script_funcs_2`
- `battle_anim/script_funcs_3`
- `battle_anim/script_funcs_4`
- `battle_anim/script_func_examples`
- `battle_anim/script_func_mimic`
- `battle_anim/script_funcs_stat_change`

Additionally, (for the sake of documenting the script funcs) I also documented most of
- `battle_anim/battle_anim_helpers`
- `battle_anim/battle_anim_util`

I also added a ton of new constants in `include/constants/battle/battle_anim.h` and a couple in `include/constants/graphics.h`.

If you're wondering about the `script_funcs_0/1/2/3/4` names, it is because I could not find any pattern in how these functions are divided up. If you have any ideas for what these could be named instead, suggestions are welcome.

I tried to stick to common naming convention for struct fields/constants/enums for each of these script functions but as I have been working on this PR for that last 3 weeks there might be some inconsistencies. I'll be happy to fix them if you see any.

This PR does **not** contain any documentation on the system aside from occasional comments. There will likely be 2 more large PRs, including unpacking the battle anim narcs, before I get to that.

### DIFF
--- a/include/battle_anim/battle_anim_helpers.h
+++ b/include/battle_anim/battle_anim_helpers.h
@@ -104,14 +104,14 @@ typedef struct AfterimageContext {
     u8 unk_AF;
 } AfterimageContext;
 
-typedef struct {
+typedef struct AlphaFadeContext {
     XYTransformContext lerp;
     int done;
 } AlphaFadeContext;
 
 typedef void (*VBlankDMAFunc)(void *);
 
-typedef struct {
+typedef struct VBlankDMAController {
     SysTask *onVBlankTask;
     SysTask *postVBlankTask;
     BOOL doDMA;
@@ -121,13 +121,13 @@ typedef struct {
     void *param;
 } VBlankDMAController;
 
-typedef struct CustomBgScrollContext {
+struct CustomBgScrollContext {
     VBlankDMAController dmaController;
     BufferManager *bufferManager;
     u32 buffer1[HW_LCD_HEIGHT];
     u32 buffer2[HW_LCD_HEIGHT];
     u32 offsetReg;
-} CustomBgScrollContext;
+};
 
 typedef struct BgScrollContext {
     VBlankDMAController dmaController;
@@ -142,7 +142,7 @@ typedef struct SpriteShakeInfo {
     int targets;
 } SpriteShakeInfo;
 
-typedef struct UnkStruct_ov12_022267D4_t {
+typedef struct PaletteFadeContext {
     BOOL active;
     SysTask *task;
     enum PaletteBufferID bufferID;
--- a/include/battle_anim/battle_anim_system.h
+++ b/include/battle_anim/battle_anim_system.h
@@ -131,6 +131,8 @@ typedef struct BattleAnimContext {
     BOOL transformed; // Flag for if transform is currently active
 } BattleAnimContext;
 
+typedef struct CustomBgScrollContext CustomBgScrollContext;
+
 typedef struct BattleAnimScriptLoop {
     u32 *startPos;
     u8 curIteration;
@@ -153,7 +155,6 @@ typedef struct {
     int unk_08;
 } UnkStruct_ov12_022224F8_sub1;
 
-struct CustomBgScrollContext;
 typedef struct {
     UnkStruct_ov12_022224F8_sub1 unk_00[16];
     struct CustomBgScrollContext *unk_C0;
--- a/src/particle_system.c
+++ b/src/particle_system.c
@@ -1,7 +1,7 @@
 #include "particle_system.h"
 
-#include "nitro/fx/fx_trig.h"
 #include <nitro.h>
+#include <nitro/fx/fx_trig.h>
 #include <string.h>
 
 #include "camera.h"


include/battle_anim/battle_anim_helpers.h
```diff
@@ -0,0 +1,232 @@
...
+} AfterimageContext;
+
+typedef struct {
+    XYTransformContext lerp;
+    int done;
+} AlphaFadeContext;
+
+typedef void (*VBlankDMAFunc)(void *);
+
+typedef struct {
```
lhearachel: **polish**: Struct tags

include/battle_anim/battle_anim_helpers.h
```diff
@@ -0,0 +1,232 @@
...
+
+typedef struct SpriteShakeInfo {
+    s16 extentX;
+    s16 extentY;
+    s16 interval;
+    s16 amount;
+    int targets;
+} SpriteShakeInfo;
+
+typedef struct UnkStruct_ov12_022267D4_t {
```
lhearachel: **polish**: Mismatched struct tag

include/battle_anim/battle_anim_system.h
```diff
@@ -146,9 +153,10 @@ typedef struct {
     int unk_08;
 } UnkStruct_ov12_022224F8_sub1;
 
+struct CustomBgScrollContext;
```
lhearachel: **note**: I would prefer that we forward-declare along with the typedef, then provide a definition without a typedef. That also matches other such instances (e.g., in `sprite.h`, `render_window.h`, etc.).

include/battle_anim/battle_anim_util.h
```diff
@@ -0,0 +1,99 @@
...
+void BattleAnimUtil_Free(void *ptr);
+
+static inline BOOL BattleAnimUtil_IsMaskSet(int value, int mask)
+{
+    if ((value & mask) == mask) {
+        return TRUE;
+    }
+
+    return FALSE;
+}
```
lhearachel, line 97: **question**: Is there a similar sort of global inline function available? This feels too generic to stick just in here.

src/battle_anim/script_funcs_stat_change.c
```diff
@@ -0,0 +1,295 @@
...
+
+void BattleAnimScriptFunc_StatChangeMetal(BattleAnimSystem *system)
+{
+    StatChangeParam param;
+
+    StatChangeParam_Init(system, &param);
+    StatChangeParam_SetStepSize(&param, STAT_CHANGE_STEP_X, -2 * STAT_CHANGE_STEP_Y);
+    StatChangeParam_SetInitAlpha(&param, STAT_CHANGE_OVERLAY_START_ALPHA, STAT_CHANGE_OTHER_START_ALPHA);
+    BattleAnimScriptFunc_StatChangeCommon(system, &param);
+}
```
lhearachel, line 295: **question**: What are the contextual use-cases for this vs. the typical up/down?

src/particle_system.c
```diff
@@ -1,5 +1,6 @@
 #include "particle_system.h"
 
+#include "nitro/fx/fx_trig.h"
```
lhearachel: **polish**: Use `<>` for SDK includes.

src/battle_anim/battle_anim_helpers.c
```diff
@@ -0,0 +1,1137 @@
...
+    *px = (1.0f * ctx->x) / MON_AFFINE_SCALE(1);
+    *py = (1.0f * ctx->y) / MON_AFFINE_SCALE(1);
+}
+
+// This function calculates a vertical offset to be applied to a pokemon sprites
+// Y position when the sprite is scaled, to ensure that the sprite appears
+// anchored to the ground. This is necessary because scaling a sprite
+// scales it from its center, which would otherwise cause the sprite to
+// appear to float above the ground when scaled up, or sink below the ground
+// when scaled down.
```
lhearachel, line 507: **praise**: Thanks for this context! 🙏

lhearachel (commented):
I've gone through all but `script_funcs_0`, `2`, and `3`; here's some feedback for now until I get to those.

On the whole, this looks great so far. Thanks for digging into this. ❤️

include/battle_anim/battle_anim_util.h
Fexty12573: This function was originally in `inlines.h` but it was only used in this part of the code so I moved it here. Should I move it back?

src/battle_anim/script_funcs_stat_change.c
Fexty12573: `StatChangeHeal` is used directly by the move Present, as well as when you use a healing/status recovery item in battle.
`StatChangeMetal` is used by Harden, Iron Tail, Metal Claw and Iron Defense.

include/battle_anim/battle_anim_system.h
Fexty12573: Does that still work if there are multiple locations where a forward declaration is necessary? Also to me it feels a bit odd for the typedef to be in a completely different place from the actual definition of the struct.

include/battle_anim/battle_anim_util.h
lhearachel: I think it's fine to have here; I was just curious if anywhere else used it. Probably not as big of a problem as I initially implied.

include/battle_anim/battle_anim_system.h
lhearachel: Is it not possible for the header with the forward-declaration to be included elsewhere? Or does that make a mess of the namespace?

An alternative could be to have a separate `battle_anim_defs.h`, if we really wanted.

Fexty12573: It *can* be included elsewhere, I'm just saying logically you would expect the typedef to be in the same place as the struct itself. Also just for myself, if I end up going over this file again sometime in the future and then I end up thinking "oh, I forgot a typedef here, let me add it". And suddenly there's errors and all that.

Idk I just feel using a non-typedefd forward declaration makes more sense but I guess if it's already like this everywhere else.... yeah

lhearachel (approved):
No further comments. As we discussed on Discord, feel free to move the struct definitions closer to their respective functions in a future PR.

Fexty12573:
rebased

PR #669, opened 2025-08-09
Continue Documenting Naming Screen

ljtpetersen:
I've been splitting this into more bite-sized PRs. Should I instead do the whole file first, and then make a PR with that?

### DIFF
--- a/include/applications/naming_screen.h
+++ b/include/applications/naming_screen.h
@@ -20,14 +20,14 @@ enum NamingScreenType {
 
 typedef struct NamingScreenArgs {
     enum NamingScreenType type;
-    int unk_04;
-    int unk_08;
+    int playerGenderOrMonSpecies;
+    int monForm;
     int maxChars;
-    int unk_10;
+    int monGender;
     int unk_14;
     Strbuf *textInputStr;
     u16 unk_1C[20];
-    int unk_44;
+    int battleMsgID;
     PCBoxes *pcBoxes;
     Options *options;
 } NamingScreenArgs;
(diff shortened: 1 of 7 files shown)


include/applications/naming_screen.h
```diff
@@ -20,14 +20,14 @@ enum NamingScreenType {
 
 typedef struct NamingScreenArgs {
     enum NamingScreenType type;
-    int unk_04;
-    int unk_08;
+    int playerGenderOrMonSpecies;
```
lhearachel, line 23: **question**: Would this make more semantic sense as an anonymous union? Or is that overdoing it?

src/applications/naming_screen.c
```diff
@@ -82,30 +88,38 @@ enum ChangeCharsState {
...
+    int maxChars;
+    int monGender;
+    BOOL printedFromBattleGMM;
     Options *options;
-    UnkStruct_02087A10_sub1 unk_1C;
-    u16 unk_38;
-    u16 unk_3A[6][13];
+    NamingScreenCursor keyboardCursor;
+    u16 plttGlowEffectAngle;
+    u16 keyboardChars[6][13];
```
lhearachel, line 122: **thought**: These might deserve constants for moddability, but I wouldn't block on it right now.

include/applications/naming_screen.h
ljtpetersen: I tried that, but it was a bit of a hassle (it is passed from a function, so it would need to be declared inline), and then the checksum failed. Though I may have made a mistake, I believe it is less hassle in this case to leave it as-is.

src/applications/naming_screen.c
ljtpetersen: I'll look into it after I finish documenting the file.

PR #670, opened 2025-08-09
Document boat cutscene

metallicity:
Documented the cutscene (and surrounding animations/map transitions/etc.) that plays when traveling by boat, initiated by `ScrCmd_23D`. I've renamed this to `PlayBoatCutscene`, but admit that there's a lot more than just a cutscene going on here. Not sure where `unk_0206C784.c` should be placed in the file hierarchy, but it's more or less fully documented now, so should probably get an actual name.


res/field/scripts/scripts_canalave_city.s
```diff
@@ -541,7 +541,7 @@ _0796:
     CallIfEq VAR_0x8004, 1, _0871
     CallIfEq VAR_0x8004, 3, _088B
     CallIfEq VAR_0x8004, 0, _08A5
-    ScrCmd_23D 3, 3, 0x120, 100, 0x1F6
+    PlayBoatCutscene DIR_EAST, DIR_EAST, MAP_HEADER_IRON_ISLAND, 100, 0x1F6
```
lhearachel: **polish**: Let's convert the hex numbers here to decimal, since we know that they are just coordinates.

src/field_map_change.c
```diff
@@ -1550,7 +1550,7 @@ static BOOL FieldTask_ChangeMapColosseum(FieldTask *task)
     return FALSE;
 }
 
-void sub_02054800(FieldTask *task, int mapId, int param2, int x, int z, int param5)
+void sub_02054800(FieldTask *task, int mapId, int warpId, int x, int z, int dir)
```
lhearachel: **suggestion**: Can we name this function, since it's fully documented?

src/unk_0206C784.c
```diff
@@ -27,323 +28,338 @@
...
-    } else if (param1 == 2) {
-        v4 = MAP_PROP_MODEL_SCREW_STEAMSHIP_SPIRAL;
-        TerrainCollisionHitbox_Init(Player_GetXPos(fieldSystem->playerAvatar), Player_GetZPos(fieldSystem->playerAvatar), -2, 2, 6, 3, &v3);
-        v1 = 1;
-        v5 = (12 * 16 * FX32_ONE);
+    if (startDir == DIR_EAST) {
+        targetMapPropModelID = MAP_PROP_MODEL_REGULAR_SHIP;
+        TerrainCollisionHitbox_Init(Player_GetXPos(fieldSystem->playerAvatar), Player_GetZPos(fieldSystem->playerAvatar), 1, -3, 3, 6, &hitbox);
+        moveBeforeFadeOut = TRUE;
+        goalDistance = (25 * 16 * FX32_ONE);
```
lhearachel: **question**: Any idea what `25` and `16` refer to here? My hunch is that they are related to map coordinates...?

src/unk_0206C784.c
```diff
@@ -27,323 +28,338 @@
...
     default:
-        GF_ASSERT(0);
-        return 1;
+        GF_ASSERT(FALSE);
+        return TRUE;
     }
 
-    if (param0->unk_7C < (FX32_ONE)) {
-        param0->unk_03++;
+    if (boatCutscene->speed < (FX32_ONE)) {
```
lhearachel: **polish**: Remove extraneous parens (throughout the file)

lhearachel (commented):
> Not sure where `unk_0206C784.c` should be placed in the file hierarchy, but it's more or less fully documented now, so should probably get an actual name.

I don't think we have a dedicated place for cutscenes at the moment; I would just give the files good names (probably just `boat_cutscene` or `cutscene_boat`, take your pick).

src/unk_0206C784.c
```diff
@@ -27,323 +28,338 @@
...
 
-    FieldTask_InitCall(fieldSystem->task, sub_0206C964, v0);
+    FieldTask_InitCall(fieldSystem->task, FieldSystem_PlayBoatCutsceneStep, boatCutscene);
 }
 
-static void sub_0206C8D4(FieldSystem *fieldSystem, const u8 param1, ModelAttributes *param2)
+static void FieldSystem_PlayBoatTravelCutscene(FieldSystem *fieldSystem, const u8 startDir, ModelAttributes *areaModelAttrs)
 {
-    UnkStruct_0206C8D4 *v0 = Heap_AllocFromHeapAtEnd(HEAP_ID_FIELDMAP, sizeof(UnkStruct_0206C8D4));
+    UnkStruct_0206C8D4 *taskEnv = Heap_AllocFromHeapAtEnd(HEAP_ID_FIELDMAP, sizeof(UnkStruct_0206C8D4));
```
lhearachel: **thought**: I think that we can also document this struct, and maybe pull it into this module's header file?

src/unk_0206C784.c
metallicity: Seems like this is better written as `25 * MAP_OBJECT_TILE_SIZE` (i.e., the number of pixels in 25 map tiles).

src/unk_0206C784.c
metallicity: I originally stopped shy of documenting the internals of playing the actual cutscene itself, and this was the boundary to that section of the code. However, in verifying the behavior of the fields of this struct, I realized it actually wasn't that bad, since all it was really doing was loading model/animation data from a .narc file, using different members of the file based on the value of `startDir`. So I went ahead and documented all of `overlay93` and moved those two files (which are nearly identical) into a new `boat_cutscene` folder, along with the top level file that calls into them.

include/boat_cutscene/boat_cutscene.h
```diff
@@ -0,0 +1,29 @@
+#ifndef POKEPLATINUM_BOAT_CUTSCENE_H
+#define POKEPLATINUM_BOAT_CUTSCENE_H
+
+#include "field/field_system_decl.h"
+#include "overlay005/model_attributes.h"
+
+#define BOAT_TRAVEL_CUTSCENE_HEAP_SIZE 0x20000
```
lhearachel, line 11: **polish**: This belongs in `constants/heap.h`, if it's to be publicly exposed.

src/boat_cutscene/ship_demo.c
```diff
@@ -0,0 +1,267 @@
...
+static void BoatCutscene_ShipDemo_Init3D(void);
+static void BoatCutscene_ShipDemo_SetGXBanks(void);
+static void BoatCutscene_ShipDemo_LoadAnimations(BoatCutscene_ShipDemo *cutsceneData);
+
+static const CameraConfiguration cameraConfig = {
+    .distance = BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_DISTANCE,
+    .cameraAngle = {
+        BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_ANGLE_X,
+        BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_ANGLE_Y,
+        BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_ANGLE_Z },
```
lhearachel: **polish**: Insert a trailing comma after the final element of this array so that it auto-formats a bit more cleanly:

```suggestion
    .cameraAngle = {
        BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_ANGLE_X,
        BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_ANGLE_Y,
        BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_ANGLE_Z,
     },
```

src/boat_cutscene/boat_cutscene.c
```diff
@@ -115,63 +116,63 @@ void FieldSystem_PlayBoatCutscene(FieldSystem *fieldSystem, const u8 startDir, c
...
-    case 0:
-        if ((taskEnv->unk_00 == DIR_SOUTH) || (taskEnv->unk_00 == DIR_WEST)) {
-            sub_0203E2D4(fieldSystem, taskEnv);
+    case BOAT_TRAVEL_CUTSCENE_TASK_STATE_START:
+        if ((taskEnv->startDir == DIR_SOUTH) || (taskEnv->startDir == DIR_WEST)) {
+            FieldTask_PlayBoatCutscene_ShipDemoPl(fieldSystem, taskEnv);
         } else {
-            sub_0203E2AC(fieldSystem, taskEnv);
+            FieldTask_PlayBoatCutscene_ShipDemo(fieldSystem, taskEnv);
         }
```
lhearachel: **thought**: I feel that this naming is inaccurate; why the `Pl` suffix vs. not? The implication of the suffix is that the former would be Platinum-exclusive, but it's not clear to me why this routine would not be available in Diamond and Pearl.

include/boat_cutscene/boat_cutscene.h
```diff
@@ -0,0 +1,29 @@
...
+#define BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_ANGLE_Y        0
+#define BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_ANGLE_Z        0
+#define BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_DISTANCE       0x29AEC1
+#define BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_PROJECTION_MTX 0
+#define BOAT_TRAVEL_CUTSCENE_CAMERA_BASE_FOVY           1473
+
+#define BOAT_TRAVEL_CUTSCENE_NUM_LIGHTS 4
+
+#define BOAT_TRAVEL_CUTSCENE_STATE_SFX      0
+#define BOAT_TRAVEL_CUTSCENE_STATE_FADE_OUT 1
```
lhearachel, line 23: **note**: I don't see any reason why these constants need to be exposed on first glance; is there a reason to introduce them into the global namespace?

include/boat_cutscene/ship_demo.h
lhearachel, line 1: **thought**: Could this be a single-header application? i.e., are `ship_demo.h` and `ship_demo_pl.h` always included alongside `boat_cutscene.h`? If so, feel free to unify them.

include/boat_cutscene/ship_demo.h
```diff
@@ -0,0 +1,10 @@
+#ifndef POKEPLATINUM_BOAT_CUTSCENE_SHIP_DEMO_H
+#define POKEPLATINUM_BOAT_CUTSCENE_SHIP_DEMO_H
+
+#include "overlay_manager.h"
+
+int BoatCutscene_ShipDemo_Init(ApplicationManager *appMan, int *state);
+int BoatCutscene_ShipDemo_Main(ApplicationManager *appMan, int *state);
+int BoatCutscene_ShipDemo_Exit(ApplicationManager *appMan, int *state);
```
lhearachel: **polish**: Use `BOOL` for the return type for these function-signatures.

src/boat_cutscene/ship_demo.c
lhearachel, line 1: **question**: Could you post a diff of this file compared to `ship_demo_pl.c`? I'm curious to see what the precise differences are.

src/boat_cutscene/boat_cutscene.c
metallicity: There was a [new ship design](https://bulbapedia.bulbagarden.net/wiki/S.S._Sinnoh#Appearance) introduced in Platinum, where previously the same ship model was recycled for both Canavale and Snowpoint. Though I admit I misunderstood the directions here a little. Once I paid more attention to how the ports were actually positioned, it became obvious that the "regular" ship in Canalave was saying it went N->S and S->N, while the new "drill" ship in Snowpoint was saying it went W->E and E->W, despite this direction variable not lining up with the typical `DIR_` constants. I also decided to use the names of their port cities to refer to the ships, since that seemed the clearest to me.

On a secondary note, it seems that `pokediamond` is a little further along in breaking down the `ship_demo.narc` file here, which is identical between the two games. It's already split up into the individual model and animation members, though I'm unsure how hard it'd be to adopt their build pipeline for it.

include/boat_cutscene/boat_cutscene.h
metallicity: These are all shared between both files for the individual ships, though if those files were unified somehow then they could be moved out of this header, since `boat_cutscene.c` doesn't reference any of them, nor do any other files elsewhere.

src/boat_cutscene/ship_demo.c
metallicity: The diffs are a little noisy due to differently named functions/structs/constants, but the only real differences are having `BOAT_TRAVEL_CUTSCENE_NUM_ANIMATIONS` be 3 for the Snowpoint ship rather than 4 for the Canalave ship, and then the contents of `narcMemberIndexes` which access different members from their respective `.narc` files. I took a brief stab of trying to reduce the redundancy here, but couldn't figure out anything that I was particularly happy with.

lhearachel: No need to reduce the redundancy here, I think; I was more just curious about what the tangible differences were. 😁

asm/macros/scrcmd.inc
```diff
@@ -3215,15 +3216,31 @@
...
+    PlayBoatCutscene BOAT_TRAVEL_DIR_WEST_TO_EAST, \exitDir, \mapID, \x, \z
+    .endm
+
+    .macro TakeShipToCanalave
+    PlayBoatCutscene BOAT_TRAVEL_DIR_SOUTH_TO_NORTH, DIR_WEST, MAP_HEADER_CANALAVE_CITY, 44, 750
+    .endm
+
+    .macro TakeShipToSnowpoint
+    PlayBoatCutscene BOAT_TRAVEL_DIR_EAST_TO_WEST, DIR_NORTH, MAP_HEADER_SNOWPOINT_CITY, 356, 246
+    .endm
```
lhearachel: **praise**: Great addition! ❤️

generated/boat_travel_dirs.txt
lhearachel, line 1: **suggestion**: If this is never used as an `enum`, then there's no need to have this be `generated` like this; instead, prefer to use plain `#define`s in some file in `include/constants`.

The only reason for this folder is to facilitate using `enum`s in C code while also having `#define`s that can be understood by the script assembler.

PR #671, opened 2025-08-09
document some underground inventory and spheres code

scbroede:
- left documenting the messages out for now due to the impending conversion to json
- not a fan of the functions with `2` at the end of the name. will look into finding a different prefix for those once i properly tackle that file (something like "undergroundinventory" or "undergroundmenus" maybe?)


src/overlay023/underground_spheres.c
```diff
@@ -0,0 +1,584 @@
...
+    if (buriedSpheresEnv->sysTask) {
+        SysTask_Done(buriedSpheresEnv->sysTask);
+        Heap_Free(buriedSpheresEnv->sphereRadarTimer);
+
+        buriedSpheresEnv->sysTask = NULL;
+        buriedSpheresEnv->sphereRadarTimer = NULL;
+    }
+}
+
+int SphereRadar_GetXCoordOfBuriedSphere(int param0)
```
lhearachel, line 560: **polish**: Document `param0`, if a name comes to mind.

lhearachel (approved):
One minor polish comment below, non-blocking.

src/overlay023/underground_spheres.c
scbroede: i was going to hold off on that until the radar code was better documented because i'm not really sure why this parameter even exists at the moment

lhearachel: No sweat, we can leave it for later. :grin:

PR #678, opened 2025-08-12
Document misc commands

metallicity:
Documented a handful of more standalone commands, which largely didn't call anywhere deeper into undocumented areas of the repo. Again, if it's preferable to have these as separate PRs, just let me know, but I tried to omit any commands which felt too large to get bundled together like this.

## Commands Documented

- **SetPosition** (from `ScrCmd_187`): Sets the position and facing direction of a map object
- **InitTurnbackCave** (from `ScrCmd_285`): Manages the pillar room navigation logic for Turnback Cave
- **SelectMoveTutorPokemon** (from `ScrCmd_191`): Shows party screen to select a move tutor pokemon
- **Duplicate_IncrementTrainerScore** (from `ScrCmd_260`): Duplicate of `IncrementTrainerScore`
  - Both commands are in used, called for non-overlapping trainer score events
  - Also documented all trainer score event constants called with it
- **Show/HideObject** (from `ScrCmd_1B1/1B2`): Controls status flag for whether to hide a map object
- **TryRevertPokemonForm** (from `ScrCmd_31E`): Attempts to revert a pokemon to its base form
  - Can fail for Origin Forme Giratina when trying to place its Griseous Orb into a full bag
- **TrySaveGame** (from `ScrCmd_12D`): Prompts the player to save the game
- **Show/HideSavingIcon** (from `ScrCmd_18D/18E`): Shows or hides saving spinner, called surrounding `TrySaveGame`
- **CheckHeapMemory** (from `ScrCmd_313`): Asserts free memory size of certain heaps goes unchanged
  - When passed `FALSE`, stores current free memory of both heaps (plus an extraneous one)
  - When passed `TRUE`, asserts current free memory of both heaps equal last stored values


res/field/scripts/scripts_battle_tower_battle_room.s
```diff
@@ -18,12 +18,12 @@ _0035:
     End
 
 _0037:
-    ScrCmd_1B2 0xFF
+    HideObject 0xFF
```
lhearachel: **polish**: Replace references to object `0xFF` with `LOCALID_PLAYER` from `include/constants/scrcmd.h`.

res/field/scripts/scripts_canalave_city_gym.s
```diff
@@ -30,7 +30,7 @@ _0016:
     PlaySound SEQ_BADGE
     WaitSound
     GiveBadge BADGE_ID_MINE
-    ScrCmd_260 23
+    Duplicate_IncrementTrainerScore TRAINER_SCORE_EVENT_BADGE_EARNED
```
lhearachel: **suggestion**: Our usual notation for such instances is something like `IncrementTrainerScore2`. It sucks, but it is what it is.

src/scrcmd.c
```diff
@@ -7875,28 +7875,28 @@ static BOOL ScrCmd_ResetDistortionWorldPersistedCameraAngles(ScriptContext *ctx)
     return FALSE;
 }
 
-u32 Unk_021C07E8;
-u32 Unk_021C07E4;
-u32 Unk_021C07EC;
+u32 fieldMapLastFreeMemory;
+u32 fieldLastFreeMemory;
+u32 fieldTaskLastFreeMemory;
```
lhearachel: **polish**: Mark these as `static` and prefix them with `s` accordingly.

lhearachel (commented):
This looks good! A few comments below.

PR #680, opened 2025-08-12
Name item use functions

ItsLeaze:
There's more to do, but this names the entrypoints and collects type definitions.


src/item_use_functions.c
```diff
@@ -68,147 +65,119 @@
...
+    { NULL, UseGenericItemInField, NULL },
+    { UseEvoStoneFromMenu, NULL, NULL },
+    { UseEscapeRopeFromMenu, NULL, CanUseEscapeRope },
+    { UseAzureFluteFromMenu, UseAzureFluteInField, CanUseAzureFlute },
+    { UseVsRecorderFromMenu, UseVsRecorderInField, NULL },
+    { UseGracideaFromMenu, UseGracideaInField, NULL }
 };
 
-u32 sub_020683F4(u16 param0, u16 param1)
+u32 GetItemUseFunction(u16 discriminator, u16 functionIdx)
```
lhearachel, line 169: **suggestion**: Make an `enum` for use as `functionIdx` (though it likely must stay as a `u16` here to match) and use that `enum`'s values as array-initializers in the data table above.

src/item_use_functions.c
```diff
@@ -68,147 +65,119 @@
...
-    { sub_02068E38, sub_02068E74, sub_02068E94 },
-    { NULL, sub_02068EFC, NULL },
-    { sub_02068FEC, NULL, NULL },
-    { sub_02069080, NULL, sub_020690C4 },
-    { sub_02069110, sub_02069120, sub_02069130 },
-    { sub_02069188, sub_020691BC, NULL },
-    { sub_020691E0, sub_02069218, NULL }
+static void PrintRegisteredKeyItemError(ItemFieldUseContext *param0, u32 param1);
+
+static const ItemUseFuncDat itemFieldUseFuncs[] = {
```
lhearachel: **polish**: `sItemUseFuncs`

src/item_use_functions.c
```diff
@@ -518,39 +487,39 @@ static u32 sub_02068948(const UnkStruct_020684D0 *param0)
...
         MapObject *v1 = Player_MapObject(param0->playerAvatar);
 
-        if ((MapObject_IsOnBikeBridgeNorthSouth(v1, param0->unk_0E) == 1) || (MapObject_IsOnBikeBridgeEastWest(v1, param0->unk_0E) == 1)) {
+        if ((MapObject_IsOnBikeBridgeNorthSouth(v1, param0->currTileBehavior) == 1) || (MapObject_IsOnBikeBridgeEastWest(v1, param0->currTileBehavior) == 1)) {
             return 1;
         }
     }
 
-    if ((TileBehavior_IsVeryTallGrass(param0->unk_0E) == 1) || (TileBehavior_IsMud(param0->unk_0E) == 1) || (TileBehavior_IsMudWithGrass(param0->unk_0E) == 1)) {
+    if ((TileBehavior_IsVeryTallGrass(param0->currTileBehavior) == 1) || (TileBehavior_IsMud(param0->currTileBehavior) == 1) || (TileBehavior_IsMudWithGrass(param0->currTileBehavior) == 1)) {
```
lhearachel: **polish**: Use `TRUE` / `FALSE` for bool-like values.

---

**polish**: Remove extraneous parens.

src/item_use_functions.c
```diff
@@ -844,26 +813,26 @@ static void sub_02068E38(UnkStruct_02068630 *param0, const UnkStruct_020684D0 *p
...
 {
     void *v0 = ov5_021F08CC(param0->fieldSystem, HEAP_ID_FIELD, 2);
 
     FieldSystem_CreateTask(param0->fieldSystem, ov5_021F08F8, v0);
-    return 0;
+    return FALSE;
 }
 
-static u32 sub_02068E94(const UnkStruct_020684D0 *param0)
+static u32 CanUseFishingRod(const ItemUseContext *param0)
```
lhearachel: **polish**: Name `param0` (applies throughout the file, where applicable).

src/item_use_functions.c
```diff
@@ -844,26 +813,26 @@ static void sub_02068E38(UnkStruct_02068630 *param0, const UnkStruct_020684D0 *p
...
 {
-    if (param0->unk_04 == 1) {
+    if (param0->hasPartner == 1) {
         return 2;
     }
 
-    if ((param0->unk_00 == 573) || (param0->unk_00 == 574) || (param0->unk_00 == 575) || (param0->unk_00 == 576) || (param0->unk_00 == 577) || (param0->unk_00 == 578) || (param0->unk_00 == 579) || (param0->unk_00 == 580) || (param0->unk_00 == 581) || (param0->unk_00 == 582) || (param0->unk_00 == 583)) {
+    if ((param0->mapHeaderID == 573) || (param0->mapHeaderID == 574) || (param0->mapHeaderID == 575) || (param0->mapHeaderID == 576) || (param0->mapHeaderID == 577) || (param0->mapHeaderID == 578) || (param0->mapHeaderID == 579) || (param0->mapHeaderID == 580) || (param0->mapHeaderID == 581) || (param0->mapHeaderID == 582) || (param0->mapHeaderID == 583)) {
```
lhearachel: **polish**: Use `MAP_HEADER_*` constants for these values (`generated/map_headers.h`). You can view the values themselves by opening `build/generated/map_headers.h` in your editor.

I would also check if this gets optimized to a range-comparison in the assembly. Otherwise, separate this into one check per line, with `||` operators at the start of each new line.

lhearachel (commented):
Nice work! This one has been waiting to be documented for a while. 🙂

src/item_use_functions.c
ghost: I'm not sure I see the benefit, since the indices are only used in the `pl_item_data.csv` file if I'm not mistaken.

src/item_use_functions.c
ghost: I gave them trivial names. If case insufficient or misleading, feel free to criticize. Though my goal with this iteration is not perfection.

src/item_use_functions.c
ghost: Nice, looks way better.
`usageContext->mapHeaderID >= MAP_HEADER_DISTORTION_WORLD_1F && usageContext->mapHeaderID <= MAP_HEADER_DISTORTION_WORLD_TURNBACK_CAVE_ROOM` results in checksum failures

src/item_use_functions.c
lhearachel: Ah! They should actually be `#define`s, then, but that's a detail of how `csv2bin` works.

Their usage in `pl_item_data.csv` is actually a benefit, because it creates a semantic link between records in that file and the table entries in this file. There is also a tertiary benefit in that it makes it more difficult for a user to mistakenly reorder the table and create unexpected behavior (which can be difficult to debug).

PR #681, opened 2025-08-13
Use JSON for text banks

lhearachel:
This is a relatively large PR in terms of number of files touched, so I've split the PR up into multiple commits in an attempt to make it easier to review.

The two primary bits here are the changes to tooling, which add support for JSON as a text-bank format. The gist of the format, as we previously discussed in the Discord channel, is this:

```json
{
    "key": 65535,
    "messages": [
      {
          "id": "message_00000",
          "en_US": "This is a single-line message."
      },
      {
          "id": "message_00001",
          "en_US": "This is also a single-line message.\n"
      },
      {
          "id": "message_00002",
          "en_US": [
              "This is a multi-line message.\f",
              "All strings will be joined together\n",
              "before encoding the full message.\r"
          ]
      }
    ]
}
```

This would also output the following C header:

```c
// This is a single-line message.
#define message_00000 0
// This is also a single-line message.
#define message_00001 1
// This is a multi-line message.
// All strings will be joined together
// before encoding the full message.
#define message_00002 2
```

## Other Changes

- Clarify `msgenc`'s program arguments and options (which were inaccurately documented in the help-text)
- Apply standard warnings to `msgenc` and fix them
- New patches for `rapidjson`; **projects should remove their existing `subprojects/rapidjson-1.1.0` directory before rebuilding**


res/text/ability_names.json
```diff
@@ -0,0 +1,375 @@
+{
+  "key": 7280,
+  "pl_msg_00000610_00000": {
```
Kuruyia: **question:** Do we want to mix entries statically defined in msgenc (here, `key`) with the message entries (here, `pl_msg_x`) at the same level?

Thinking about something like this instead:
```json5
{
  "key": 7280,
  "messages": {
    "pl_msg_00000610_00000": {
      "en_US": " -"
    },
    // ...
  }
}
```

This is more of a nitpick. I wouldn't expect any conflicts between the names of those entries.

This would be useful if we want to add other (non-message) entries to this file later on tho', as we'd have a clear separation between both type of entries. This would also allow to easily infer the number of messages in a text bank.

lhearachel: This was brought up during discussions about the format, but my opinion is that it creates an unnecessary nesting. I don't think that we will realistically ever add more fields to these objects (famous last words).

If we were being strict to-the-letter about our JSON inputs, we would probably want something like this, but it would be an array, and individual messages would be objects that define their key as an inner-member. The current structure maintains ordering for most major libraries, as far as I'm aware, since they maintain key-order; this is true of Python's `json` module (which uses the default-ordered `dict`) and RapidJSON for C++.

res/text/acuity_cavern.json
```diff
@@ -0,0 +1,12 @@
...
+  "AcuityCavern_Text_UxieCry": {
+    "en_US": "Kyouuuun!\\r"
+  },
+  "AcuityCavern_Text_UxieDisappeared": {
+    "en_US": [
+      "UXIE disappeared deep into its\\n",
+      "cavern..."
+    ]
+  }
+}
```
cbt6:
```suggestion
{
  "key": 9291,
  "messages": [
    {
      "id": "AcuityCavern_Text_UxieCry",
      "message": "Kyouuuun!\\r"
    },
    {
      "id": "AcuityCavern_Text_UxieDisappeared",
      "message": [
        "UXIE disappeared deep into its\\n",
        "cavern..."
      ]
    }
  ]
}
```
The message ids cannot be used as keys since JSON does not preserve ordering of keys, so they have to be placed in an array.

lhearachel: It is implementation-dependent, as explained above. Some libraries will maintain key-ordering by default. Some may not, but still support doing so with configuration. Python's native `json` library maintains key-ordering by default, as does RapidJSON for C++. Even `nlohmann/json` can be configured to maintain key-ordering.

There's, of course, an argument to force ordering with an array. We would need to fix other data structures if we want to make that a policy. [`res/pokemon/move_tutors.json`](https://github.com/pret/pokeplatinum/blob/65d0e9daff6347c5bb95e291a8489e0578143dfb/res/pokemon/move_tutors.json) comes to mind.

cbt6: Sorry, missed out on the earlier comment. But that's exactly the point - we should follow the spec instead of relying on implementations of individual libraries.

lhearachel: I can get behind spec-compliance as a general-purpose rule.

We'll need to do an audit of existing files to update them to be spec-compliant. I can make an issue for that as follow-up to this PR. For this PR specifically, I'll update the existing code to handle this proposed spec:

```json
{
  "key": 9291,
  "messages": [
    {
      "id": "AcuityCavern_Text_UxieCry",
      "en_US": "Kyouuuun!\\r"
    },
    {
      "id": "AcuityCavern_Text_UxieDisappeared",
      "en_US": [
        "UXIE disappeared deep into its\\n",
        "cavern..."
      ]
    }
  ]
}
```

(Replacing `message` with `en_US`, as requested for future localization extensions)

lhearachel:
Formatting of the message-banks has been updated as discussed above. An additional feature has been added to the charmap reader that permits reading ASCII control characters (namely, `\n`, `\r`, and `\f`) in either escaped or unescaped form. This allows the JSON banks to use unescaped forms in the string input while maintaining compatibility with the existing XML format (as `pugixml` will ingest the escaped-form).

lhearachel:
This PR shall not be merged until the queue is devoid of PRs touching existing text banks.

PR #682, opened 2025-08-13
Finish Documenting Naming Screen and Extract `namein` NARC

ljtpetersen:


res/graphics/naming_screen/meson.build
```diff
@@ -0,0 +1,96 @@
...
+    'naming_screen_dummy_0.png',
+    'naming_screen_unk_tiles.png',
+)
+
+naming_screen_sprites_pngs = files(
+    'naming_screen_sprites.png',
+    'naming_screen_unk_sprite.png',
+)
+
+naming_screen_pals = files('naming_screen.pal', 'naming_screen_sprites.pal')
```
h2o-DS: I think it would be slightly preferable to give each file its own line

res/text/generic_names.gmm
```diff
@@ -1,147 +1,147 @@
 <?xml version="1.0"?>
 <body language="English">
 	<key value="59361" />
-	<row id="pl_msg_00000427_00000" index="0">
+	<row id="GenericNames_PlayerMale_First" index="0">
```
h2o-DS, line 4: Are you doing some sort of parse later? Why is this not just name 0?

src/applications/naming_screen.c
```diff
@@ -2230,7 +2325,7 @@ static void NamingScreen_InitWindows(NamingScreen *namingScreen, ApplicationMana
     if (namingScreen->type == NAMING_SCREEN_TYPE_UNK4) {
         Graphics_LoadTilemapToBgLayerFromOpenNARC(
             narc,
-            6 + 3,
+            naming_screen_chars_bg_0_NSCR_lz + 3,
```
h2o-DS: What does the 3 represent?

src/unk_0203D1B8.c
```diff
@@ -1170,23 +1170,23 @@ static BOOL sub_0203DE98(FieldTask *param0)
...
-                v2->unk_0C->unk_14 = 1;
+                v2->unk_0C->noInput = 1;
             }
         } else if (v2->unk_0C->type == NAMING_SCREEN_TYPE_GROUP) {
             const u16 *v3 = Strbuf_GetData(v2->unk_0C->textInputStr);
             RecordMixedRNG *v4 = SaveData_GetRecordMixedRNG(fieldSystem->saveData);
 
             if (RecordMixedRNG_DoesCollectionContainGroup(v4, v3)) {
-                v2->unk_0C->unk_14 = 2;
+                v2->unk_0C->noInput = 2;
```
h2o-DS: What does 2 mean? I think this should get an enum.

h2o-DS (changes requested):

res/graphics/naming_screen/meson.build
ljtpetersen: Sure.

res/text/generic_names.gmm
ljtpetersen: The code doesn't refer to each generic name individually. Instead it picks a random number between the first and the last one, inclusive. So I used first to denote the start of the range and last to denote the end. I could definitely use 0 as well though.

src/applications/naming_screen.c
ljtpetersen: It doesn't reflect anything in particular. I have folded it in and replaced it with the appropriate index constant.

src/unk_0203D1B8.c
ljtpetersen: I'm not sure exactly what 2 means, in this context. The `noInput` field is never actually read by the naming screen. It serves as a return value for the result of the naming screen, `TRUE` if no name was input, `FALSE` otherwise. However, this function re-uses the `noInput` field to contain a different value, which, as best I can tell, is returned for certain script commands that open the naming screen.

The reason I hesitate to give this an enum is because the value 2 will never be returned by the naming screen, so it would be a bit misleading. The value 2 here is also only read from the `noInput` field by the same function that assigns it, after which the structure is freed. So, I would treat it more as a special value for this function than a value that should be associated with the naming screen.

lhearachel: > The `noInput` field is never actually read by the naming screen. It serves as a return value for the result of the naming screen, `TRUE` if no name was input, `FALSE` otherwise. However, this function re-uses the `noInput` field to contain a different value, which, as best I can tell, is returned for certain script commands that open the naming screen.

This sounds like it is actually just a return code, then...? What script commands reference it?

ljtpetersen: `ScrCmd_0BB`, `ScrCmd_21D`, and `ScrCmd_271` use it. It is a return code for these script commands.

lhearachel: I think we might want to re-classify this as an `exitCode`, similar to the bag interface, then.

h2o-DS (approved):

PR #687, opened 2025-08-16
Common script POC

metallicity:
This is just a quick example of how a few simple macros could allow `scripts_common.s` and calls to `CallCommonScript` to share the same script name, despite one wanting to use a label and the other wanting to use a constant for the script ID. This is achieved by replacing all `ScriptEntry` and their labels with macros that instead use `Common_\scriptName` in place of those labels. This ends up with `CommonScriptEntry HandleSignpostInput` at the top, with `DefCommon HandleSignpostInput:` at the actual label itself, and then `CallCommonScript HandleSignpostInput` to actually call into the script.

(note that under the hood I believe these labels actually resolve to, e.g. `Common_2000`, using the given script ID, which shouldn't matter too much, and I couldn't figure out a better way to do it)


asm/macros/scrcmd.inc
```diff
@@ -152,6 +153,14 @@
     .short \scriptID
     .endm
 
+    .macro CommonScriptEntry scriptName
```
CharlesFolz4, line 156: This pair of macros solves an issue I ran into while documenting a few of the common scripts, do you mind if I take them?

metallicity: Fine with me. I made a few tweaks for better safety against namespace collisions while working on with it on a [test case](https://github.com/metallicity/pokeplatinum/commit/a65dffd3fb93883ffee2095309ad572ab30369c3) (using the common scripts for pokedex ratings found in a separate scripts file). It's really just a matter of ensuring all common script ID constants have a prefix (e.g. `CommonScript_HandleSignpostInput` rather than just `HandleSignpostInput`), and changing the actual macros so that the common script label prefix is a bit longer, since it isn't very visible anyway:

```
    .macro CommonScriptEntry scriptName
    ScriptEntry CommonScript_Label_\scriptName
    .endm

    .macro DefCommonScript scriptName
    CommonScript_Label_\scriptName
    .endm
```

The other catch (as I brought up in discord) is that you need to give an additional label to use for any references to a common script that gets called (or jumped to) within the same file. E.g., [PokedexCommon_Internal_CheckNationalDex](https://github.com/metallicity/pokeplatinum/blob/a65dffd3fb93883ffee2095309ad572ab30369c3/res/field/scripts/scripts_pokedex_common.s#L61) placed directly after the `DefCommonScript PokedexCommon_CheckNationalDex:` macro.

metallicity:
Just wanted to add for anyone reviewing this, I originally tried implementing this the other way around, where the labels were the actual symbol and the script IDs had a hidden prefix. This would have eliminated the need for any macros for declaring common scripts, and instead just require a wrapper macro over `CallCommonScript` to add the prefix, at the marginal cost of having any references to these scripts within C use the full prefixed script ID.

However, before any tradeoffs could be considered there, this approach simply didn't work. While the macros could successfully assemble, e.g., `ScriptID_CommonScript_Unk01`, this occurred only after the pass which replaced all C defines with their literal values, and so was treated as a non-existent label rather than that constant. Which means that this current approach where a symbol trying to become part of a prefixed label is instead treated as a constant and replaced is actually the easier behavior to work around here.

metallicity:
Since #688 has mostly taken on this approach, I'm closing this to avoid any potential confusion. I left it up mainly because I felt it demonstrated a better set of names for the common script constants, compared to what is currently being proposed to be merged, but that's probably not enough of a reason on its own.

PR #688, opened 2025-08-16
Document several common scripts called from CallCommonScript

CharlesFolz4:
This was originally going to be just documenting the common scripts for adding item(s) to the bag, but ended up documenting most of the common scripts called by `CallCommonScript`.  Several of the scripts in `scripts_common` are not called by that, and those ones have not been looked at.

~~Macro suggestions from PR 687 has not been implemented, pending further feedback.~~
Macro suggestions from PR 687 have now been implemented.

Message IDs have not been documented, to reduce potential friction with ongoing json conversion effort.

### DIFF
--- a/asm/macros/scrcmd.inc
+++ b/asm/macros/scrcmd.inc
@@ -4030,7 +4030,7 @@
     .short \level
     .endm
 
-    .macro GetTrainerCardStars destVar
+    .macro GetTrainerCardLevel destVar
     .short 702
     .short \destVar
     .endm
--- a/res/field/scripts/scripts_common.s
+++ b/res/field/scripts/scripts_common.s
@@ -4,64 +4,64 @@
 #include "res/text/bank/common_strings.h"
 #include "res/text/bank/menu_entries.h"
 
-    ScriptEntry CommonScript_HandleSignpostInput
-    ScriptEntry CommonScript_CheckBagPocketForItem
-    ScriptEntry CommonScript_Pokecenter_SpeakWithNurse
-    ScriptEntry _03E8
-    ScriptEntry _043B
-    ScriptEntry _0479
-    ScriptEntry CommonScript_SaveGame
-    ScriptEntry CommonScript_EmptyScript2
-    ScriptEntry _05EA
-    ScriptEntry CommonScript_ObtainPoketchApp
-    ScriptEntry CommonScript_EmptyScript1
-    ScriptEntry CommonScript_EmptyScript3
-    ScriptEntry CommonScript_SendToUndergroundPC
-    ScriptEntry CommonScript_ObtainUndergroundTrap
-    ScriptEntry CommonScript_ObtainUndergroundSphere
-    ScriptEntry CommonScript_ObtainAccessoryWaitForConfirm
-    ScriptEntry CommonScript_AddItemQuantityNoLineFeed
-    ScriptEntry CommonScript_BagIsFull
-    ScriptEntry _0BEE
-    ScriptEntry CommonScript_VendorGreetingGeneric
-    ScriptEntry _0FCA
-    ScriptEntry _103A
-    ScriptEntry CommonScript_EmptyScript4
-    ScriptEntry _0FA7
-    ScriptEntry CommonScript_EmptyScript5
-    ScriptEntry _1282
-    ScriptEntry _12A8
-    ScriptEntry _12BA
-    ScriptEntry _12CD
-    ScriptEntry _12E0
-    ScriptEntry _12F3
-    ScriptEntry _1361
-    ScriptEntry _138C
-    ScriptEntry _139D
-    ScriptEntry _048B
-    ScriptEntry _13AB
-    ScriptEntry CommonScript_ObtainContestBackdropWaitForConfirm
-    ScriptEntry CommonScript_EmptyScript6
-    ScriptEntry CommonScript_PrintPlateObtainedMessage
-    ScriptEntry _1477
-    ScriptEntry CommonScript_SetCounterpartBGM
-    ScriptEntry CommonScript_FadeToDefaultMusic
-    ScriptEntry CommonScript_SetRivalBGM
-    ScriptEntry CommonScript_FadeToDefaultMusic
-    ScriptEntry CommonScript_AddItemQuantity
-    ScriptEntry CommonScript_ObtainContestBackdrop
-    ScriptEntry CommonScript_SendToUndergroundPCWithLinefeed
-    ScriptEntry CommonScript_ObtainAccessory
-    ScriptEntry _15BF
-    ScriptEntry CommonScript_FadeToDefaultMusic
-    ScriptEntry _15E7
-    ScriptEntry _1636
-    ScriptEntry _164A
-    ScriptEntry _165E
-    ScriptEntry CommonScript_Frontier_CheckAllFrontierGoldPrintsObtained
-    ScriptEntry CommonScript_SetLookerBGM
-    ScriptEntry CommonScript_FadeToDefaultMusic
-    ScriptEntry _170A
+    ScriptEntry CommonScript_HandleSignpostInput @ 0x7D0
+    ScriptEntry CommonScript_CheckBagPocketForItem @ 0x7D1
+    ScriptEntry CommonScript_PokecenterNurse @ 0x7D2
+    ScriptEntry _03E8 @ 0x7D3
+    ScriptEntry _043B @ 0x7D4
+    ScriptEntry _0479 @ 0x7D5
+    ScriptEntry CommonScript_SaveGame  @ 0x7D6
+    ScriptEntry CommonScript_EmptyScript2 @ 0x7D7
+    ScriptEntry _05EA @ 0x7D8
+    ScriptEntry CommonScript_ObtainPoketchApp @ 0x7D9
+    ScriptEntry CommonScript_EmptyScript1 @ 0x7DA
+    ScriptEntry CommonScript_EmptyScript3 @ 0x7DB
+    ScriptEntry CommonScript_SendToUndergroundPC @ 0x7DC
+    ScriptEntry CommonScript_ObtainUndergroundTrap @ 0x7DD
+    ScriptEntry CommonScript_ObtainUndergroundSphere @ 0x7DE
+    ScriptEntry CommonScript_ObtainAccessoryWaitForConfirm @ 0x7DF
+    ScriptEntry CommonScript_AddItemQuantityNoLineFeed @ 0x7E0
+    ScriptEntry CommonScript_BagIsFull @ 0x7E1
+    ScriptEntry _0BEE @ 0x7E2
+    ScriptEntry CommonScript_VendorGreetingGeneric @ 0x7E3
+    ScriptEntry _0FCA @ 0x7E4
+    ScriptEntry _103A @ 0x7E5
+    ScriptEntry CommonScript_EmptyScript4 @ 0x7E6
+    ScriptEntry _0FA7 @ 0x7E7
+    ScriptEntry CommonScript_EmptyScript5 @ 0x7E8
+    ScriptEntry _1282 @ 0x7E9
+    ScriptEntry _12A8 @ 0x7EA
+    ScriptEntry _12BA @ 0x7EB
+    ScriptEntry _12CD @ 0x7EC
+    ScriptEntry _12E0 @ 0x7ED
+    ScriptEntry _12F3 @ 0x7EE
+    ScriptEntry _1361 @ 0x7EF
+    ScriptEntry _138C @ 0x7F0
+    ScriptEntry _139D @ 0x7F1
+    ScriptEntry _048B @ 0x7F2
+    ScriptEntry _13AB @ 0x7F3
+    ScriptEntry CommonScript_ObtainContestBackdropWaitForConfirm @ 0x7F4
+    ScriptEntry CommonScript_EmptyScript6 @ 0x7F5
+    ScriptEntry CommonScript_PrintPlateObtainedMessage @ 0x7F6
+    ScriptEntry _1477 @ 0x7F7
+    ScriptEntry CommonScript_SetCounterpartBGM @ 0x7F8
+    ScriptEntry CommonScript_FadeToDefaultMusic @ 0x7F9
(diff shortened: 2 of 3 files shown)


include/constants/scripts_common.h
cbt6, line 1: Is this how we want to do this? Ideally we'd have the constants generated from the script file itself (using that as a single source of truth), instead defining the constants in an external file, because this would mean a user would need to edit both this file and `scripts_common.s` when adding a new script.

generated/vars_flags.txt
```diff
@@ -103,9 +103,9 @@ FLAG_UNK_0x0065
 FLAG_UNK_0x0066
 FLAG_UNK_0x0067
 FLAG_UNK_0x0068
-FLAG_UNK_0x0069
-FLAG_UNK_0x006A
-FLAG_UNK_0x006B
+FLAG_POKECENTER_5STAR_CARD
```
cbt6:
```suggestion
FLAG_POKECENTER_5_STAR_TRAINER_CARD
```

src/unk_020474B8.c
```diff
@@ -491,14 +491,14 @@ BOOL ScrCmd_BufferItemNamePlural(ScriptContext *ctx)
     return FALSE;
 }
 
-BOOL ScrCmd_33E(ScriptContext *param0)
+BOOL ScrCmd_BufferUndergroundGoodName(ScriptContext *param0)
```
cbt6:
```suggestion
BOOL ScrCmd_BufferUndergroundGoodName(ScriptContext *ctx)
```
And in the header file.

src/unk_020474B8.c
```diff
@@ -491,14 +491,14 @@ BOOL ScrCmd_BufferItemNamePlural(ScriptContext *ctx)
...
-    StringTemplate **v1 = FieldSystem_GetScriptMemberPtr(fieldSystem, SCRIPT_MANAGER_STR_TEMPLATE);
-    u8 v2 = ScriptContext_ReadByte(param0);
-    u16 v3 = ScriptContext_GetVar(param0);
+    StringTemplate **strTemplate = FieldSystem_GetScriptMemberPtr(fieldSystem, SCRIPT_MANAGER_STR_TEMPLATE);
+    u8 strFormatIdx = ScriptContext_ReadByte(param0);
+    u16 undergroundGood = ScriptContext_GetVar(param0);
 
-    StringTemplate_SetUndergroundGoodsNameWithArticle(*v1, v2, v3);
+    StringTemplate_SetUndergroundGoodsNameWithArticle(*strTemplate, strFormatIdx, undergroundGood);
     return 0;
```
cbt6:
```suggestion
    return FALSE;
```

asm/macros/scrcmd.inc
```diff
@@ -2604,10 +2613,10 @@
     .endif
     .endm
 
-    .macro ScrCmd_1D2 arg0, arg1
+    .macro AddAccessoryToFashionCase accessoryID, amount
```
cbt6:
```suggestion
    .macro AddAccessory accessoryID, amount
```

asm/macros/scrcmd.inc
```diff
@@ -2625,7 +2634,7 @@
     .short \arg2
     .endm
 
-    .macro ScrCmd_1D5 arg0
+    .macro AddContestBackdrop arg0
```
cbt6:
```suggestion
    .macro AddContestBackdrop backdropID
```

res/field/scripts/scripts_common.s
```diff
@@ -4,122 +4,122 @@
...
+    CommonScriptEntry _1361
+    CommonScriptEntry _138C
+    CommonScriptEntry _139D
+    CommonScriptEntry _048B
+    CommonScriptEntry _13AB
+    CommonScriptEntry ObtainContestBackdropWaitForConfirm
+    CommonScriptEntry EmptyScript6
+    CommonScriptEntry PrintPlateObtainedMessage
+    CommonScriptEntry _1477
+    CommonScriptEntry SetGenderBasedBGM
```
cbt6:
```suggestion
    CommonScriptEntry SetCounterpartBGM
```

res/field/scripts/scripts_common.s
```diff
@@ -4,122 +4,122 @@
...
+    CommonScriptEntry FadeToDefaultMusic
+    CommonScriptEntry SetRivalBGM
+    CommonScriptEntry FadeToDefaultMusic
+    CommonScriptEntry AddItemQuantity
+    CommonScriptEntry ObtainContestBackdrop
+    CommonScriptEntry SendToUndergroundPCWithLinefeed
+    CommonScriptEntry ObtainAccessory
+    CommonScriptEntry SetTSURETEKEBGM
```
cbt6: Let's leave this unnamed for now since we definitely would not be using that BGM name.

res/field/scripts/scripts_common.s
```diff
@@ -4,122 +4,122 @@
...
+    CommonScriptEntry FadeToDefaultMusic
+    CommonScriptEntry _15E7
+    CommonScriptEntry _1636
+    CommonScriptEntry _164A
+    CommonScriptEntry _165E
+    CommonScriptEntry Frontier_CheckAllFrontierGoldPrintsObtained
+    CommonScriptEntry SetHandsomeBGM
```
cbt6:
```suggestion
    CommonScriptEntry SetLookerBGM
```

cbt6 (commented):
Ensure that the script command names in `src/scrcmd.c` are same as those in `asm/macros/scrcmd.inc`.

asm/macros/scrcmd.inc
```diff
@@ -3865,10 +3874,10 @@
     .short \price
     .endm
 
-    .macro ScrCmd_2A7 arg0, arg1
+    .macro CheckItemIsPlate item, isPlate
     .short 679
-    .short \arg0
-    .short \arg1
+    .short \item
+    .short \isPlate
```
scbroede: imo this should be destVar for consistency

asm/macros/scrcmd.inc
```diff
@@ -3998,9 +4007,9 @@
     .short \level
     .endm
 
-    .macro ScrCmd_2BE arg0
+    .macro GetTrainerCardStars stars
     .short 702
-    .short \arg0
+    .short \stars
```
scbroede: ditto

include/constants/scripts_common.h
lhearachel: > Is this how we want to do this? Ideally we'd have the constants generated from the script file itself (using that as a single source of truth), instead defining the constants in an external file, because this would mean a user would need to edit both this file and `scripts_common.s` when adding a new script.

The following `awk` program accomplishes the simplest version of this:

```bash
awk 'BEGIN { i = 0 } /ScriptEntry \w+/ { print "#define " $2 " " i; i++ }' < res/field/scripts/scripts_common.s
```

One of the caveats with this approach is that the common-scripts entry table uses the same label for a few different entries (`_15D7` is the offender), but we can circumvent that by defining multiple fall-through labels.

metallicity: Note that we still do need different names for these duplicate entries, since different `CallCommonScript` calls will use different script IDs to refer to it and need to match the original bytes exactly. So rather than fallthrough labels, it makes more sense to me to just add some special handling for `_15D7` (which is the only such example of duplicate `ScriptEntry` labels in the repo, as best I can tell).

CharlesFolz4: Do we really want these to be generated though?  The perspective I approached this from was to treat the `ScriptEntry` labels and matching header values not as constants, but instead as though they were effectively equivalent to functions that can be externally called, just in a .s file instead of in a .c file.  With that in mind, manually maintaining the function/label definitions in a matching .h file, instead of generating them, becomes the intuitive approach.

metallicity: Having generated constants seems like the strictly better dev experience long term. To carry on from your "header file" analogy, you already have the "header declarations" in the form of the `ScriptEntry` section at the top. I'm not sure why we'd want to make people trying to add a single common script edit three different places in the code, if we had the option to have everything stay in the same file.

That said, I don't see any reason why it needs to get done right at this particular moment. Even having hard coded constants is a still big step up from no constants, and much of this work (i.e., tracking down and identifying all these raw hex/decimal values and making their references consistent between different contexts) will be helpful to already have done if/when we do decide to start generating constants.

CharlesFolz4: When put like that, I certainly see that point and agree with it.  That being said, after some trial and error (emphasis on error), integrating it into the build is beyond my current skill level.  Best I can do for now is put a TODO: comment in the file.

ljtpetersen: > Note that we still do need different names for these duplicate entries, since different `CallCommonScript` calls will use different script IDs to refer to it and need to match the original bytes exactly. So rather than fallthrough labels, it makes more sense to me to just add some special handling for `_15D7` (which is the only such example of duplicate `ScriptEntry` labels in the repo, as best I can tell).

Check out `scripts_unk_0408.s` (hidden items scripts). these are almost *all* the same script entry (284 duplicates), so special care will be needed if we want to apply this in general. Fall-through labels would work, though inelegant.

metallicity: Not sure how I missed the other 8 files with duplicate entries (best guess is that none of them have more than a one entry gap between them, and my search somehow accidentally required a larger gap). But since we already want at least one reference to each distinct duplicate constant in the script file for discoverability, I think `scripts_unk_0408.s` (and `scripts_unk_1114.s`, which does the same thing for what looks like trainer dialog) is pretty easy to work with here, since all the duplicate entries are consecutive:

```as
    RepeatCommonScriptEntry CommonScript_HiddenItems_Interact 284

    LabelCommonEntryScript CommonScript_HiddenItems_Interact:
    LabelCommonEntryScript CommonScript_HiddenItems_Interact2:
    LabelCommonEntryScript CommonScript_HiddenItems_Interact3:
    // ...
    LabelCommonEntryScript CommonScript_HiddenItems_Interact284:
        // shared definition
```

With `RepeatCommonScriptEntry` being:

```as
    .macro RepeatCommonScriptEntry scriptID, count
    .rept \count
         CommonScriptEntry \scriptID
    .endr
    .endm
```

(note: I'm not actually sure whether any of these other files are actually called via `CallCommonScript`, but the point stands in any case, since we eventually will want to have labels for all these scripts just for map references and the like)

lhearachel: None of these should be called by `CallCommonScript`. There's technically nothing *stopping* them from being used that way, but I think we are overcomplicating systems that aren't actually used in this manner.

Hidden items and trainer dialogue are similar such systems, but the process there is a bit different: they are invoked by special IDs, which are offset by a range value. If we take the "single-source-of-truth" component to its logical extreme, then we should actually be generating those script-files alongside hidden item data / trainer data, as no user should ever want to nor be expected to edit them; they are just delegate entry-points.

CharlesFolz4:
Scale of this PR has been reduced to be pure documentation, and not make any changes to how the scripts are handled in general.

res/field/scripts/scripts_mt_coronet_2f.s
```diff
@@ -67,7 +67,7 @@ _00E5:
 
 _00F2:
     Call _0100
-    CallCommonScript 0x808
+   CallCommonScript 0x808
```
lhearachel: **polish**: Revert this change.

res/field/scripts/scripts_oreburgh_city_east_house_2f.s
```diff
@@ -40,7 +40,7 @@ _006C:
     End
 
 _0077:
-    CallCommonScript 0x7E1
+   CallCommonScript 0x7E1
```
lhearachel: **polish**: Revert this change.

res/field/scripts/scripts_common.s
```diff
@@ -40,86 +40,86 @@
...
     LockAll
     FacePlayer
-    ScrCmd_2BE VAR_RESULT
-    GoToIfGe VAR_RESULT, 4, _027A
+    GetTrainerCardStars VAR_RESULT
+    GoToIfGe VAR_RESULT, 4, Common_Pokecenter_4StarCard
     SetVar VAR_0x8004, 0
     GetTimeOfDay VAR_RESULT
     Dummy1F9 VAR_RESULT
     SetVar VAR_0x8004, 120
```
lhearachel, line 82: **suggestion**: Identify and use message IDs from `res/text/bank/common_strings.h` where contextually appropriate.

lhearachel (commented):
I'll leave the majority of the feedback to others, but here are some bits that popped out. Thanks for doing this!

res/field/scripts/scripts_common.s
CharlesFolz4: PR 690 identifies and applies the message ID for this file, no need to duplicate that effort here.

lhearachel (approved):
Sticking my approval; I'd like to get approval from @cbt6 and @scbroede before we merge.

scbroede (approved):

res/field/scripts/scripts_common.s
```diff
@@ -4,31 +4,31 @@
 #include "res/text/bank/common_strings.h"
 #include "res/text/bank/menu_entries.h"
 
-    ScriptEntry Common_HandleSignpostInput
-    ScriptEntry _034C
-    ScriptEntry _00EE
+    ScriptEntry CommonScript_HandleSignpostInput
+    ScriptEntry CommonScript_CheckBagPocketForItem
+    ScriptEntry CommonScript_Pokecenter_SpeakWithNurse
```
cbt6: Is there a reason for having `_Pokecenter_` in some of these?

cbt6 (commented):
Many related PRs have been merged since this PR was opened, will need to review this again after resolving conflicts.

res/field/scripts/scripts_common.s
CharlesFolz4: Looks like some of those were a refactoring error, my bad.

metallicity:
I tested out a local branch rebasing the changes from #690 on top of the latest changes here. Still some merge conflicts, but mostly just the usual overly cautious issues from adjacent lines of code getting touched by both sides. I'd be happy to deal with those, if this PR were merged first.

PR #689, opened 2025-08-16
document scroll prompts

scbroede:
ie. these arrows you see in certain underground menus:
image

ran into a bug in nitrogfx while unpacking the narc for these and fixed it, hence the update to it


res/graphics/scroll_prompts/meson.build
```diff
@@ -0,0 +1,35 @@
+scroll_prompts_png = files('scroll_prompts.png')
+scroll_prompts_ncgr = ncgr_gen.process(scroll_prompts_png,
+    extra_args: [ '-version101', '-clobbersize', '-cell', files('scroll_prompts_cell.json')[0].full_path(), ],
```
scbroede: was there a better way to pass in this file as an argument?

h2o-DS: I think `'@BASENAME@_cell.json'` would have also worked and been more general. The problem is I do not believe that adds it explicitly as input for the meson generation.

scbroede: that gives me `Failed to open "@BASENAME@_cell.json" for reading.`

lhearachel: Template strings only work in the generator definition: https://mesonbuild.com/Reference-manual_functions.html#generator

h2o-DS: Oh, right. You would need to define a new generator such as:
```
ncgr_cell_gen = generator(nitrogfx_exe,
    arguments: [ '@INPUT@', '@OUTPUT@', '-cell', '@BASENAME@_cell.json', '@EXTRA_ARGS@', ],
    output: '@BASENAME@.NCGR',
)
```
This is what I meant about it being kinda awkward in meson.

scbroede: that gets the filename right but still gives me `Failed to open "scroll_prompts_cell.json" for reading.`. i think it needs the full path for whatever reason

h2o-DS: I could have sworn I tested that at one point...
I would leave it for now since your use case is only a single file. I will let you know what I come up with. Does this image need the cell?

scbroede: it also comes out fine with `-width 2`. is that preferred over using cells?

h2o-DS: If it is the same I would go with `width` because it reduces the dependencies on input. If they are identical cell wouldn't even do anything

src/scroll_prompts.c
```diff
@@ -0,0 +1,109 @@
...
+ScrollPrompts *ScrollPrompts_New(enum HeapId heapID)
+{
+    SpriteResourceCapacities capacities = {
+        .asStruct = {
+            .charCapacity = 1,
+            .plttCapacity = 1,
+            .cellCapacity = 1,
+            .animCapacity = 1,
+            .mcellCapacity = 0,
+            .manimCapacity = 0 }
```
lhearachel: **polish**: Add a trailing-comma to enforce better formatting.

lhearachel (approved):
Looks good, just one minor polish fix for formatting.

src/scroll_prompts.c
scbroede: done

lhearachel (approved):

PR #690, opened 2025-08-16
Document pokecenter common scripts and animation

metallicity:
Documented a few common scripts all related to healing the party, both when visiting a pokecenter and also after losing a battle (with separate scripts for landing at a pokecenter or at home). Along with this, I also documented the animation used for the pokecenter healing machine, which has a near identical copy (minus the flashing screen and sound effects) when entering a player into the hall of fame. This is the second such pair of animations I've encountered with extremely similar implementation, which does lead me to wonder if Game Freak just loves copy/pasting or had some sort of shared build process that produced the logic for both these animations.

I've also avoided the question of how to reference common scripts in this PR, instead just adding a few constants for the exact common scripts that were documented. See draft PR #687 for my current ideas of how this could be managed for all common scripts. I did rename `Common_` to `CommonScript_`, since I felt that would be a better prefix to use consistently both inside and outside this file, but don't feel too strongly about it. It just seemed best to consider a change like this now, when only a handful of common scripts are documented and/or referenced by name.


src/trainer_card_screen/trainer_card_screen.c
```diff
@@ -483,23 +484,23 @@ static void TrainerCard_LoadCardPalette(u8 stars, u8 pokedexObtained, NARC *narc
 
     if (pokedexObtained) {
         switch (stars) {
-        case 0:
-            nclrBuffer = Graphics_GetPlttDataFromOpenNARC(narc, 0, &paletteData, HEAP_ID_TRAINER_CARD_SCREEN);
+        case TRAINER_CARD_LEVEL_NORMAL:
+            nclrBuffer = Graphics_GetPlttDataFromOpenNARC(narc, TRAINER_CARD_LEVEL_NORMAL, &paletteData, HEAP_ID_TRAINER_CARD_SCREEN);
```
scbroede: the narc member indices should have their own constants (ie. something like trainer_card_NCLR). but leave these as numbers for now, those constants will be generated once the relevant narc is unpacked

src/trainer_card_screen/trainer_card_screen.c
```diff
@@ -483,23 +484,23 @@ static void TrainerCard_LoadCardPalette(u8 stars, u8 pokedexObtained, NARC *narc
 
     if (pokedexObtained) {
         switch (stars) {
```
scbroede: the `stars` variable should probably be changed to `trainerCardLevel` or something similar to match the constant names

lhearachel: This seems like a note for a future PR that revisits this screen.

src/trainer_card_screen/trainer_card_screen.c
lhearachel: +1 here; just keep the raw values for now; feel free to leave comments denoting what they are, as breadcrumbs for future contribution.

src/unk_020528D0.c
```diff
@@ -245,11 +244,11 @@ BOOL sub_02052B2C(FieldTask *task)
     case 5:
         BrightnessController_SetScreenBrightness(0, GX_BLEND_PLANEMASK_BG0 | GX_BLEND_PLANEMASK_BG1 | GX_BLEND_PLANEMASK_BG2 | GX_BLEND_PLANEMASK_BG3 | GX_BLEND_PLANEMASK_OBJ | GX_BLEND_PLANEMASK_BD, BRIGHTNESS_BOTH_SCREENS);
 
-        if (sub_0203A7EC()
+        if (FieldOverworldState_GetDefaultWarpID()
             == FieldOverworldState_GetWarpId(SaveData_GetFieldOverworldState(fieldSystem->saveData))) {
```
lhearachel: **polish**: Pull the right-hand-side up to the previous line. Splitting the equality-operator is a bit confusing.

lhearachel (changes requested):

src/trainer_card_screen/trainer_card_screen.c
scbroede: i'm not going to insist upon it or anything, but it seems reasonable to me that if we're changing from "is this 1, 2, 3, etc. stars" to "what level of trainer card is this", that should be accompanied with changing the name of the relevant parameter to reflect that.

metallicity: It's also possible to change the constant names to match "stars". It's just that with two possible names for each constant, and one of them would just append each number to some prefix, I leaned towards the descriptive, non-numerical names instead, since the former is already implicitly documented by the values. I do admit thought that in the case of a strictly incrementing constant like this, whose values are also used in less-than/greater-than operations, it seems more reasonable to put the numerical values more up front.

lhearachel: I think we can just leave them for now until we unpack the NARC itself, since we're just keeping the raw literals here. Is that agreeable?

lhearachel: I'm realizing now that the constants and the NARC files are implicitly coupled; do we want to just trash the `constants` added in this PR and revisit them later when we unpack the NARC...?

metallicity: There is an extra "doesn't have the pokedex" palette index `6` for this narc, which is not a possible result from `TrainerCard_CalculateStars`, for however much that matters to you. Loading up a new save file, the only difference is that the area for the pokedex count is omitted, but this does mean that trainer card levels and trainer card palettes do not have a 1:1 relationship.

More broadly though, I wouldn't expect a constant showing up in `scripts_common.s` to be called something like `TRAINER_CARD_PALETTE_INDEX_GOLD`, when it isn't dealing with palettes at all. Having two sets of constants or the palette indexes borrowing the corresponding `TRAINER_CARD_LEVEL_GOLD` (or `TRAINER_CARD_STARS_4`, or what have you) both seem more natural options.

Here are some screenshots for comparison, though the "Pokedex" text and count are missing since the pokedex isn't actually acquired yet.

### Normal Trainer Card (Without Pokedex) Palette

trainer_card_normal_no_pokedex

### Normal Trainer Card (With Pokedex) Palette

trainer_card_normal_pokedex

metallicity:
Rebased on main to switch from .gmm to .json for the common strings text bank.

metallicity:
@lhearachel I went ahead and removed all the script constants here, and just left them undocumented in `CallCommonScript` and the like for now. Shouldn't be too hard to track them down again once we actually have the constants ready, since there aren't many too literals around the range of `0x7D0`. I also rebased on the latest version of main, since the merge conflicts were starting to pile up a bit. Let me know if there's anything else that still needs to be addressed before this can be merged.

include/party.h
```diff
@@ -7,10 +7,12 @@
 
 #include "savedata.h"
 
+#define PARTY_MAX_POKEMON 6
+
 typedef struct Party {
     int capacity;
     int currentCount;
-    Pokemon pokemon[6];
+    Pokemon pokemon[PARTY_MAX_POKEMON];
```
lhearachel: **suggestion**:

```c
#include "constants/pokemon.h"
```

then:

```suggestion
    Pokemon pokemon[MAX_PARTY_SIZE];
```

lhearachel (approved):
One more piece of polish-feedback and then fix merge conflicts.

asm/macros/scrcmd.inc
```diff
@@ -3220,9 +3220,14 @@
     .short \partySlot
     .endm
 
-    .macro ScrCmd_23B arg0
+    .macro PlayPokecenterAnimation pokeballCount
```
cbt6:
```suggestion
    .macro PlayPokecenterHealingAnimation pokeballCount
```

src/scrcmd.c
```diff
@@ -7380,12 +7379,12 @@ static BOOL ScrCmd_2B6(ScriptContext *ctx)
...
+static BOOL ScrCmd_TrainerCardLevel(ScriptContext *ctx)
 {
     FieldSystem *fieldSystem = ctx->fieldSystem;
-    u16 *v1 = ScriptContext_GetVarPointer(ctx);
+    u16 *destVar = ScriptContext_GetVarPointer(ctx);
 
-    *v1 = TrainerCard_CalculateStars(fieldSystem);
+    *destVar = TrainerCard_CalculateStars(fieldSystem);
     return FALSE;
 }
```
cbt6: The naming here is inconsistent (card level vs stars) - regardless, script command function names should begin with a verb.

include/constants/rtc.h
```diff
@@ -1,13 +1,7 @@
 #ifndef POKEPLATINUM_CONSTANTS_RTC_H
 #define POKEPLATINUM_CONSTANTS_RTC_H
 
-enum TimeOfDay {
```
cbt6, line 4: Is there a reason for moving this out to its own file?

generated/player_request_states.txt
cbt6, line 1: I hate the choice of name here (what even is a player request state?) but that's to be settled in another PR.

include/constants/rtc.h
metallicity: Now that it's being included in scripts, `TimeOfDay` needs to become a generated enum instead. If you tried to include `constants/rtc.h` in a script, it will add a bunch of C enum syntax that neither the preprocessor nor the assembler know what to do with.

```
../include/constants/rtc.h: Assembler messages:
../include/constants/rtc.h:6: Error: bad instruction `enum MonthOfYear {'
../include/constants/rtc.h:7: Error: junk at end of line, first unrecognized character is `,'
../include/constants/rtc.h:8: Error: bad instruction `month_feb,'
../include/constants/rtc.h:9: Error: bad instruction `month_mar,'
```

For this reason, at some point it might be worth asking if `enum` declarations should be separated from header files with purely `#define` constants, since `constants` contains mostly the latter but there's no way to be sure if a particular header file within it is safe to include in a script without examining it in full.

generated/player_request_states.txt
metallicity: Same. I think the issue might be that `PlayerAvatar_RequestState[PlayerState]` was instead interpreted as `PlayerAvatar_RequestState[PlayerRequestState]`, since there were a bunch more states than the 3 possible values `PlayerState` appeared to have (when it could be there are just 3 states specifically where the player is actionable, as all the additional documented states aren't). With the added wrinkle that "request state" is a mask which can theoretically set multiple states at once for whatever reason (fishing + surfing?). Then this was recently made more confusing by some (but not all) of these references being renamed to `PlayerTransition`. Which is maybe a little more clear when taken in isolation, but still not particularly so, especially now that `PlayerState`, `PlayerTransition` and `RequestState` are all floating around together.

A good starting point for cleaning things up might look something like:

- `PlayerAvatar_SetRequestStateFlag` -> `PlayerAvatar_SetStateTransition`
- `PlayerAvatar_TurnOnRequestStateBit` -> `PlayerAvatar_SetStateTransitionFlagOn`
- `PlayerAvatar_RequestStateFlag` -> `PlayerAvatar_GetStateTransition`
- `ScrCmd_SetPlayerState` -> `ScrCmd_SetPlayerStateTransition`
- `ScrCmd_ChangeRequestState` -> `ScrCmd_TransitionPlayerState`

Possibly with a combined `ChangePlayerState` macro, since the two script commands are never called separately and there's no real reason to want to do so.

---

All that said, I'm just going to stick with `PLAYER_TRANSITION_` for this PR (and also fix the merge issue where these were unintentionally split into separate sets of constants). That's what these constants are currently called in `main`, and these changes just want to refer to one of those constants in a script.

metallicity: Actually, looking at what it's doing more closely, it seems like these are all mostly for a "temporary state" than an actual transition. The player's original state (walking/cycling/surfing) is retained, and all the places I see it used eventually return the player to this state (or assume walking).

There's still the factor that it's only set during the transition animation itself, and walking/cycling/surfing do end up changing the player's actual state. So `PlayerTransition` is in some way a decent name, but I don't know what would best capture what it does. We might be missing some distinction like "state vs. pose" or something, since a player can have a state of "walking" or "cycling" but be in the "fishing" or "berry watering" pose/animation/whatever.

include/constants/rtc.h
lhearachel: > at some point it might be worth asking if `enum` declarations should be separated from header files with purely `#define` constants

We could, in theory, implement a custom preprocessor *just* for use by scripts, if we really wanted to, but it would be annoying to maintain. I would like to avoid that, if possible.

I don't think the burden of needing to examine any given header file is that terrible, honestly; maybe we could make a special macro for `enum` that creates a better error message?

metallicity:
@narf0708 I put our `scripts_common.s` changes side by side and took on some of your terminology in places, just to ease the process of landing on a single set of names. The base prefix I went with was `CommonScript_PokecenterNurse`, which is more or less what was discussed on Discord, just without the additional underscore for `Pokecenter_Nurse`, which felt a little extraneous. I think all (or at least nearly all) `CommonScript_SomeName` script entries should aim have just a single phrase after the prefix, so that their own sub-labels can use `CommonScript_SomeName_` as a prefix without things getting too unwieldy.

Leave some review comments if there are any names/naming patterns here you'd still want to see changed.

metallicity:
@lhearachel Rebased this with the changes added to main by #688. Should be fine to merge in now.

asm/macros/movement.inc
```diff
@@ -500,7 +500,7 @@
     .short \length
     .endm
 
-    .macro NurseJoyBow length=1
+    .macro PokecenterNurseBow length=1
     .short MOVEMENT_ACTION_NURSE_JOY_BOW
```
cbt6:
```suggestion
    .short MOVEMENT_ACTION_POKECENTER_NURSE_BOW
```

PR #693, opened 2025-08-18
Update Map Headers to Use Event File Names Instead of Numbers

ljtpetersen:


include/data/map_headers.h
```diff
@@ -7,10 +7,13 @@
 
 #include "map_header.h" // Only for Intellisense/LSPs
 
+#include "res/field/encounters/pl_enc_data.naix.h"
 #include "res/field/events/zone_event.naix.h"
 #include "res/field/scripts/scr_seq.naix.h"
 #include "res/text/bank/location_names.h"
 
+#define ENCOUNTERS_NONE 0xFFFF
```
ravepossum: suggestion: I could this being useful elsewhere - would it maybe make more sense to define this in something like `include/map_header.h` that isn't a header with data?

ravepossum (commented):
One suggestion, and then a question - is there any particular reason we need to generate the encounter files with a `.bin` extension? I know that was done well before this PR though. My thought is we could drop that and avoid including them in the NAIX constants which would tidy them up a bit.

PR #694, opened 2025-08-19
Create simpler FadeScreenIn/Out macros

metallicity:
Nearly all calls to `FadeScreen` in scripts use `FadeScreen 6, 1, 0, 0` to fade out and `FadeScreen 6, 1, 1, 0` to fade in, with a small handful that swap `3` or `6` in as the second argument for some degree of slower fade, or changing the last argument to fade to white instead of black. This says to me there should be more specific macros, so I've made `FadeScreenOut` and `FadeScreenIn` which set some sensible defaults based on this existing usage. The only screen transition in any script these macros do not cover is the pair of fade types `20` in [res/field/scripts/scripts_unk_0212.s](https://github.com/pret/pokeplatinum/blob/3a2b65ab4aa8ae1f325f4ce1a80503bfce06643d/res/field/scripts/scripts_unk_0212.s#L707), which appears to be handling contests.

I also had to do a bit of fiddling to get the color constants accessible inside asm. Notably, the color constants formerly declared in `include/constants/graphics.h` have switched from `GX_RGB` to the simpler `RGB` macro taken from `include/palette.h`. As best I could tell, nothing necessitated using the `GX_` version here, and it did not seem feasible to try to include that version for the asm side.


metallicity:
One of the merge conflicts related to changes in `palette.h` now referencing `GX_RGB_G_SHIFT` and `GX_RGB_B_SHIFT`, which the creation of `color.h` was intentionally avoiding. I've added the following local defines to `color.h`, which mirror the `GX_` equivalents:

```c
// Equivalent to GX_RGB_ constants
#define COLOR_RGB_R_MASK  0x001F
#define COLOR_RGB_G_SHIFT 5
#define COLOR_RGB_B_SHIFT 10
```

As a side note, I realized that while `GX_RGB()` and `RGB()` compile to the same constants, they flip the order of their bitwise OR operations, the former going `r | g | b`, the latter going `b | g | r`. So while they are always semantically equivalent, they are only byte-equivalent when all parameters are constants, or maybe in a few other odd cases (e.g. `RGB(0, someVal, 0)`). As scripts can only reference compiled constants anyway, this shouldn't be an issue.

PR #696, opened 2025-08-19
document sunyshore market scripts

scbroede:


res/field/scripts/scripts_sunyshore_market.s
```diff
@@ -1,158 +1,159 @@
...
     BufferPartyMonSpecies 0, VAR_0x8000
-    Message 5
+    Message SunyshoreMarket_Text_EffortRibbonLooksGood
     WaitABXPadPress
     CloseMessage
     ReleaseAll
     End
 
-_0096:
+SunyshoreMarket_NPC1:
```
ravepossum: suggestion: Out of the loop on some of the recent developments so we may have already decided on a standard for this and correct me if I'm wrong. From working with pokeemerald, I think I prefer how it refers to the NPCs' scripts as their object gfx ID (e.g. Man/NinjaBoy/whatever) because it makes it easier to associate the script to the corresponding object event at a glance. I think I'd prefer that rather than just numbering them (or numbering them in addition to that if there's multiple with the same gfx).

ravepossum (approved):

PR #697, opened 2025-08-20
Document Play time manager. unk_02017428

ljm862:
Documents `unk_02017428.c` as my first task for #325.
Not super happy about the name `play_time_manager` but couldn't think of a better one off the top of my head! 

There's a variable I didn't touch called `Unk_021BF668` that I couldn't find any other references to beyond it gets set in the constructor. I'm not sure if it's preferred to leave it in or remove it in this case so for now I've left it.


src/play_time_manager.c
```diff
@@ -0,0 +1,39 @@
+#include "play_time_manager.h"
+
+#include <nitro.h>
+#include <string.h>
+
+#include "play_time.h"
+#include "unk_02022844.h"
+
+static BOOL sStarted = FALSE;
+static u64 Unk_021BF668 = 0;
```
lhearachel: **suggestion**: It looks like this isn't used, even though it's assigned to a value; you can give this a simple name like `sDummy` to illustrate that it has no purpose.

ljm862: Thanks! I wasn't sure how best to approach these ones, that makes sense to me.

PR #701, opened 2025-08-23
Revert "Fully sync Heap functions with pokeheartgold"

PikalaxALT:
Reverts pret/pokeplatinum#621

The enum should be typedeffed.


lhearachel:
`enum` is not typedef'd

PR #704, opened 2025-08-25
Document scrcmd_strings

cbt6:


asm/macros/scrcmd.inc
```diff
@@ -2122,17 +2123,17 @@
...
-    .macro ScrCmd_17B arg0, arg1, arg2
+    .macro BufferBerryName templateArg, item
     .short 379
-    .byte \arg0
-    .short \arg1
-    .short \arg2
+    .byte \templateArg
+    .short \item
+    .short VAR_0x8001
     .endm
```
lhearachel: **suggestion**: Expose `\destVar` as a parameter and give it a default value of `VAR_0x8001` instead of enforcing that it must always use `VAR_0x8001`. Refer to `DrawSignpostInstantMessage` for the syntax.

include/scrcmd_names.h
```diff
@@ -0,0 +1,53 @@
+#ifndef POKEPLATINUM_SCRCMD_NAMES_H
```
lhearachel: **question**: Would this make more sense as `scrcmd_strings`? That might be too nitpicky, given that the only functions here that don't deal with `names` are `BufferNumber` (etc.) and `CapitalizeFirstLetter`.

cbt6: Sure, I don't really mind either way.

PR #707, opened 2025-08-25
move frontier trainer names to individual files

scbroede:


res/text/meson.build
```diff
@@ -43,8 +43,12 @@ trainer_text_bank_names = [
     'npc_trainer_names.json'
 ]
 
+frontier_text_bank_names = [
+    'frontier_trainer_names.json'
+]
+
 # This list will be what's used hereafter when referring to any-and-all generated banks.
-generated_text_banks = species_text_bank_names + trainer_text_bank_names
+generated_text_banks = species_text_bank_names + trainer_text_bank_names + frontier_text_bank_names
```
lhearachel: **question**: Does this work as:

```meson
generated_text_banks = [
    species_text_bank_names,
    trainer_text_bank_names,
    frontier_text_bank_names,
]
```

If not, we could see if [the `.flatten()` function added to `1.9.0`](https://mesonbuild.com/Release-notes-for-1-9-0.html#array-flatten-method) is of use.

lhearachel (approved):

res/text/meson.build
scbroede: it works that way. pushed an update

PR #713, opened 2025-08-26
Document init scripts

metallicity:
While putting together a proposal for [documenting script constants](https://gist.github.com/metallicity/30874211f16ffcb4a377c198805d7999), I noted that init scripts are still largely undocumented. But I already mostly figured out how they were shaped while documenting some functions in `script_manager.c`, and so I decided to write up a few macros to better communicate this structure.

All init script entry macros, including the conditional entries pointed to by `InitScriptEntry_ConditionMet`, can take either a local script ID for their corresponding map, or a full script ID as recognized by `ScriptContext_LoadAndOffsetID`. This is in line with the expected behavior from the proposal above, and so init scripts shouldn't have any issues using the same constants its describes, if they were implemented.

### New Macros

- `InitScriptEntry_ConditionMet`: Entry for the first type of init script, which takes a label containing one or more conditional init script entries.
- `InitScriptEntry_EnterLocation/MapReady/MapLoad`: The three types of fixed init scripts, associated with a single script ID. Only the first entry of each type can be run, though enforcing this as a hard limit through macros felt excessive.
- `InitScriptEntryEnd`: Signals the end of the list of entries, and should still be included if that list is empty.
- `InitScriptGoToIfEqual`: Not quite as straightforward as a `GoTo`, but it gets the message across in terms of behavior.
- `InitScriptConditionsEnd`: Signals the end of the list of `InitScriptGoToIfEqual` entries, omitted if `InitScriptEntry_ConditionMet` not present.
- `InitScriptEnd`: Could just be left as a literal `.balign 4, 0`, but includes a few extra sanity checks

Note: In introducing these macros globally, I also identified several dozen `scripts_unk_###.s` which seem to be init scripts. I felt it was worth renaming them to `scripts_init_`, but didn't look into them much beyond that.


metallicity:
I decided to give actual names to all the init script types, even if I couldn't speak to the precise reasons why one might prefer to run an init script during `MapLoad` (init script type 4, called from `state = 0` of `FieldMap_Init`) vs. `MapReady` (init script type 3, called from `state = 1` of `FieldMap_Init`). The others are much more straightforward, either only called when setting a new location for `EnterLocation` (init script type 2), or called repeatedly on each input for `ConditionMet` (init script type 1). The latter two are also by far the more common types of init script.

lhearachel (approved):
This looks good to me, but I would like to get a review from @cbt6, too, whenever they have time.

src/script_manager.c
```diff
@@ -41,7 +41,7 @@ static void *ScriptContext_LoadScripts(int headerID);
 static u32 MapHeaderToMsgArchive(int headerID);
 static BOOL ScriptManager_SetHiddenItem(ScriptManager *scriptManager, u16 scriptID);
 static u16 FieldSystem_GetFixedInitScriptID(const u8 *param0, u8 param1);
-static u16 FieldSystem_GetFirstMatchInitScriptID(FieldSystem *fieldSystem, const u8 *param1, u8 param2);
+static u16 FieldSystem_GetConditionMetInitScriptID(FieldSystem *fieldSystem, const u8 *param1, u8 param2);
```
cbt6: Rename parameters and for all other documented function declarations.

cbt6 (commented):
As I've said previously, I'd prefer the different init script types to be named only after the map loading functions get documented. At the least, can you explain / document in `include/constants/init_script_types.h` how or at which points are the various init scripts called?

metallicity:
@cbt6 I spent a bit of time looking into each individual state of `FieldMap_Init` (which is the only place where the two init scripts types I was less sure about were invoked), and I eventually stopped documenting things just because it was pretty clear my initial assumptions were broadly correct. Everything up until the first init script call is just clearing out/initializing data, and only after does it start to actually load in anything from the specific map/objects/etc., with the second init script call being essentially the last thing to happen (aside from initializing the touch screen UI).

I updated the init script types to be named `_MAP_LOAD_START` and `_MAP_LOAD_FINISH`, but they mean basically the same thing as before, just maybe a little more clear. I'm still not sure how to capture "these can also run after using field moves or bag items or whatever" (as various places in the code eventually call into `FieldMap_Init` to return to the overworld), but I'm not sure there's a succinct way to convey that nuance without making it sound like they don't run when the map first loads. I felt "ready" did an okay job at that, but it didn't really give a good impression of how tightly coupled these two init script types were.

If you still prefer, I can change them back to `INIT_SCRIPT_TYPE_UNK_3` and `INIT_SCRIPT_TYPE_UNK_4`, but I don't think any issues with the names will be solved simply by better documenting the details of `FieldMap_Init`, when it's more about accurately describing the different entrypoints into this function, which are already fairly clear.

cbt6:
Ok `_MAP_LOAD_START` and `_MAP_LOAD_FINISH` are slight improvements, but it's still a little vague to me - `MAP_LOAD_FINISH` definitely comes after `MAP_LOAD_START` at some point, but what about `ENTER_LOCATION` and `CONDITION_MET`? What's the sequence for each of them? (see https://github.com/pret/pokeemerald/blob/9ab03f454ddf6f9627e6252fe974897749aa2f3f/include/constants/map_scripts.h for example of what I'm looking for)

metallicity:
While at first I wasn't sure it made sense, given half the old modes didn't seem to have a newer equivalent and a few names had issues similar to what I was talking about above, I decided it was probably best to just adopt Gen 3 terminology in general here. There is genuine merit in using the same terms to mean the same things across repos, and all of the behavior described for the different types of map script seems to line up exactly. Game Freak must have just dropped a few of the less necessary options between Gen 3 and Gen 4.

To that end, I've made the following changes:

- `scripts_init_` -> `map_scripts_`
- `initScript` -> `mapScript`
- `CONDITION_MET` -> `ON_FRAME_TABLE`
- `ENTER_LOCATION` -> `ON_TRANSITION`
- `MAP_LOAD_START` -> `ON_LOAD`
- `MAP_LOAD_FINISH` -> `ON_RESUME`

The only name I particularly disagree with here is "map script". I still find it a less misleading name than "init script", in that `ON_FRAME_TABLE` checks on every frame of input processing rather than just when the map is initialized. But it is a little weird that every map also has its non-map "scripts" file, which includes scripts that are only called as "map scripts". That said, I do like the simplicity of just following Gen 3's lead until/unless there's some better alternative.

I've also added the requested documentation for each type of map script, including common use cases for each like pokeemerald does.

lhearachel:
I am a bit cautious about using the term `map` as a prefix for this sort of scripts, because a "map" in Gen4 has a different structure than a "map" in Gen3. My understanding of the Gen3 structure is that the header-to-map association is essentially 1-to-1 (with few exceptions like the Battle Frontier). This is not the case in Gen4; headers instead are associated to a single matrix, which then translates player coordinates to an individual map-model.

Are we sure that we want to call these "map" scripts, given that context? The binary folks for Gen4 call these "level" scripts, but I don't think that that is a great name, either.

metallicity:
@lhearachel As I said, I don't think "map script" is a great name, but I don't think "init script" is either. So naming it the same as Gen 3 seemed like the better option between the two until some better option arises. Interpreted in the most generous way, a "map script" is a script that runs simply by virtue of the player being on a given map, rather than being tied to a specific event within it (or invoked by some offset ID call). So far, I haven't come up with anything I feel both does a decent job describing these scripts and also avoids being mistaken for other kind(s) of script.

Some alternatives I sort of like (in rough order of preference):
- "header script": Has the double meaning of "map header" and "these don't actually define scripts, just declare them", but that also makes it a bit cryptic.
- "transition script": scripts that run either when there is a screen transition or when some change causes a "frame table" script to run.
- "state script": scripts that run when a particular game state is reached (?), rather than the player interacting with some particular part of the map (but warps don't really fit with that definition cleanly).
- "field script": Attempted improvement over "map script", trying to better imply the interpretation I gave above, and without raising the "what is a map?" question. But other scripts still also run on the "field".

If none of those seem worth switching to, I'd be fine keeping "init script" for now and just adopting the Gen 3 terminology for the individual script types.

cbt6:
Let's just keep them called as init scripts for now. If we're going to use a different name then can be done in a separate PR. As soon as that change is reverted then this PR is good to go.

metallicity:
@cbt6 I reverted all the instances of "map script" back to "init script" (including the file renames) and rebased on the latest version of main just to check everything still build alright. Seems good to merge.

PR #714, opened 2025-08-27
Update Debug Macros to Allow a Single Argument

ljtpetersen:
With
```c
#define EmulatorPrintf(text, ...)
```
calls of `EmulatorPrintf` with a single parameter fail. e.g., `EmulatorPrintf("text\n");`. To remedy this, replace it with
```c
#define EmulatorPrintf(...)
```

### DIFF
--- a/include/debug.h
+++ b/include/debug.h
@@ -15,8 +15,8 @@ __attribute__((format(printf, 1, 2))) void EmulatorLog(const char *text, ...);
 
 #else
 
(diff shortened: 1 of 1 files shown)


lhearachel:
Does our compiler support `__VA_OPT__` or the extended use of `##` before a comma during preprocessing?

Reference: https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html

PR #717, opened 2025-08-28
Document elevator scripts and animation

metallicity:
Documents all scripts that call `PlayElevatorAnimation`, as well as the internals of this script command. This also identified two related commands, `BufferFloorNumber` and `SetExitLocation`, the latter of which is also used for returning from the Great Marsh.

I also want to make a particular note of the Battle Tower elevator scripts, which are the only ones which were left partially undocumented. The entire Battle Tower appears to take place more or less inside these elevator scripts through some manner or another, aside from going down to pick your partner for double battles, which does release control flow to the "Battle Salon" map. Otherwise, the scripts all start with taking an elevator up, and then just call into a few Battle Tower script commands before warping the player back to their appropriate starting location, with the different x positions helping to identify which type of Battle Room the player entered in the first place.

### DIFF
--- a/res/field/scripts/scripts_jubilife_tv_elevator.s
+++ b/res/field/scripts/scripts_jubilife_tv_elevator.s
@@ -42,10 +42,10 @@ JubilifeTVElevator_SelectFloor:
     ShowCurrentFloor 20, 1, VAR_ELEVATOR_FLOORS_ABOVE
     Message JubilifeTVElevator_Text_Hello
     InitGlobalTextMenu 1, 1, 0, VAR_RESULT
-    AddMenuEntryImm MenuEntries_Text_Floor4, 0
-    AddMenuEntryImm MenuEntries_Text_Floor3, 1
-    AddMenuEntryImm MenuEntries_Text_Floor2, 2
-    AddMenuEntryImm MenuEntries_Text_Floor1, 3
+    AddMenuEntryImm MenuEntries_Text_4F, 0
+    AddMenuEntryImm MenuEntries_Text_3F, 1
+    AddMenuEntryImm MenuEntries_Text_2F, 2
+    AddMenuEntryImm MenuEntries_Text_1F, 3
     AddMenuEntryImm MenuEntries_Text_ElevatorExit, 4
     ShowMenu
     SetVar VAR_0x8008, VAR_RESULT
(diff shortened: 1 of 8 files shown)


res/field/scripts/scripts_jubilife_tv_elevator.s
```diff
@@ -1,153 +1,154 @@
...
-    GetFloorsAbove VAR_UNK_0x40CE
-    ScrCmd_11B 11, 2, 18, 2, 1
-    GoToIfEq VAR_UNK_0x40CE, 3, _026C
-    CallIfGt VAR_UNK_0x40CE, 3, _022D
-    CallIfLt VAR_UNK_0x40CE, 3, _0238
-    Call _0243
-    SetVar VAR_UNK_0x40CE, 3
-    GoTo _0261
+    GetFloorsAbove VAR_ELEVATOR_FLOORS_ABOVE
+    SetExitLocation MAP_HEADER_JUBILIFE_TV_1F, 2, 18, 2, 1
```
cbt6:
```suggestion
    SetExitLocation MAP_HEADER_JUBILIFE_TV_1F, 2, 18, 2, DIR_SOUTH
```
And for all others

res/text/menu_entries.json
```diff
@@ -490,39 +490,39 @@
       "en_US": "EXIT"
     },
     {
-      "id": "pl_msg_00000361_00116",
+      "id": "MenuEntries_Text_Floor1",
```
cbt6:
```suggestion
      "id": "MenuEntries_Text_1F",
```

src/string_template.c
```diff
@@ -35,6 +35,8 @@
 #include "res/text/bank/mystery_gift_event_names.h"
 #include "res/text/bank/special_met_location_names.h"
 
+#define MENU_ENTRIES_DEPARTMENT_STORE_FLOOR_OFFSET MenuEntries_Text_Floor1 - 1
```
cbt6: Is this necessary? Why create a define that is used only once? I don't think the extra "documentation" that it's an offset is needed - seems quite self explanatory.

src/scrcmd.c
```diff
@@ -4746,7 +4746,7 @@ static BOOL ScrCmd_Unused_0F7(ScriptContext *ctx)
     return TRUE;
 }
 
-static BOOL ScrCmd_11B(ScriptContext *ctx)
+static BOOL ScrCmd_SetExitLocation(ScriptContext *ctx)
```
cbt6: I know that this is used for setting the "exit"s from elevators / great marsh, but this calls `FieldOverworldState_SetSpecialLocation`, and `exit` already means something else in the `FieldOverworldState` struct. Let's rename it to `SetSpecialLocation` for now to keep it consistent.

src/unk_020474B8.c
```diff
@@ -592,13 +592,13 @@ BOOL ScrCmd_346(ScriptContext *param0)
     return 0;
 }
 
-BOOL ScrCmd_347(ScriptContext *param0)
+BOOL ScrCmd_BufferFloorNumber(ScriptContext *ctx)
```
cbt6, line 595: 👍Good catch that this for floor number generally, not just department store, updated #704 to reflect that.

src/scrcmd.c
metallicity: I was trying to give it a more helpful name for how it's actually used in scripts specifically, but didn't realize "exit" was an overloaded term here. Agreed to just go with the consistent name instead.

PR #720, opened 2025-08-31
Unpack Item Icons Narc

ljtpetersen:


scbroede:
imo instead of naming the tm palettes after a single representative tm, it would make more sense to name them after the type. for example, something along the lines of `dragon_type_tm.pal` instead of `tm02.pal`

PR #721, opened 2025-08-31
document party-related scrcmds

scbroede:
also renames a couple of previously-documented scrcmds

- `GetPartyCountHatched` -> `CountPartyNonEggs` (previous name implied pokemon that were received from other means than hatching an egg wouldn't count)
- `CheckPartyHasFatefulEncounter` -> `FindPartySlotWithFatefulEncounterSpecies` (more accurate; this doesn't put a boolean value in destVar indicating whether there's a fateful encounter pokemon, it puts the party slot of a fateful encounter pokemon of a specific species)


src/scrcmd_party.c
```diff
@@ -0,0 +1,828 @@
...
+        Egg_CreateEgg(egg, species, 1, trainer, 3, specialMetLoc);
+
+        Party_AddPokemon(party, egg);
+        Heap_Free(egg);
+    }
+
+    return FALSE;
+}
+
+BOOL ScrCmd_ResetPartyMonMoveSlot_Unused(ScriptContext *ctx)
```
lhearachel, line 110: **question**: Is the `_Unused` prefix valuable?

src/scrcmd_party.c
```diff
@@ -0,0 +1,828 @@
...
+
+    u8 slot;
+    for (slot = 0, *destVar = MAX_PARTY_SIZE; slot < partyCount; slot++) {
+        Pokemon *mon = Party_GetPokemonBySlotIndex(SaveData_GetParty(fieldSystem->saveData), slot);
+
+        if (Pokemon_GetValue(mon, MON_DATA_IS_EGG, NULL) != FALSE) {
+            continue;
+        }
+
+        if ((Pokemon_GetValue(mon, MON_DATA_MOVE1, NULL) == move) || (Pokemon_GetValue(mon, MON_DATA_MOVE2, NULL) == move) || (Pokemon_GetValue(mon, MON_DATA_MOVE3, NULL) == move) || (Pokemon_GetValue(mon, MON_DATA_MOVE4, NULL) == move)) {
```
lhearachel: **polish**: Break lines on the conjunction operators.

src/scrcmd_party.c
scbroede: you could probably make a strong argument that it isn't. adding it here for consistency because we already have a lot of scrcmds labeled as "unused", but if we want to get rid of all of that i wouldn't necessarily be opposed

lhearachel: Let's leave that to a separate PR; I don't think it's particularly valuable for scripting commands, if we expect users to make use of anything available to them.

src/scrcmd_party.c
scbroede: done

PR #722, opened 2025-08-31
Document Timer

ljm862:
Documenting the timer `unk_02022844`.

A continuation of #697 and part of #325.

The `sShouldResetTimer` seems to never get set to TRUE, so the check is basically dead code. I'm not sure if that's an issue to keep the dead code or if that's even intentional.
I changed the TicksToX functions to be wrappers for the Nitro calls instead of redoing the same calculation logic that we have elsewhere. Not sure if that's the preferred approach here or not!
I'm also unsure if some of the steps in the `CalculateCurrentTimestamp` function may have calls within Nitro that do something similar too.


src/timer.c
```diff
@@ -0,0 +1,71 @@
...
+    OSIntrMode irqState = OS_DisableInterrupts();
+
+    timerLow = *(REGType16 *)((u32)REG_TM0CNT_L_ADDR + (OS_TIMER_3) * 4);
+    timerHigh = sTimerValue & 0xffffffffffffULL;
+
+    if (reg_OS_IF & (OS_IE_TIMER3) && !(timerLow & 0x8000)) {
+        timerHigh++;
+    }
+
+    (void)OS_RestoreInterrupts(irqState);
```
lhearachel: **polish**: Remove the `void` cast, if not needed to match.

PR #723, opened 2025-09-01
Document initialization routine for the Party Menu application

lhearachel:
This PR documents the initialization routine, associated structure members, and associated data tables for the Party Menu.

The application is typically initialized by a client with a requested Mode and Type; the latter, as best as I can tell, is functionally always 0 (called `BASIC` in the enumeration), and the other two are wholly unused (though they *are* checked for). More interesting is the requested mode, which controls:

1. What elements are shown on each party member's "panel,"
2. Comment strings that can be shown for various modes, e.g., whether a party member is compatible with a TM or evolution stone
3. What "buttons" -- i.e., "Confirm", "Cancel", or neither -- are shown on the bottom of the screen, and
4. The precise movement capabilities of the cursor on the menu.

The client is also expected to provide a callback for consuming the result when the party menu closes. This is wholly undocumented at the moment, and will be covered in a future PR.

> [!NOTE]
> There is likely some messiness in this PR with missing constants and the like; any misses not caught by my replacement scripts will be tidied up as I go through the rest of the code.


lhearachel:
Marking as ready for review; this is not the full application by any means, but this PR is large enough as-is, and there is still much more to do. I'd rather tackle this over a handful of medium-to-large PRs than one massive PR, especially since there are some valuable breadcrumbs here for other systems.

src/start_menu.c
```diff
@@ -1260,23 +1260,23 @@ static BOOL sub_0203BC5C(FieldTask *taskMan)
...
-        partyMan->unk_21 = 0;
-        partyMan->unk_20 = 9;
-        partyMan->usedItemID = BagSystem_GetItem(v2);
-        partyMan->fieldSystem = fieldSystem;
-
-        FieldSystem_StartChildProcess(fieldSystem, &gPokemonPartyAppTemplate, partyMan);
-        menu->taskData = partyMan;
+        PartyMenu *partyMenu;
+
+        partyMenu = Heap_Alloc(HEAP_ID_FIELD2, sizeof(PartyMenu));
```
scbroede: merge declaration and assignment (also there's a couple more of these below)

lhearachel: I mostly left these alone save for standardizing the names, but I will go ahead and document these routines since I'm in this section of the code anyways.

asm/macros/scrcmd.inc
```diff
@@ -2252,16 +2252,21 @@
...
+    ReturnToField
+    .endm
+
     .macro GetSelectedPartySlot destVar
     .short 403
     .short \destVar
     .endm
 
-    .macro ScrCmd_194 arg0, arg1, arg2, arg3
+    .macro OpenPartyMenuForContest arg0, arg1, arg2, arg3
```
scbroede: label arguments

src/applications/party_menu/main.c
```diff
@@ -0,0 +1,2993 @@
...
+    },
+    {
+        { .panelX = 0,  .panelY = 0,  .speciesIconX = 30,  .speciesIconY = 16,  .ballSpriteX = 16,  .ballSpriteY = 14  },
+        { .panelX = 16, .panelY = 0,  .speciesIconX = 158, .speciesIconY = 16,  .ballSpriteX = 144, .ballSpriteY = 14  },
+        { .panelX = 0,  .panelY = 6,  .speciesIconX = 30,  .speciesIconY = 64,  .ballSpriteX = 16,  .ballSpriteY = 62  },
+        { .panelX = 16, .panelY = 6,  .speciesIconX = 158, .speciesIconY = 64,  .ballSpriteX = 144, .ballSpriteY = 62  },
+        { .panelX = 0,  .panelY = 12, .speciesIconX = 30,  .speciesIconY = 112, .ballSpriteX = 16,  .ballSpriteY = 110 },
+        { .panelX = 16, .panelY = 12, .speciesIconX = 158, .speciesIconY = 112, .ballSpriteX = 144, .ballSpriteY = 110 },
+    },
+};
```
scbroede, line 199: should there be designated initializers for the array indices here?

src/applications/party_menu/main.c
```diff
@@ -0,0 +1,2993 @@
...
+    0xE6,
+    0x1C0,
+    0xD0,
+    0x87,
+};
+
+static BOOL PartyMenu_Init(ApplicationManager *appMan, int *state)
+{
+    PartyMenuApplication *application;
+    NARC *narc;
```
scbroede, line 268: merge declaration and assignment

src/applications/party_menu/main.c
```diff
@@ -0,0 +1,2993 @@
...
+    InitializeTouchPad(4);
+    Sound_SetSceneAndPlayBGM(SOUND_SCENE_SUB_57, SEQ_NONE, 0);
+
+    PartyMenu_AddWindows(application);
+    PartyMenu_InitSpriteResources(application);
+    PartyMenu_InitSprites(application);
+
+    SetupRequestedMode(application);
+    SetupRequestedModePanels(application);
+    InitSelectedSlot(application, application->currPartySlot, 1);
```
scbroede: should the last argument for calls to this function be TRUE/FALSE instead of 1/0?

src/applications/party_menu/sprites.c
```diff
@@ -0,0 +1,369 @@
...
+#include "sprite.h"
+#include "sprite_system.h"
+#include "sprite_util.h"
+#include "unk_0208C098.h"
+#include "vram_transfer.h"
+
+void PartyMenu_InitSpriteResources(PartyMenuApplication *application)
+{
+    GXLayers_EngineAToggleLayers(GX_PLANEMASK_OBJ, 1);
+    GXLayers_EngineBToggleLayers(GX_PLANEMASK_OBJ, 1);
```
scbroede: use TRUE instead of 1

src/item_use_functions.c
```diff
@@ -1124,7 +1120,7 @@ static BOOL RegisteredItem_GoToApp(FieldTask *task)
     switch (v1->unk_2A) {
     case 0:
         MapObjectMan_PauseAllMovement(fieldSystem->mapObjMan);
-        ov5_021D1744(0);
+        FieldMap_FadeScreen(FADE_TYPE_BRIGHTNESS_OUT);
```
scbroede, line 1123: would it be better to pass boolean values or something along those lines into this function instead of fade types? i wonder if this might make someone incorrectly think that fade types other than brightness could be passed in.

then again the parameter being called `fadeInOrOut` might take care of that

src/applications/party_menu/main.c
```diff
@@ -0,0 +1,2993 @@
...
+static void InitBgMainLayer0(BgConfig *bgConfig);
+static void InitBgs(BgConfig *bgConfig);
+static void sub_0207EA24(BgConfig *param0);
+static void LoadGraphics(PartyMenuApplication *application, NARC *narc);
+static PartyMenuApplication *NewPartyMenuApplication(ApplicationManager *appMan);
+static void SetupRequestedMode(PartyMenuApplication *application);
+static void SetupRequestedModePanels(PartyMenuApplication *application);
+static u8 sub_0207F984(PartyMenuApplication *application, u8 param1);
+static u8 sub_0207FE98(PartyMenuApplication *application);
+static void InitSelectedSlot(PartyMenuApplication *application, u8 param1, u8 param2);
```
scbroede: copy parameter names from definition

src/applications/party_menu/main.c
lhearachel: I don't think it would make anything clearer or safer; the outer dimension does not align with the `PartyMenuType` definitions in terms of usage, and the inner dimension is just an index on party slot.

src/item_use_functions.c
lhearachel: That was why we called it `fadeInOrOut`, yeah; I do see where it could make it harder to make a mistake, but I'm not convinced that it's worth it to effectively couple these two fade values to `TRUE` and `FALSE`.

src/applications/party_menu/main.c
lhearachel: Doesn't match; I'll leave a comment.

src/applications/party_menu/main.c
lhearachel: Yeah, it should; I'll get the other instances, too.

scbroede (approved):

PR #725, opened 2025-09-02
document safari game scripts and text

scbroede:
the tram constants list is kind of bad because it's really 3 sets of constants, but i'm working with what game freak gave me here


src/scrcmd.c
```diff
@@ -233,6 +233,11 @@
 #include "constdata/const_020EAC58.h"
 #include "constdata/const_020F8BE0.h"
 
+enum SafariGameActiveState {
+    SAFARI_GAME_ACTIVE = 0,
+    SAFARI_GAME_INACTIVE,
+};
```
lhearachel: **question**: Should these just be simple `#define`s that can be shared with the `scrcmd.inc`?

scbroede: yeah, i don't know why i didn't just do it that way to begin with

PR #730, opened 2025-09-04
Use constant identifiers for CoordEvent variables in events files

lhearachel:


tools/scripts/migration/events.py
cbt6, line 1: To what extent do we want to support migrations? This definitely would not be the last refactoring and I doubt having a migration script for every refactor in the future is the way to go.

tools/json2bin/convert.py
```diff
@@ -63,17 +63,20 @@ def from_movement_type(s: str) -> int:
 def from_trainer_type(s: str) -> int:
     return trainer_types.TrainerType[s].value
 
-def from_var_flag(s: str) -> int:
-    if s.isnumeric():
-        return int(s)
-    return vars_flags.VarFlag[s].value
+def from_var_flag(v: str | int) -> int:
```
cbt6: Any reason why it should accept `int`s too?
(In fact I'm thinking of probably asserting that `^(VAR_)|(FLAG_)` but that might be too far)

lhearachel: Accepting `int` is just for backwards compatibility; if we want to mandate `str`, then we should probably expect users to write (or use) a script to convert all of their existing files, if they merge downwards.

tools/scripts/migration/events.py
lhearachel: Good question. I don't have a good answer for it. We might be at a stage where we expect any users to either be frozen-in-time and cherry-picking things that they want, or technically-minded enough to write their own tools for things like this. I'm open to saying that we don't want scripts like these in-repo, but I will probably still publish the ones that I write to a separate source that I can point users to if they come asking questions.

cbt6: I think if you want to offer some kind of support, the PR description would be a better place, instead of having them tracked in git history. But there again personally I don't think any kind of migration / backward compatibility is necessary.

lhearachel: That's fair; I'll remove the script from the history and place it into the PR description. For the existing scripts, I'll post a separate PR to remove them from the tree, if desired.

tools/json2bin/convert.py
lhearachel: With the context above of not providing migration tools, do we want to maintain backward-compatibility here?

cbt6: I would vote no, at least not at this point in time.

lhearachel:
Removed the script from the history + backwards-compatibility.

tools/json2bin/convert.py
```diff
@@ -63,17 +63,20 @@ def from_movement_type(s: str) -> int:
...
     return trainer_types.TrainerType[s].value
 
-def from_var_flag(s: str) -> int:
-    if s.isnumeric():
-        return int(s)
-    return vars_flags.VarFlag[s].value
+def from_var_flag(v: str) -> int:
+    if v.isnumeric():
+        return int(v)
+    return vars_flags.VarFlag[v].value
```
cbt6:
```suggestion
def from_var_flag(v: str) -> int:
    return vars_flags.VarFlag[v].value
```

tools/json2bin/convert.py
```diff
@@ -63,17 +63,20 @@ def from_movement_type(s: str) -> int:
...
 
 def from_map_header(s: str) -> int:
     return map_headers.MapHeader[s].value
 
 def from_map(s: str) -> int:
     return maps.MapID[s].value
 
+def from_variable(v: int | str) -> int:
+    return v if type(v) is int else vars_flags.VarFlag[v].value
```
cbt6:
```suggestion
```

tools/json2bin/convert.py
lhearachel: This doesn't work, because we use `"0"` for some objects to denote that they do not have flags associated with them. e.g., in `wayward_cave_1f`:

```json
{
    "graphics_id": "OBJ_EVENT_GFX_CAMPER",
    "movement_type": "MOVEMENT_TYPE_LOOK_RIGHT",
    "trainer_type": "TRAINER_TYPE_NORMAL",
    "flag": "0",
    "script": 3378,
    "initial_dir": 3,
    "data": [
        2
    ],
    "movement_range_x": 0,
    "movement_range_z": 0,
    "x": 2,
    "z": 14,
    "y": 0
}
```

Do you have a preference on keeping the JSON field vs. removing the JSON field vs. making it `null`able? I think `null` makes the most sense here, personally.

cbt6:
```py
def from_var_flag(v: str) -> int:
    return 0 if v == '0' else vars_flags.VarFlag[v].value
```
Let's just go with this for now then. We can change the schema in another PR.

PR #732, opened 2025-09-04
Document Party Menu windows

lhearachel:
There is one lingering `TODO` in here that I will address when I next have time, but the rest of this is reviewable.


src/applications/party_menu/main.c
```diff
@@ -1758,10 +1758,10 @@ static void sub_0207FFC8(PartyMenuApplication *application)
         v1 = sub_020801F0(application, v0);
     }
 
-    sub_02081CF4(application, v0, v1);
+    PartyMenu_DrawContextMenu(application, v0, v1);
     Heap_FreeExplicit(HEAP_ID_PARTY_MENU, v0);
-    sub_02081E08(application);
-    sub_020826F4(application, 0xffffffff, 1);
+    PartyMenu_LoadContextMenuPrompt(application);
+    PartyMenu_PrintMediumMessage(application, 0xffffffff, 1);
```
scbroede: should this have its own constant similar to `PRINT_LONG_MESSAGE_PRELOADED`? (or maybe that one could be generalized; it seems to behave the same way here)

PR #735, opened 2025-09-05
nitrogfx: various cell fixes

scbroede:
- fixes the calculation of pixelOffset and cleans up the various mappingTypes
credit goes to @h2o-DS for the following:
- handles overlapping oams by rounding to the nearest multiple of 8
- fixes height calculations to prevent extra separator rows

### DIFF
--- a/tools/nitrogfx/gfx.c
+++ b/tools/nitrogfx/gfx.c
@@ -10,8 +10,6 @@
 #include "json.h"
 #include "util.h"
 
-static int SnapToTile(int val);
-
 static unsigned int FindNitroDataBlock(const unsigned char *data, const char *ident, unsigned int fileSize, unsigned int *blockSize_out)
 {
     unsigned int offset = 0x10;
--- a/tools/nitrogfx/gfx.c
+++ b/tools/nitrogfx/gfx.c
@@ -10,6 +10,8 @@
 #include "json.h"
 #include "util.h"
 
+static int SnapToTile(int val);
+
 static unsigned int FindNitroDataBlock(const unsigned char *data, const char *ident, unsigned int fileSize, unsigned int *blockSize_out)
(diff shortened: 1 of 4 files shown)


tools/nitrogfx/main.c
```diff
@@ -302,6 +304,24 @@ void HandleNtrToPngCommand(char *inputPath, char *outputPath, int argc, char **a
...
+                if (strcmp(argv[i + 1], "-nosnap") == 0)
+                {
+                    options.cellSnap = false;
+                    i++;
+                }
+            }
+
+            if (i + 1 < argc)
+            {
+                if (strcmp(argv[i + 1], "-noskip") == 0)
```
h2o-DS: I believe this does not allow for `-nosnap` to follow `-noskip`. Is that acceptable?

You could put them in a for loop.

tools/nitrogfx/main.c
```diff
@@ -302,6 +304,26 @@ void HandleNtrToPngCommand(char *inputPath, char *outputPath, int argc, char **a
...
+            {
+                if (i + 1 < argc)
+                {
+                    if (strcmp(argv[i + 1], "-nosnap") == 0)
+                    {
+                        options.cellSnap = false;
+                        i++;
+                    }
+
+                    if (strcmp(argv[i + 1], "-noskip") == 0)
```
h2o-DS: Make this an else if or it could go out of bounds.

tools/nitrogfx/gfx.c
```diff
@@ -10,6 +10,8 @@
 #include "json.h"
 #include "util.h"
 
+static int SnapToTile(int val);
```
lhearachel: **polish**: Remove unnecessary forward-declare

PR #736, opened 2025-09-05
Document overlay 102 (diploma)

cbt6:


include/diploma/diploma.h
lhearachel, line 1: **quibble**: Follow the pattern set by `options_menu` and `naming_screen`, which are also single-file applications/scenes:

```
include/applications/naming_screen.h
include/applications/options_menu.h
src/applications/naming_screen.c
src/applications/options_menu.c
```

lhearachel (commented):
Looks good; one comment about where the files should live in the source tree, extremely minor.

PR #737, opened 2025-09-06
nitrogfx: add options for converting to and from 8bpp

scbroede:
intended to eliminate the need for .pal files when a 4bpp NCGR's associated NCLR has more than 16 colours.

sample usage:

`nitrogfx file.NCGR file.png -palette file.NCLR -convertTo8Bpp` produces an 8bpp png with the entire palette

`nitrogfx file.png file.NCGR -bitdepth 8 -convertTo4Bpp` recreates the original NCGR

currently not implemented for scanned files. seems to work fine with cells

### DIFF
--- a/tools/nitrogfx/gfx.c
+++ b/tools/nitrogfx/gfx.c
@@ -83,20 +83,21 @@ static void ConvertFromTiles1Bpp(unsigned char *src, unsigned char *dest, int nu
     }
 }
 
-static void ConvertFromTiles4Bpp(unsigned char *src, unsigned char *dest, int numTiles, int chunksWide, int colsPerChunk, int rowsPerChunk, bool invertColors)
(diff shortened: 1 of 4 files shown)


lhearachel:
The code changes look fine to me.

As a general thought, how do we want to communicate to users that individual images are actually 4bpp and only support 16 colors? Do we need to communicate that at all?

scbroede:
good question, and i'm honestly not sure. can we expect them to infer it from the fact that only a single row of the palette is actually used in the image? do we want to put something in the file name to indicate it? i think in some images the implication exists (ie. one row is greyscale and another is in colour), but maybe not all of them

lhearachel:
My bigger worry is for files that have palettes with more than 16 colors yet only ever *use* the first 16. I'm not sure how we can effectively communicate that to end-users, but maybe it's not a big deal.

h2o-DS:
As long as the color indexes are modulo'd down to 16 I think it's probably fine. Likely more identifiable than the cells.
The additional colors and cells have rules that can't realistically be enforced without a gui of some sort.

PR #743, opened 2025-09-07
Document Poketch Dot Artist (Overlay 31)

VicSevenT:
There are two aspects of this app that I found interesting/unique from other poketch apps, so I want to document them here.
* The dot art is made up of a 24x20 grid, with each cell having a value of 1-4, indicating its color. The image is saved and will be reloaded when poketch apps are changed or the game is restarted. To save the image, the cells are packed into a 120-byte array, where each cell value is represented by 2-bits: value of 1->00, 2->01, 3->10, 4->11. 
* Each "dot" takes up the space of one tile on the display. The app does not use a NCGR file to load the tiles, but rather generates them manually. In `GenerateDotTiles()`, 16 tiles are generated, where there is one tile for every value 0-15 and the every nibble in the tile has that same value, effectively creating tiles made up of entirely one color.  In `UpdateTilemap()`, the cell values are mapped to a tile/palette value with the `tileAssignments` array.


src/applications/poketch/dot_artist/graphics.c
```diff
@@ -0,0 +1,158 @@
...
+        4, 15, 8, 1
+    };
+    int x, y;
+
+    u16 *tilemapBuffer = Bg_GetTilemapBuffer(bgConfig, BG_LAYER_SUB_2);
+    tilemapBuffer += ((POKETCH_SCREEN_TILE_OFFSET_Y * POKETCH_WIDTH_TILES) + POKETCH_SCREEN_TILE_OFFSET_X);
+
+    for (y = 0; y < CANVAS_HEIGHT; y++) {
+        for (x = 0; x < CANVAS_WIDTH; x++) {
+            tilemapBuffer[x] = tileAssignments[(dotArt->dots[y][x] - 1)];
```
lhearachel: **polish**: Remove extraneous parens.

src/applications/poketch/dot_artist/main.c
```diff
@@ -0,0 +1,299 @@
...
+#include "bg_window.h"
+#include "heap.h"
+#include "poketch.h"
+#include "sys_task.h"
+#include "sys_task_manager.h"
+#include "touch_screen.h"
+
+#define PACKED_SIZE ((CANVAS_WIDTH * CANVAS_HEIGHT) / 4)
+
+typedef struct {
```
scbroede: could you tag this struct? ie. `typedef struct PoketchDotArtist {`

PR #744, opened 2025-09-07
nitrogfx: Output Converted Images with OAM Colors

h2o-DS:
Updates `NCGR` -> `png` conversion to output OAMs with the palette used in-game if both `-cell` and `-convertTo8Bpp` are passed

Ex:
type_icons
versus
type_icons

Also includes another bugfix with pl_font.narc's ignore file.

### DIFF
--- a/tools/nitrogfx/gfx.c
+++ b/tools/nitrogfx/gfx.c
@@ -260,7 +260,7 @@ static void ConvertFromTiles8Bpp(unsigned char *src, unsigned char *dest, int nu
     }
 }
 
-static void ConvertFromTiles8BppCell(unsigned char *src, unsigned char *dest, int oamWidth, int oamHeight, int imageWidth, int startX, int startY, bool hFlip, bool vFlip, bool hvFlip, bool toPNG)
+static void ConvertFromTiles8BppCell(unsigned char *src, unsigned char *dest, int oamWidth, int oamHeight, int imageWidth, int startX, int startY, bool hFlip, bool vFlip, bool hvFlip, int palette, bool toPNG)
 {
     int tilesSoFar = 0;
     int rowsSoFar = 0;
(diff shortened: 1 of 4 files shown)


tools/nitrogfx/gfx.c
```diff
@@ -290,10 +290,21 @@ static void ConvertFromTiles8BppCell(unsigned char *src, unsigned char *dest, in
...
                 {
-                    *dest++ = src[idxComponentY * pitch + idxComponentX];
+                    if (palette == -1)
+                    {
+                        *dest++ = src[idxComponentY * pitch + idxComponentX];
+                    }
+                    else
+                    {
+                        *dest++ = src[idxComponentY * pitch + idxComponentX] % 16;
+                    }
```
scbroede: i'm not sure this code would ever actually be reached? i believe when converting to 4bpp it's `ConvertFromTiles4BppCell` that ends up getting called, though i could be wrong. either way i'm not sure this needs to be here since what it's doing should be taken care of by the `& 0xF`s in `ConvertToTiles8Bpp`

h2o-DS:
Addressed.

tools/nitrogfx/main.c
```diff
@@ -155,6 +155,10 @@ void ConvertPngToNtr(char *inputPath, char *outputPath, struct PngToNtrOptions *
     struct Image image;
 
     image.bitDepth = options->bitDepth == 0 ? 4 : options->bitDepth;
+    if (options->convertTo4Bpp)
+    {
+        image.bitDepth = 8;
+    }
```
scbroede, line 161: this is probably better than expecting people to pass in `-bitdepth 8` every time lol

tools/nitrogfx/main.c
```diff
@@ -177,7 +181,7 @@ void ConvertPngToNtr(char *inputPath, char *outputPath, struct PngToNtrOptions *
 
     if (options->cellFilePath != NULL)
     {
-        ApplyCellsToImage(options->cellFilePath, &image, false, options->cellSnap, false);
+        ApplyCellsToImage(options->cellFilePath, &image, false, options->cellSnap, false, options->convertTo4Bpp);
```
scbroede: nitpick: this doesn't do anything regardless of the value of `options->convertTo4Bpp`. should it just be `false` instead?

h2o-DS: Sure

PR #747, opened 2025-09-09
Document Poketch Friendship Checker (Overlay 33)

VicSevenT:
This is the most complicated Poketch app I've done yet, with multiple state machines and a bunch of vector arithmetic to make things interesting. Here's a summary of the app's behavior:

* By default, all the pokemon in the player's party will move around the screen in random directions. The pokemon will bounce off of wall and each other (more on this below). I've called this default state `ACTION_WANDER`.
* When a pokemon is touched, it will stop. If it has enough friendship, it will display hearts depending on how high the friendship is (`ACTION_SHOW_LIKE` and `ACTION_SHOW_DISLIKE`).
* When the screen is touched, not on a pokemon, nearby friendly pokemon will approach the touched point, unfriendly pokemon will move away from it (`ACTION_GATHER` and `ACTION_RUN_AWAY`).
* When the screen is double-tapped, all pokemon will jump in place (`ACTION_JUMP`).

* There are a series of functions named `TapState*()` that manage a state machine tracking progress towards a double-tap. When the final state is reached, `ACTION_JUMP` is triggered.

* Movement and collision detection is handled in `UpdateIconPositions()`. It does the following on each update loop:
  *  Attempt to move each pokemon 16 steps, that is, multiply it's velocity by 16 and calculate the new position.
  *  Check if the pokemon has hit a wall or each other. If so determine at how many steps the collision would have first occurred. The collision that would happen the earliest is recorded.
  * Update all the mons positions up to the point of collision, and update the velocities for any pokemon involved in a collision. Subtract the elapsed steps from the remaining steps (starts as 16), and repeat the above steps. This loop continues until no collision are detected in the remaining steps, or a full 16 steps are processed.
  * Only after the collision-detection loop is completed, commit the position changes to the display.

Some expressions got very lengthy as I tried to make the code and names as clear as possible. Any suggestions for naming adjustments are welcome.


src/applications/poketch/friendship_checker/main.c
```diff
@@ -0,0 +1,250 @@
...
+
+        return TRUE;
+    }
+
+    return FALSE;
+}
+
+static u32 GetFriendshipLevel(u32 friendship)
+{
+    static const u8 friendshipTiers[] = {
```
lhearachel, line 132: **thought**: I wonder if we could parameterize these with some formulae applied to the base and max friendship values, in case a modder changes them.

src/applications/poketch/friendship_checker/graphics.c
```diff
@@ -0,0 +1,963 @@
...
+}
+
+static void PrintDebugLog(const char *text, ...)
+{
+    if (gSystem.heldKeys & PAD_BUTTON_B) {
+        va_list args;
+        va_start(args, text);
+        va_end(args);
+    }
+}
```
lhearachel, line 482: **thought**: Very surprising that this made it into the retail ROM!

src/applications/poketch/friendship_checker/graphics.c
```diff
@@ -0,0 +1,963 @@
...
+{
+    int x, y;
+
+    for (int slot = 0; slot < graphics->appState->monCount; slot++) {
+        x = graphics->pokemon[slot].position.x >> FX32_SHIFT;
+        y = graphics->pokemon[slot].position.y >> FX32_SHIFT;
+        x = (graphics->touchX - x) * (graphics->touchX - x);
+        y = (graphics->touchY - y) * (graphics->touchY - y);
+
+        if ((x + y) < (16 * 16)) {
```
lhearachel: **polish**: Remove extraneous parens.

---

**question**: What does `256` here mean?

src/applications/poketch/friendship_checker/graphics.c
```diff
@@ -0,0 +1,963 @@
...
+        NARC_dtor(pokeIconNarc);
+    }
+}
+
+static void SetRandomVelocity(VecFx32 *velocity)
+{
+    fx32 x, y;
+
+    x = (-32 + (MTRNG_Next() & 63)) * FX32_ONE;
+    y = (-32 + (MTRNG_Next() & 63)) * FX32_ONE;
```
lhearachel: **question**: Do these match as `MTRNG_Next() % 64`?

src/applications/poketch/friendship_checker/graphics.c
```diff
@@ -0,0 +1,963 @@
...
+            SysTask_Done(graphics->updateTask);
+        }
+
+        Heap_Free(graphics);
+    }
+}
+
+static const PoketchTask sFrienshipCheckerTasks[] = {
+    { 0x0, Task_InitGraphics, 0x0 },
+    { 0x1, Task_FreeBackground, 0x0 },
```
lhearachel: **polish**: Use decimal values unless hex confers special meaning.

lhearachel (commented):
This looks good! Thank you for the documentation in the description, it was very helpful to understand what was going on.

src/applications/poketch/friendship_checker/graphics.c
VicSevenT: Yup, updated. Also rearranged them to a straight subtraction instead of adding a negative.

src/applications/poketch/friendship_checker/graphics.c
VicSevenT: Oops, those are supposed to use enums anyway, updated.

src/applications/poketch/friendship_checker/graphics.c
VicSevenT: This is a distance calculation. `x` and `y` at this point actually contain the squares of the distance between the touched point and the icon's center point. In Pythagorean terms, the conditional is essentially `if (a^2 + b^2 < c^2)`, where c is 16. It's essentially checking if the screen was touched within a 16 unit circle around the icons center.

I added in another constant `ICON_RADIUS` to make this a little clearer.

lhearachel (approved):

PR #749, opened 2025-09-09
Unpack Poffin Case Graphics poru_gra.narc

h2o-DS:
Wasn't able to find where the sprites are loaded.

Documented the function `App_LoadGraphicMember` while I was there.


res/graphics/poffin_case/sub_sprites.png
lhearachel, line 1: **question**: Should this file not use `-cell`?

h2o-DS: There was overlap between the stars and the button. I can write a custom cell for it if desired.

lhearachel: I don't think there's a need; up to you.

lhearachel (approved):

PR #754, opened 2025-09-09
Document `software_sprite.c`

lhearachel:
This is a thin wrapper around the base SDK library for billboarding sprites into 3D space. Such sprites support a handful of additional features, which should be familiar to those with experience handling GBA-type sprites:

1. Per-object alpha blending, controlled by the `alpha` value on the `SoftwareSprite`.
2. Color modulations via a basic diffusion against an RGB-555 color.
3. Flipping across horizontal and vertical axes.
4. Z-rotation about an arbitrary "center" point.
5. Arbitrary vector-scaling along the horizontal and vertical axes.
6. Rendering priority via a "Z" coordinate.

Despite these functionalities, it is rarely used; the only uses of this code that I could find were limited to the Pokedex, the Starter Briefcase, and the Poffin Case.

### DIFF
--- a/src/software_sprite.c
+++ b/src/software_sprite.c
@@ -202,10 +202,10 @@ NNSG2dSVec2 SoftwareSprite_GetPosition(SoftwareSprite *sprite)
     return sprite->position;
 }
 
-void SoftwareSprite_SetCenter(SoftwareSprite *sprite, s16 param1, s16 param2)
+void SoftwareSprite_SetCenter(SoftwareSprite *sprite, s16 x, s16 y)
 {
-    sprite->center.x = param1;
-    sprite->center.y = param2;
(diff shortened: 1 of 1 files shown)


src/software_sprite.c
```diff
@@ -0,0 +1,467 @@
...
+    sprite->position.x = x;
+    sprite->position.y = y;
+}
+
+NNSG2dSVec2 SoftwareSprite_GetPosition(SoftwareSprite *sprite)
+{
+    return sprite->position;
+}
+
+void SoftwareSprite_SetCenter(SoftwareSprite *sprite, s16 param1, s16 param2)
```
scbroede: name these params

lhearachel:
Feel free to squash this one on merge.

PR #756, opened 2025-09-10
unpack ug_parts.narc

scbroede:
rather than using pal files for the images with rotated versions, i just generate the nclr from the non-rotated one. i figure this is fine since anyone editing those images would probably just edit that one and then rotate it as necessary to create the others.

the revive image is also used to generate the nclr for both it and max revive, which is probably less defensible; willing to use a pal file instead if deemed necessary.


res/graphics/mining/claw_fossil.png
lhearachel, line 1: **thought**: I wonder if we could write up a flag to transform these files from a single source input and still generate a match...

res/graphics/mining/max_revive.png
lhearachel, line 1: **thought**: I do think generating this image's palette from `revive.png` is a bit quirky, but still lean towards it being fine.

lhearachel:
Leaving this open in case @scbroede decides to implement a rotation function within `nitrogfx` for this PR and rework the build configuration.

scbroede:
updated to generate the rotated images from a single source (except the moon stone, which, annoyingly, is different by 2 pixels)

lhearachel (approved):

PR #758, opened 2025-09-10
Document Poketch Calculator (Overlay 28)

VicSevenT:
Another rather complicated app. Here's a breakdown of its components

* `main.c` is largely composed of a series of state handlers. They are all very similar, and the general flow is `enter operand 1`->`enter operator`->`enter operator 2`->`show result`->`enter operator`->`enter operand 2`->`show result`..., just like a real calculator. 
* All the calculator buttons and output are displayed in the background. `graphics.c` contains arrays of the tiles to display the buttons in the pressed position; the NSCR contains the arrangements for unpressed buttons. Most of the graphics task focus on updating the background, either by toggling a button, or displaying one of the calculators values (operand 1, 2, or result)
* This is the first (and only I think) poketch app with a third source file. `value.c` is dedicated to the `CalculatorValue`, a structure designed to hold a calculator friendly number. The value is stored as a whole number, with an additional flag to indicate whether it's negative, and a counter to if and where the decimal point is located
* The calculator can display up to 10 "symbols", which includes digits, a decimal point, and a minus sign for negative numbers. Decimal digits that don't fit get truncated away. If the value is too large or there is a divide by zero error, the output will show '????'.

While working on this app, two pieces came up that I intend to follow up on in
future PRs.
* I believe that the values of `BUTTON_TOUCH_PRESSED` and `BUTTON_TOUCH_RELEASED` in `poketch_button.h` had been reversed, based on what I was seeing in the calculator app and after looking a little in `poketch_system.c`. I flipped them in this PR, but now I think some other variables might be misnamed, so those need to be reviewed.
* I broke with a naming pattern that's been used in the poketch apps up to this point. I've changed the name of the state handlers in `main.c` from `Task_*` to `State_*` as they aren't actually tasks, and are really more the different pieces of the app's state machine. This also led to changing the naming convention of the associated enums and several struct fields. I'm planning to update the other already documented apps with the same style.

### DIFF
--- a/src/applications/poketch/calculator/graphics.c
+++ b/src/applications/poketch/calculator/graphics.c
@@ -44,147 +44,145 @@ static void UpdateValueGraphic(BgConfig *bgConfig, const u16 *displayValues);
 // clang-format off
 // These are all indices of tiles in the background tile NCGR that draw the
 // button in a pressed state
-static const u16 pressed0Tiles[BUTTON_SIZE_TILES_LARGE] = {
+static const u16 sPressed0Tiles[BUTTON_SIZE_TILES_LARGE] = {
     251, 252, 252, 252, 252, 252, 252, 253,
     291,  85,  86, 292, 292, 292, 292, 293,
     291, 125, 126, 292, 292, 292, 292, 293,
-    337, 338, 338, 338, 338, 338, 338, 339
+    337, 338, 338, 338, 338, 338, 338, 339,
 };
 
-static const u16 pressed1Tiles[BUTTON_SIZE_TILES_SMALL] = {
+static const u16 sPressed1Tiles[BUTTON_SIZE_TILES_SMALL] = {
     251, 252, 252, 253,
     291,  87,  88, 293,
     291, 127, 128, 293,
-    331, 332, 332, 333
+    331, 332, 332, 333,
 };
 
-static const u16 pressed2Tiles[BUTTON_SIZE_TILES_SMALL] = {
+static const u16 sPressed2Tiles[BUTTON_SIZE_TILES_SMALL] = {
     251, 252, 252, 253,
     291,  89,  90, 293,
     291, 129, 130, 293,
-    331, 332, 332, 333
+    331, 332, 332, 333,
 };
 
-static const u16 pressed3Tiles[BUTTON_SIZE_TILES_SMALL] = {
+static const u16 sPressed3Tiles[BUTTON_SIZE_TILES_SMALL] = {
     251, 252, 252, 253,
     291,  91,  92, 293,
     291, 131, 132, 293,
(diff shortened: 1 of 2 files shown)


src/applications/poketch/calculator/graphics.c
```diff
@@ -0,0 +1,476 @@
...
+    pressedMinusTiles,
+    pressedPlusTiles,
+    pressedTimesTiles,
+    pressedDivideTiles,
+    pressedEqualsTiles,
+    pressedClearTiles
+};
+
+// clang-format off
+static const struct {
```
lhearachel: **style**: I don't think there's harm in enveloping this array in the `clang-format off/on` tags.

src/applications/poketch/calculator/graphics.c
```diff
@@ -0,0 +1,476 @@
...
+    291, 113, 114, 293,
+    291, 153, 154, 293,
+    331, 332, 332, 333
+};
+
+static const u16 pressedEqualsTiles[BUTTON_SIZE_TILES_LARGE] = {
+    251, 252, 252, 252, 252, 252, 252, 253,
+    291, 115, 116, 292, 292, 292, 292, 293,
+    291, 155, 156, 292, 292, 292, 292, 293,
+    337, 338, 338, 338, 338, 338, 338, 339
```
lhearachel: **style**: Add a trailing comma to the final element of each of these arrays.

src/applications/poketch/calculator/graphics.c
```diff
@@ -0,0 +1,476 @@
...
+// These are all indices of tiles in the background tile NCGR that draw the
+// button in a pressed state
+static const u16 pressed0Tiles[BUTTON_SIZE_TILES_LARGE] = {
+    251, 252, 252, 252, 252, 252, 252, 253,
+    291,  85,  86, 292, 292, 292, 292, 293,
+    291, 125, 126, 292, 292, 292, 292, 293,
+    337, 338, 338, 338, 338, 338, 338, 339
+};
+
+static const u16 pressed1Tiles[BUTTON_SIZE_TILES_SMALL] = {
```
lhearachel: **style**: `sPressed1Tiles`, etc.

src/applications/poketch/calculator/main.c
```diff
@@ -0,0 +1,654 @@
...
+    }
+
+    return FALSE;
+}
+
+static void PlayResultSpeciesCry(PoketchCalculator *appData, const CalculatorValue *result)
+{
+    s64 lookupValue = CalculatorValue_GetSignedInt(result);
+
+    if ((lookupValue > 0) && (lookupValue <= NATIONAL_DEX_COUNT)) {
```
lhearachel: **polish**: Remove extraneous parens.

PR #759, opened 2025-09-11
Document hall of fame sequence (overlay086)

Viperio19:
This PR documents overlay086, which contains one file, `ov86_0223B140.c`, which I renamed to `league_hall_of_fame.c`. This file handles the hall of fame sequence that gets played after beating Cynthia, where first the individual pokemon are shown, and then the full party. `league_hall_of_fame.c` is mostly documented now, except for mostly some graphics stuff. The names could once again use some work probably, and the file can also probably be moved outside of overlay086, but I wanted to already make this PR to get some thoughts. For context, here is an overview of all the hall of fame related files at the moment:

- `hall_of_fame.c`: This mainly handles adding and getting hall of fame entries
- `pc_hall_of_fame/display.c` + `pc_hall_of_fame/manager.c`: Handles the hall of fame screen that you can access via the PC
- `league_hall_of_fame.c`: As said, this handles the hall of fame sequence at the end of the pokemon league

I ended up at `league_hall_of_fame.c` for this mainly to make it different from `hall_of_fame.c`, but I still just used the `HallOfFame_` prefix in `league_hall_of_fame.c`. I'd like to at least propose renaming `hall_of_fame.c` to `hall_of_fame_entries.c`. Maybe a batter name for `league_hall_of_fame.c` would be `hall_of_fame_sequence.c`, but yeah I just wanted to discuss this before moving forward.


lhearachel:
Of note, there is a checksum failure in CI.

This, as I understand, is a "cutscene"-type overlay; there is another instance of this for overlay 93, which is used for the cutscene that plays when riding the *S.S. Sinnoh*. I think that there is potential here for us to group these types of overlays into a uniform structure in the source tree, similar to the `applications` folders in `include` and `src`. Maybe the right approach is to have a pair of `cutscenes` folders that house these types of overlays?

With that sort of structure, this file could instead be `src/cutscenes/hall_of_fame.c`.

Viperio19:
> Of note, there is a checksum failure in CI.

```
../../../home/runner/work/pokeplatinum/pokeplatinum/subprojects/NitroSDK-4.2.30001/tools/makelcf/makelcf.c:95:17: warning: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 3 has type ‘char *’ [-Wformat=]
   95 |                 "NITRO-SDK Development Tool - makelcf - Make linker command file\n"
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
......
   98 |                 __DATE__);
      |                 ~~~~~~~~
      |                 |
      |                 char *
../../../home/runner/work/pokeplatinum/pokeplatinum/subprojects/NitroSDK-4.2.30001/tools/makelcf/makelcf.c:96:26: note: format string is defined here
   96 |                 "Build %lu\n\n"
      |                        ~~^
      |                          |
      |                          long unsigned int
      |                        %s
```      

The build passes locally, this is the only warning I see when going through the build and I'm not sure what to do with this

Viperio19:
> This, as I understand, is a "cutscene"-type overlay; there is another instance of this for overlay 93, which is used for the cutscene that plays when riding the _S.S. Sinnoh_. I think that there is potential here for us to group these types of overlays into a uniform structure in the source tree, similar to the `applications` folders in `include` and `src`. Maybe the right approach is to have a pair of `cutscenes` folders that house these types of overlays?
> 
> With that sort of structure, this file could instead be `src/cutscenes/hall_of_fame.c`.

I like this idea, but I would like to ask what we exactly define as a cutscene in that case. At the end of the hall of fame sequence/cutscene, there is a button press required to continue for example. I do still think this could be called a cutscene, and I'm in favor of having `.../cutscenes/` folders

Viperio19:
> > This, as I understand, is a "cutscene"-type overlay; there is another instance of this for overlay 93, which is used for the cutscene that plays when riding the _S.S. Sinnoh_. I think that there is potential here for us to group these types of overlays into a uniform structure in the source tree, similar to the `applications` folders in `include` and `src`. Maybe the right approach is to have a pair of `cutscenes` folders that house these types of overlays?
> > With that sort of structure, this file could instead be `src/cutscenes/hall_of_fame.c`.
> 
> I like this idea, but I would like to ask what we exactly define as a cutscene in that case. At the end of the hall of fame sequence/cutscene, there is a button press required to continue for example. I do still think this could be called a cutscene, and I'm in favor of having `.../cutscenes/` folders

There was another structure suggested on discord by cbt:
```
src/hall_of_fame_entries.c
src/hall_of_fame/pc.c (overlay87 -> hall_of_fame_pc)
src/hall_of_fame/pokemon_league.c (overlay86 -> hall_of_fame_pokemon_league)
```
I think both that or putting the league hall of fame in `.../cutscenes/` could be possible. Which structure would be preferred?

src/cutscenes/hall_of_fame.c
```diff
@@ -0,0 +1,1926 @@
...
+    SysTask *taskConfetti;
+    SysTask *tasks[4];
+    StringTemplate *strTemplate;
+    Strbuf *strbuf_1C30;
+    Strbuf *strbuf_1C4C;
+    MessageLoader *msgLoaderHallOfFame;
+    NARC *narc;
+} HallOfFameMan;
+
+typedef struct HallOfFameConfetto {
```
lhearachel: **typo**: `Confetti`

Viperio19: This was actually not a typo; confetto means a singular piece of confetti and that struct represents a singular piece of confetti, which is why I named it that way. If something else would be clearer though, I'll ofc still rename it.

lhearachel: That's fair, but I think `Confetti` gets the point across well enough and doesn't instill confusion / doubt.

PR #760, opened 2025-09-11
nitrogfx: allow applying cells without boundingRect

scbroede:
when the min and max x and y aren't explicitly provided we can still calculate them based on coordinates and dimensions.

sample output (using files 5, 6, and 7 in ug_trap.narc):

bubble

### DIFF
--- a/tools/nitrogfx/gfx.c
+++ b/tools/nitrogfx/gfx.c
@@ -669,35 +669,44 @@ static int SnapToTile(int val)
     return val;
 }
 
-static void CalculateOAMDimensions(struct OAM *oam, int *oamHeightOut, int *oamWidthOut)
+
+struct Dimensions {
+    int width;
+    int height;
+};
+
+static struct Dimensions CalculateOAMDimensions(struct OAM *oam)
 {
-    *oamHeightOut = 0;
-    *oamWidthOut = 0;
+    struct Dimensions oamdim = {
+        .width = 0,
+        .height = 0,
+    };
+
     int oamSize = oam->attr1.Size;
     switch (oam->attr0.Shape)
     {
     case 0:
-        *oamHeightOut = 1 << oamSize;
-        *oamWidthOut = *oamHeightOut;
+        oamdim.height = 1 << oamSize;
+        oamdim.width = oamdim.height;
         break;
     case 1:
         switch (oamSize)
         {
         case 0:
-            *oamHeightOut = 1;
-            *oamWidthOut = 2;
+            oamdim.height = 1;
+            oamdim.width = 2;
             break;
         case 1:
-            *oamHeightOut = 1;
-            *oamWidthOut = 4;
+            oamdim.height = 1;
+            oamdim.width = 4;
             break;
         case 2:
-            *oamHeightOut = 2;
-            *oamWidthOut = 4;
+            oamdim.height = 2;
+            oamdim.width = 4;
             break;
         case 3:
-            *oamHeightOut = 4;
-            *oamWidthOut = 8;
+            oamdim.height = 4;
+            oamdim.width = 8;
             break;
         }
         break;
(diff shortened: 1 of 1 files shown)


tools/nitrogfx/gfx.c
```diff
@@ -716,17 +775,59 @@ void ApplyCellsToImage(char *cellFilePath, struct Image *image, bool toPNG, bool
...
-            FATAL_ERROR("No bounding rectangle. Incompatible NCER\n");
+            int minX = 0;
+            int minY = 0;
+            int maxX = 0;
+            int maxY = 0;
+            for (int j = 0; j < options->cells[i]->oamCount; j++)
+            {
+                int oamHeight;
+                int oamWidth;
+                CalculateOAMDimensions(&options->cells[i]->oam[j], &oamHeight, &oamWidth);
```
lhearachel: **note**: You could return a `struct` here that contains the two values, if you prefer:

```c
struct Dimensions {
    int width;
    int height;
};

struct Dimensions CalculateOAMDimensions(struct OAM *oam) { /* snip */ }
```

```suggestion
struct Dimensions oamdim = CalculateOAMDimensions(&options->cells[i]->oam[j]);
```

tools/nitrogfx/gfx.c
```diff
@@ -669,6 +669,62 @@ static int SnapToTile(int val)
     return val;
 }
 
+static void CalculateOAMDimensions(struct OAM *oam, int *oamHeightOut, int *oamWidthOut)
+{
+    *oamHeightOut = 0;
+    *oamWidthOut = 0;
+    int oamSize = oam->attr1.Size;
```
h2o-DS, line 685: Maybe add a check for oamSize <4, especially for shape 0

tools/nitrogfx/gfx.c
```diff
@@ -669,6 +669,62 @@ static int SnapToTile(int val)
...
+    switch (oam->attr0.Shape)
+    {
+    case 0:
```
h2o-DS, line 688: Maybe this could wait for the refactor, but if we want this code more legible we could start adding enums for stuff like these.
Ie: square, wide rectangle, and tall rectangle. That could be output into the json too, but that's probably going too far for now.

tools/nitrogfx/gfx.c
```diff
@@ -716,17 +775,59 @@ void ApplyCellsToImage(char *cellFilePath, struct Image *image, bool toPNG, bool
...
+            int maxY = 0;
+            for (int j = 0; j < options->cells[i]->oamCount; j++)
+            {
+                int oamHeight;
+                int oamWidth;
+                CalculateOAMDimensions(&options->cells[i]->oam[j], &oamHeight, &oamWidth);
+                int xCoord = options->cells[i]->oam[j].attr1.XCoordinate;
+                if (xCoord & (1 << 8))
+                {
+                    xCoord |= ~0x1FF;
```
h2o-DS, line 804: Do we still need to do this after the signed output fix?

tools/nitrogfx/gfx.c
scbroede: what kind of check? it seems to run fine with shape 0 and size 1 (eg. files 1, 2, and 3 in ug_trap.narc)
petal

h2o-DS: I meant an out of bounds check to make sure the OAM isn't a weird size

scbroede: well < 4 is valid and this code handles it fine afaik, so what are the bounds, and are we worried about encountering an NCER that goes outside them?

h2o-DS: You're right, the check for validity should happen during the read step, not here. My bad

tools/nitrogfx/gfx.c
scbroede: yes. that only applied to converting NCERs to and from JSON. when providing an NCER file via `-cell` this still needs to happen

tools/nitrogfx/gfx.c
scbroede: yeah, let's do it that way in the rewrite

PR #385, opened 2025-02-15
Partially document Poketch Party Status app

TheSylphIsIn:
Documents portions of ov32, which is the Party Status poketch app. I documented most of the struct members and functions. However, there are some shared structs from ov25 that seem to be graphics-related but I feel are out of my depth to fully document. I figured I would go ahead and submit what I've done so far.

Also documents the PoketchTask structs/functions from ov25_02255090 which are shared among all poketch apps. There are other helper functions in that file that I didn't touch.

Note: I chose not to rename the ov32 files, but if this amount of documentation is sufficient, I was planning to call them `poketch_party_status_system`/`poketch_party_status_graphics`.

PR #761, opened 2025-09-12
Document Poketch Party Status (Overlay 32)

VicSevenT:
Much of this app was already documented in https://github.com/pret/pokeplatinum/pull/385. There were some graphics-related things undocumented, and then I also updated a bunch of names to be consistent with all the other apps.

This app shows the status of all the pokemon currently in the player's party. It shows their health, whether they are holding an item, and will show the mon as grayed out if it has a status condition.

The healthbar borders are part of the background and are drawn with 3 tiles that rotated as needed to cover all four sides. The  healthbar contents are in windows overlaying those borders.

If a pokemon icon is tapped, the icon will bounce a little. That bounce is slower if the pokemon has a status condition. If the screen is pressed somewhere besides a mon icon, the party data and all the graphics will be reloaded.


src/applications/poketch/party_status/graphics.c
```diff
@@ -0,0 +1,444 @@
...
+
+#include "graphics.h"
+#include "heap.h"
+#include "item.h"
+#include "narc.h"
+#include "pokemon_icon.h"
+#include "sys_task.h"
+#include "sys_task_manager.h"
+
+#define POKE_ICON_SIZE       16
```
lhearachel: **suggestion**: We've started using `_TILE_COUNT` to refer to constants like these that denote how many tiles a given graphics resource has, e.g., [in `include/render_window.h`](https://github.com/pret/pokeplatinum/blob/30b07a565f0be4e1ed51748b69e66d29f0e6fbd7/include/render_window.h#L13).

```suggestion
#define POKE_ICON_TILE_COUNT 16
```

src/applications/poketch/party_status/graphics.c
```diff
@@ -0,0 +1,444 @@
...
+#define POKE_ICON_SIZE_BYTES (POKE_ICON_SIZE * TILE_SIZE_4BPP)
+
+#define HP_BAR_WIDTH_TILES 8
+#define HP_BAR_SEGMENTS    ((HP_BAR_WIDTH_TILES * TILE_WIDTH_PIXELS) / 2)
+
+#define FLIP_H (1 << 10)
+#define FLIP_V (1 << 11)
```
lhearachel: **suggestion**: These are standard tile-flip flags for background tiles; they belong in `include/constants/graphics.h`, if there is not a related constant that we can pull from the SDK. When you move them, prefix them with `BG_TILE_` to denote that they are used by background tiles.

src/applications/poketch/party_status/graphics.c
```diff
@@ -0,0 +1,444 @@
...
+#include "narc.h"
+#include "pokemon_icon.h"
+#include "sys_task.h"
+#include "sys_task_manager.h"
+
+#define POKE_ICON_SIZE       16
+#define POKE_ICON_SIZE_BYTES (POKE_ICON_SIZE * TILE_SIZE_4BPP)
+
+#define HP_BAR_WIDTH_TILES 8
+#define HP_BAR_SEGMENTS    ((HP_BAR_WIDTH_TILES * TILE_WIDTH_PIXELS) / 2)
```
lhearachel: **suggestion**: Prefer `HEALTHBAR` rather than `HP_BAR`. This also applies to function names, where it would be `Healthbar` rather than `HpBar`.

src/applications/poketch/party_status/graphics.c
```diff
@@ -0,0 +1,444 @@
...
+static void SetupMonIconSprites(PartyStatusGraphics *graphics, const PartyStatus *partyData);
+static void StartMonIconBounceTask(PartyStatusGraphics *graphics);
+
+static void EndTask(PoketchTaskManager *taskMan);
+static void Task_DrawAppScreen(SysTask *task, void *taskMan);
+static void Task_HandleMonIconBounce(SysTask *task, void *bounceAnimData);
+static void Task_RedrawAppScreen(SysTask *task, void *taskMan);
+static void Task_FreeBackground(SysTask *task, void *taskMan);
+
+static const struct MonIconPosition {
```
lhearachel: **polish**: No need to tag this `struct` if it's never referenced by name or `typedef`.

src/applications/poketch/party_status/graphics.c
```diff
@@ -0,0 +1,444 @@
...
+{
+    u32 top, bottom, left, right;
+
+    for (u32 slot = 0; slot < partyCount; slot++) {
+        top = 8 + sMonPosition[slot].y - 16;
+        bottom = 8 + sMonPosition[slot].y + 16;
+        left = sMonPosition[slot].x - 16;
+        right = sMonPosition[slot].x + 16;
+
+        if ((x - left < right - left) & (y - top < bottom - top)) {
```
lhearachel: **question**: Interesting construct. Does this also match as a `&&`? If so, prefer to use that instead of a bitwise `&`, and remove the unnecessary parentheses around the comparisons.

src/applications/poketch/party_status/graphics.c
VicSevenT: Hmm, didn't find anything in the SDK after a brief lookaround. But I did rename and move the constants to that header.

src/applications/poketch/party_status/graphics.c
VicSevenT: Unfortunately, `&&` is non-matching. Removed the parentheses though.

lhearachel: Mmmmmmh, no, keep the parentheses if `&&` is non-matching. Odd that it is, though.

lhearachel (approved):

PR #762, opened 2025-09-13
Document Poketch Analog Watch (Overlay 30)

VicSevenT:
A pretty simple app. The watch face updates every minute, the hour hand updates every other minute. Every position of each hand has a separate animation, so there's 60 for the minute hand and 360 (12 hours * 30 positions per hour) for the hour hand.

When touched, the screen shifts palettes to become a little brighter.

### DIFF
--- a/src/applications/poketch/analog_watch/main.c
+++ b/src/applications/poketch/analog_watch/main.c
@@ -46,12 +46,15 @@ static BOOL State_LoadApp(PoketchAnalogWatch *appData);
 static BOOL State_UpdateApp(PoketchAnalogWatch *appData);
 static BOOL State_UnloadApp(PoketchAnalogWatch *appData);
 
-static const TouchScreenRect screenHitBox[] = {
-    { .rect = {
-          .top = POKETCH_SCREEN_MIN_Y,
-          .bottom = POKETCH_SCREEN_MAX_Y,
-          .left = POKETCH_SCREEN_MIN_X,
-          .right = POKETCH_SCREEN_MAX_X } }
+static const TouchScreenRect sScreenHitBox[] = {
+    {
+        .rect = {
+            .top = POKETCH_SCREEN_MIN_Y,
+            .bottom = POKETCH_SCREEN_MAX_Y,
+            .left = POKETCH_SCREEN_MIN_X,
+            .right = POKETCH_SCREEN_MAX_X,
+        },
+    },
 };
 
 static void NitroStaticInit(void)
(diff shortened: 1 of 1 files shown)


src/applications/poketch/analog_watch/main.c
```diff
@@ -0,0 +1,232 @@
...
+static BOOL State_UpdateApp(PoketchAnalogWatch *appData);
+static BOOL State_UnloadApp(PoketchAnalogWatch *appData);
+
+static const TouchScreenRect screenHitBox[] = {
+    { .rect = {
+          .top = POKETCH_SCREEN_MIN_Y,
+          .bottom = POKETCH_SCREEN_MAX_Y,
+          .left = POKETCH_SCREEN_MIN_X,
+          .right = POKETCH_SCREEN_MAX_X } }
+};
```
lhearachel: *polish*: `sScreenHitBox`

---

**polish**: Add trailing commas to force `clang-format` to render this correctly:

```suggestion
static const TouchScreenRect screenHitBox[] = {
    { 
        .rect = {
            .top = POKETCH_SCREEN_MIN_Y,
            .bottom = POKETCH_SCREEN_MAX_Y,
            .left = POKETCH_SCREEN_MIN_X,
            .right = POKETCH_SCREEN_MAX_X,
        },
     },
};
```

Ordinarily, I'd prefer this to be in a tabular format, but this is fine since it's the only hitbox.

src/applications/poketch/analog_watch/graphics.c
```diff
@@ -0,0 +1,218 @@
...
+    CP_SetDiv32_32(graphics->watchData->time.hour, 10);
+
+    // Hour hand - index changes every 2 minutes, 360 possible positions
+    animIdx = 60 + ((graphics->watchData->time.hour % 12) * 30) + (graphics->watchData->time.minute / 2);
+    PoketchAnimation_UpdateAnimationIdx(graphics->animSpriteData[0], animIdx);
+
+    // Minute hand
+    animIdx = 0 + graphics->watchData->time.minute;
+    PoketchAnimation_UpdateAnimationIdx(graphics->animSpriteData[1], animIdx);
+}
```
lhearachel, line 174: **question**: How does the rotation work? The surrounding code implies that file 29 of this NARC is the character-set for the sprites, and then 27 and 28 are the NCER and NANR, respectively. The implication that I take from this code is that the index in the However, when I open these files in Tinke, the angle doesn't seem to change as I play with the animation index.

lhearachel (commented):
Looks good! One polish comment and a leading question.

src/applications/poketch/analog_watch/graphics.c
VicSevenT: You are correct about the NARC indices for the assets. I hadn't used tinke before, but opened it up and saw the same thing as you, that nothing changed animation to animation. But I initially inspected the files with NitroPaint, and this is what I see there (disregard the messed up UI):
image

Note that there are only 2 cells, one for each hand. Each animation then sets a different rotation value. Again, I'm not very familiar with tinke, but maybe it doesn't support that rotation property?

lhearachel: Seems like it doesn't, yeah; good to know that NitroPaint does, though! 👍

PR #763, opened 2025-09-13
Document scripts for pokedex ratings

metallicity:
Documents a few scripts which display all the various messages for a professor rating the player's pokedex, which are invoked by `CallCommonScript`.


generated/vars_flags.txt
```diff
@@ -270,7 +270,7 @@ FLAG_UNK_0x010C
 FLAG_UNK_0x010D
 FLAG_UNK_0x010E
 FLAG_UNK_0x010F
-FLAG_UNK_0x0110
+FLAG_COMPLETE_LOCAL_DEX_RATING
```
cbt6:
```suggestion
FLAG_COMPLETED_LOCAL_DEX
```

src/scrcmd.c
```diff
@@ -4826,20 +4826,20 @@ static BOOL ScrCmd_Unused_122(ScriptContext *ctx)
...
-static BOOL ScrCmd_123(ScriptContext *ctx)
+static BOOL ScrCmd_LoadPokedexRating(ScriptContext *ctx)
 {
-    const Pokedex *v0 = SaveData_GetPokedex(ctx->fieldSystem->saveData);
-    const TrainerInfo *v1 = SaveData_GetTrainerInfo(ctx->fieldSystem->saveData);
-    u8 v2 = ScriptContext_ReadByte(ctx);
-    u16 *v3 = ScriptContext_GetVarPointer(ctx);
-    u16 v4;
+    const Pokedex *pokedex = SaveData_GetPokedex(ctx->fieldSystem->saveData);
+    const TrainerInfo *player = SaveData_GetTrainerInfo(ctx->fieldSystem->saveData);
```
```suggestion
    const TrainerInfo *trainerInfo = SaveData_GetTrainerInfo(ctx->fieldSystem->saveData);
```

generated/vars_flags.txt
metallicity: This flag is only set when the pokedex is actually rated, after having completing the local dex, which then skips a section of the script when next speaking to Rowan in order to obtain the national dex. If I were changing the name at all, it would be to something more explicit, like `FLAG_ROWAN_RATED_COMPLETE_LOCAL_DEX` or `FLAG_COMPLETE_LOCAL_DEX_ALREADY_RATED`.

The logic essentially looks like this (when the player talks to Rowan and has yet to receive the national dex):

```c
if (CheckLocalDexCompleted() == TRUE) {
  if (COMPLETE_LOCAL_DEX_ALREADY_RATED == FALSE) {
    if (PromptPokedexRating() == MENU_NO) return;
    // Rowan rates the complete local dex
  }
  // Oak shows up and gives the player the national dex
}
```

To be even more precise, the flag is only set when the player specifically gets their complete local dex "rated" somewhere else before this interaction (e.g., through the PC), since this uses its own special dialog rather than calling the shared rating script which sets the flag. Which isn't really a mistake, since it's never checked anywhere else and gaining the national dex afterwards prevents this branch from ever being reached again, so it's probably safe to omit that level of detail when naming the flag.

cbt6: I'll be honest this reply left me more confused than before. Is the distinction about having the dex rated important? In fact I'm now leaning towards simply `FLAG_SHOULD_RECEIVE_NATIONAL_DEX`.

metallicity: It's entirely possible to receive the national dex without ever setting this flag. All it does is skip the part where Rowan asks to check your dex progress in person if he hasn't already learned you completed the Sinnoh dex. The flag does not reflect the player's dex progress or ability to upgrade to the national dex, just that Rowan already knows when you next speak to him.

Both the names you've suggested imply that Oak should not show up unless the flag is set, but the flag not being set only gives the player the option for Oak to not show up, by declining to have their dex rated (though they can just immediately speak to Rowan again after that). Furthermore, the flag is only checked in a place where those conditions (local dex completion/national dex eligibility) have just been verified by `CheckLocalDexCompleted`, so it would give the appearance of a dummy branch when in fact it is the much more typical branch for players to take.

lhearachel: I think there's good argumentation to keep naming simple even if it's not 100% perfectly descriptive of what the flag controls. `FLAG_SHOWN_COMPLETE_LOCAL_DEX`?

metallicity: `SHOWN_COMPLETE_LOCAL_DEX` sounds fine to me.

generated/vars_flags.txt
```diff
@@ -270,7 +270,7 @@ FLAG_UNK_0x010C
 FLAG_UNK_0x010D
 FLAG_UNK_0x010E
 FLAG_UNK_0x010F
-FLAG_UNK_0x0110
+FLAG_COMPLETE_LOCAL_DEX_ALREADY_RATED
```
lhearachel: **nit**: Update this as noted in commentary above: `SHOWN_COMPLETE_LOCAL_DEX`

lhearachel (approved):

metallicity:
@lhearachel Ready to merge.

PR #765, opened 2025-09-14
Document Berry patch management, graphics, and scripting commands

MrKumaPants:
#325 Documents unk_0204B64C.c - Berry-related script commands along with the rest of the berry functionality


include/overlay005/const_ov5_021FB67C.h
```diff
@@ -3,6 +3,6 @@
 
 #include "overlay005/struct_ov5_021FB67C.h"
 
-extern const UnkStruct_ov5_021FB67C Unk_ov5_021FB67C[];
+extern const BerryGraphicsData BerryGraphicsTable[];
```
Gudf: Prefix globals with `g`.
```suggestion
extern const BerryGraphicsData gBerryGraphicsTable[];
```

src/berry_patch_graphics.c
```diff
@@ -0,0 +1,234 @@
...
+            BerryPatches_GetBerryID(MapObject_FieldSystem(mapObject), mapObject), currentGrowthStage);
+
+        if (graphicsData->graphicsResourceID != 0xffff) {
+            if (graphicsData->lastGrowthStage != BERRY_GROWTH_STAGE_NONE) {
+                ov5_021F22BC(mapObject);
+            }
+
+            ov5_021ECEB4(mapObject, &graphicsData->graphicsObject, graphicsData->graphicsResourceID);
+        } else {
+            if ((graphicsData->lastGrowthStage != BERRY_GROWTH_STAGE_NONE) && (patchData->needsUpdate == 0)) {
```
Gudf: Remove extraneous parentheses.

src/berry_patch_graphics.c
```diff
@@ -0,0 +1,234 @@
...
+        }
+    }
+
+    return 0xffff;
+}
+
+void BerryPatch_MarkForUpdate(MapObject *mapObject)
+{
+    BerryPatchData *patchData = sub_02062A78(mapObject);
+    patchData->needsUpdate = 1;
```
Gudf: This should probably use `TRUE` and `FALSE` instead of `1` and `0` as well.

src/berry_patch_graphics.c
```diff
@@ -0,0 +1,234 @@
...
+    enum BerryGrowthStage lastGrowthStage;
+    UnkStruct_020216E0 *graphicsObject;
+    UnkStruct_ov5_021ED01C graphicsState;
+} BerryPatchGraphics;
+
+static int BerryPatch_GetGraphicsResourceID(int berryID, enum BerryGrowthStage growthStage);
+
+BOOL BerryPatch_IsBerryPatch(int graphicsID)
+{
+    if (graphicsID == 0x64) {
```
Gudf: Use decimal literals when hex provides no benefit.

src/berry_patch_graphics.c
```diff
@@ -0,0 +1,234 @@
...
+                return 0x1000 + 0x0;
+            case BERRY_GROWTH_STAGE_GROWING:
+                return graphicsData->stage3ResourceID;
+            case BERRY_GROWTH_STAGE_BLOOMING:
+                return graphicsData->stage4ResourceID;
+            case BERRY_GROWTH_STAGE_FRUIT:
+                return graphicsData->stage5ResourceID;
+            }
+        }
+    }
```
Gudf: Remove the extraneous block delimiters.

src/berry_patch_graphics.c
```diff
@@ -0,0 +1,234 @@
...
+        return 0xffff;
+    default: {
+        berryID--;
+
+        {
+            const BerryGraphicsData *graphicsData = &BerryGraphicsTable[berryID];
+
+            switch (growthStage) {
+            case BERRY_GROWTH_STAGE_SPROUTED:
+                return 0x1000 + 0x0;
```
Gudf: Calculations like this can usually be folded:
```suggestion
                return 0x1000;
```

src/berry_patch_manager.c
```diff
@@ -0,0 +1,475 @@
...
+    int timer;
+    SysTask *animationTask;
+} BerryWateringTask;
+
+static void BerryPatchManager_Init3DRendering(FieldSystem *fieldSystem, BerryPatchManager *manager);
+static void BerryPatchManager_Cleanup3DRendering(BerryPatchManager *manager);
+
+static u16 BerryPatches_ConvertBerryIDToItemID(int berryID)
+{
+    if (berryID == 0) {
```
Gudf: Replace literals with the appropriate constants.

include/overlay005/struct_ov5_021FB67C.h
```diff
@@ -2,9 +2,9 @@
 #define POKEPLATINUM_STRUCT_OV5_021FB67C_H
 
 typedef struct {
```
Gudf: Tag the structs with their `typedef`'d name as well.
```suggestion
typedef struct BerryGraphicsData {
```
Also applies to the other files.

include/overlay005/struct_ov5_021FB67C.h
Gudf, line 1: Can this be moved to an appropriate code header? Otherwise, the file should be renamed.

src/berry_patch_manager.c
```diff
@@ -0,0 +1,475 @@
...
+    Heap_Free(manager);
+}
+
+void BerryPatches_ElapseTime(FieldSystem *fieldSystem, int minutes)
+{
+    BerryPatch *berryPatches;
+    BerryGrowthData *growthData;
+
+    if (fieldSystem->unk_04 == NULL) {
+        growthData = BerryGrowthData_Init(11);
```
Gudf: Is this 11 an heap ID? Same with the 17 a bit further down.

src/berry_patch_manager.c
```diff
@@ -0,0 +1,475 @@
...
+    const VecFx32 scale = { FX32_ONE, FX32_ONE, FX32_ONE };
+    MtxFx33 transform;
+
+    MTX_Identity33(&transform);
+
+    if (GFXBoxTest_IsModelInView(fieldSystem->unk_04->berryPatchManager->model, position, &transform, &scale) != 0) {
+        return TRUE;
+    } else {
+        return FALSE;
+    }
```
Gudf: Does this match?
```suggestion
    return GFXBoxTest_IsModelInView(fieldSystem->unk_04->berryPatchManager->model, position, &transform, &scale);
```

src/berry_patch_manager.c
```diff
@@ -0,0 +1,475 @@
...
+    MapObject *mapObject;
+    BerryPatch *berryPatches = MiscSaveBlock_GetBerryPatches(fieldSystem->saveData);
+
+    while (sub_020625B0(fieldSystem->mapObjMan, &mapObject, &objectIndex, (1 << 0)) == 1) {
+        if (BerryPatch_IsBerryPatch(MapObject_GetGraphicsID(mapObject)) == 1) {
+            if (BerryPatches_IsInView(fieldSystem, MapObject_GetPos(mapObject))) {
+                int patchID = MapObject_GetDataAt(mapObject, 0);
+                BerryPatches_SetIsPatchGrowing(berryPatches, patchID, 1);
+            }
+        }
```
Gudf: Can you combine these two `if`s?
```suggestion
        if (BerryPatch_IsBerryPatch(MapObject_GetGraphicsID(mapObject)) == TRUE && BerryPatches_IsInView(fieldSystem, MapObject_GetPos(mapObject))) {
            int patchID = MapObject_GetDataAt(mapObject, 0);
            BerryPatches_SetIsPatchGrowing(berryPatches, patchID, TRUE);
        }
```

src/berry_patch_graphics.c
```diff
@@ -0,0 +1,234 @@
...
+
+int BerryPatch_GetCurrentGraphicsResourceID(const MapObject *mapObject)
+{
+    {
+        BerryPatchData *patchData = sub_02062A78((MapObject *)mapObject);
+
+        if (patchData->growthStage == BERRY_GROWTH_STAGE_NONE) {
+            return 0xffff;
+        }
+    }
```
Gudf: Extraneous block delimiters.

src/berry_patch_manager.c
```diff
@@ -0,0 +1,475 @@
...
+{
+    BerryPatchManager_Cleanup3DRendering(manager);
+    Heap_Free(manager->growthData);
+    Heap_Free(manager);
+}
+
+void BerryPatches_ElapseTime(FieldSystem *fieldSystem, int minutes)
+{
+    BerryPatch *berryPatches;
+    BerryGrowthData *growthData;
```
Gudf: Try and inline the variable declarations with their definitions where it matches.

include/overlay005/struct_ov5_021FB67C.h
```diff
@@ -2,9 +2,9 @@
...
-    u32 unk_00;
-    u32 unk_04;
-    u32 unk_08;
-} UnkStruct_ov5_021FB67C;
+    u32 stage3ResourceID;
+    u32 stage4ResourceID;
+    u32 stage5ResourceID;
```
Gudf: These should probably be using the growth stages names, e.g `growing`, `blooming`, `fruit`.

src/overlay005/ov5_021F17B8.c
```diff
@@ -340,8 +339,7 @@ static void ov5_021F1B4C(UnkStruct_ov5_021F1AD8 *param0, VecFx32 *param1)
     fx32 v2;
     fx32 v3[3] = {
         (FX32_ONE * 12),
-        (FX32_ONE * 16),
-        (FX32_ONE * 12)
+        FX32_ONE * 16, (FX32_ONE * 12)
```
Gudf: Add a comma at the end to make the formatter behave properly.

include/scrcmd_berry.h
```diff
@@ -0,0 +1,21 @@
+#ifndef POKEPLATINUM_SCRCMD_BERRY_H
+#define POKEPLATINUM_SCRCMD_BERRY_H
+
+#include "field_script_context.h"
+
+enum BerryWateringCommand {
+    BERRY_WATERING_START = 0, // Start berry watering mode
+    BERRY_WATERING_END = 1, // End berry watering mode
+};
```
lhearachel: **issue**: I see that these constants are used in `ScrCmd_BerryWatering`, so this requires some extra consideration for compatibility with the assembler used to generate byte-code scripts. The assembler we use—`arm-none-eabi-gcc`, with `-x assembler-with-cpp`—does not handle C `enum`s like this, but it *does* handle C preprocessor directives.

We (currently) get around this for some constants that we *want* to have a `enum` in the C code by generating both the `enum` and a matching set of `#define`s from a text file (see `generated` for the implementation). However, given the scope of these specific constants, I would go simpler: port these definitions to `include/constants/scrcmd.h` and translate them to `#define`s instead of an `enum`.

src/scrcmd.c
```diff
@@ -1150,15 +1150,15 @@ const ScrCmdFunc Unk_020EAC58[] = {
...
-    ScrCmd_185,
+    ScrCmd_GetBerryGrowthStage,
+    ScrCmd_GetBerryItemID,
+    ScrCmd_GetBerryMulchType,
+    ScrCmd_GetBerryMoisture,
+    ScrCmd_GetBerryYield,
+    ScrCmd_SetBerryMulch,
+    ScrCmd_PlantBerry,
+    ScrCmd_BerryWatering,
+    ScrCmd_HarvestBerry,
```
lhearachel: **todo**: Rename the corresponding macros in `asm/macros/scrcmd.inc`—as well as their usage in any of the field scripts (`res/field/scripts/*.s`)—and document any arguments passed to them. When doing so, remove the `ScrCmd_` prefix from the macros; that is:

| This C function              | Becomes this macro    |
| :--------------------------- | :-------------------- |
| `ScrCmd_GetBerryGrowthStage` | `GetBerryGrowthStage` |
| `ScrCmd_GetBerryItemID`      | `GetBerryItemID`      |
| `ScrCmd_GetBerryMulchType`   | `GetBerryMulchType`   |
| `ScrCmd_GetBerryMoisture`    | `GetBerryMoisture`    |
| `ScrCmd_GetBerryYield`       | `GetBerryYield`       |
| `ScrCmd_SetBerryMulch`       | `SetBerryMulch`       |
| `ScrCmd_PlantBerry`          | `PlantBerry`          |
| `ScrCmd_BerryWatering`       | `BerryWatering`       |
| `ScrCmd_HarvestBerry`        | `HarvestBerry`        |

src/scrcmd.c
```diff
@@ -1150,15 +1150,15 @@ const ScrCmdFunc Unk_020EAC58[] = {
...
-    ScrCmd_184,
-    ScrCmd_185,
+    ScrCmd_GetBerryGrowthStage,
+    ScrCmd_GetBerryItemID,
+    ScrCmd_GetBerryMulchType,
+    ScrCmd_GetBerryMoisture,
+    ScrCmd_GetBerryYield,
+    ScrCmd_SetBerryMulch,
+    ScrCmd_PlantBerry,
+    ScrCmd_BerryWatering,
```
lhearachel: **suggestion**: We favor leading verbs in function names, especially for script-commands like these. I suggest `ScrCmd_WaterBerry`; this would also apply to the macro, as described above.

src/berry_patch_manager.c
```diff
@@ -0,0 +1,470 @@
...
+static void BerryPatchManager_Init3DRendering(FieldSystem *fieldSystem, BerryPatchManager *manager);
+static void BerryPatchManager_Cleanup3DRendering(BerryPatchManager *manager);
+
+static u16 BerryPatches_ConvertBerryIDToItemID(int berryID)
+{
+    if (berryID == 0) {
+        return 0;
+    }
+
+    return berryID + 149 - 1;
```
lhearachel: **polish**: `149` here refers to `ITEM_CHERI_BERRY`; in `include/constants/items.h`, there is a special constant `FIRST_BERRY_IDX` that points to this item ID.

src/berry_patch_manager.c
```diff
@@ -0,0 +1,470 @@
...
+    return itemID - 149 + 1;
+}
+
+static u16 BerryPatches_ConvertMulchTypeToItemID(int mulchType)
+{
+    if (mulchType == 0) {
+        return 0;
+    }
+
+    return mulchType + 95 - 1;
```
lhearachel: **suggestion**: Similar to the above comment re:`149`, `95` here refers to `ITEM_GROWTH_MULCH`. We don't have a `FIRST_MULCH_IDX` constant at the moment, but maybe we should. Feel free to add one to `include/constants/items.h`. 🙂

src/berry_patch_manager.c
```diff
@@ -0,0 +1,470 @@
...
+static const MapObjectAnimCmd BerryWatering_LeftAnimation[] = {
+    { 0xA, 0x1 }, // Play animation frame 0xA for 1 frame
+    { 0xfe, 0x0 } // End of animation sequence
+};
+
+// Animation sequence for watering berries while facing right
+static const MapObjectAnimCmd BerryWatering_RightAnimation[] = {
+    { 0xB, 0x1 }, // Play animation frame 0xB for 1 frame
+    { 0xfe, 0x0 } // End of animation sequence
+};
```
lhearachel: **polish**: Use decimal `0`, `1`, `10`, and `11` instead of the equivalent hexadecimal values here, but keep `0xfe`.

src/berry_patch_manager.c
```diff
@@ -0,0 +1,470 @@
...
+    if (direction == FACE_LEFT) {
+        playerX--;
+    } else if (direction == FACE_RIGHT) {
+        playerX++;
+    } else if (direction == FACE_UP) {
+        playerZ--;
+    } else if (direction == FACE_DOWN) {
+        playerZ++;
+    } else {
+        GF_ASSERT(0);
```
lhearachel: **polish**: Use `GF_ASSERT(FALSE)` instead of `GF_ASSERT(0)`.

lhearachel (commented):
Great work, and thanks for digging into this! 🙏 Feel free to ask any questions regarding these comments in-line or on Discord.

src/berry_patch_manager.c
fuddlesworth: Technically it should yeah. I'll simplify.

src/berry_patch_manager.c
fuddlesworth: 11 is a heap id. Updated to be HEAP_ID_FIELD2

Looks like 17 is an index into a NARC file

lhearachel: Don't worry about documenting indices into NARC files; semantic constants for those get generated for member files during the build process.

src/berry_patch_manager.c
fuddlesworth: Simplified but kept the != 0 else hash verification fails.

Gudf: Ah, that makes sense, I forgot that function doesn't return a `BOOL`.

include/scrcmd_berry.h
fuddlesworth: Moved it to `include/constants/scrcmd.h` and converted to defines.

include/overlay005/berry_graphics_data.h
```diff
@@ -0,0 +1,10 @@
+#ifndef POKEPLATINUM_BERRY_GRAPHICS_DATA_H
+#define POKEPLATINUM_BERRY_GRAPHICS_DATA_H
+
+typedef struct {
```
scbroede: tag this struct

res/field/scripts/scripts_unk_0413.s
```diff
@@ -112,10 +112,10 @@ _01F2:
     ShowYesNoMenu VAR_RESULT
     GoToIfEq VAR_RESULT, MENU_NO, _053A
     CloseMessage
-    ScrCmd_184 0
+    WaterBerry 0
```
scbroede: use the `BERRY_WATERING_START` and `BERRY_WATERING_END` constants instead of 0 and 1 (also applies to all other uses of `WaterBerry`).

also given how it works, maybe `WaterBerry` isn't the best name? something along the lines of `SetBerryWateringState` would be more clear imo

src/berry_patch_graphics.c
```diff
@@ -0,0 +1,229 @@
...
+#include "berry_patch_manager.h"
+#include "berry_patches.h"
+#include "map_header_data.h"
+#include "map_object.h"
+#include "unk_02020AEC.h"
+
+typedef struct {
+    u16 growthStage;
+    u16 needsUpdate;
+} BerryPatchData;
```
scbroede, line 25: tag this struct (and the one below)

src/berry_patch_manager.c
```diff
@@ -0,0 +1,468 @@
...
+
+struct BerryPatchManager {
+    int heapID;
+    BerryGrowthData *growthData;
+    NNSG3dRenderObj renderObj;
+    NNSG3dResMdl *model;
+    NNSG3dResFileHeader *resource;
+};
+
+typedef struct {
```
scbroede: tag this struct

src/berry_patch_manager.c
```diff
@@ -0,0 +1,468 @@
...
+    TVBroadcast_RecordBerryHarvest(fieldSystem, BerryPatches_ConvertTagNumberToItemID(berryID), BerryPatches_GetPatchYieldRating(berryPatches, patchID), yieldAmount);
+    BerryPatches_HarvestPatch(berryPatches, patchID);
+    BerryPatch_MarkForUpdate(mapObject);
+
+    return Bag_TryAddItem(SaveData_GetBag(fieldSystem->saveData), BerryPatches_ConvertTagNumberToItemID(berryID), yieldAmount, HEAP_ID_FIELD1);
+}
+
+void BerryPatches_SetMulchType(FieldSystem *fieldSystem, MapObject *mapObject, u16 mulchItemID)
+{
+    int patchID;
```
scbroede: does this still match if you merge the declaration with the assignment? (ditto for all similar functions below)

res/field/scripts/scripts_unk_0413.s
fuddlesworth: Yeah, that does make more sense. Renamed.

src/berry_patch_manager.c
```diff
@@ -48,40 +50,40 @@ typedef struct {
...
-        return 0;
+    if (itemID == ITEM_NONE) {
+        return MULCH_TYPE_NONE;
     }
 
-    return itemID - 95 + 1;
+    return itemID - FIRST_MULCH_IDX + 1;
 }
 
 BerryPatchManager *BerryPatchManager_Create(FieldSystem *fieldSystem, int heapID)
```
lhearachel: **polish**: The common terminology-set we use for functions like this and others is:

- `Alloc` -> allocates memory for some `struct`
- `Init` -> initializes an allocated `struct`
- `New` -> `Alloc` + `Init`
- `Free` -> teardown procedures

---

**suggestion**: Prefer `enum HeapId heapID` to `int heapID`, if it produces a match.

fuddlesworth: Addressed both issues.

PR #766, opened 2025-09-15
Document Poketch Pedometer (Overlay 36)

VicSevenT:
This is another very simple app, and it is almost identical to the Counter app, both visually and in source code.

On every update loop iteration, it check the step count, and if it is updated, it triggers an update of the step count graphically. When the button is pressed, the step count is reset, which involves a call to update the internal step count variable, a call to update the step count graphics, and calls to animate the button being pressed.


src/applications/poketch/pedometer/main.c
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    return FALSE;
+}
+
+static BOOL Init(PoketchPedometer *appData, PoketchSystem *poketchSys, BgConfig *bgConfig, u32 appID)
+{
+    static const TouchScreenRect buttonHitBox[] = {
+        { .rect = { .top = 100, .bottom = 156, .left = 82, .right = 146 } },
+    };
```
lhearachel, line 74: **question**: Would it make sense to use the same paradigm for all the applets? There's some value in consistency here; my initial opinion is that either *all* Poketch applets should use this paradigm of "the touch-screen hitboxes are declared and defined inside `Init` to limit scope-access," or *none* of them do.

Whatever the answer here is, I would say alignment is out-of-scope for this PR, though the decision should be applied to this applet as needed.

src/applications/poketch/pedometer/main.c
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    return FALSE;
+}
+
+static void ButtonCallback(u32 buttonID, u32 buttonState, u32 touchState, void *appData)
+{
+    PoketchPedometer *pedometer = appData;
+
+    if (buttonID == 0) {
```
lhearachel, line 98: **question**: Is there value in constants for button IDs at all?

src/applications/poketch/pedometer/main.c
```diff
@@ -0,0 +1,235 @@
...
+    Heap_Free(appData);
+}
+
+static void Task_Main(SysTask *task, void *appData)
+{
+    static BOOL (*const stateFuncs[])(PoketchPedometer *) = {
+        State_LoadApp,
+        State_UpdateApp,
+        State_UnloadApp
+    };
```
lhearachel, line 116: **note**: The previous question for touch-screen hit-boxes applies to these state-function arrays, as well.

src/applications/poketch/pedometer/main.c
VicSevenT: Hmm, I can get behind that. I didn't realize that these static objects could be moved in or out of functions without breaking matching, I just carried it over from the undocumented version. But I agree on the value of consistency. 

Regarding the note below, I think all of the state-function arrays to this point have been defined in function, so those are already consistent and it feels like it makes sense to keep that as is. In which case it probably makes sense to follow that pattern for the hitboxes as well. The one point I have to consider against the hitboxes being all in function is demonstrated by the calculator app, where the hitbox array is 17 elements, that it may unnecessarily break up the function declaration from the body more that is ideal. But that's only one case, and if you don't have a problem with it, I don't either.

I hope I am not speaking too much in circles here. Ultimately, I favor moving towards a consistent structure, and I think I lean more towards defining in function in that case. After this app I was planning on taking a break and going back to make some consistency updates anyway, so whatever the decision, applying it to all the other apps will be coming in short order.

src/applications/poketch/pedometer/main.c
VicSevenT: I don't see a ton of value, particularly for an app like this where there's only one button. The ID is just based on the button's index in the hitbox table, so it's not too magic of a number, plus creating a constant wouldn't have the benefit of creating a single point in code to modify in the future - any change to the buttons would also require a change to the hitbox array. Also, there's enough variance in how the button callbacks behave across apps, some check buttonID immediately, some just save it for later use in a state function, some ignore it completely, that enforcing a consistent button ID naming structure for all apps wouldn't really result in any improvement in app structure consistency.

lhearachel: Sounds reasonable to me!

VicSevenT:
I think this should all be good to go then. Follow ups will align the other apps with the format in this PR.

PR #768, opened 2025-09-16
Document berry tree interaction script

MrKumaPants:
Also renamed some of the function chains as part of it so that the script is fully named.

Feel free to suggest better naming.


lhearachel (approved):
No feedback. Nice work!

PR #769, opened 2025-09-16
Patch generated compile_commands.json with correct DWC include-paths and explicit wchar_t typedef

lhearachel:
This does some hacky stuff just to make silly compilation errors go away when using an editor with `clangd` integration.


include/pch/global_pch.h
```diff
@@ -1,6 +1,10 @@
 #ifndef POKEPLATINUM_GLOBAL_H
 #define POKEPLATINUM_GLOBAL_H
 
+#ifdef EXPLICIT_WCHAR_TYPEDEF
+typedef unsigned short wchar_t;
+#endif
```
lhearachel, line 6: This explicit `typedef` addresses the canonical `typedef` being hidden behind an option-check that is built into the toolchain, which `clangd` cannot safely emulate.

tools/devtools/gen_compile_commands.py
```diff
@@ -155,11 +160,14 @@ def find_option(opt: str):
...
             f"-I{cwextrasdir}",
             f"-I{cwlibcarmdir}",
+            f"-I{cwextrasarmdir}",
             f"-I{homedir}/subprojects/NitroSDK-4.2.30001/include",
             f"-I{builddir}/subprojects/NitroSDK-4.2.30001/gen",
             f"-I{homedir}/subprojects/NitroSystem-071126.1/include",
             f"-I{homedir}/subprojects/NitroWiFi-2.1.30003/include",
             f"-I{homedir}/subprojects/NitroDWC-2.2.30008/include",
+            f"-I{homedir}/subprojects/NitroDWC-2.2.30008/include/gs",
+            f"-I{homedir}/subprojects/NitroDWC-2.2.30008/include/base",
```
lhearachel, line 170: These directories are present in the DWC subproject's build configuration, but were not present here. This fixes the `nonport.h` compilation error encountered by `clangd`.

src/main.c
```diff
@@ -1,16 +1,11 @@
 #include "main.h"
 
 #include <dwc.h>
-#include <nitro.h>
-#include <string.h>
```
lhearachel, line 5: Here I am just cleaning up headers that are not used in this file, since this is what I used for my testing.

scbroede (approved):

PR #770, opened 2025-09-16
Document berry patch effects

MrKumaPants:


src/overlay005/berry_patch_effects.c
```diff
@@ -0,0 +1,370 @@
...
+    sizeof(BerryPatchSparkleEffect),
+    BerryPatchSparkleEffect_Init,
+    BerryPatchSparkleEffect_Free,
+    BerryPatchSparkleEffect_Update,
+    BerryPatchSparkleEffect_Render
+};
+
+// Resource IDs for different moisture levels of berry patches
+// These correspond to different visual states: dry, moist, wet
+static const u32 BerryPatchMoistureResourceIDs[3] = {
```
lhearachel: **polish**: Denote static data with a prefixed `s`.

fuddlesworth: Prefixed all the static structs with s

src/overlay005/berry_patch_effects.c
```diff
@@ -0,0 +1,370 @@
...
+
+    do {
+        sub_0207395C(&manager->resources[i]);
+        i++;
+    } while (i < 3);
+}
+
+void BerryPatchGraphics_NewMoistureEffect(MapObject *mapObject)
+{
+    int priority, effectPriority;
```
scbroede: merge declaration and assignment whenever possible (same goes for the rest of the file)

src/overlay005/berry_patch_effects.c
```diff
@@ -0,0 +1,370 @@
...
+// These correspond to different visual states: dry, moist, wet
+static const u32 sBerryPatchMoistureResourceIDs[3] = {
+    77, // Dry berry patch moisture indicator
+    76, // Moist berry patch moisture indicator
+    75 // Wet berry patch moisture indicator
+};
+
+// Sparkle effect configuration data for berry patches
+// Each entry defines: { model_variant, animation_frame_count, effect_type }
+static const UnkStruct_020217F4 sBerryPatchSparkleEffectData[] = {
```
scbroede, line 361: it seems like you have enough information to document this struct?

src/overlay005/berry_patch_effects.c
```diff
@@ -0,0 +1,370 @@
...
+{
+    // Animation speed multipliers for different sparkle effect frames
+    fx32 animationSpeeds[5] = { 4096, 4096, 8192, 8192, 4096 };
+    BerryPatchSparkleEffect *sparkleEffect = context;
+
+    if (sparkleEffect->isAnimating == TRUE) {
+        sparkleEffect->isAnimating = FALSE;
+        sparkleEffect->animationFrame++;
+
+        if (sparkleEffect->animationFrame >= 5) { // Animation has 5 frames total
```
scbroede: this could be `(int)NELEMS(animationSpeeds)`. or as an alternative you could define a constant with a name like `NUM_SPARKLE_EFFECT_FRAMES` as 5 and use it both here and in the `animationSpeeds` definition.

either way i think there are better ways to document this than a comment

src/overlay005/berry_patch_effects.c
```diff
@@ -0,0 +1,370 @@
...
+{
+    int priority, effectPriority;
+    BerryPatchMoistureEffectContext context;
+    UnkStruct_ov5_021DF47C *renderManager;
+    VecFx32 position;
+
+    renderManager = ov5_021DF578(mapObject);
+
+    context.renderManager = renderManager;
+    context.graphicsManager = ov5_021DF55C(renderManager, 12); // Get graphics manager from overlay 12
```
scbroede: why not just rename the relevant parameter in `ov5_021DF55C` to `overlay`? it's going to happen eventually anyway once that function is documented, and once it does this comment won't really serve any purpose imo (same goes for a lot of other comments in this file)

src/overlay005/berry_patch_effects.c
```diff
@@ -0,0 +1,370 @@
...
+{
+    return;
+}
+
+static void BerryPatchMoistureEffect_Update(UnkStruct_ov101_021D5D90 *effectTask, void *effectData)
+{
+    BerryPatchMoistureEffect *berryPatchEffect = effectData;
+    MapObject *mapObject = berryPatchEffect->context.mapObject;
+
+    if (sub_02062764(mapObject, berryPatchEffect->localID, berryPatchEffect->mapID) == FALSE) {
```
scbroede: we prefer `!func()` over `func() == FALSE`

src/overlay005/berry_patch_effects.c
```diff
@@ -0,0 +1,370 @@
...
+
+static void BerryPatchMoistureEffect_Render(UnkStruct_ov101_021D5D90 *effectTask, void *effectData)
+{
+    BerryPatchMoistureEffect *berryPatchEffect = effectData;
+
+    if (berryPatchEffect->isHidden != TRUE) {
+        VecFx32 effectPosition;
+
+        sub_020715E4(effectTask, &effectPosition);
+        effectPosition.z += (FX32_ONE * 0);
```
scbroede: remove this line (it does nothing and the compiler optimizes it out)

src/overlay005/berry_patch_effects.c
```diff
@@ -0,0 +1,370 @@
...
+    BerryPatchSparkleEffect_Update,
+    BerryPatchSparkleEffect_Render
+};
+
+// Resource IDs for different moisture levels of berry patches
+// These correspond to different visual states: dry, moist, wet
+static const u32 sBerryPatchMoistureResourceIDs[3] = {
+    77, // Dry berry patch moisture indicator
+    76, // Moist berry patch moisture indicator
+    75 // Wet berry patch moisture indicator
```
scbroede: define constants with this info in the names instead of using comments imo

src/overlay005/berry_patch_effects.c
fuddlesworth: Mostly, but renaming would significantly increase this PRs scope

scbroede: fair enough

src/overlay005/berry_patch_effects.c
fuddlesworth: updated all the function declaration parameters that have been figured out and removed comments

src/overlay005/berry_patch_effects.c
fuddlesworth: combined when it was possible to maintain hash

PR #773, opened 2025-09-16
Documented mining museum scripts and messages

MrKumaPants:
Takes care of #325 unk_0204E240.c and related


src/unk_0204E240.c
```diff
@@ -3,87 +3,89 @@
...
-    for (v1 = 0, v3 = 0; v1 < 7; v1++) {
-        v3 += Bag_GetItemQuantity(SaveData_GetBag(fieldSystem->saveData), Unk_020EBFFC[v1][0], HEAP_ID_FIELD1);
+    for (i = 0, totalCount = 0; i < 7; i++) {
+        totalCount += Bag_GetItemQuantity(SaveData_GetBag(fieldSystem->saveData), sFossilItemToTypeMapping[i][0], HEAP_ID_FIELD1);
     }
 
-    *v4 = v3;
+    *destVar = totalCount;
 
     return 0;
```
scbroede: use FALSE instead of 0 (also applies to other functions in this file)

src/unk_0204E240.c
```diff
@@ -3,87 +3,89 @@
...
+            *outIndex = i;
             break;
         }
     }
 
     return 0;
 }
 
-BOOL ScrCmd_1F2(ScriptContext *param0)
+BOOL ScrCmd_Unused_1F2(ScriptContext *ctx)
```
scbroede: use `Dummy` instead of `Unused` when it doesn't actually do anything (same with below function)

src/unk_0204E240.c
```diff
@@ -3,87 +3,89 @@
...
-
-    for (v1 = 0, v3 = 0; v1 < 7; v1++) {
-        v3 += Bag_GetItemQuantity(SaveData_GetBag(fieldSystem->saveData), Unk_020EBFFC[v1][0], HEAP_ID_FIELD1);
-
-        if (v3 >= v6) {
-            *v4 = Unk_020EBFFC[v1][0];
-            *v5 = v1;
+    FieldSystem *fieldSystem = ctx->fieldSystem;
+    u8 i;
+    u16 dummy, currentCount;
```
scbroede: you should be able to delete `dummy` and still have it match

src/unk_0204E240.c
```diff
@@ -3,87 +3,89 @@
...
-    { 0x68, 0x19A },
-    { 0x69, 0x198 }
+const u16 sFossilItemToTypeMapping[7][2] = {
+    { ITEM_OLD_AMBER, FOSSIL_TYPE_OLD_AMBER }, // dataID 0x67 → 0x8E
+    { ITEM_HELIX_FOSSIL, FOSSIL_TYPE_HELIX }, // dataID 0x65 → 0x8A
+    { ITEM_DOME_FOSSIL, FOSSIL_TYPE_DOME }, // dataID 0x66 → 0x8C
+    { ITEM_ROOT_FOSSIL, FOSSIL_TYPE_ROOT }, // dataID 0x63 → 0x159
+    { ITEM_CLAW_FOSSIL, FOSSIL_TYPE_CLAW }, // dataID 0x64 → 0x15B
+    { ITEM_ARMOR_FOSSIL, FOSSIL_TYPE_ARMOR }, // dataID 0x68 → 0x19A
+    { ITEM_SKULL_FOSSIL, FOSSIL_TYPE_SKULL } // dataID 0x69 → 0x198
```
scbroede: why the fossil type enum? wouldn't it make more sense to just do `SPECIES_AERODACTYL`, `SPECIES_OMANYTE`, etc.?

on a similar note, all uses of "type" in this context should be replaced with "species"

scbroede:
also, given that you've now documented all the functions in `unk_0204E240.c`, it's probably ready for a proper name now (something along the lines of `scrcmd_fossil`, maybe)

src/unk_0204E240.c
fuddlesworth: Thought it would cause a hash failure, but that does make sense since compiler should optimize out unused variables.

src/unk_0204E240.c
fuddlesworth: Thank you. Changed to species. That definitely makes more sense..

src/scrcmd_fossil.c
```diff
@@ -0,0 +1,91 @@
...
+
+const u16 sFossilItemToSpeciesMapping[7][2] = {
+    { ITEM_OLD_AMBER, SPECIES_AERODACTYL },
+    { ITEM_HELIX_FOSSIL, SPECIES_OMANYTE },
+    { ITEM_DOME_FOSSIL, SPECIES_KABUTO },
+    { ITEM_ROOT_FOSSIL, SPECIES_LILEEP },
+    { ITEM_CLAW_FOSSIL, SPECIES_ANORITH },
+    { ITEM_ARMOR_FOSSIL, SPECIES_SHIELDON },
+    { ITEM_SKULL_FOSSIL, SPECIES_CRANIDOS }
+};
```
lhearachel: **polish**: Prefer to map such tables to a `struct`, if it still matches:


```suggestion
static const struct {
    u16 item;
    u16 species;
} sFossilItemToSpeciesMapping[7] = {
    { .item = ITEM_OLD_AMBER,    .species = SPECIES_AERODACTYL },
    { .item = ITEM_HELIX_FOSSIL, .species = SPECIES_OMANYTE },
    { .item = ITEM_DOME_FOSSIL,  .species = SPECIES_KABUTO },
    { .item = ITEM_ROOT_FOSSIL,  .species = SPECIES_LILEEP },
    { .item = ITEM_CLAW_FOSSIL,  .species = SPECIES_ANORITH },
    { .item = ITEM_ARMOR_FOSSIL, .species = SPECIES_SHIELDON },
    { .item = ITEM_SKULL_FOSSIL, .species = SPECIES_CRANIDOS },
};
```

fuddlesworth: Interesting that works.

fuddlesworth:
Reverted script defines

PR #774, opened 2025-09-17
nitrogfx: rotate flag fixes

scbroede:


tools/nitrogfx/gfx.c
```diff
@@ -1252,26 +1252,19 @@ void WriteNtrImage(char *path, int numTiles, int bitDepth, int colsPerChunk, int
...
-        }
-        char filename[sizeof(path) + 30];
-        snprintf(filename, sizeof(filename), "%s%s%d%s", stem, "_", (i + 1) * 90, "deg.NCGR");
+        char *stem = calloc(length, sizeof(char));
+        strcpy(stem, path);
+        stem[length - 5] = 0;
+
+        char *filename = calloc(length + 10, sizeof(char));
+        snprintf(filename, length + 10, "%s%s%d%s", stem, "_", (i + 1) * 90, "deg.NCGR");
+        free(stem);
```
lhearachel: **suggestion**: No need to double-alloc here; you can request a variable length from `path` in `snprintf`.


```suggestion
        char *filename = calloc(length + 10, sizeof(char));
        snprintf(filename, length + 10, "%.*s_%ddeg.NCGR", length - 5, stem, (i + 1) * 90);
```

PR #775, opened 2025-09-17
documented battle regulation code

fuddlesworth:


src/battle_regulation_validation.c
```diff
@@ -0,0 +1,266 @@
...
+{
+    Pokemon *pokemon;
+    int ruleValue, selectedCount = 0, i, j, totalLevel = 0;
+    u16 species[6], heldItems[6];
+
+    if (regulation == NULL) {
+        return 0;
+    }
+
+    for (i = 0; i < 6; i++) {
```
fuddlesworth: Is there a constant somewhere for max party size?

scbroede: yes. `MAX_PARTY_SIZE` in `constants/pokemon.h`

include/struct_defs/battle_regulation_data.h
```diff
@@ -0,0 +1,10 @@
+#ifndef POKEPLATINUM_STRUCT_BATTLE_REGULATION_DATA_H
+#define POKEPLATINUM_STRUCT_BATTLE_REGULATION_DATA_H
+
+#include "struct_defs/battle_regulation.h"
+
+struct BattleRegulationData {
+    BattleRegulation regulations[1];
+};
```
lhearachel, line 8: **suggestion**: This seems odd. Can we decompose this to:

```c
typedef BattleRegulation* BattleRegulationPtr;
```

Or, better yet, get rid of it entirely and replace existing instances with `BattleRegulation*`?

include/struct_defs/battle_regulation.h
```diff
@@ -0,0 +1,18 @@
...
+    u8 teamSize; // Required team size
+    u8 maxLevel; // Maximum level per Pokémon
+    s8 heightRestriction; // Height restriction value (positive/negative)
+    s8 weightRestriction; // Weight restriction value (positive/negative)
+    u8 speciesRestriction : 1; // Enable species restrictions
+    u8 itemRestriction : 1; // Enable item restrictions
+    u8 speciesUniqueness : 1; // Require unique species (no duplicates)
+    u8 itemUniqueness : 1; // Require unique items (no duplicates)
+    u8 moveRestrictions : 1; // Move restrictions flag for regulated battles
+} BattleRegulation;
```
lhearachel: **polish**: Move this struct to an owning header.

---

**suggestion**: I personally prefer longer names for fields rather than comments describing what they do.

src/battle_regulation.c
```diff
@@ -0,0 +1,116 @@
...
+}
+
+void RegulationBattles_Init(BattleRegulationData *regulationData)
+{
+    memset(regulationData, 0, sizeof(BattleRegulationData));
+}
+
+void BattleRegulation_GetName(const BattleRegulation *regulation, Strbuf *strbuf)
+{
+    Strbuf_CopyNumChars(strbuf, regulation->name, (11 + 1));
```
lhearachel: **suggestion**: Seems that 11 characters is the true max length of a name, and the last is just a null-terminator.

src/battle_regulation_data.c
```diff
@@ -0,0 +1,143 @@
...
+#include "struct_defs/battle_regulation_data.h"
+
+#include "battle_regulation.h"
+#include "message.h"
+#include "savedata.h"
+#include "strbuf.h"
+
+BattleRegulation BattleRegulation_PredefinedRules[] = {
+    {
+        .name = { 0xffff },
```
lhearachel: **polish**: `0xFFFF` here is `CHAR_EOS` from `include/constants/charcode.h`

src/battle_regulation_data.c
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    for (i = 0; i < NELEMS(BattleRegulation_PredefinedRules); i++) {
+        if (memcmp(regulation, &BattleRegulation_PredefinedRules[i], sizeof(BattleRegulation)) == 0) {
+            return i;
+        }
+    }
+
+    return 5;
+}
```
lhearachel: **question**: Does moving `int i` into the `for` loop declaration not match?

include/struct_defs/battle_regulation_data.h
fuddlesworth: Breaks hash sadly :(

include/struct_defs/battle_regulation.h
fuddlesworth: I'm confused. It's already in it's own header. This header is included in lots of files.

src/battle_regulation.c
fuddlesworth: added define

include/struct_defs/battle_regulation.h
lhearachel: Sorry, I should clarify: we have an aim to leave as few files in `include/struct_defs` as possible, so we tend to merge them into other headers so that they live alongside their primary API. The usual exceptions are if there is a circular-reference that cannot be untangled or if there is no such primary API.

In this particular instance, my thought is that adding `battle_regulation.h` should include:

1. The definition + `typedef` of `struct BattleRegulation`,
2. The definition of `enum BattleRegulationRule`,
3. The declarations of `extern` functions that operate on `BattleRegulation`.

I'm also seeing some opportunity here to merge `battle_regulation` and `battle_regulation_data` into the same translation unit, since they are located next to each other in the compiled ROM, per this change in `platinum.us/main.lsf`:

```diff
-	Object main.nef.p/src_unk_0202602C.c.o
-	Object main.nef.p/src_unk_02026150.c.o
+	Object main.nef.p/src_battle_regulation.c.o
+	Object main.nef.p/src_battle_regulation_data.c.o
```

Such a merge would simplify things a bit more. Does the build produce a hash-match if you concatenate these files into one?

src/battle_regulation.c
```diff
@@ -32,12 +32,12 @@ void RegulationBattles_Init(BattleRegulationData *regulationData)
 
 void BattleRegulation_GetName(const BattleRegulation *regulation, Strbuf *strbuf)
 {
-    Strbuf_CopyNumChars(strbuf, regulation->name, (11 + 1));
+    Strbuf_CopyNumChars(strbuf, regulation->name, (BATTLE_REGULATION_NAME_LENGTH));
```
lhearachel, line 35: **polish**: Remove unnecessary parens around these constants.

PR #777, opened 2025-09-18
Document player avatar movement

MrKumaPants:


asm/macros/movement.inc
```diff
@@ -525,248 +525,248 @@
     .short \length
     .endm
 
-    .macro MoveAction_105 length=1
-    .short MOVEMENT_ACTION_105
+    .macro MoveAction_DistortionWestWallWalkNorth length=1
+    .short MOVEMENT_ACTION_DISTORTION_WEST_WALL_WALK_NORTH
```
lhearachel: **question**: Do we want to follow the same rules as with `ScrCmd_`, where we remove the prefix in assembler macros and always start with verbs?

include/constants/player_avatar.h
```diff
@@ -36,4 +36,38 @@ enum AvatarDistortionState {
...
+enum ElevationChange {
+    ELEVATION_CHANGE_NONE = 0,
+    ELEVATION_CHANGE_UP = 1,
+    ELEVATION_CHANGE_DOWN = 2,
+};
+
+enum CyclingGear {
+    CYCLING_GEAR_LOW = 0,
+    CYCLING_GEAR_HIGH = 1,
+};
```
lhearachel: **polish**: Remove manual increments for `enum` values.

include/struct_defs/struct_020EDF0C.h
```diff
@@ -1,17 +1,17 @@
...
-    UnkFuncPtr_020EDF0C_1 unk_08;
-    UnkFuncPtr_020EDF0C_2 unk_0C;
-    UnkFuncPtr_020EDF0C_3 unk_10;
-} UnkStruct_020EDF0C;
+    int movementType;
+    UnkFuncPtr_020EDF0C initFunc;
+    UnkFuncPtr_020EDF0C_1 updateFunc;
+    UnkFuncPtr_020EDF0C_2 completeFunc;
+    UnkFuncPtr_020EDF0C_3 cleanupFunc;
+} MovementActionDescriptor;
```
lhearachel, line 15: **polish**: Tag this struct and find an owning header for it. Probably `player_avatar_movement.h`?

---

**polish**: Let's name these function pointer typedefs.

include/unk_020EDBAC.h
```diff
@@ -1,3 +1,15 @@
...
+ * @file unk_020EDBAC.h
+ * @brief Movement Action System - Header definitions for map object movement
+ *
+ * This header defines the core movement action system used by all map objects
+ * in Pokemon Platinum. It provides function tables and lookup arrays for
+ * converting movement commands into specific action implementations.
+ *
+ * @author Pokemon Platinum Development Team
+ * @date 2008
+ */
```
lhearachel: **question**: Where did this come from...? We shouldn't have file-headers like this in the code, IMO.

src/unk_020EDBAC.c
```diff
@@ -12,551 +32,566 @@
...
+// Each descriptor contains an ID and 4 function pointers for different phases:
+// - movementType: Movement type/ID (0x0-0x42)
+// - initFunc: Initialization function
+// - updateFunc: Update function
+// - completeFunc: Completion function
+// - cleanupFunc: Cleanup function
+//
+// Known descriptors:
+// - sMovementActionDescriptor_BerryPatch: Berry patch growth stage management (ID 0x2F)
+// - All others: Unknown purposes, kept as Unk_ names until functionality is determined
```
lhearachel: **suggestion**: I personally feel that this is superfluous; the `struct` itself should document what the descriptors do, and it should be clear from names which ones we know and which ones we don't.

src/unk_020EDBAC.c
```diff
@@ -12,551 +32,566 @@
...
 };
 
-static const UnkStruct_020EDF0C Unk_020EDE30 = {
+static const MovementActionDescriptor sMovementActionDescriptor_BerryPatch = {
     0x2f,
     BerryPatchGraphics_NewData,
     BerryPatchGraphics_UpdateGrowthStage,
     BerryPatchGraphics_NoOp,
     sub_020633EC
 };
```
lhearachel, line 425: **question**: I'm curious what about this is Movement-related now, because these seem to just update some graphics...?

lhearachel (commented):
I haven't gotten into the changes to `src/player_avatar_movement.c` yet, but here are some preliminary thoughts on the remainder.

asm/macros/movement.inc
fuddlesworth: I would assume so. Fixed.

src/unk_020EDBAC.c
fuddlesworth: I think i went on a weird path and was hitting movement stuff. Let me rename.

include/unk_020EDBAC.h
fuddlesworth: Removed. Was playing with some tools and missed this.

src/unk_020EDBAC.c
fuddlesworth: I see why. `sub_0206320C(MapObject_GetMovementType(mapObj))` there's this in map_object.c and this function returns `gMovementActionDescriptors[param0];` (which is what i renamed the global to).

Something isn't lining up. Suggestions? Or should i rename it back to Unk?

lhearachel: Oooh, I had a thought. Is this for the "sparkles" that play after you water a berry?

fuddlesworth: Indirectly? Looking into it further. This structure 1) Initializes the graphics manager 2) Calls update to update the data needed for the graphics render functions.

The UnkStruct_ov5_021FB0F0 seems to be the struct that uses the data from 1) and 2) in order to actually render. This render function does actually show the sparkle effects based off the growth stage returned by 1) and 2)

lhearachel: Ah! Right, I have a vague recollection that berry trees repurpose their "movement type" field, since they obviously cannot actually *move* like other NPCs. I will follow-up on this later today after I have a chance to research further.

PR #779, opened 2025-09-19
Document bottom screen

MrKumaPants:


include/overlay005/poketch_system.h
lhearachel, line 1: **thought**: Is this name potentially confusing with `applications/poketch/poketch_system.h`? It seems like this header moreso is a bottom-screen manager / messenger of sorts (but I hesitate to call it either of those given the presence of `bottom_screen_manager.h` above). Maybe just `include/overlay005/bottom_screen.h`?

include/overlay056/ov56_022561C0.h
```diff
@@ -4,7 +4,7 @@
 #include "field/field_system_decl.h"
 #include "overlay056/struct_ov56_02256468_decl.h"
 
-UnkStruct_ov56_02256468 *ov56_02256410(FieldSystem *fieldSystem);
-void ov56_02256468(UnkStruct_ov56_02256468 *param0);
+UnkStruct_ov56_02256468 *ColosseumSystem_Init(FieldSystem *fieldSystem);
```
lhearachel, line 7: **suggestion**: Name this `struct` if we know what it does.

include/struct_defs/map_load_mode.h
```diff
@@ -1,8 +1,10 @@
 #ifndef POKEPLATINUM_STRUCT_MAP_LOAD_MODE_H
 #define POKEPLATINUM_STRUCT_MAP_LOAD_MODE_H
 
+#include "overlay005/bottom_screen_manager.h"
+
 typedef struct MapLoadMode {
-    u32 fieldBottomScreen : 4;
+    u32 bottomScreenMode : BOTTOM_SCREEN_MODE_POKETCH_UNAVAILABLE;
```
lhearachel, line 7: **suggestion**: I think that this is an unintentional replacement.

include/overlay056/ov56_022561C0.h
fuddlesworth: Not sure exactly. It's used a lot.

include/struct_defs/map_load_mode.h
fuddlesworth: This was intentional. I had traced where it's used and it does look related.

lhearachel: I'm not sure that I agree; this is for defining how many bits of a single `u32` are used for this field. If I changed the `enum` to include an additional bottom-screen mode before `BOTTOM_SCREEN_MODE_POKETCH_UNAVAILABLE`, then that would destroy the alignment of this `struct`, as this field would shift from 4 bits to 5.

include/overlay056/ov56_022561C0.h
lhearachel: Fair; there might be a justification to leaving this file undocumented, then. I personally don't mind, if you're certain that this is related to the Union Room Colosseum.

PR #783, opened 2025-09-22
Switch to a relative path for the text bank headers directory

Gudf:
This fixes build on systems where the absolute path to the repository contains spaces or other characters that are not allowed in C identifiers.

Invalid characters caused an issue because msgenc didn't replace them when creating the header guard constant, creating invalid preprocessor directives. In addition to that, the path wasn't properly enclosed in quotes when passed to `mkdir`, which means the directory wasn't created.


lhearachel (approved):
Fixes issue: #379

PR #784, opened 2025-09-22
Generate event object ids

CharlesFolz4:
Generate event objects from newly added "id" field in events files
Rename "local_id" field to "clone_id" as discussed in discord
Document Twinleaf Town event object ids using newly generated defines, to give an example of how this will look in a script.

Default ids have been generated according to event file name, object graphics id, and json array index, in snake case.


res/field/events/events_acuity_cavern.json
```diff
@@ -13,7 +13,8 @@
             "movement_range_z": 0,
             "x": 14,
             "z": 10,
-            "y": 0
+            "y": 0,
+            "id": "acuity_cavern_uxie_0"
```
lhearachel: **suggestion**: I would move these values to the top of each object, if that is possible, as it allows for a more naturally-ordered reading of each object-event. I can help with cooking up a Python script to reorder them. 😁

---

**suggestion**: I think that these should be in `ALL_CAPS_SNAKE_CASE` to reflect other constants; what do you think?

The primary reason that we don't do that for generated NAIX files is because they are derived from file-system names, and the POSIX standard mandates that filenames are case-sensitive; this isn't the case on Windows or macOS (at least by default).

lhearachel:
I forgot to mention: thanks for doing this! This will be a big step forward for the field-scripting section of the repo. 🙏

PR #785, opened 2025-09-24
Document the town map overlay

Gudf:
A few things to note:
- The town map relies on the main map matrix to determine which map is currently hovered, but includes a number of maps that don't appear on the main matrix, namely the Mt Coronet and the Battle Frontier Gate. Instead, these use a coordinates to header array in `LoadMapName` (`graphics.c`) that's checked against the cursor's coordinates.
- Most fly locations span an entire map header, but some of them (the Pal Park, the Victory Road and the Pokemon League) don't, instead spanning a single map within a larger map header block. These are once again checked against the cursor's coordinates in `TownMap_GetFlyLocationAtPos`.
- For hidden locations, such as the the Newmoon Island or the Spring Path, the game uses a tilemap separate from the main screens' tilemaps and copies some tiles to show the hidden locations once they're unlocked. This is defined and handled in `LoadMapGraphics` and `ShowHiddenLocation`.
- The maps in the Distorsion World have their origin offset from (0,0,0). As part of figuring out where the player icon should be shown, the game substracts the offsets from the player's coordinates to have in-bounds coordinates when retrieving the map header ID from the player's coordinates.

There is some more stuff, but that'll be better suited for a followup PR unpacking `data/tmap_block.dat`, `data/tmap_flags.dat` and moving `sFlyLocations` to use datagen instead of being in-code.

### DIFF
--- a/include/applications/town_map/defs.h
+++ b/include/applications/town_map/defs.h
@@ -102,7 +102,7 @@ typedef struct TownMapGraphicsManager {
     int unused_0C;
     u8 showingZoomedInMap;
     u8 queuedMovement;
-    u8 zommedInMapMvtDone;
+    u8 zoomedInMapMvtDone;
     u8 zoomedMapMvtStage;
     u8 zoomedInMapTransitionStage;
     u8 unk_15_0 : 4;
--- a/src/applications/town_map/graphics.c
+++ b/src/applications/town_map/graphics.c
@@ -258,10 +258,10 @@ BOOL TownMap_UpdateBottomScreen(TownMapAppData *appData)
 {
     TownMapGraphicsManager *graphicsMan = appData->graphicsMan;
 
-    if (graphicsMan->zommedInMapMvtDone & TRUE) {
+    if (graphicsMan->zoomedInMapMvtDone & TRUE) {
         if (graphicsMan->unk_15_0 >= 2) {
             UpdateBottomScreenText(appData);
-            graphicsMan->zommedInMapMvtDone = 0x0;
+            graphicsMan->zoomedInMapMvtDone = 0x0;
             graphicsMan->zoomedInMapMvtFrame = 0;
             graphicsMan->unk_15_0 = 0;
         }
@@ -424,13 +424,13 @@ static void HandleInput(TownMapAppData *appData, int heldKeys)
     graphicsMan->zoomedMapMvtStage = 3;
     graphicsMan->queuedMovement |= 0x80;
 
-    if (graphicsMan->zommedInMapMvtDone != 0) {
+    if (graphicsMan->zoomedInMapMvtDone != 0) {
         graphicsMan->zoomedInMapMvtFrame++;
     } else {
         graphicsMan->zoomedInMapMvtFrame = 0;
     }
 
-    graphicsMan->zommedInMapMvtDone = 0;
+    graphicsMan->zoomedInMapMvtDone = 0;
     graphicsMan->unk_15_0 = 1;
 
     return;
@@ -594,7 +594,7 @@ static void DoZoomedMapMvt(TownMapAppData *appData)
         return;
     }
 
-    graphicsMan->zommedInMapMvtDone = TRUE;
+    graphicsMan->zoomedInMapMvtDone = TRUE;
     graphicsMan->queuedMovement = 0;
 
     if (graphicsMan->prevLocationMapHeader != graphicsMan->hoveredLocationMapHeader) {


include/applications/town_map/app_data_decl.h
```diff
@@ -0,0 +1,6 @@
+#ifndef POKEPLATINUM_APPLICATIONS_TOWN_MAP_APP_DATA_DECL_H
+#define POKEPLATINUM_APPLICATIONS_TOWN_MAP_APP_DATA_DECL_H
+
+typedef struct TownMapAppData TownMapAppData;
```
lhearachel: **question**: Can this not be merged into `main.h`?

include/applications/town_map/main.h
lhearachel, line 1: **suggestion**: This seems more like `defs.h` to me, and `appliciation.h` would be `main.h`, similar to the structure for the party menu. Thoughts?

include/struct_defs/union_0206B878.h
```diff
@@ -1,14 +1,11 @@
...
-typedef union {
-    struct {
-        u8 val1_0 : 2;
-        u8 val1_2 : 2;
-        u8 val1_4 : 2;
-        u8 val1_6 : 2;
-    };
-    u8 val2;
-} UnkUnion_0206B878;
+typedef struct TownMapBlockDescriptionCheckResults {
```
lhearachel: **question**: Does this fit anywhere in the town map application folder?

src/applications/town_map/context.c
```diff
@@ -0,0 +1,213 @@
...
+            checksResult->landmarkDescHasCheck = TRUE;
+            break;
+        case TOWN_MAP_DESC_FLAG_GENERAL:
+            checksResult->landmarkCheckResult = FieldSystem_CheckFlag(fieldSystem, descFlags->landmarkDescriptionFlag);
+            checksResult->landmarkDescHasCheck = TRUE;
+            break;
+        }
+    }
+
+    (void)FS_CloseFile(&flagsFile);
```
lhearachel: **polish**: Remove the `(void)` cast.

src/applications/town_map/graphics.c
```diff
@@ -0,0 +1,1134 @@
...
+#include "system.h"
+#include "text.h"
+#include "touch_screen.h"
+
+#define TOWN_MAP_RIGHT (1 << 0)
+#define TOWN_MAP_LEFT  (1 << 1)
+#define TOWN_MAP_DOWN  (1 << 2)
+#define TOWN_MAP_UP    (1 << 3)
+
+typedef struct {
```
lhearachel: **polish**: Tag structs.

src/applications/town_map/graphics.c
```diff
@@ -0,0 +1,1134 @@
...
+    Window_CopyToVRAM(window);
+    Strbuf_Free(string);
+}
+
+static void MakeAppWindows(TownMapAppData *appData)
+{
+    TownMapGraphicsManager *graphicsMan = appData->graphicsMan;
+
+    Window_Add(appData->bgConfig, &graphicsMan->windows[TOWN_MAP_WINDOW_LOCATION_NAME], BG_LAYER_MAIN_1, 3, 21, 29, 3, PLTT_15, 1023 - 29 * 3);
+    Window_Add(appData->bgConfig, &graphicsMan->windows[TOWN_MAP_WINDOW_SIGNPOST_LOCATION_NAME_1], BG_LAYER_SUB_0, 9, 3, 21, 4, PLTT_14, (1023 - (21 * 4)));
```
lhearachel: **thought**: These math constants seem like they should be computable from base-tile definitions.

include/struct_defs/union_0206B878.h
Gudf: ~~Yeah, I must've missed it, oops~~ Actually, I just forgot to remove the file, it's already in `context.h`.

include/applications/town_map/app_data_decl.h
Gudf: That would end up causing an include loop with `graphics.h`, which requires `TownMapAppData` for its functions, but contains the definition for `TownMapGraphicsManager` which is needed for `TownMapAppData`.
Though of course that could be solved by having a single header for all non-function defs.

include/applications/town_map/defs.h
```diff
@@ -0,0 +1,180 @@
...
+} TownMapScreenScroll;
+
+typedef struct TownMapGraphicsManager {
+    int graphicsLoadStage;
+    int taskState;
+    int taskFrameCount;
+    int unused_0C;
+    u8 showingZoomedInMap;
+    u8 queuedMovement;
+    u8 zommedInMapMvtDone;
```
scbroede:
```suggestion
    u8 zoomedInMapMvtDone;
```

scbroede (approved):

PR #786, opened 2025-09-25
Unpack Trainer Graphics

h2o-DS:
unpacks narcs
`/res/prebuilt/poketool/trgra/trfgra.narc`
`/res/prebuilt/poketool/trgra/trbgra.narc`
`/res/prebuilt/graphic/field_encounteffect.narc`
to
`/res/trainers/classes`

This required two `nitrogfx` changes:
1. `ucat` and `uaat` chunk compatibility for `NCER` and `NANR` files, respectively
2. Passing `-preservepath` as the cell path when converting `png`->`NCGR` will look for a file in the same directory named `@basename@_cell.json`. This was required due to the meson preserve path generator argument not passing the sub-directory. This does also make the argument generally simpler and no longer requires `ncgr_cell_gen` to be its own generator, but this PR is already quite large and I though that cleanup would be best done on its own later.

All images are output as 8bpp and store the respective palette.

The trainer classes are used when adding the files to meson instead of each class having its own `meson.build`. This also adds optional sprites, such as back-sprites, based solely upon their existence in the class' directory.

### DIFF
--- a/tools/nitrogfx/options.h
+++ b/tools/nitrogfx/options.h
@@ -125,7 +125,7 @@ struct JsonToCellOptions {
     struct CellVramTransferData **transferData;
     char **labels;
     int labelCount;
-    int *ucatCellAttribtes;
+    int *ucatCellAttributes;
 };
 
 struct JsonToScreenOptions {
--- a/res/trainers/classes/meson.build
+++ b/res/trainers/classes/meson.build
@@ -195,7 +195,8 @@ trainer_front_NCLR = nclr_gen.process(trainer_front_png,
 )
 trainer_front_scan_NCGR = ncgr_gen.process(trainer_front_scan_png,
     extra_args: [
-        '-scanfronttoback',
+        '-encodefronttoback',
+        '-scan',
     ],
     preserve_path_from: trainer_classes_root,
 )
@@ -258,7 +259,8 @@ trainer_back_NCLR = nclr_gen.process(trainer_back_png,
 )
 trainer_back_scan_NCGR = ncgr_gen.process(trainer_back_scan_png,
     extra_args: [
-        '-scanfronttoback',
+        '-encodefronttoback',
+        '-scan',
     ],
     preserve_path_from: trainer_classes_root,
 )
--- a/tools/nitrogfx/gfx.c
+++ b/tools/nitrogfx/gfx.c
@@ -1640,10 +1640,10 @@ void ReadNtrCell_CEBK(unsigned char * restrict data, unsigned int blockOffset, u
     {
         offset = blockOffset + 0x18 + ucatOffset + 0x04 * options->cellCount;
 
-        options->ucatCellAttribtes = malloc(sizeof(uint32_t) * options->cellCount);
+        options->ucatCellAttributes = malloc(sizeof(uint32_t) * options->cellCount);
         for (int i = 0; i < options->cellCount; i++)
         {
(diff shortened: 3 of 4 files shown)


res/trainers/classes/.shared/frontier_vs.png
lhearachel, line 1: **question**: How does this meaningfully differ from `league_vs.png`?

tools/nitrogfx/main.c
```diff
@@ -550,6 +550,23 @@ void HandlePngToNtrCommand(char *inputPath, char *outputPath, int argc, char **a
...
+                options.cellFilePath[strlen(inputPath) - 2] = 'e';
+                options.cellFilePath[strlen(inputPath) - 1] = 'l';
+                options.cellFilePath[strlen(inputPath)] = 'l';
+                options.cellFilePath[strlen(inputPath) + 1] = '.';
+                options.cellFilePath[strlen(inputPath) + 2] = 'j';
+                options.cellFilePath[strlen(inputPath) + 3] = 's';
+                options.cellFilePath[strlen(inputPath) + 4] = 'o';
+                options.cellFilePath[strlen(inputPath) + 5] = 'n';
+                options.cellFilePath[strlen(inputPath) + 6] = 0;
+            }
```
lhearachel, line 568: **suggestion**:

```suggestion
            if (strcmp(options.cellFilePath, "-preservepath") == 0)
            {
                const char *suffix = "_cell.json";
                size_t inputStemSize = strlen(inputPath) - sizeof(".png");
                options.cellFilePath = calloc(inputStemSize + sizeof(suffix) + 1, sizeof(char));
                sprintf(options.cellFilePath, "%.*s%s" , (int)inputStemSize, inputPath, suffix);
            }
```

res/trainers/classes/.shared/frontier_vs.png
h2o-DS: Palette and use-location. I need to confirm, but this probably could be a single image and two (one?) pal files if it is worth entangling them.

h2o-DS: I was able to combine the image and anim files, but the cell file needed to remain due to the frontier version using a different mapping type (still not entirely sure what the significance is). Luckily no pal files were created, I forgot the frontier just uses the trainer palettes.

tools/nitrogfx/main.c
h2o-DS: Thank you. However, needed to replace `sizeof` with `strlen`

scbroede:
are the scanned pngs always just the first two frames of the regular png side by side (or an empty frame if there's only one)? if so i feel like we could probably generate them from the regular png instead of expecting someone to create/edit both.

though maybe that's better as a future enhancement rather than part of this pr

tools/nitrogfx/options.h
```diff
@@ -116,13 +116,15 @@ struct JsonToCellOptions {
...
     bool vramTransferEnabled;
+    bool ucatEnabled;
     int mappingType;
     int cellCount;
     struct Cell **cells;
     int vramTransferMaxSize;
     struct CellVramTransferData **transferData;
     char **labels;
     int labelCount;
+    int *ucatCellAttribtes;
```
scbroede:
```suggestion
    int *ucatCellAttributes;
```

h2o-DS:
> are the scanned pngs always just the first two frames of the regular png side by side (or an empty frame if there's only one)? if so i feel like we could probably generate them from the regular png instead of expecting someone to create/edit both.
> 
> though maybe that's better as a future enhancement rather than part of this pr

I do think that is probably a good idea, but I don't think we currently have tooling to accomplish that (that I know of). Should likely be addressed on its own.

PR #787, opened 2025-09-25
Unpack battle anim narcs (wazaeffect)

Fexty12573:
Unpacks all NARCs previously in `res/prebuilt/wazaeffect` which includes:
```
- we.arc
- we_sub.narc
- waza_particle.narc
- ball_particle.narc
- wechar.narc
- wecell.narc
- wecellanm.narc
- wepltt.narc
```

Additionally, adds a new sub-flag to nitrogfx for `pal -> NCLR` and `png -> NCLR` conversions in the form of `-pcmp -start N` where `N` is the index the `PCMP` block should start counting at.

Also adjusts `NCER -> json -> NCER` conversions to maintain a lack of `KBEC` block padding. (No explicit options)

Here is how the unpacked NARCs were distributed:

- `we.arc`
Contains scripts for move animations. They were put into their respective subdirectory in `res/battle/moves/<name>/anim/<name>.s`.

- `we_sub.narc`
Contains battle animation subscripts such as for passive effects like burn, healing, paralysis, etc. All of these were put into `res/battle/scripts/subscripts/anims`.

- `waza_particle.narc`
Contains particle files (`.spa`) for moves as well as encounters. All were put into `res/battle/particles`.

- `ball_particle.narc`
Contains particle files for pokeballs and ball seals. All were put into `res/battle/particles/pokeballs`. Was debating on that vs. `res/battle/particles/balls` but I have the humor of a 12 year old sooo yeah.

- `wechar.narc`, `wecell.narc`, `wecellanm,narc`, `wepltt.narc`
Contains additional textures used by some moves. All were put into `res/graphics/battle/moves`. Explicit `.pal` files are only used where necessary (i.e. one NCGR used with multiple NCLRs).

The most interesting part of this PR is probably the move anim scripts. They are not yet 100% mapped out but they make use of naix entries where possible and some commands already use defines for their parameters. I will continue documenting these as I continue documenting the rest of ov12.

Also 2 new asm macro files were added:
```
- asm/macros/btlanimcmd.inc
- asm/macros/btlanimfunc.inc
```
The first one is macros for battle anim script commands. The second is specializations for the `CallFunc` script command.


Fexty12573:
> Is there a reason that these can't just be res/battle/moves/<name>/anim.s? I don't see the value of an additional subdirectory here.

Good point. I was initially unsure whether there would be more files I would be placing in there. I'll change it like you suggested.

> Would it be clearer if we propagate these up a subdirectory to make them "first-class", e.g. res/battle/scripts/common_anims or somesuch?

I don't have a preference here, either is fine with me.

> Should `waza_particle` files go to their own subfolder? e.g., `res/battle/particles/moves`

I like that idea, gonna change that.

Fexty12573:
Actually I just remembered; my reasoning for the first one was that I didn't want to call all of them `anim.s` to make fuzzy finding easier.

asm/macros/btlanimcmd.inc
```diff
@@ -0,0 +1,809 @@
...
+    // Note: This is usually used if some advanced scrolling effect should be applied to a battler.
+    // The sprite is loaded into BATTLE_BG_BASE (BG2).
+    .macro LoadPokemonSpriteIntoBg battlerRole, trackBattler
+    .long 31
+    .long \battlerRole
+    .long \trackBattler
+    .endm
+
+    // Removes a Pokemon sprite from the background layer
+    // unused: Unused parameter, set to 0
```
Gudf: **suggestion**: It looks like it's only ever set to 0, so we could probably hardcode the 0 instead of providing a useless parameter

asm/macros/btlanimcmd.inc
```diff
@@ -0,0 +1,809 @@
...
+    .if \argCount > 5
+        .long \arg5
+    .endif
+    .if \argCount > 6
+        .long \arg6
+    .endif
+    .if \argCount > 7
+        .long \arg7
+    .endif
+    .endm
```
Gudf: **suggestion**: GNU `as` supports macros with variadic arguments, so we could probably use these with helper macros to unpack the argument list to support an arbitrary amount of arguments out of the box (and have it look a little nicer):
```suggestion
    .macro CountArgs count, first, rest:vararg
    .ifnb \first
    CountArgs \count+1, \rest
    .else
    .long \count
    .endif
    .endm
    
    .macro UnpackArgs first, rest:vararg
    .ifnb \first
    .long \first
    UnpackArgs \rest
    .endif
    .endm
    
    .macro UnpackArgsWithCount args:vararg
    CountArgs \args
    UnpackArgs \args
    .endm
    
    .macro BtlAnimCmd_055 args:vararg
    .long 55
    UnpackArgsWithCount \args
    .endm
```

asm/macros/btlanimcmd.inc
```diff
@@ -0,0 +1,809 @@
...
+    .endif
+    .if \argCount > 7
+        .long \arg7
+    .endif
+    .if \argCount > 8
+        .long \arg8
+    .endif
+    .if \argCount > 9
+        .long \arg9
+    .endif
```
Gudf: Same thing here, we could use a variadic argument.

asm/macros/btlanimfunc.inc
```diff
@@ -0,0 +1,745 @@
...
+    .endif
+    .if \argCount > 7
+        .long \arg7
+    .endif
+    .if \argCount > 8
+        .long \arg8
+    .endif
+    .if \argCount > 9
+        .long \arg9
+    .endif
```
Gudf: Same thing here, we could use a variadic argument. It seems like it could be especially useful here since you could then turn the `Func_` macros into simply `CallFunc \funcID, <named argument list>` — though we would lose the information of which argument combinations are allowed, I don't know if that's a significant concern?

asm/macros/btlanimfunc.inc
```diff
@@ -0,0 +1,745 @@
...
+    .ifnb \pivotX
+        CallFunc 4, 6, \startAngle, \endAngle, \frames, \mode, \pivotX, \pivotY
+    .else
+        .ifnb \mode
+            CallFunc 4, 4, \startAngle, \endAngle, \frames, \mode
+        .else
+            CallFunc 4, 3, \startAngle, \endAngle, \frames
+        .endif
+    .endif
+    .endm
```
Gudf: **question**: For the functions that take a variable number of arguments, would it make sense to mark the first ones as required in the macro definition directly? And ideally we would then also do it for the functions with a fixed number of arguments, for the sake of consistency.

include/constants/battle/battle_anim.h
```diff
@@ -149,14 +157,53 @@
 
 #define BASE_SCALE_XY 100
 
+// For use in scripts in place of DEG_TO_IDX because said macro can't be included
+// in script files because the assembler doesn't understand the other code inside that file.
+#define BATTLE_ANIM_DEGREES(degrees) (((degrees) * 0xFFFF) / 360)
```
Gudf: **question**: Is there some particular reason not to name this macro `DEG_TO_IDX` as well? The preprocessor should accept an identical macro redefinition.

Gudf (commented):
Great work, thank you! 
I've only given the nitrogfx changes a cursory glance but they seem sound.

Two extra things:

**question**: I assume documenting the move anim scripts more thoroughly (e.g. using constants from the NAIX for CreateEmitter) is out of scope?

**polish**: the NARCs could probably benefit from being renamed? `waza_effect` and `we` aren't particularly clear if you don't know Japanese, and it would avoid having to keep that `we.arc` typo, clarifying that this is, in fact, a NARC, and not some other special archive format.

asm/macros/btlanimcmd.inc
Fexty12573: I mainly did it like this so you don't have extra unneeded data in each script when you use those commands. I guess it's only a couple bytes tho so it ain't that bad.

Fexty12573:
> **question**: I assume documenting the move anim scripts more thoroughly (e.g. using constants from the NAIX for CreateEmitter, naming each label...) is out of scope?

`CreateEmitter` doesn't take any NARC indices, the `resourceID` it takes is an index *inside* a .spa file. I realize my comment says otherwise, blame in on me spamming tab with copilot running, will fix. 😅

And about the labels, yeah I could probably do those. Most of em are just `start` or something anyway.

> polish: the NARCs could probably benefit from being renamed? waza_effect and we aren't particularly clear if you don't know Japanese, and it would avoid having to keep that we.arc typo, clarifying that this is, in fact, a NARC, and not some other special archive format.

Will do. Name suggestions? My idea would be
- `waza_particle` -> `battle_particles`
- `we` -> `anim_scripts` or `battle_anim_scripts`
- `we_sub` -> `anim_subscripts` or `battle_anim_subscripts`

Gudf:
> > **question**: I assume documenting the move anim scripts more thoroughly (e.g. using constants from the NAIX for CreateEmitter, naming each label...) is out of scope?
> 
> `CreateEmitter` doesn't take any NARC indices, the `resourceID` it takes is an index _inside_ a .spa file. I realize my comment says otherwise, blame in on me spamming tab with copilot running, will fix. 😅

Ah, lol.

> > polish: the NARCs could probably benefit from being renamed? waza_effect and we aren't particularly clear if you don't know Japanese, and it would avoid having to keep that we.arc typo, clarifying that this is, in fact, a NARC, and not some other special archive format.
> 
> Will do. Name suggestions? My idea would be
> 
>     * `waza_particle` -> `battle_particles`
> 
>     * `we` -> `anim_scripts` or `battle_anim_scripts`
> 
>     * `we_sub` -> `anim_subscripts` or `battle_anim_subscripts`

These look fine to me, I don't think the `battle_` prefix is necessary for the last two since they're already in the battle subdirectory.
If you go for `battle_particles`, you should probably also pluralize `ball_particle` -> `ball_particles` for consistency.
I'm not sure what to do with the `we(cell|cellanm|char|pltt)` ones, I think I would do 
* `wecell` -> `anim_cells`
* `wechar` -> `anim_char`
* `wepltt` -> `anim_pltt`

But then `wecellanm` would be... `anim_anim`? `anim_cell_anim`? looks a bit weird. Alternatively we could use the files formats' names, so `anim_ncgr`, `anim_nanr`, etc.?

Fexty12573:
I like the file extension approach

asm/macros/btlanimfunc.inc
Fexty12573: You want to get rid of the `Func_` macros entirely?

asm/macros/btlanimcmd.inc
Fexty12573: Not a huge fan of varargs here mainly because you lose information on how many args are allowed at most. For example there is a limit on 10 max extra args for any given script command. Not gonna die on this hill tho so if you say the explicit args bother you I can change it.

asm/macros/btlanimfunc.inc
Gudf: Oh, no, having the argument names is good, but macros such as `Func_Flail` could be simplified by not having these `if` chains is what I meant.

asm/macros/btlanimcmd.inc
Gudf: Ah, didn't check the code, in that case I'm good with having the 10 separate args. Though I think there could still be value in using the macro and simply passing it all 10 args? This would avoid the potential footgun of having to specify the number of arguments.

asm/macros/btlanimfunc.inc
```diff
@@ -0,0 +1,613 @@
...
+    // Rotates a battler sprite on the Z axis
+    // startAngle: The starting angle (0-65535)
+    // endAngle: The ending angle (0-65535)
+    // frames: The number of frames to rotate over
+    // mode: The rotation mode (optional, see ROTATE_MON_MODE_*)
+    // pivotX: The X coordinate of the pivot point (optional, default: center of sprite)
+    // pivotY: The Y coordinate of the pivot point (optional, default: center of sprite)
+    .macro Func_RotateMon startAngle:req, endAngle:req, frames:req, mode, pivotX, pivotY
+    CallFunc 4, \startAngle, \endAngle, \frames, \mode, \pivotX, \pivotY
+    .endm
```
Gudf, line 62: Would it make sense to add a check for the case where `pivotX` is provided without `pivotY`? Since that would now result in having 5 args instead of the 3, 4 or 6 that are presumably expected by the function.
```suggestion
    // Rotates a battler sprite on the Z axis
    // startAngle: The starting angle (0-65535)
    // endAngle: The ending angle (0-65535)
    // frames: The number of frames to rotate over
    // mode: The rotation mode (optional, see ROTATE_MON_MODE_*)
    // pivotX: The X coordinate of the pivot point (optional, default: center of sprite)
    // pivotY: The Y coordinate of the pivot point (optional, default: center of sprite)
    .macro Func_RotateMon startAngle:req, endAngle:req, frames:req, mode, pivotX, pivotY
    .ifnb \pivotX
        .ifb \pivotY
            .error "Can't call Func_RotateMon with pivotX but no pivotY" // Or invoke CallFunc with a 0 instead of pivotY
        .endif
    .endif
    CallFunc 4, \startAngle, \endAngle, \frames, \mode, \pivotX, \pivotY
    .endm
```

Fexty12573: I understand the concern but **personally** I feel that if someone does that they *deserve* to get hard-to-debug issues 🤬 
And if I do it here I'll have to do it for the 15 other cases as well and sometimes there are 3 or more optional args which results in the macro getting just as complicated as before the varargs macros 😥

Anyway a whole lot of yap to say I don't think it's really necessary?

Gudf: Ah, I see. I only suggested this because it seemed to me like this was the only one which wouldn't accept cutting off the argument list anywhere, but if that's not the case then it's fine. Also gave the code a look, and it seems like the worse it would do is treat pivotY as 0 anyway.
Though, that being said, it looks like it reads pivotX and pivotY if and only if the mode is 1? But that's more of a documentation thing and it would probably be significant work to go through all the functions and document the exact valid arg combinations, not to mention validation in the macros of course...
So yeah, I'm good with leaving this as-is.

PR #790, opened 2025-10-02
Update documented Poketch apps to consistent format

VicSevenT:
This is a sort of checkpoint PR where I've gone back through all the poketch apps documented up to this point and tried to make them consistent as best as possible. This largely entailed renaming things in apps that were done first to match conventions that were changed later on. A rough summary of this PR is
* Fix various bits of polish that were missed in initial PRs
* Ensure that the state in `main.c` are referred to as "states" instead of "tasks", as first done in #758 
* Ensure that button hitbox structs are declared in `Init()`, as discussed in #766 
* Make better use of constants, a result of me better knowing what's in the repo and seeing patterns across apps

I probably missed some things, and some more improvements with constants can surely be done, but I will save those for after all poketch apps are done, when I intend to repeat this review process.


src/applications/poketch/counter/graphics.c
```diff
@@ -64,7 +64,7 @@ static void SetupSprites(CounterGraphics *graphics, const CounterData *counterDa
         .hasAffineTransform = FALSE,
     };
     static const PoketchAnimation_AnimationData digitsAnimData = {
```
h2o-DS: `sDigitsAnimData`

src/applications/poketch/counter/graphics.c
```diff
@@ -73,14 +73,14 @@ static void SetupSprites(CounterGraphics *graphics, const CounterData *counterDa
     };
 
     Graphics_LoadObjectTiles(NARC_INDEX_GRAPHIC__POKETCH, 2, DS_SCREEN_SUB, 0, 0, TRUE, HEAP_ID_POKETCH_APP);
-    Graphics_LoadObjectTiles(NARC_INDEX_GRAPHIC__POKETCH, 47, DS_SCREEN_SUB, 80 * TILE_SIZE_4BPP, 0, TRUE, HEAP_ID_POKETCH_APP);
+    Graphics_LoadObjectTiles(NARC_INDEX_GRAPHIC__POKETCH, 47, DS_SCREEN_SUB, POKETCH_DIGITS_NCGR_SIZE_TILES * TILE_SIZE_4BPP, 0, TRUE, HEAP_ID_POKETCH_APP);
```
h2o-DS: `POKETCH_DIGITS_NCGR_NUM_TILES` may be more clear.

src/applications/poketch/daycare_checker/graphics.c
```diff
@@ -21,42 +20,44 @@
...
 static void Task_UpdateMosaicSize(SysTask *task, void *taskMan);
+
+static void TriggerMosaicUpdate(DaycareCheckerGraphics *graphics);
 static void LoadDaycareMonIcons(u32 offset, const DaycareStatus *daycareStatus);
 static void DrawSprites(DaycareCheckerGraphics *graphics, const DaycareStatus *daycareStatus);
 static void SetLevelSprites(PoketchAnimation_AnimatedSpriteData **digitSprites, u32 level);
 static void SetGenderSprites(PoketchAnimation_AnimatedSpriteData *genderSymbol, enum Gender gender);
 
-BOOL DaycareCheckerGraphics_New(DaycareCheckerGraphics **graphics, const DaycareStatus *daycareStatus, BgConfig *bgConfig)
+BOOL PoketchDaycareCheckerGraphics_New(DaycareCheckerGraphics **dest, const DaycareStatus *daycareStatus, BgConfig *bgConfig)
```
h2o-DS, line 36: Adding `Poketch` probably isn't necessary here.

include/applications/poketch/move_tester/graphics.h
```diff
@@ -35,18 +36,18 @@ typedef struct PoketchMoveTesterGraphics {
     Strbuf *strBuf;
 } PoketchMoveTesterGraphics;
 
-enum MoveTesterGraphicsTasks {
-    TASK_DRAW_APP_SCREEN = 0,
-    TASK_FREE_WINDOWS_AND_BG,
-    TASK_BUTTON_PRESSED,
-    TASK_BUTTON_RELEASED,
-    TASK_UPDATE_GRAPHICS
+enum MoveTesterGraphicsTask {
```
h2o-DS, line 39: Should this be `State` instead of `Task`?

src/applications/poketch/friendship_checker/graphics.c
```diff
@@ -203,7 +203,7 @@ static void Task_InitGraphics(SysTask *task, void *taskMan)
     EndTask(taskMan);
 }
 
-static void SetupSprites(FriendshipCheckerGraphics *graphics, const FriendshipCheckerState *appState)
+static void SetupSprites(FriendshipCheckerGraphics *graphics, const FriendshipCheckerData *appState)
```
h2o-DS: Update the variable names.

h2o-DS (changes requested):
Thanks for the polish. I have some minor questions, but it looks good.

include/applications/poketch/move_tester/graphics.h
VicSevenT: No, the state convention is for the states that called from within `Task_Main()` in `main.c`. The graphics tasks are indeed actual tasks and are kept named as such.

src/applications/poketch/daycare_checker/graphics.c
VicSevenT: While I agree that the prefix isn't strictly necessary here, the goal with this is to make the names consistent with all the other poketch apps, which also use the `Poketch` prefix.

PR #791, opened 2025-10-02
document some object events

scbroede:


scbroede:
updated. sorry, thought i had already done this a week ago for some reason

PR #793, opened 2025-10-04
Document Gym scripts

CharlesFolz4:
Document Gym scripts
Add macro for GoToIfBadgeAcquired


asm/macros/scrcmd.inc
```diff
@@ -2101,7 +2106,7 @@
     .byte \floorID
     .endm
 
-    .macro ScrCmd_176 arg0
+    .macro SunnyshoreGymButton arg0
```
lhearachel: **polish**: `PressSunyshoreGymButton` (note that it's Sunyshore, not Sunnyshore).

asm/macros/scrcmd.inc
```diff
@@ -4126,7 +4131,7 @@
     .short 713
     .endm
 
-    .macro ScrCmd_2CA
+    .macro EternaGymAdvanceClockState
```
lhearachel: **suggestion**: `AdvanceEternaGymClock`

res/field/events/events_canalave_city.json
```diff
@@ -186,11 +186,11 @@
             "y": 1
         },
         {
-            "id": "CANALAVE_CITY_BARRY_11",
+            "id": "CANALAVE_CITY_RIVAL_BRIDGE",
             "graphics_id": "OBJ_EVENT_GFX_BARRY",
             "movement_type": "MOVEMENT_TYPE_LOOK_NORTH",
             "trainer_type": "TRAINER_TYPE_NONE",
-            "flag": "FLAG_UNK_0x01B2",
+            "flag": "FLAG_CANALAVE_RIVAL",
```
lhearachel: **thought**: Should flags have some verbiage attached to them? e.g., does this flag designate that you've spoken to your rival? Fought your rival? etc.

res/field/events/events_oreburgh_city.json
```diff
@@ -559,13 +559,14 @@
     ],
     "coord_events": [
         {
-            "script": 3,
+            "_comment": "this is right above the kid at the bottom of the stairs",
+            "script": 3,
```
lhearachel: **suggestion**: I don't think that this comment adds much value long-term; the goal is for events to be edited in some kind of GUI, anyways.

res/field/scripts/scripts_acuity_cavern.s
```diff
@@ -14,13 +14,13 @@ _000E:
     End
 
 AcuityCavern_CheckShouldUxieBeRemoved:
-    GoToIfSet FLAG_UNK_0x008E, AcuityCavern_RemoveUxie
+    GoToIfSet FLAG_MAP_LOCAL, AcuityCavern_RemoveUxie
```
lhearachel, line 17: **question**: Is this really local...? If so, what does "local" mean?

res/field/events/events_canalave_city.json
CharlesFolz4: I wasn't entirely sure how to document the "flag" fields, because "flag" seems to be an active-low indicator for the object's presence.  To me, that means that this field would make more sense to rename as "hidden", rather than append `_HIDDEN` to every one of these flag variables.  I'm not sure if that would have any further reaching implications though.

res/field/events/events_oreburgh_city.json
CharlesFolz4: Whoops, that was supposed to be a debug/note thing, forgot to delete it before committing.

res/field/scripts/scripts_acuity_cavern.s
CharlesFolz4: As far as I can tell, this is being used in the same way as the `VAR_MAP_LOCAL_` variables are, except it's a boolean value written to and read from like a flag, instead of being an integer like the `VAR_MAP_LOCAL_` variables, so that terminology was copied.

res/field/events/events_canalave_city.json
lhearachel: Yes, the flag being set to non-`0` means that the object should be hidden. What if we just use a verbiage of `FLAG_HIDE_X`?

CharlesFolz4: Updated, that does make sense. Although I think it may also be desirable to consider renaming this "flag" field to something that's actually descriptive. Naming a field "flag" makes about as much sense as naming a field "integer", for basically all of the same reasons.  Making that change would probably be outside of the scope for this PR though.

lhearachel (approved):

PR #794, opened 2025-10-05
Document Poketch Roulette (Overlay 41)

VicSevenT:
The drawing functionality of this app is nearly identical to the memo pad, except that it substitutes the eraser for a clear button. The roulette spinner is a rotating sprite with a start and stop button. When the stop button is pressed, the spinner waits a random number of frames between 0 and 7 before it starts slowing down in order to randomize the stop position.


src/applications/poketch/roulette/graphics.c
```diff
@@ -0,0 +1,395 @@
...
+}
+
+static void Task_UpdateDrawing(SysTask *task, void *taskMan)
+{
+    s32 x, y, tileIdx;
+    RouletteGraphics *graphics = PoketchTask_GetTaskData(taskMan);
+    x = graphics->rouletteData->x * 2;
+    y = graphics->rouletteData->y * 2;
+    // Dividing x and y by 8 (TILE_WIDTH/HEIGHT_PIXELS) here
+    tileIdx = ((y >> 3) * WINDOW_WIDTH_TILES) + (x >> 3);
```
VicSevenT: I would have liked to changed this line to
```
tileIdx = ((y / TILE_HEIGHT_PIXELS) * WINDOW_WIDTH_TILES) + (x / TILE_WIDTH_PIXELS)
```
but I couldn't get it to match, so I left the comment to explain the significance.

src/applications/poketch/roulette/graphics.c
```diff
@@ -0,0 +1,395 @@
...
+        .screenBase = GX_BG_SCRBASE_0x7800,
+        .charBase = GX_BG_CHARBASE_0x00000,
+        .bgExtPltt = GX_BG_EXTPLTT_01,
+        .priority = 3,
+        .areaOver = 0,
+        .mosaic = FALSE,
+    };
+    GXSDispCnt dispCnt;
+    RouletteGraphics *graphics;
+    u32 bgTileCount;
```
scbroede: merge declaration and assignment if possible

src/applications/poketch/roulette/graphics.c
```diff
@@ -0,0 +1,395 @@
...
+static void Task_Unused(SysTask *task, void *taskMan)
+{
+    RouletteGraphics *graphics = PoketchTask_GetTaskData(taskMan);
+    Window_LoadTiles(&graphics->window);
+    EndTask(taskMan);
+}
+
+static void Task_UpdateDrawing(SysTask *task, void *taskMan)
+{
+    s32 x, y, tileIdx;
```
scbroede: merge declaration and assignment if possible

src/applications/poketch/roulette/main.c
```diff
@@ -0,0 +1,428 @@
...
+                appData->subState = 3;
+                break;
+            }
+
+            appData->buttonState = BUTTON_MANAGER_STATE_NULL;
+            break;
+        }
+
+        if (appData->currentlyDrawing) {
+            u32 prevX, prevY;
```
scbroede: merge declaration and assignment if possible

src/applications/poketch/roulette/main.c
```diff
@@ -0,0 +1,428 @@
...
+            return TRUE;
+        }
+    }
+    return FALSE;
+}
+
+static void UpdatePixelsOnPath(PoketchRoulette *appData, u32 x, u32 y, u32 endX, u32 endY)
+{
+    u32 deltaX, deltaY, majorAxisStep, minorAxisRounded;
+    fx32 minorAxis, minorAxisStep;
```
scbroede: merge declaration and assignment if possible

VicSevenT: I was able to merge all but one of them.

src/applications/poketch/roulette/main.c
```diff
@@ -0,0 +1,425 @@
...
+static void Free(PoketchRoulette *appData)
+{
+    PoketchButtonManager_Free(appData->buttonManager);
+    PoketchRouletteGraphics_Free(appData->graphics);
+    Heap_Free(appData);
+}
+
+static void Task_Main(SysTask *task, void *appData)
+{
+    static BOOL (*const stateFuncs[])(PoketchRoulette *) = {
```
Gudf: **suggestion**: This function pointer type could probably be `typedef`d, to make this easier to parse.

VicSevenT: * What do you mean by easier to parse?

* I think I am generally resistant to making this a typedef, primarily because it would only be used once, and I don't know that splitting the format from the usage makes sense if there's no reuse. Then I got to thinking about whether there's a way to generalize the typedef and make it apply to all apps. I arrived at this:
```
#define POKETCH_STATE_FUNC_TYPEDEF(appStruct) typedef BOOL (*StateFunc)(appStruct *);
```
which would be called in each main.c like
```
POKETCH_STATE_FUNC_TYPEDEF(PoketchRoulette)
```
and then the usage changes to
```
   static const StateFunc stateFuncs[] = {
```
I made a test commit to demonstrate the usage here: https://github.com/VicSevenT/pokeplatinum/commit/5652bc3eb9bfc851405c655aabf7c823cb57b9bc.

But I question whether this is too indirect/too much abuse of the preprocessor. Thoughts?

Gudf: Easier to parse for a human, I mean. C function pointer types are kinda complex to decipher, so it's a bit unclear at a glance that this is a const array of pointers to functions that take a `PoketchRoulette*` and return a `BOOL`, imo?
I don't think a single-use typedef is that bad, in a way that's already what we do for function pointers in structs.

As for the preproc stuff I do think it's a bit too indirect. If this is a common idiom for poketch apps in general, would it make sense to have a single typedef for all poketch state functions (taking a `void*`)? I.e. `typedef BOOL (*PoketchStateFunc)(void *);`. Though I see how this might be confusing

VicSevenT: I did think about creating a generic typedef with `void *`, but then all the `State_` functions would have to be updated to match that signature and then would have to do a typecast inside, which I think is overkill (plus I'm not 100% sure that would still match). 

I ultimately agree that the macro is probably not the right move, but I wanted to put it out there if nothing else then as a curiosity. Knowing that single-use typedefs are used elsewhere in this code already is good enough for me, I'll make the update. In this case I'm planning to update all the apps with the typedef to keep the format consistent, so the PR will be inflated a bit, hopefully that's not an issue.

Gudf: Yeah I probably should have suggested this at the very least before your previous Poketch formatting harmonization PR, my bad 😅

VicSevenT: No problem, it was easy enough to replicate.

scbroede (approved):
will merge once @Gudf approves

PR #798, opened 2025-10-11
Town map: unpack descriptions data & generate text bank

Gudf:
Created a new subdirectory in `res` because of dependency order: my script depends on the NAIX for the signpost graphics and is depended on by the text banks narc.

This is the data the game pulls from when displaying the descriptions on the bottom screen in the town map. It is separated into two files, `tmap_block.dat` and `tmap_flags.dat`. 
`tmap_block.dat` holds each block's position, which signpost graphic to display, which string to use for each of the two parts of the tile's description and where to place them, as well as whether the tile belongs to any of the hidden locations.
`tmap_flags.dat` holds which flag, if any, to check for to decide whether or not each part of the description should be shown. It supports using a first arrival flag as well as using an arbitrary flag (up to flag 0xff, since that's held in a single byte), though the game only ever uses first arrival flags.

These, along with the town map text bank, are moved to a single JSON file with the following:
- `text_bank`: This is the start of the town map text bank in the standard JSON town map format. Used to have the key and a few strings that aren't part of a location's description.
- `areas` and `landmarks`: These contain essentially the definitions for each part of a location description: `areas` is the top halves, `landmarks` is the bottom halves. Each one is comprised of a `condition` and a `description`. The `description` is what ends up in `tmap_flags.dat`, and contains the strings to use depending on the ROM language.
- `blocks`: This is the tiles proper. Each one has its position on the map, the signpost graphic to use, which hidden location it's a part of, if any, and references to an `area` or a `landmark` to pull the description string indices and flags from.

This almost works to get matching files, but would fail for the victory road block which uses the same string as the pokemon league while having a different condition. Programmatically deduplicating the strings isn't an option since the three lakes all use the exact same string but in different indices, so I opted to instead add a `copy_description` property to the areas and landmarks, replacing `description` and indicating that it should reuse another area/landmark's description, including using the same string index. This is very much only a workaround solely for matching and is not intended to be used by modders.

Note: the game uses a fixed-size array of 200 elements in #785's `TownMapContext` to store the results of the description checks, which means this array needs to be resized if there's a need for more blocks.


ravepossum:
This looks great. The decisions you've made to organize it this way all make sense to me. I don't really have any feedback - I'll hold off a bit to see if any other collaborators have thoughts, and if not I think we can merge this.

lhearachel (approved):
Looks great to me. Let's merge. 😁

PR #801, opened 2025-10-14
Document distortion world floating platform jump points

Kuruyia:
This continues documentation of the distortion world overlay by documenting the contents of the second section of files in the `tw_arc.narc` archive.

This section contains the data for the jump points used by the player to enter or exit a floating platform. It contains info about where the jump point is located, where the player will be transported to (both position and sprite direction), what the destination floating platform is and how fast the animation is.

The code itself is mostly about animating the player movement from the jump point to its destination, and animating the player sprite so it corresponds to the floating platform orientation.

### DIFF
--- a/src/overlay009/ov9_02249960.c
+++ b/src/overlay009/ov9_02249960.c
@@ -5,6 +5,7 @@
 
 #include "constants/field/dynamic_map_features.h"
 #include "constants/field/map.h"
+#include "constants/map_object.h"
 #include "constants/species.h"
 #include "constants/types.h"
 #include "generated/map_headers.h"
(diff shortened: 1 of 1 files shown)


src/overlay009/ov9_02249960.c
```diff
@@ -9516,15 +9540,18 @@ static u32 DistWorldSystem_GetMapHeaderID(DistWorldSystem *system)
     return system->fieldSystem->location->mapId;
 }
 
-static int ov9_022510D8(u32 param0)
+static enum AvatarDistortionState GetAvatarDistortionStateForFloatingPlatformKind(u32 platformKind)
 {
-    switch (param0) {
+    switch (platformKind) {
     case 0:
```
scbroede: shouldn't these values use the `FloatingPlatformKind` enum?

Kuruyia: Indeed, fixed it

scbroede (approved):

PR #802, opened 2025-10-14
Document battle tower lobby

Viperio19:
This PR documents the text, scripts and script commands for the lobby of the battle tower. There are still some parts that are unclear, but this should be a good start.

There are three different receptionists:
- one for the single and double batlle rooms
- one for the multi battle room
- one for the wifi battle room

Their scripts make up most of the documentation.

### DIFF
--- a/asm/macros/scrcmd.inc
+++ b/asm/macros/scrcmd.inc
@@ -1444,20 +1444,20 @@
     .long \offset-.-4
     .endm
 
-    .macro StartBattleClient arg0, arg1, arg2, arg3
+    .macro StartBattleClient commType, arg1, arg2, destVarID
     .short 242
-    .short \arg0
+    .short \commType
     .short \arg1
     .short \arg2
-    .short \arg3
+    .short \destVarID
     .endm
 
-    .macro StartBattleServer arg0, arg1, arg2, arg3
+    .macro StartBattleServer commType, arg1, arg2, destVarID
     .short 243
-    .short \arg0
+    .short \commType
     .short \arg1
     .short \arg2
-    .short \arg3
+    .short \destVarID
     .endm
 
     .macro ScrCmd_Unused_0F4 arg0
@@ -2946,9 +2946,9 @@
     .short 518
     .endm
 
-    .macro GetCurNetID arg0
+    .macro GetCurNetID destVarID
     .short 519
-    .short \arg0
+    .short \destVarID
     .endm
 
     .macro DrawPokemonPreview species, gender
@@ -4714,7 +4714,7 @@
     .short \destVar
     .endm
 
-    .macro LogConnectedPlayersInfoInWiFiHistory
+    .macro LogLinkInfoInWiFiHistory
     .short 816
     .endm
 
--- a/include/overlay005/ov5_021F6454.h
+++ b/include/overlay005/ov5_021F6454.h
@@ -23,7 +23,7 @@ BOOL ScrCmd_331(ScriptContext *param0);
 BOOL ScrCmd_332(ScriptContext *param0);
 BOOL ScrCmd_338(ScriptContext *param0);
 BOOL ScrCmd_339(ScriptContext *param0);
-BOOL ScrCmd_LogConnectedPlayersInfoInWiFiHistory(ScriptContext *param0);
+BOOL ScrCmd_LogLinkInfoInWiFiHistory(ScriptContext *ctx);
 BOOL ScrCmd_333(ScriptContext *param0);
 
 #endif // POKEPLATINUM_OV5_021F6454_H
--- a/src/unk_020363E8.c
+++ b/src/unk_020363E8.c
@@ -178,9 +178,7 @@ void CommList_Refresh(void)
 
 void CommTool_ClearReceivedTempDataAllPlayers(void)
 {
-    int i;
-
-    for (i = 0; i < MAX_CONNECTED_PLAYERS; i++) {
+    for (int i = 0; i < MAX_CONNECTED_PLAYERS; i++) {
         sCommTool->hasRecievedTempData[i] = 0;
     }
 }
--- a/src/unk_02038F8C.c
+++ b/src/unk_02038F8C.c
@@ -10,23 +10,20 @@
 #include "trainer_info.h"
 #include "unk_0202C858.h"
 
-void WiFiHistory_FlagConnectedPlayersInfo(WiFiHistory *wiFiHistory)
+void WiFiHistory_FlagGeonetLinkInfo(WiFiHistory *wiFiHistory)
 {
-    int i;
-    int country, region, regionCode;
-    TrainerInfo *trainerInfo;
     int connectedCnt = CommSys_ConnectedCount();
 
     if (CommInfo_TrainerInfo(0) == NULL) {
         return;
     }
 
-    for (i = 0; i < connectedCnt; i++) {
+    for (int i = 0; i < connectedCnt; i++) {
         if (CommSys_CurNetId() != i) {
-            country = CommInfo_PlayerCountry(i);
-            region = CommInfo_PlayerRegion(i);
-            trainerInfo = CommInfo_TrainerInfo(i);
-            regionCode = TrainerInfo_RegionCode(trainerInfo);
+            int country = CommInfo_PlayerCountry(i);
+            int region = CommInfo_PlayerRegion(i);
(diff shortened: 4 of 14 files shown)


asm/macros/scrcmd.inc
```diff
@@ -1441,15 +1444,15 @@
     .long \offset-.-4
     .endm
 
-    .macro ScrCmd_0F2 arg0, arg1, arg2, arg3
+    .macro StartBattleClient arg0, arg1, arg2, arg3
```
scbroede: document the args you've identified. same goes for other cmds

generated/battle_tower_functions.txt
```diff
@@ -0,0 +1,43 @@
+BATTLE_TOWER_FUNCTION_CHECK_ENOUGH_VALID_POKEMON = 1
```
scbroede, line 1: do these files need to be generated, or can they go in a `constants` file instead?

include/overlay005/ov5_021F6454.h
```diff
@@ -23,7 +23,7 @@ BOOL ScrCmd_331(ScriptContext *param0);
 BOOL ScrCmd_332(ScriptContext *param0);
 BOOL ScrCmd_338(ScriptContext *param0);
 BOOL ScrCmd_339(ScriptContext *param0);
-BOOL ScrCmd_330(ScriptContext *param0);
+BOOL ScrCmd_LogConnectedPlayersInfoInWiFiHistory(ScriptContext *param0);
```
scbroede:
```suggestion
BOOL ScrCmd_LogConnectedPlayersInfoInWiFiHistory(ScriptContext *ctx);
```

asm/macros/scrcmd.inc
```diff
@@ -1795,12 +1798,12 @@
...
 
-    .macro ScrCmd_135 arg0
+    .macro ScrCmd_135 syncNo
     .short 309
-    .short \arg0
+    .short \syncNo
     .endm
 
-    .macro ScrCmd_136
+    .macro ClearReceivedTempDataAllPlayers
```
scbroede, line 1806: is it possible to indicate what this "temp data" is more specifically?

src/unk_020363E8.c
```diff
@@ -176,12 +176,12 @@ void CommList_Refresh(void)
...
 
-void sub_020365F4(void)
+void CommTool_ClearReceivedTempDataAllPlayers(void)
 {
-    int v0;
+    int i;
 
-    for (v0 = 0; v0 < MAX_CONNECTED_PLAYERS; v0++) {
-        sCommTool->hasRecievedTempData[v0] = 0;
+    for (i = 0; i < MAX_CONNECTED_PLAYERS; i++) {
```
scbroede: does this match as `for (int i = 0; i < MAX_CONNECTED_PLAYERS; i++) {`?

src/unk_02038F8C.c
```diff
@@ -10,25 +10,25 @@
...
-void sub_02038F8C(WiFiHistory *wiFiHistory)
+void WiFiHistory_FlagConnectedPlayersInfo(WiFiHistory *wiFiHistory)
 {
-    int v0;
-    int v1, v2, v3;
-    TrainerInfo *v4;
-    int v5 = CommSys_ConnectedCount();
+    int i;
+    int country, region, regionCode;
+    TrainerInfo *trainerInfo;
```
scbroede: merge declaration and assignment if possible

src/unk_020494DC.c
```diff
@@ -28,331 +29,329 @@
...
+    return FALSE;
 }
 
-BOOL ScrCmd_1DD(ScriptContext *param0)
+BOOL ScrCmd_CallBattleTowerFunction(ScriptContext *ctx)
 {
-    u16 v0, v1, v2;
-    u16 *v3, *v4;
+    u16 functionIndex, functionArgument, varID;
+    u16 *destVar, *v4;
```
scbroede: merge declaration and assignment if possible

src/unk_02049D08.c
```diff
@@ -436,42 +434,40 @@ BOOL sub_0204A378(UnkStruct_0204AFC4 *param0, void **param1, SaveData *saveData)
...
-    u16 v0 = 0, v1 = 0;
-    u16 v2[4], v3[4];
-    Party *v4;
-    Pokemon *v5;
-
-    v4 = SaveData_GetParty(saveData);
+    u16 i = 0, j = 0;
+    u16 species[4], heldItems[4];
+    Party *party = SaveData_GetParty(saveData);
+    Pokemon *mon;
```
scbroede: merge declaration and assignment if possible

src/unk_0204AEE8.c
```diff
@@ -102,38 +103,38 @@ static const u16 sTrainerClassToObjectID[][2] = {
...
-    u16 v1;
-    Strbuf *v2, *v3;
+    u8 i;
+    u16 species;
+    Strbuf *strbufSpeciesName, *strbufUnused;
     Pokedex *pokedex;
-    StringTemplate *v5;
-    MessageLoader *v6;
+    StringTemplate *strTemplate;
+    MessageLoader *msgLoader;
```
scbroede: merge declaration and assignment if possible

asm/macros/scrcmd.inc
```diff
@@ -4711,7 +4714,7 @@
     .short \destVar
     .endm
 
-    .macro ScrCmd_330
+    .macro LogConnectedPlayersInfoInWiFiHistory
```
scbroede: iirc we use `Link` as shorthand for "connected player" (same goes for all other uses of the term)

asm/macros/scrcmd.inc
Viperio19: I'm personally not sure, this is what it ends up getting called:
```c
void CommTool_ClearReceivedTempDataAllPlayers(void)
{
    int i;

    for (i = 0; i < MAX_CONNECTED_PLAYERS; i++) {
        sCommTool->hasRecievedTempData[i] = 0;
    }
}
```

asm/macros/scrcmd.inc
Viperio19: So would you suggest `LogLinkInfoInWiFiHistory`, `LogLink(ed)PlayersInfoInWiFiHistory`, or something else?

scbroede: the former imo

generated/battle_tower_functions.txt
Viperio19: From what I remembered, in order to use constants in scripts, they need to be generated, but I have may misremembered that. Either way, I just tested using a `constants` file instead and that resulted in a bunch of bad instruction errors:
```
FAILED: res/field/scripts/scr_seq.narc.p/scripts_init_sandgem_town_pokemon_research_lab
/mnt/c/Users/jesse/Documents/Decomp-local/pokeplatinum/pokeplatinum-main/tools/scripts/make_script_bin.sh -i ../include -i ../asm -i ./res/text -i ./res -i . --depfile --assembler /usr/bin/arm-none-eabi-gcc --objcopy /usr/bin/arm-none-eabi-objcopy --depfile --out-dir res/field/scripts/scr_seq.narc.p ../res/field/scripts/scripts_init_sandgem_town_pokemon_research_lab.s
../include/constants/battle_tower_functions.h: Assembler messages:
../include/constants/battle_tower_functions.h:4: Error: bad instruction `enum BattleTowerFunc {'
../include/constants/battle_tower_functions.h:5: Error: junk at end of line, first unrecognized char
```

Gudf: You can use a `constants` file, but only as long as it only has `#define`d constants, since the assembler doesn't understand C enums.

Viperio19: Ah, check. I think here it makes sense to keep it as an enum because of the mostly sequential structure, but feel free to let me know if there is a good reason to turn it into `#define`d constants

Viperio19: I see that this is a more general conversation about what does and does not belong in `generated/`, I'd like to suggest keeping this like it is for now (unless there are strong opinions against that) and maybe then creating an issue for going through `generated/` to determine what stays there and what gets turned into `#define`d constants.

scbroede: fair enough

scbroede (approved):
just the one thing left

PR #805, opened 2025-10-16
document traps

scbroede:
- some radar-related stuff is left undocumented for now; will be easier once some stuff in other files is documented (also not sure if i did enough to distinguish the touch screen radar from the radar you use from the secret base)
- not a fan of the names i gave to `PauseFieldSystemWithContextBit` and its resume equivalent, but couldn't think of anything better
- the "floor texture" name i used in the overlay 5 file is technically wrong because those are models rather than textures, but i couldn't come up with a better name for a completely flat model

### DIFF
--- a/src/overlay023/underground_traps.c
+++ b/src/overlay023/underground_traps.c
@@ -154,7 +154,7 @@ typedef struct TrapsEnv {
     void *currentTrapContext;
     UnusedFunc unused4; // never anything other than null
     BuriedTrap playerPlacedTraps[MAX_PLACED_TRAPS];
-    OverworldAnimManager *unk_368[MAX_PLACED_TRAPS];
+    OverworldAnimManager *trapTextureManager[MAX_PLACED_TRAPS];
     BuriedTrap buriedTraps[MAX_BURIED_TRAPS];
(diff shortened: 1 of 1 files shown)


src/overlay023/underground_traps.c
```diff
@@ -0,0 +1,5011 @@
...
+enum TrapSpriteResourceType {
+    TRAP_RESOURCES = 0,
+    RADAR_RESOURCES,
+    RESOURCE_TYPE_COUNT,
+};
+
+typedef void (*UnusedFn)(BgConfig *bgConfig);
+typedef void (*EndTrapEffectClientFn)(int netID, BOOL allowToolStepBack);
+typedef void (*TrapEffectClientFn)(int netID, BOOL isTool, int toolInitialDir);
+typedef void (*TrapServerFn)(int netID);
```
lhearachel: **nit**: The rest of the codebase uses `Func` as the suffix for function pointer typedefs.

lhearachel:
Resolve conflicts and we can merge

scbroede:
conflicts resolved (again). also added a bit of documentation for the overworld anim manager stuff

PR #808, opened 2025-10-24
Document remaining unused Poketch apps (Overlays 38, 39, 51, 55)

VicSevenT:
These are the remaining unused Poketch apps. Unlike the backlight toggle, these all appear to be either unfinished or just for testing purposes, and don't have any functional use. I still want to document them for the sake of completeness and to get them out of undocumented code, but I kept it simple and just named them `PoketchUnusedApp1-4`. I've gone through and named and polished everything, but I did forego some of the other steps done in the other apps, such as adding extra macros/enums. If we really want those additional constants, I can add those in. I've also grouped all 4 apps under one directory `src/applications/poketch/unused` to keep the parent directory a little less cluttered.

Here's what the apps look like if they are loaded.
* Unused apps 1, 3, and 4 (Overlays 38, 51, and 55) just show the text アプリ (app in Japanese) and a number. The numbers in unused apps 1 and 3 are hardcoded; unused app 4 shows its AppID, which will change depending on which app slot it is substituted into. Beside the numbers, the source code of these apps are virtually identical.

Unused app 1
unused_1
Unused app 3
unused_3
Unused app 4
unused_4

* Unused app 2 appears to be some sort of unfinished menu alternative. It has buttons for what could be the bag, party, trainer ID and more. The buttons do animate if pressed, but don't do anything beyond that.
unused_2


src/applications/poketch/unused/2/graphics.c
```diff
@@ -0,0 +1,168 @@
...
+    static const u16 tilemapButtonIndex[] = {
+        66,
+        84,
+        290,
+        308,
+        514,
+        532,
+    };
+    u16 *tilemapBuffer;
+    int j, i, tile;
```
scbroede, line 151: do these match with declaration and assignment merged?

VicSevenT: I can't get it to match if I merge any of the declarations

scbroede (approved):

PR #809, opened 2025-10-25
battle_cursor documentation

froggestspirit:
Documenting some battle cursor/menu stuff

### DIFF
--- a/src/battle/battle_cursor.c
+++ b/src/battle/battle_cursor.c
@@ -600,7 +600,7 @@ __attribute__((aligned(4))) static const u8 Unk_ov16_02270A0C[NELEMS(Unk_ov16_02
     0x4
 };
 
-__attribute__((aligned(4))) static const u8 sBattleMonMenuButtonLayout[3][2] = {
+__attribute__((aligned(4))) static const u8 sSelectTargetMenuButtonLayout[3][2] = {
     { 0x3, 0x1 },
     { 0x0, 0x2 },
     { 0x4, 0x4 }
@@ -4046,7 +4046,7 @@ static int BattleSystem_Cursor_Battler(UnkStruct_ov16_02268A14 *param0, BOOL cur
     cursor = &param0->cursor;
     v2 = &Unk_ov16_02270670[param0->unk_66B];
 
-    MI_CpuCopy8(sBattleMonMenuButtonLayout, v8, 3 * 2);
+    MI_CpuCopy8(sSelectTargetMenuButtonLayout, v8, 3 * 2);
 
     ov16_0226B20C(param0, v6, 1);
     ov16_0226B20C(param0, v7, 0);
@@ -4199,7 +4199,7 @@ static int BattleSystem_Cursor_Battler(UnkStruct_ov16_02268A14 *param0, BOOL cur
                 cursor->y = 1;
             }
 
-            v4 = sBattleMonMenuButtonLayout[cursor->y][cursor->x];
+            v4 = sSelectTargetMenuButtonLayout[cursor->y][cursor->x];
             BattleSystem_DrawCursor(param0->unk_6B8, v2->unk_14[v4].rect.left + 8, v2->unk_14[v4].rect.right - 8, v2->unk_14[v4].rect.top + 8, v2->unk_14[v4].rect.bottom - 8, (192 + 80) << FX32_SHIFT);
         } else {
             cursor->x = 0;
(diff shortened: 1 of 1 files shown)


scbroede:
don't have time for a full review at the moment, but you need to update `src/meson.build` to reflect the rename of `ov16_0226871C.c` to `battle_cursor.c` for it to build properly. also try running `make format` and committing the results to fix the pr-lint job (or set up `pre-commit` as outlined in contributing.md)

froggestspirit:
I set up pre-commit, I think that's all set now, but let me know if anything else is needed. I'm a little new to this

scbroede:
weird that it's not catching the problem then. which version of clang-format are you using?

lhearachel:
You can always run `pre-commit run --all-files` to force every file to be re-touched.

In this particular case, all of the failures are related to spacing between the last semantic token of a line and the start of a comment token, or `include` ordering.

src/battle/battle_cursor.c
```diff
@@ -1610,8 +1611,8 @@ int ov16_0226925C(UnkStruct_ov16_02268A14 *param0)
     } else {
         v1 = TouchScreen_CheckRectanglePressed(v4->unk_14);
 
-        if (v1 == 0xffffffff) {
-            v1 = ov16_0226BE48(param0);
+        if (v1 == 0xffffffff) { // Nothing was selected with touch
```
scbroede: we have a constant for this (`TOUCHSCREEN_INPUT_NONE`)

src/battle/battle_cursor.c
```diff
@@ -492,7 +492,7 @@ __attribute__((aligned(4))) static const u8 Unk_ov16_02270A08[NELEMS(Unk_ov16_02
     0x4
 };
 
-__attribute__((aligned(4))) static const u8 Unk_ov16_022701EC[2][3] = {
+__attribute__((aligned(4))) static const u8 battleMenuButtonLayout[2][3] = {
```
scbroede:
```suggestion
__attribute__((aligned(4))) static const u8 sBattleMenuButtonLayout[2][3] = {
```

same goes for the other static consts below

src/battle/battle_cursor.c
```diff
@@ -3782,26 +3783,26 @@ static void ov16_0226BD74(SysTask *param0, void *param1)
...
-    v0 = &param0->unk_6BC;
+    cursor = &param0->cursor;
     v1 = &Unk_ov16_02270670[param0->unk_66B];
 
     if (v1->unk_20 == NULL) {
         return 0xffffffff;
     }
 
-    if (v0->unk_00 == 0) {
+    if (cursor->unk_00 == 0) { // Check if the cursor is inactive
```
scbroede: can `unk_00` be renamed to something like `isActive` then? if so this could be `if (!cursor->isActive)` and the below line could be `cursor->isActive = TRUE`

src/battle/battle_cursor.c
```diff
@@ -3909,86 +3911,86 @@ static void ov16_0226C0A0(UnkStruct_ov16_02268A14 *param0, int param1)
     }
 }
 
-static int ov16_0226C100(UnkStruct_ov16_02268A14 *param0, int param1)
+static int BattleSystem_Cursor_PalPark(UnkStruct_ov16_02268A14 *param0, int cursorHidden)
```
scbroede:
```suggestion
static int BattleSystem_Cursor_PalPark(UnkStruct_ov16_02268A14 *param0, BOOL cursorHidden)
```

same goes for other functions with a `cursorHidden` param

src/battle/battle_cursor.c
```diff
@@ -3909,86 +3911,86 @@ static void ov16_0226C0A0(UnkStruct_ov16_02268A14 *param0, int param1)
...
-    v0 = &param0->unk_6BC;
+    cursor = &param0->cursor;
     v2 = &Unk_ov16_02270670[param0->unk_66B];
 
-    if (param1 == 1) {
-        v0->unk_02 = 0;
-        v0->unk_01 = 0;
-        v4 = Unk_ov16_022701CC[v0->unk_01][v0->unk_02];
-        ov16_0226DD9C(param0->unk_6B8, v2->unk_14[v4].rect.left + 8, v2->unk_14[v4].rect.right - 8, v2->unk_14[v4].rect.top + 8, v2->unk_14[v4].rect.bottom - 8, (192 + 80) << FX32_SHIFT);
+    if (cursorHidden == 1) {
```
scbroede:
```suggestion
    if (cursorHidden == TRUE) {
```

same goes for other functions with a `cursorHidden` param

src/battle/battle_cursor.c
```diff
@@ -600,7 +600,7 @@ __attribute__((aligned(4))) static const u8 Unk_ov16_02270A0C[NELEMS(Unk_ov16_02
     0x4
 };
 
-__attribute__((aligned(4))) static const u8 Unk_ov16_022701DC[3][2] = {
+__attribute__((aligned(4))) static const u8 battlerMenuButtonLayout[3][2] = {
```
scbroede: is there a way to distinguish this from `battleMenuButtonLayout` better?

src/battle/battle_cursor.c
froggestspirit: Only exception here, is that using BOOL on the struct definition doesn't compile matching, otherwise the other changes work

src/battle/battle_cursor.c
froggestspirit: I could try something like sBattleMonMenuButtonLayout. It's for doubles, when selecting the targets on moves

scbroede: how about `sSelectTargetMenuButtonLayout` or something along those lines?

PR #810, opened 2025-10-25
Document Poketch Kitchen Timer (Overlay 46)

VicSevenT:
The kitchen timer is very similar in design to the stopwatch and alarm clock. It stores timestamps for the start time and any pausing to track how much time has elapsed. When the time expires, the alarm sound blares until the timer is reset or paused. 

The app stores its data struct, which includes things like timestamps and display state, in memory so that it will persist when the poketch app is changed and changed back to the timer.

### DIFF
--- a/src/applications/poketch/kitchen_timer/main.c
+++ b/src/applications/poketch/kitchen_timer/main.c
@@ -29,7 +29,7 @@
 
 typedef struct KitchenTimerData {
     KitchenTimerDisplay display;
-    BOOL buttonUnpressed[3];
+    BOOL buttonUnpressed[NUM_STATE_BUTTONS];
     u64 startTimestamp;
     u64 pauseTimestamp;
     u64 elapsedTime;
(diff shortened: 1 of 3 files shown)


src/applications/poketch/kitchen_timer/main.c
```diff
@@ -0,0 +1,698 @@
...
+    appData->timer.isTickingDown = FALSE;
+    appData->timer.display.minutesTensDigit = 0;
+    appData->timer.display.minutesOnesDigit = 0;
+    appData->timer.display.secondsTensDigit = 0;
+    appData->timer.display.secondsOnesDigit = 0;
+}
+
+static void UpdateDisplayDigits(PoketchKitchenTimer *appData, u64 totalSeconds)
+{
+    u32 minutes, remainingSeconds;
```
scbroede: does this match with the declarations and assignments merged?

scbroede (commented):
imo add a constant for the number of buttons somewhere, and define all the animation indices

src/applications/poketch/kitchen_timer/main.c
VicSevenT: Indeed it does

PR #813, opened 2025-10-29
Document most remaining message script commands

Gudf:
Documents remaining script commands using `ScriptMessage`, except for one.
The remaining `ScrCmd_1FE` is used in the battle tower and seems to be for pre/post-battle messages, from what I can see in text bank 613? But it looks like it would require looking into battle frontier stuff more in-depth so I chose to left it undocumented for now.

### DIFF
--- a/src/unk_0204AEE8.c
+++ b/src/unk_0204AEE8.c
@@ -28,6 +28,7 @@
 #include "save_player.h"
 #include "savedata.h"
 #include "strbuf.h"
+#include "string_padding_mode.h"
 #include "string_template.h"
 #include "trainer_info.h"
 #include "unk_0202D05C.h"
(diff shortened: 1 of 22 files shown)


src/unk_0204AEE8.c
```diff
@@ -102,38 +103,39 @@ static const u16 sTrainerClassToObjectID[][2] = {
     { TRAINER_CLASS_POKE_KID, OBJ_EVENT_GFX_PIKACHU }
 };
 
-StringTemplate *sub_0204AEE8(SaveData *saveData, u16 param1, u16 param2, u8 param3, u8 *param4)
+StringTemplate *BattleFrontier_MakeCaughtBannedSpeciesList(SaveData *saveData, u16 numPokemonNeeded, u16 unused2, u8 unused3, u8 *outNumBannedCaught)
```
Viperio19: I also documented this (and ScrCmd_1FF) in my battle tower lobby PR, so there will be some conflicts. I like your naming more overall, except that this function doesn't make a list with the _caught_ banned species, but with the _seen_ banned species: `Pokedex_HasSeenSpecies(pokedex, bannedSpecies)`.

Gudf: Oops, not sure how I missed that.

lhearachel:
Resolve conflicts and we can merge.

PR #818, opened 2025-11-01
Document Poketch Calendar (Overlay 50)

VicSevenT:
The calendar has only a single sprite to mark the current day, everything else is rendered in backgrounds, with the numbers being on a layer in front of the main background. Most of the complexity of this app comes from laying out the month based on what day of the week it starts on.


src/applications/poketch/calendar/graphics.c
```diff
@@ -0,0 +1,254 @@
...
+#include "applications/poketch/poketch_graphics.h"
+#include "applications/poketch/poketch_task.h"
+
+#include "bg_window.h"
+#include "graphics.h"
+#include "heap.h"
+#include "sys_task_manager.h"
+
+#define MARKED_CELL_TILE_IDX   143
+#define UNMARKED_CELL_TILE_IDX 131
```
lhearachel, line 15: **question**: At what point are we planning to unpack the graphics archive and replace all of these instances with generated constants?

VicSevenT: I'm planning to unpack the poketch NARC very soon. I wanted to finish documenting all the apps first, as along the way I've been building a list of what index is what/what app uses each asset/index. 

Note though, that these particular constants are not NARC indices, but rather indices of specific tiles within a single NCGR file, which are used to update the tilemap for marked and unmarked calendar cells, respectively. So I don't think these could be replaced by anything generated from asset unpacking.

PR #819, opened 2025-11-01
Document Poketch Pokemon History (Overlay 54)

VicSevenT:
A very simple app. It just pulls a list of 12 pokemon and then displays their icons. Tapping on one plays its cry.

### DIFF
--- a/src/applications/poketch/pokemon_history/graphics.c
+++ b/src/applications/poketch/pokemon_history/graphics.c
@@ -15,6 +15,17 @@
 #include "sys_task_manager.h"
 #include "text.h"
 
+#define MON_ANIM_DATA(r, c)                                        \
+    {                                                              \
+        .translation = { FX32_CONST(48 + HISTORY_ICON_STEP_X * c), \
+            FX32_CONST(48 + HISTORY_ICON_STEP_Y * r) },            \
+        .animIdx = 4,                                              \
+        .flip = NNS_G2D_RENDERERFLIP_NONE,                         \
+        .oamPriority = 2,                                          \
+        .priority = MAX_HISTORY_SIZE - (4 * r) - c,                \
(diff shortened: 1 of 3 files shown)


src/applications/poketch/pokemon_history/graphics.c
```diff
@@ -0,0 +1,263 @@
...
+        {
+            .translation = { FX32_CONST(48 + 40 * 3), FX32_CONST(48 + 48 * 0) },
+            .animIdx = 4,
+            .flip = NNS_G2D_RENDERERFLIP_NONE,
+            .oamPriority = 2,
+            .priority = 9,
+            .hasAffineTransform = TRUE,
+        },
+        {
+            .translation = { FX32_CONST(48 + 40 * 0), FX32_CONST(48 + 48 * 1) },
```
lhearachel: **question**: Is it worth keeping the products of these `const`s?

VicSevenT: I don't quite understand your question. Are you asking whether these expressions should be parametrized with constants? Or if they should be evaluated in place so that it's just a single number inside each `FX_CONST()`? Or something else?

lhearachel: The latter; here, these would be assigned as:

```c
            .translation = { FX32_CONST(48), FX32_CONST(96) },
```

I do see that there are some macros later that codify how the multipliers on `+ 40` and `+ 48` are used; maybe we should make use of those here?

VicSevenT: Hmm, that's an interesting idea. I don't know if I had realized before that `MON_HITBOX` was using those same values. 

I'm pushing a change with a new macro to generate the animation data structs similar to the hitboxes, It might be a little overkill, but I do think it demonstrates nicely how the sprites and hitboxes are related.

PR #821, opened 2025-11-02
Rename message loader type constants

Gudf:
Also renames `MessageLoaderType` to `MessageLoaderMode`


lhearachel (dismissed):

include/message.h
```diff
@@ -30,9 +30,9 @@ typedef struct MessageLoader {
     };
 } MessageLoader;
 
-enum MessageLoaderType {
-    MESSAGE_LOADER_BANK_HANDLE,
-    MESSAGE_LOADER_NARC_HANDLE,
+enum MessageLoaderMode {
+    PRELOAD_ENTIRE_BANK,
+    LOAD_ON_DEMAND,
 };
```
lhearachel, line 36: **thought**: Should these be prefixed somehow to assist with auto-complete?

Gudf: Probably, yeah, didn't think of that... Do we think `MSG_LOADER_` is good enough?

lhearachel: I think so, yeah. Maybe an additional `M_` after that if we really care about specificity.

Gudf: These are the only message loader-related constants afaict, I don't think it's necessary

PR #822, opened 2025-11-03
Unpack Move Tutor interface graphics

VicSevenT:
These are the assets for the Move Tutor interface, used by the tutors for Draco Meteor and Frenzy Plant/Blast Burn/Hydro Cannon, and by the Move Reminder. 

This is my first time unpacking graphics. I think I did everything in line with how other NARCs have been unpacked, but am not 100%. Doing this now as a trial before unpacking the bigger poketch.narc, and because I've been looking at the tutor interface code so I figured it would be good to get the naix.h file built ahead of time.


res/graphics/move_tutor/select_box.png
lhearachel, line 1: **question**: Can we generate a match for this using the `-cell` option? It would make the graphics in-repo look a bit nicer.

res/graphics/move_tutor/meson.build
```diff
@@ -0,0 +1,69 @@
...
+        narc_exe, 'create',
+        '--naix',
+        '--order', move_tutor_order,
+        '--output', '@OUTPUT0@',
+        '@PRIVATE_DIR@',
+    ]
+)
+
+nitrofs_files += move_tutor_narc
+naix_headers += move_tutor_narc[1]
```
scbroede, line 73: could you clean up the indentation in this file?

scbroede:
as a piece of general guidance when unpacking graphics: pal files are bad, but are sometimes necessary for clarity. my personal rule of thumb is that if an NCLR only applies to a single NCGR, and that NCGR only ever uses that NCLR, generate the NCLR from the png instead of using a pal file

res/graphics/move_tutor/select_box.png
VicSevenT: Yes, that can be done. I left that out to keep a single meson rule `move_tutor_tiles` for generating NCGRs. In order to use the `-cell` arg with `select_box.png` I would need to split it into a separate rule, correct? I have no problem doing that, just want to confirm that that is preferred.

VicSevenT:
> as a piece of general guidance when unpacking graphics: pal files are bad, but are sometimes necessary for clarity. my personal rule of thumb is that if an NCLR only applies to a single NCGR, and that NCGR only ever uses that NCLR, generate the NCLR from the png instead of using a pal file

Got it, thanks. I'll keep that in mind in the future, but I think in this PR both pals are still needed, right? `sprites.pal` is associated with all 3 NCGRs, so the NCLR can't be built from just one. And while the background has just a single NCGR and NCLR, that NCLR has several palettes in it, so the png doesn't have the full set of information needed to regenerate the NCLR.

scbroede:
> And while the background has just a single NCGR and NCLR, that NCLR has several palettes in it, so the png doesn't have the full set of information needed to regenerate the NCLR.

i added flags to nitrogfx specifically to handle this kind of scenario. use `-convertTo8Bpp` when converting to png, and add `-convertTo4Bpp` to the list of flags when converting back to NCGR

VicSevenT:
Cool, those flags worked and I was able to get rid of the background pal.

res/graphics/move_tutor/select_box.png
VicSevenT: Oh I just found `ncgr_cell_gen` and that works with all the graphics. Updated to that and now all the sprites PNGs are well formed.

scbroede (approved):

PR #825, opened 2025-11-05
Document battle tower battle salon and multi corridor

Viperio19:
This PR is mainly focused on `scripts_battle_tower_battle_salon.s` and `scripts_battle_tower_corridor_multi.s`, it started getting a bit out of scope with different parts of the battle tower so I decided to just keep this a relatively small PR and continue with more battle tower documentation later.


res/field/scripts/scripts_battle_tower_battle_salon.s
```diff
@@ -2,119 +2,126 @@
...
 
-
-    ScriptEntry _01AC
-    ScriptEntry _0275
-    ScriptEntry _02FE
-    ScriptEntry _0365
-    ScriptEntry _03C6
-    ScriptEntry _0427
-    ScriptEntry _002A
+#define LOCALID_GUIDE  0
```
CharlesFolz4: Reminder that you should be renaming and using the id field from the events file instead of redefining duplicate values for these.  You'll have to include the generated `res/field/events/events_{location}.h` file though.
Same goes for the instance of these constants being redefined in `include/constants/battle_tower.h`

Viperio19: Oh I didn't know those could be included, thanks! The constants are different though, you can see those start with cheryl at 0 instead of at 1 for the local IDs because those are used for the partner related arrays in the `BattleTower` struct.

About the ids, there is for example `BATTLE_TOWER_BATTLE_SALON_RILEY_3`. Do we want to keep the numbers at the end that indicate the id value?

CharlesFolz4: You're right; I missed that the IDs didn't quite match up with `battle_tower.h`

I don't think the numbers at the end are necessary here; they were only initially included to prevent duplicate values when the default values were generated based on the graphics ID.  You can see how that would be relevant in the `events_battle_tower_corridor_multi` file, but even there, replacing the number with your documented `LEFT_GUIDE` and `RIGHT_GUIDE` names would be much more descriptive and useful.

Viperio19: Also to add on to the id names, the guide/attendant is always called `BATTLE_TOWER_BATTLE_SALON_TEALA_0`. I've mostly called her "Guide" in script names so far, is there a preference between Teala and Guide? And if Guide is preferred in script names, would we then also want `BATTLE_TOWER_BATTLE_SALON_GUIDE_0`?

CharlesFolz4: As discussed in the discord, let's prefer more descriptive names that are easily identifiable in context.

Viperio19: Forgot to mention it, but I fixed this in my last PR

asm/macros/scrcmd.inc
```diff
@@ -690,11 +691,11 @@
...
-    .macro ScrCmd_06B arg0, arg1, arg2
+    .macro MoveCamera x, y, z
     .short 107
-    .short \arg0
-    .short \arg1
-    .short \arg2
+    .short \x
+    .short \y
+    .short \z
     .endm
```
lhearachel, line 699: **question**: Is the ordering of these correct?

generated/frontier_trainers.txt
```diff
@@ -0,0 +1,315 @@
+FRONTIER_TRAINER_YOUNGSTER_JIM
```
lhearachel, line 1: **thought**: I think these should not be included here, because now a user has to add an entry to this file if they want to add a new trainer. I would argue the same is true for the other trainers, maybe. Do we want or need this to be generated before the NARC is built (and thus can emit a header of its own, which will include all files and not just those listed here)? Further, do we need these to be an `enum` in C?

asm/macros/scrcmd.inc
Viperio19: yeah, it's based on the `VecFx32` in here:
```diff
- static BOOL ScrCmd_06B(ScriptContext *ctx)
+ static BOOL ScrCmd_MoveCamera(ScriptContext *ctx)
{
    FieldSystem *fieldSystem = ctx->fieldSystem;

-   u16 v0 = ScriptContext_GetVar(ctx);
-   u16 v1 = ScriptContext_GetVar(ctx);
-   u16 v2 = ScriptContext_GetVar(ctx);
+   u16 x = ScriptContext_GetVar(ctx);
+   u16 y = ScriptContext_GetVar(ctx);
+   u16 z = ScriptContext_GetVar(ctx);

-   VecFx32 v3;
-   v3.x = FX32_CONST(v0);
-   v3.y = FX32_CONST(v1);
-   v3.z = FX32_CONST(v2);
+   VecFx32 pos;
+   pos.x = FX32_CONST(x);
+   pos.y = FX32_CONST(y);
+   pos.z = FX32_CONST(z);

-   sub_020630AC(Player_MapObject(ctx->fieldSystem->playerAvatar), &v3);
-   Camera_Move(&v3, ctx->fieldSystem->camera);
+   sub_020630AC(Player_MapObject(ctx->fieldSystem->playerAvatar), &pos);
+   Camera_Move(&pos, ctx->fieldSystem->camera);

    return FALSE;
}

generated/frontier_trainers.txt
Viperio19: I'm not sure what other approach you are currently suggesting and whether this thought is relevant for this PR or more something to do in the future. Either way, the only instance of these constants being used in this PR is in `unk_02049D08.c`, but I've already stashed some more documentation where I use these constants a lot more in `unk_0204AEE8.c`.

lhearachel: I think it's fine for now, at least. We can revisit it later, if we decide it's worthwhile to do so.

PR #827, opened 2025-11-06
Document move reminder

cbt6:


res/text/move_reminder.json
h2o-DS, line 1: Suggestion: rename this file

src/applications/move_reminder.c
```diff
@@ -0,0 +1,1708 @@
...
+#include "type_icon.h"
+#include "unk_0208C098.h"
+#include "unk_02094EDC.h"
+#include "vram_transfer.h"
+
+#include "constdata/const_020F410C.h"
+#include "res/text/bank/unk_0645.h"
+
+enum {
+    MOVE_REMINDER_WIN_LABEL_BATTLE_MOVES,
```
h2o-DS: Nitpick: enum first entries should have `=0`

src/applications/move_reminder.c
```diff
@@ -0,0 +1,1708 @@
...
+    ManagedSprite *managedSprites[MAX_MOVE_REMINDER_SPRITE];
+    PokemonSummary summary;
+    ApplicationManager *appMan;
+    int nextState;
+    u8 numMoves;
+    u8 textPrinterID;
+    u8 yesNoCallback;
+} MoveReminderController;
+
+static void MoveReminder_Setup(MoveReminderController *controller);
```
h2o-DS, line 103: Polish: static functions don't require a prefix

src/applications/move_reminder.c
```diff
@@ -0,0 +1,1708 @@
...
+#include "text.h"
+#include "type_icon.h"
+#include "unk_0208C098.h"
+#include "unk_02094EDC.h"
+#include "vram_transfer.h"
+
+#include "constdata/const_020F410C.h"
+#include "res/text/bank/unk_0645.h"
+
+enum {
```
h2o-DS, line 46: Suggestion: you can name enum sets and use them to specify parameters if desired.

src/applications/move_reminder.c
```diff
@@ -0,0 +1,1708 @@
...
+
+    MessageLoader_GetStrbuf(controller->messageLoader, MoveReminder_Text_AppealPts, controller->strbuf);
+    MoveReminder_DrawText(controller, MOVE_REMINDER_WIN_LABEL_APPEAL_PTS, FONT_SYSTEM, TEXT_COLOR(15, 14, 0), ALIGN_CENTER);
+}
+
+static u32 MoveReminder_GetNumMoves(MoveReminderController *controller)
+{
+    u32 i;
+
+    for (i = 0; i < 256; i++) {
```
h2o-DS, line 1077: Question: does this `256` have a constant?

src/applications/move_reminder.c
```diff
@@ -0,0 +1,1708 @@
...
+    MoveReminder_SetStringTemplate(controller, str);
+    RenderControlFlags_SetCanABSpeedUpPrint(TRUE);
+    RenderControlFlags_SetAutoScrollFlags(AUTO_SCROLL_DISABLED);
+
+    controller->textPrinterID = Text_AddPrinterWithParams(&controller->windows[MOVE_REMINDER_WIN_MESSAGE_BOX], FONT_MESSAGE, controller->strbuf, 0, 0, Options_TextFrameDelay(controller->data->options), MoveReminder_TextPrinterCallback);
+}
+
+static BOOL MoveReminder_TextPrinterCallback(TextPrinterTemplate *printer, u16 param1)
+{
+    switch (param1) {
```
h2o-DS, line 1327: Suggestion: identify this parameter

src/applications/move_reminder.c
```diff
@@ -0,0 +1,1708 @@
...
+    for (u16 i = 0; i < NELEMS(controller->managedSprites); i++) {
+        Sprite_UpdateAnim(controller->managedSprites[i]->sprite, FX32_ONE);
+    }
+}
+
+static void MoveReminder_LoadSpriteResources(MoveReminderController *controller, NARC *narc)
+{
+    SpriteSystem_LoadCharResObjFromOpenNarc(controller->spriteSystem, controller->spriteManager, narc, 5, FALSE, NNS_G2D_VRAM_TYPE_2DMAIN, 47272);
+    SpriteSystem_LoadCharResObjFromOpenNarc(controller->spriteSystem, controller->spriteManager, narc, 8, FALSE, NNS_G2D_VRAM_TYPE_2DMAIN, 47273);
+    SpriteSystem_LoadCharResObjFromOpenNarc(controller->spriteSystem, controller->spriteManager, narc, 2, FALSE, NNS_G2D_VRAM_TYPE_2DMAIN, 47274);
```
h2o-DS, line 1513: Polish: if this only expects a single narc replace values with narc member indexes. This applies to the calls lower in this function as well

src/applications/move_reminder.c
```diff
@@ -0,0 +1,1708 @@
...
+}
+
+static void MoveReminder_Scroll(MoveReminderController *controller, u16 prevListPos, u16 listPos)
+{
+    u32 i;
+    s16 x, y;
+
+    if (prevListPos < listPos) {
+        // scroll down
+        for (i = 0; i < 7; i++) {
```
h2o-DS: Question: this `7` is used for each loop. Do you think a constant is justified?

src/applications/move_reminder.c
cbt6: Unnecessary since these are used only in this file.

src/applications/move_reminder.c
cbt6: I don't see a reason against prefixing static functions.

src/applications/move_reminder.c
cbt6: Params in text printer callbacks are still undocumented at the moment - tracking them down and finding out what they do is out of scope of this PR.

src/applications/move_reminder.c
cbt6: The NARC is still not yet unpacked, and unpacking it is out of scope of this PR.

src/applications/move_reminder.c
cbt6: Documenting list menu constants is out of scope of this PR.

src/applications/move_reminder.c
cbt6: No.

src/applications/move_reminder.c
Gudf: This really isn't a list menu constant, the 7 element display limit is specific to this app, and set in `sListMenuTemplate` (as `maxDisplay`).

h2o-DS (approved):

PR #830, opened 2025-11-14
Unpack `tmap_gra.narc`

Gudf:
Created a new tool to combine multiple NCLRs into one, optionally preserving the PCMP values. This allows splitting large NCLRs containing palettes for multiple NCGRs or for NCGRs that use multiple palettes, either to pull the different palettes from the corresponding PNGs or to store sets of related palettes in separate JASC-PAL files.
The option to preserve PCMP values was added because nitrogfx only allows generating NCLRs with sequential PCMP values, but one of the palettes in this NARC skips over value 0xA.


res/graphics/town_map/fly_destination_blocks.png
h2o-DS, line 1: Suggestion: store palettes in images when possible.

Gudf: I'd rather not split this file, it holds all 5 palettes the fly destination blocks use depending on their state.

h2o-DS: You could convert the image to 8bpp.

Gudf: Ah, right. Let me try that

Gudf: Managed to make it work

res/graphics/town_map/cursor_player_female.pal
h2o-DS, line 1: Nitpick: `player_male`/`player_female` are preferred to `lucas`/`dawn`

res/graphics/town_map/fly_destination_blocks.png
h2o-DS: This applies to other images as well.

res/graphics/town_map/meson.build
```diff
@@ -0,0 +1,158 @@
...
+    ],
+    input: [
+        tilemaps_nscr,
+        sprite_anims_nanr,
+        sprite_cells_ncer,
+        sprite_ncgr_cells,
+        fly_destination_blocks_ncgr,
+        tilesets_ncgr,
+        palettes_nclr,
+        copy_gen.process(top_screen_sprites_nclr),
```
h2o-DS, line 174: Can this not directly be input?

Gudf: No, it needs to be explicitly copied to the NARC target's private directory

res/graphics/town_map/fly_destination_blocks.png
Gudf: I don't think it makes sense for the tilesets without having to manually assign the right palette to each tile, which I can't find an appropriate tool to do, and it definitely doesn't make sense for the remaining sprite palettes.

tools/scripts/nclr_concat.py
lhearachel, line 1: **quibble**: Does this need to be a standalone tool-directory? Could it be in `tools/scripts`?

Gudf: I guess it doesn't, I think I initially put it here because I was thinking about expanding it and rewriting it in C/C++ in the process but I ended up not doing it

PR #831, opened 2025-11-16
Rename object event field "flag" to "hidden"

CharlesFolz4:
Rename object event field "flag" to "hidden"
This is a flag field that determines if the object should be hidden or displayed


cbt6:
Disagree, the name "flag" should be retained since the value is the name of the flag. Naming it "hidden" implies that the field should be a boolean instead.

lhearachel:
Akin to the above, would prefer `hidden_flag` to just `hidden`, personally. I do think `flag` by itself implies that the flag being set to `1` makes the object visible, which is the opposite of the actual behavior. This historically has tripped up a number of people in the binary sphere (myself included), and I wouldn't be surprised if it continued to do so in the decomp.

CharlesFolz4:
I have no objection to using `hidden_flag` if the datatype is considered important enough to retain and also isn't considered obvious enough to be implicit.

cbt6:
I think this is just an issue of inconsistent / poor naming of the flags. Ideally the flags should be named as `FLAG_HIDE_<map_header_name>_<object_event_name>` so that it's obvious whether the flag hides or shows.

Regardless, if you're renaming the field in the data files, you should also rename it in the `ObjectEvent` struct in C (and thus also `MapObject`, `MapObjectSave`, etc.)

PR #832, opened 2025-11-17
Accept trainer ids as object event scripts

metallicity:
Updates the definition of the `script` field in `object_events` items to allow passing in a trainer ID, rather than requiring a `3XXX` script offset (or `5XXX` for the second trainer in double battle pairs). Along the way, all generic trainers defined in this manner (rather than triggering their battle within some other script) have had their "id" field updated to reflect their actual name and trainer class, rather than simply being named after their overworld sprite.

As "Galactic Grunts" are all unnamed, when multiple generic trainers of this class share a map, they have been numbered to avoid collisions, but these numbers now match how their trainer IDs are differentiated instead of just by their index in `object_events`. These also remain distinct from non-trainer Grunts (or those with scripted battles), as their IDs are of the form `GRUNT_M` or `GRUNT_F` rather than the trainer class `GALACTIC_GRUNT`.

One additional field has been added to `object_events` items, `double_battle_id`. This corresponds to their dialog in `res/trainers/data` (e.g. `TRMSG_POST_DOUBLE_BATTLE_1` vs. `TRMSG_POST_DOUBLE_BATTLE_2`), with a value of 1 resolving to a `3XXX` script offset and a value of 2 resolving to a `5XXX` script offset. I also manually checked each of these pairs (either by dialog or in-game) to ensure their names were correctly assigned.

### DIFF
--- a/tools/json2bin/convert.py
+++ b/tools/json2bin/convert.py
@@ -67,13 +67,11 @@ def from_trainer_type(s: str) -> int:
 def from_var_flag(v: str) -> int:
     return 0 if v == "0" else vars_flags.VarFlag[v].value
 
-def from_trainer_id(v: str, double_battle_id: int) -> int:
-    return trainers.TrainerID[v].value - 1 + (5000 if double_battle_id == 2 else 3000)
-
 def from_script(v: str, double_battle_id: int) -> int:
-    if v.startswith("TRAINER_"):
-        return from_trainer_id(v, double_battle_id)
-    return int(v)
+    if v in trainers.TrainerID.__members__:
+        return trainers.TrainerID[v].value - 1 + (5000 if double_battle_id == 2 else 3000)
+    else:
+        return int(v)   
 
 def from_map_header(s: str) -> int:
     return map_headers.MapHeader[s].value


tools/json2bin/convert.py
```diff
@@ -66,6 +67,14 @@ def from_trainer_type(s: str) -> int:
 def from_var_flag(v: str) -> int:
     return 0 if v == "0" else vars_flags.VarFlag[v].value
 
+def from_trainer_id(v: str, double_battle_id: int) -> int:
+    return trainers.TrainerID[v].value - 1 + (5000 if double_battle_id == 2 else 3000)
+
+def from_script(v: str, double_battle_id: int) -> int:
+    if v.startswith("TRAINER_"):
+        return from_trainer_id(v, double_battle_id)
+    return int(v)
```
cbt6:
```suggestion
def from_script(v: str, double_battle_id: int) -> int:
    if v in trainers.TrainerID.__members__:
        return trainers.TrainerID[v].value - 1 + (5000 if double_battle_id == 2 else 3000)
    else:
        return int(v)
```

cbt6 (commented):
Remove the unnecessary `"double_battle_id": 1,` in script files.

(We could probably rename this field to be a boolean with not present being `false` by default, but I've got no good suggestions for a better name)

tools/json2bin/convert.py
metallicity: I originally had this as an optional boolean, but found it quite weird that only one of member of each double battle pair would have `"double_battle": true`. Additionally, while `"double_battle_id: 1"` has no effect on the build output over leaving it implicit, this is a clarifying and informative value, not only indicating the trainer is partnered with another trainer on the map, but also that this particular object event corresponds to the `TRMSG_*DOUBLE_BATTLE*_1` dialog messages (rather than `*_2`). It also seems more intuitive when adding new double battle trainer pairs as well, rather than trying to name the field in a way that explains why only one trainer should have it set, and which trainer this should be.

edit: meant to reply to the comment above, the suggested change here makes sense

PR #833, opened 2025-11-21
document battle regulations

scbroede:
rebase of https://github.com/pret/pokeplatinum/pull/775 with pr feedback addressed

### DIFF
--- a/src/battle_regulation.c
+++ b/src/battle_regulation.c
@@ -9,6 +9,8 @@
 #include "savedata.h"
 #include "strbuf.h"
 
+#include "res/text/bank/unk_0353.h"
+
 int BattleRegulation_Size(void)
 {
     return sizeof(BattleRegulation);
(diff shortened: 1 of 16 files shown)


src/battle_regulation.c
```diff
@@ -0,0 +1,243 @@
...
+}
+
+void BattleRegulation_GetNameByIndex(SaveData *saveData, int index, Strbuf *strbuf, int heapID)
+{
+    MessageLoader *messageLoader;
+    BattleRegulationData *regulationData = NULL;
+
+    if (index < NELEMS(BattleRegulation_PredefinedRules)) {
+        messageLoader = MessageLoader_Init(MSG_LOADER_PRELOAD_ENTIRE_BANK, NARC_INDEX_MSGDATA__PL_MSG, TEXT_BANK_UNK_0353, heapID);
+        MessageLoader_GetStrbuf(messageLoader, 70 + index, strbuf);
```
lhearachel: **polish**: Use a constant-ref from the text-bank header for `70`.

src/battle_regulation_validation.c
```diff
@@ -0,0 +1,280 @@
...
+    HeightWeightData *data = Pokedex_HeightWeightData(heapID);
+    Pokedex_HeightWeightData_Load(data, FALSE, heapID);
+    return data;
+}
+
+void HeightWeightData_Free(HeightWeightData *heightWeightData)
+{
+    Pokedex_HeightWeightData_Release(heightWeightData);
+    Pokedex_HeightWeightData_Free(heightWeightData);
+}
```
lhearachel: **thought**: Should we separate these into their own TU, if that matches?

scbroede: maybe. they're general-use functions in a vacuum, but in practice are only ever used for `HeightWeightData` that get passed into other functions in this file. an alternative could be renaming them to `BattleRegulation_(Load/Free)HeightWeightData`

splitting them out does still match, fwiw

lhearachel: I think tying them to regulation is probably a mistake.

> they're general-use functions in a vacuum, but in practice are only ever used for HeightWeightData that get passed into other functions in this file.

You certain about that? There are some instances in `party_menu/main.c`.

lhearachel (approved):
Pending merge conflicts

PR #724, opened 2025-09-01
Document TV broadcast commands and some internals

metallicity:
There's a lot of code related to TV broadcasts, and I didn't even cover half of it or look very deep into the exact narc structure. I mostly aimed to keep this as small as possible while still reaching enough understanding along the way to give accurate front-facing names to all the various script sub-commands and constants involved. There are a few things I'm a little shaky on for exactly how program "segments" are represented and retrieved, since I avoided touching that code more than necessary, but I'm fairly confident about all the function/macro/constant names I landed on. Some of the param names might have room for improvement.

The basic structure of a player interaction with a generic TV goes like this:
- Check the status of the current TV program, which resets on a cyclical week-long schedule for each 15 minute interval:
  - `TV_BROADCAST_STATUS_BEGIN` (the program has no previously played segments and isn't marked as finished):
    - Show the program's "greeting" message, then "lead segment intro" message, then the message for the first segment
  - `TV_BROADCAST_STATUS_CONTINUE` (the program has 1-3 previously played segments and isn't marked as finished):
    - Find the next segment to play, and if it exists:
      - Show the program's "next segment intro" message, then the message for the next segment
    - If there is no next segment to play (this is also checked before the first segment, but is probably unreachable):
      - Show the program's "extended farewell" message (often a player call-to-action for more program material)
      - Mark the program as finished (until the time slot rotates)
  - `TV_BROADCAST_STATUS_FINISH` (if the program has already played the maximum 4 segments):
    - Show the shorter "farewell" message
    - Mark the program as finished (until the time slot rotates)
  - `TV_BROADCAST_STATUS_FINISHED` (the program is marked as finished):
    - Show a random commercial message.
  - Any other result would technically also play a commercial, but in practice this unconditional branch is unreachable

PR #834, opened 2025-11-22
Document "Trainer Sightings" TV episode segments

metallicity:
Following up from #724, which (among other things related to TV broadcasts) documented all TV episode segments for the "Interview" category, all "Trainer Sighting" segments have now been documented as well.


asm/macros/scrcmd.inc
```diff
@@ -4693,9 +4693,9 @@
     .short \species
     .endm
 
-    .macro ScrCmd_31B arg0
+    .macro GetCurrentSafariGameCaughtNum destVar
```
scbroede: imo this should be something like `GetLastSafariGameCaughtNum` instead, since it's called in the reception area after the safari game ends

include/field_battle_data_transfer.h
```diff
@@ -31,6 +30,18 @@ typedef struct BattleRecords {
...
 
+typedef struct CaptureAttempt {
+    int resultMask;
+    u16 ballsThrown;
+    u16 species;
+    u8 gender;
+    u8 language;
+    u8 metGame;
+    u8 hasNickname;
+    u16 nickname[11];
```
scbroede:
```suggestion
    u16 nickname[MON_NAME_LEN + 1];
```

PR #841, opened 2025-11-29
Document `unk_02073838` -> `simple3d`

Gudf:
Another 3D abstraction...

Mostly just another G3D wrapper, not much to mention... Seems to be the main G3D wrapper they used for overlays 5, 9 and 104, so hopefully that should make the field graphics stuff easier to understand!


include/simple3d.h
lhearachel, line 1: I can't say that I'm thrilled about this name, but, well. I don't have a better one.

src/another3d.c
```diff
@@ -0,0 +1,337 @@
...
+    FX32_ONE,
+};
+
+static const MtxFx33 sIdentity = {
+    // clang-format off
+    FX32_ONE, 0, 0,
+    0, FX32_ONE, 0,
+    0, 0, FX32_ONE,
+    // clang-format on
+};
```
lhearachel: **suggestion**: I would just align these entirely:


```suggestion
static const MtxFx33 sIdentity = {
    // clang-format off
    FX32_ONE, 0,        0,
    0,        FX32_ONE, 0,
    0,        0,        FX32_ONE,
    // clang-format on
};
```

PR #843, opened 2025-12-02
document underground pc

scbroede:


src/overlay023/underground_pc.c
```diff
@@ -0,0 +1,968 @@
...
+static void UndergroundPC_UpdateCursorPos(UndergroundMenu *menu)
+{
+    u16 prevPos = menu->unk_2AE;
+    ListMenu_CalcTrueCursorPos(menu->unk_48, &menu->unk_2AE);
+
+    if (prevPos != menu->unk_2AE) {
+        Sound_PlayEffect(SEQ_SE_CONFIRM);
+    }
+
+    return;
```
lhearachel: **polish**: Remove.

lhearachel (approved):
Just one little change to make, the rest looks good.

src/overlay023/underground_pc.c
scbroede: done. also updated the more or less identical function in `underground_player_talk.c` to match

PR #844, opened 2025-12-02
Created macros for most calls to `CallCommonScript`

AsparagusEduardo:
Also moved the macros to their own file


asm/macros/common.inc
lhearachel, line 1: **thought**: The name of this file seems potentially confusing, as there is also `btlcmd.inc`, etc. Naming this `common` implies that it could be used for any of those, which is definitely not true.

lhearachel (commented):
I'm generally fine with these macros, but we might want something more robust that is based on the content of `scripts_common` long-term. There's been a lot of discussion about that previously, so I won't get into it here.

asm/macros/common.inc
AsparagusEduardo: > **thought**: The name of this file seems potentially confusing, as there is also `btlcmd.inc`, etc. Naming this `common` implies that it could be used for any of those, which is definitely not true.

would `fieldcmd.inc` be better?

lhearachel: Is there any reason they can't just be in `scrcmd.inc`?

AsparagusEduardo: > Is there any reason they can't just be in `scrcmd.inc`?

Sure, I have no issue

lhearachel (approved):

PR #845, opened 2025-12-02
Documented most "Sinnoh Now" TV functions

AsparagusEduardo:
The only one not handled is the Pokémon photo rating, as I believe it'll be better to document the struct and its related functions in a different PR

### DIFF
--- a/src/tv_episode_segment.c
+++ b/src/tv_episode_segment.c
@@ -2053,7 +2053,7 @@ static BOOL TVEpisodeSegment_IsEligible_BattleFrontierFrontlineNews_Multi(FieldS
     return SystemFlag_HandleFirstArrivalToZone(SaveData_GetVarsFlags(fieldSystem->saveData), HANDLE_FLAG_CHECK, FIRST_ARRIVAL_FIGHT_AREA);
 }
 
-static const u8 Unk_020EFD34[] = {
+static const u8 sGroupRNGEntries[] = {
     RECORD_MIXED_RNG_PLAYER_OVERRIDE,
     RECORD_MIXED_RNG_QUEUE_0,
     RECORD_MIXED_RNG_QUEUE_1,
@@ -2065,8 +2065,8 @@ static int RecordMixedRNG_CountValidEntries(RecordMixedRNG *rngCollection)
 {
     int i, count;
 
-    for (i = 0, count = 0; i < NELEMS(Unk_020EFD34); i++) {
-        if (RecordMixedRNG_IsEntryValid(rngCollection, Unk_020EFD34[i])) {
+    for (i = 0, count = 0; i < NELEMS(sGroupRNGEntries); i++) {
+        if (RecordMixedRNG_IsEntryValid(rngCollection, sGroupRNGEntries[i])) {
             count++;
         }
     }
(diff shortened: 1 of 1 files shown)


src/tv_episode_segment.c
```diff
@@ -2050,212 +2054,213 @@ static BOOL TVEpisodeSegment_IsEligible_BattleFrontierFrontlineNews_Multi(FieldS
 }
 
 static const u8 Unk_020EFD34[] = {
```
lhearachel: **polish**: Name this table, if possible.

src/tv_episode_segment.c
```diff
@@ -2050,212 +2054,213 @@ static BOOL TVEpisodeSegment_IsEligible_BattleFrontierFrontlineNews_Multi(FieldS
...
     int weather = FieldSystem_GetWeather(fieldSystem, mapHeaderID);
-    StringTemplate_SetLocationName(param1, 0, MapHeader_GetMapLabelTextID(mapHeaderID));
+    StringTemplate_SetLocationName(template, 0, MapHeader_GetMapLabelTextID(mapHeaderID));
 
     switch (weather) {
     case OVERWORLD_WEATHER_CLEAR:
         switch (LCRNG_RandMod(4)) {
         case 0:
-            return 1;
+            return pl_msg_00000418_00001;
```
lhearachel, line 2147: **thought**: Is it worth documenting these strings now?

src/tv_episode_segment.c
```diff
@@ -2050,212 +2054,213 @@ static BOOL TVEpisodeSegment_IsEligible_BattleFrontierFrontlineNews_Multi(FieldS
...
-        return 10;
+        return pl_msg_00000418_00010;
     case OVERWORLD_WEATHER_THUNDERSTORM:
-        return 11;
+        return pl_msg_00000418_00011;
     case OVERWORLD_WEATHER_HAILING:
-        return 12;
+        return pl_msg_00000418_00012;
     default:
         GF_ASSERT(0);
```
lhearachel: **polish**: `GF_ASSERT(FALSE);`

src/tv_episode_segment.c
```diff
@@ -3099,21 +3011,21 @@ static const TVProgramSegment sInterviewsSegments[TV_PROGRAM_TYPE_INTERVIEWS_NUM
 };
 
 static const TVProgramSegment sSinnohNowSegments[TV_PROGRAM_TYPE_SINNOH_NOW_NUM_SEGMENTS] = {
-    { sub_0206E870, sub_0206E928 },
-    { sub_0206E940, sub_0206EA0C },
-    { sub_0206EA10, NULL },
+    { MsgFunc_DiscoveringGroups, FieldSystem_RecordMixedRNGHasValidEntries },
```
metallicity: This function naming convention doesn't match the existing one for other already document program types, which instead use `TVEpisodeSegment_[LoadMessage/IsEligible]_[SegmentName]`. Also, looking at the unused messages in the Sinnoh Now text bank, it might be appropriate to give `AlwaysFalse` and `AlwaysFalse2` the more descriptive segment names `MatchupChannelDummy` and `PokemonResearchCornerDummy`, being two segments from Diamond/Pearl that were cut in Platinum (but seem to have some traces remaining).

AsparagusEduardo: > This function naming convention doesn't match the existing one for other already document program types...

Thank you for the clarification. Adjusted names to match current convention.

src/tv_episode_segment.c
AsparagusEduardo: > **polish**: `GF_ASSERT(FALSE);`

Done

src/tv_episode_segment.c
AsparagusEduardo: > **thought**: Is it worth documenting these strings now?

I'm uncertain due to the unresolved naming convention.

src/tv_episode_segment.c
AsparagusEduardo: > **polish**: Name this table, if possible.

Done. `sGroupRNGEntries`

AsparagusEduardo:
I forgot to mention, changes applied

PR #847, opened 2025-12-03
`Strbuf` to `String`

AsparagusEduardo:
Changed every instance of `Strbuf` into `String`, including functions, labels and defines.

### DIFF
--- a/include/battle_regulation.h
+++ b/include/battle_regulation.h
@@ -44,7 +44,7 @@ int BattleRegulation_SaveSize(void);
 void BattleRegulation_Copy(const BattleRegulation *src, BattleRegulation *dst);
 void RegulationBattles_Init(BattleRegulationData *regulationData);
 void BattleRegulation_GetName(const BattleRegulation *regulation, String *string);
-String *BattleRegulation_GetNameStrbuf(const BattleRegulation *regulation, int heapID);
+String *BattleRegulation_GetNameString(const BattleRegulation *regulation, int heapID);
 int BattleRegulation_GetRuleValue(const BattleRegulation *regulation, enum BattleRegulationRule rule);
 BattleRegulation *BattleRegulation_Load(SaveData *saveData, int index);
 void BattleRegulation_Save(SaveData *saveData, const BattleRegulation *regulation);
--- a/src/battle_regulation.c
+++ b/src/battle_regulation.c
@@ -36,7 +36,7 @@ void BattleRegulation_GetName(const BattleRegulation *regulation, String *string
     String_CopyNumChars(string, regulation->name, BATTLE_REGULATION_NAME_LENGTH);
 }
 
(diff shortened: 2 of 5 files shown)


include/applications/bag/defs.h
```diff
@@ -118,16 +118,16 @@ struct BagController {
...
     Menu *menu;
     ListMenu *itemList;
     StringList *itemListEntries;
-    Strbuf *itemNames[LARGEST_POCKET_SIZE];
-    Strbuf *strBuffer;
-    Strbuf *itemCountX;
-    Strbuf *itemCountNumberFmt;
-    Strbuf *pocketNames[POCKET_MAX];
+    String *itemNames[LARGEST_POCKET_SIZE];
+    String *stringBuffer;
```
scbroede, line 127: just call this one `string` like everything else imo

AsparagusEduardo: > just call this one `string` like everything else imo

Done. Since I saw other uses of String here, I assumed that this was used differently, but now it's just `string`

Gudf: Afraid to say Edu was right, this is explicitly a buffer intended to be printed with `BagUI_PrintStrBufferToWideMsgBox` (which should *also* retain the `Buffer` part).

AsparagusEduardo: > Afraid to say Edu was right, this is explicitly a buffer intended to be printed with `BagUI_PrintStrBufferToWideMsgBox` (which should _also_ retain the `Buffer` part).

Reverted and set the name of the function to `BagUI_PrintStringBufferToWideMsgBox`

scbroede: in that case strBuffer was completely fine, but if you'd rather not change it again i won't insist on it

AsparagusEduardo: > in that case strBuffer was completely fine, but if you'd rather not change it again i won't insist on it

I have no problem 👍 
Done

lhearachel (approved):
Thanks for doing this! Good to merge pending conflicts.

include/applications/bag/defs.h
scbroede: if you're okay with one more change, the field itself was fine at `strBuffer`. if not, i'm fine merging this once the conflicts are fixed

AsparagusEduardo:
> Thanks for doing this! Good to merge pending conflicts.

Solved 👍

lhearachel:
Sorry, looks like more conflicts popped up during a merge spree. :sweat: Once you fix these, I'll prioritize merging your PR to avoid another set of them.

AsparagusEduardo:
> Sorry, looks like more conflicts popped up during a merge spree. 😓 Once you fix these, I'll prioritize merging your PR to avoid another set of them.

Conflicts fixed

PR #849, opened 2025-12-08
Unpack poketch.narc

VicSevenT:
Unpacks `res/prebuilt/graphics/poketch.narc` into `res/graphics/poketch`.

There is one NCLR, `poketch_border.NCLR` that I had to use a PAL file for, as I could not find a way to build a PNG that can generate the NCLR. The problem is the the NCLR contains just 2 palettes, starting with index 14. 

There is also graphics file, `stopwatch.NCGR.lz` that I left unconverted to PNG. It has two problems: some of the sprites overlap, meaning a generated PNG is lacking information when converted back to NCGR, and the cell file has character offsets expecting the NCGR is loaded into memory behind another asset, so as is the cell file doesn't convert the graphic properly. This was discussed in Discord and it sounds like there is hope that nitrogfx support for these issues is added in the future. I will create a Github issue to make a reminder of this so that it can be addressed when nitrogfx is updated.


VicSevenT:
Issue to track unpacking of the stopwatch graphics https://github.com/pret/pokeplatinum/issues/850

res/graphics/poketch/alarm_clock.NSCR.lz
lhearachel, line 1: **issue**: We should still un-`lz` these into base `NSCR` files, so that they can be used by outside programs like Tilemap Studio.

lhearachel (commented):
Only one note below. Thanks for doing this, and for going through to use NAIX constants in the code! 🙏

res/graphics/poketch/alarm_clock.NSCR.lz
VicSevenT: Done

scbroede (approved):

PR #851, opened 2025-12-10
Document Pokemon animations

ravepossum:
This PR documents the Pokemon sprite animation script system. Similar to field scripts, they use bytecode scripts with macros that correspond to script commands. The scripts can update the sprite positions through a series of individual commands or they can use functions that apply a standard transformation like moving the sprite in a straight line or in a curved pattern.

There were a few symbols already named that have been renamed (namely `PokemonAnimationSys` -> `PokemonAnimManager` and `PokeAnimationSettings` -> `PokemonAnimTemplate`) to make them consistent with how we tend to name things nowadays.

This only covers the code that drives the scripts - in a follow-up PR I will unpack the `pokeanime/` NARCs and write macros for the animation script commands.


src/pokemon_anim.c
```diff
@@ -0,0 +1,1165 @@
...
+
+#include "constants/graphics.h"
+
+#include "heap.h"
+#include "narc.h"
+#include "pokemon_sprite.h"
+#include "sys_task.h"
+
+// The first/third & second/fourth entries here seem to be used interchangeably.
+enum AnimScriptReadType {
```
lhearachel: **thought**: I noticed that the union of all of these `enum` values is continuous as a single `enum`. Is there a particular reason to split them?

src/pokemon_anim.c
```diff
@@ -0,0 +1,1165 @@
...
+    }
+}
+
+// The final parameter here is only ever invoked with a value of 1.
+static int ReadScript_WordIndex(u32 *scriptPtr, u8 index, u8 one)
+{
+    int ret = scriptPtr[index];
+
+    if (one != 1) {
+        GF_ASSERT(0);
```
lhearachel: **polish**: `GF_ASSERT(FALSE)` (applies to other instances)

src/pokemon_anim.c
```diff
@@ -0,0 +1,1165 @@
...
+{
+    return ReadScript_WordIndex(scriptPtr, 0, one);
+}
+
+static int ReadScript_Word(u32 *scriptPtr)
+{
+    return ReadScript_WordIndexZero(scriptPtr, 1);
+}
+
+static TransformData *TransformFuncData_New(PokemonAnim *monAnim, const u8 funcType)
```
lhearachel: **polish**: Irrelevant `const` on a primitive parameter

src/pokemon_anim.c
ravepossum: Yeah based on the weird values it does seem likely that they were originally one single enum, but my thinking was that they indicate such a variety of different things that it made more sense to separate them into different ones based on how they're actually used.

ravepossum: Per discord discussion, we're going to go ahead and switch these to a contiguous `generated` enum since these will need to be used in scripts. Hoping that the unique prefixes for each will help keep them distinct.

ravepossum:
Feedback addressed!

generated/pokemon_anim_constants.txt
```diff
@@ -0,0 +1,36 @@
+TRANSFORM_FUNC_CURVE = 0
```
lhearachel, line 1: **suggestion**: Metang supports comments (prefixed with `#`, both start-of-line and in-line), and I think that this is a file that is deserving of one explaining what's actually going on here.

ravepossum: Ah I totally forgot about that. I'll definitely add some!

lhearachel: Copying over from Discord:

> I wasn't very clear in the PR but my plan was to annotate the constants in the follow up PR I've already started working on

PR #852, opened 2025-12-11
Generic data-processing library with built-in error-reporting

lhearachel:
There are two pieces to this PR:

1. Commit the `dataproc` library source code into the tree, which is a successor to `datagen` and our use of `rapidjson`.
2. A C utility to pack all species data from the relevant `res/pokemon/**/{data,sprite_data}.json` files, built using the above library.

`dataproc` is a middleware library for generic access of fields in data-files. Right now, the only implementation is for JSON, but any language could theoretically be used, if the appropriate interface is fully implemented. `yyjson` serves as the back-end for processing JSON files; a patch is included for the subproject definition to implement byte-indices of individual value-nodes in the tree.

The library comes with smart reporting of errors for free. For instance, given the following function call:

```c
dp_u8(dp_get(data, ".base_stats.defense"));
```

And the following data file snippet:

```json
{
    "base_stats": {
        "hp": 90,
        "attack": 92,
        "defense": 512
    }
}
```

The following error will be written to the console:

```text
data.json:5:20: error: .base_stats.defense: expected a value in the range [0,255]
    5 |         "defense": 512,
      |                    ^~^
```

## `speciesproc`

This is a reference implementation of a processing tool using `dataproc`. The common bits are extracted into another library which provide all the boilerplate for initializing and packing NARCs, generating header-files from templates, loading constants from C files containing `enum`s or `#define`s, and emitting dependency files. Any C file loaded for constant-extraction and any template loaded for a header-file will be included in the output dependency file for integration with the build-system.


tools/dataproc/src/resdat.c
Gudf, line 1: **Suggestion**: It would be nice to have generated constants to refer to the cldat entries.

lhearachel: I would like to defer that as part of a follow-up PR after all of these files have been unpacked.

lhearachel: I went ahead and wound up doing this anyways. Not too hard.

lhearachel:
I've included unpacked data files for the `resdat` and `cldat` used by the Party Menu, Shop Menu, Town Map, and Poffin Case. All others require unpacking NARCs that we have yet to unpack.

lhearachel:
This is now ready for review. I've modified the description according to the updated content.

tools/dataproc/src/speciesproc.c
```diff
@@ -0,0 +1,738 @@
...
+    const char *size = dp_string(dp_lookup_s(dp_objmemb(root, "size"), "FootprintSize"));
+    const char *type = dp_string(dp_lookup_s(dp_objmemb(root, "type"), "FOOTPRINT_TYPE"));
+
+    fprintf(*f_footprint_sizes, "    { %s, %s },\n",
+            has ? "TRUE" : "FALSE",
+            size);
+    fprintf(*f_footprint_types, "    [SPECIES_%s] = { %s, %s },\n",
+            species_upper,
+            type,
+            has || i == SPECIES_SPIRITOMB ? "TRUE" : "FALSE");
```
scbroede, line 615: if we're doing this check instead of using the `has_type` field, we can just remove it from spiritomb's data.json file

tools/dataproc/src/speciesproc.c
```diff
@@ -0,0 +1,738 @@
...
+        return result;
+    }
+
+    for (size_t i = 0; i < len_evos; i++) {
+        datanode_t entry  = dp_arrelem(evolutions, i);
+        datanode_t method = dp_lookup(dp_arrelem(entry, 0), "EvolutionMethod");
+
+        datanode_t param   = { .type = DATAPROC_T_MAPPED, .mapped = 0 };
+        datanode_t species = { 0 };
+        switch ((enum EvolutionMethod)dp_u16(method)) {
```
scbroede, line 372: we might want a default case here with an error message in case someone adds a new evolution method

tools/dataproc/src/speciesproc.c
lhearachel: Good catch, I had made a note of that in my clean-up pass but just forgot to do it.

tools/dataproc/src/speciesproc.c
lhearachel: The call to `lookup` above it will already result in an error when that happens, so this `u16` call would be functionally equivalent to `EVO_NONE` (which is `0`). What we could do instead is just `continue` if that `lookup` fails.

lhearachel: ...oh, wait, I see what you mean. Right, yes. We do need one of those here.

scbroede (approved):

PR #853, opened 2025-12-11
Reorganize battle party screen constants

Morglom:
Just doing some renaming and moving around to prepare for the battle party buttons pull request (which would be massive if combined with the changes in this pull request).


include/overlay013/battle_party_text.h
```diff
@@ -3,6 +3,137 @@
 
 #include "overlay013/battle_party.h"
 
+enum BattlePartyPokemonPartyScreenWindow {
+    BATTLE_PARTY_POKEMON_PARTY_SCREEN_WINDOW_PARTY_SLOT_1 = 0,
```
lhearachel: **thought**: A lot of these names are quite verbose; is it sufficient to condense the prefixes to `BATTLE_`? Or something generally shorter. Happy to bikeshed this a bit more.

include/overlay013/battle_sub_menu_buttons_defs.h
lhearachel, line 1: **note**: In the party-screen code (the base version from the Start menu), I called a header like this `defs.h`. Is that more clear than `util.h`?

include/overlay013/battle_party_text.h
Morglom: Yeah, I could do that.

include/overlay013/battle_sub_menu_buttons_defs.h
Morglom: I think that would better.

lhearachel (approved):

PR #859, opened 2025-12-15
Document Battle Party Buttons

Morglom:
It's a bit wonkly in some spots because of how the buttons are ordered.


include/overlay013/battle_party.h
```diff
@@ -97,6 +97,30 @@ enum BattleConfirmLearnMoveScreenButton {
     BATTLE_CONFIRM_LEARN_MOVE_SCREEN_BUTTON_CANCEL,
 };
 
+#define PARTY_POKEMON_BUTTON_WIDTH  16
```
lhearachel: **suggestion**: It's probably worth documenting that these are widths in tiles (assuming that that's the case, since they are used as dimensions for an array of `u16`)

src/overlay013/battle_party_buttons.c
```diff
@@ -0,0 +1,762 @@
...
+        } else {
+            if (battleParty->partyPokemon[button].curHP == 0) {
+                for (u8 i = 0; i < width * height; i++) {
+                    buttonData[i] = (buttonData[i] & 0xfff) | (2 << 12);
+                }
+            } else if (BattlePartyTask_CheckIfSwitchingWithPartnersPokemon(battleParty, button - 0) == TRUE) {
+                for (u8 i = 0; i < width * height; i++) {
+                    buttonData[i] = (buttonData[i] & 0xfff) | (1 << 12);
+                }
+            }
```
lhearachel, line 459: **question**: Is this updating palettes?

Morglom: I think so.

include/overlay013/battle_party.h
Morglom: Do you think they should be renamed or would a comment above the section like "Dimensions are in tiles" be better?

lhearachel: I think a rename is our typical paradigm. Just adding `_TILES` as a suffix should be sufficient! :grin:

Morglom: > I think a rename is our typical paradigm. Just adding `_TILES` as a suffix should be sufficient! 😁

Done

PR #860, opened 2025-12-17
Duplicate generator for field-scripts to force dependency on events

lhearachel:
This kinda sucks, but it's the only solution that I can find for establishing this link. I'm working on a patch proposal for Meson 1.11 that aims to make it possible to specify extra dependencies in the `process` call.

### DIFF
--- a/res/field/scripts/meson.build
+++ b/res/field/scripts/meson.build
@@ -1,5 +1,33 @@
-relative_source_root = fs.relative_to(meson.project_source_root(), meson.project_build_root())
 relative_build_dir = fs.relative_to(meson.current_build_dir(), meson.project_build_root())
+
+# NOTE: This is a functional copy of script_bin_gen, which is separated from the
+# base generator due to an additional dependency on headers generated for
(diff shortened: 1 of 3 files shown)


res/field/scripts/meson.build
```diff
@@ -1,6 +1,34 @@
-relative_source_root = fs.relative_to(meson.project_source_root(), meson.project_build_root())
 relative_build_dir = fs.relative_to(meson.current_build_dir(), meson.project_build_root())
 
+# NOTE: This is a functional copy of script_bin_gen, which is separated from the
+# base generator due to an additional dependency on headers generated for
+# events IDs. Meson unfortunately does not have a clean way to copy an existing
+# object and tweak a single property.
+# WARN: If you update this generator, you should also consider updating its parent
+# in res/field/scripts/meson.build
```
cbt6:
```suggestion
# WARN: If you update this generator, you should also consider updating its parent
# in res/meson.build
```

scbroede (approved):

PR #861, opened 2025-12-19
Document Battle Frontier script context functions

VicSevenT:
This is some initial documentation for the functions that run battle frontier-specific scripts found in `fr_script.narc`. `frontier_script_context.c` is akin to `field_script_context.c`, with many of the functions being the same or very similar. I've also gone through and given preliminary names to all the script commands, similar to how all the `scrcmd`s are in the main script environment. This should hopefully serve as a good jumping-off point for further documentation of the battle frontier scripts.

### DIFF
--- a/include/overlay104/frontier_script_context.h
+++ b/include/overlay104/frontier_script_context.h
@@ -21,7 +21,7 @@ struct FrontierScriptContext {
     const FrontierScrCmdFunc *cmdTable;
     u32 cmdTableSize;
     u16 data[4];
-    MessageLoader *unk_80;
+    MessageLoader *msgLoader;
     u8 *scripts;
     FrontierShouldResumeScriptFunc shouldResume;
     u8 unk_8C;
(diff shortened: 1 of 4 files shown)


include/overlay104/frontier_script_context.h
```diff
@@ -0,0 +1,42 @@
...
+    u32 unk_04;
+    u16 unk_08[8];
+    u8 stackPointer;
+    u8 state;
+    const u8 *scriptPtr;
+    const u8 *stack[20];
+    const FrontierScrCmdFunc *cmdTable;
+    u32 cmdTableSize;
+    u16 data[4];
+    MessageLoader *unk_80;
```
cbt6:
```suggestion
    MessageLoader *messageLoader;
```

src/overlay104/frontier_script_context.c
```diff
@@ -0,0 +1,146 @@
...
+    ctx->shouldResume = NULL;
+    ctx->cmdTable = cmdTable;
+    ctx->cmdTableSize = cmdTableSize;
+
+    u32 i;
+    for (i = 0; i < NELEMS(ctx->data); i++) {
+        ctx->data[i] = 0;
+    }
+
+    for (i = 0; i < 20; i++) {
```
cbt6:
```suggestion
    for (i = 0; i < NELEMS(ctx->stack); i++) {
```

PR #869, opened 2025-12-25
Document pokemon league scripts

Viperio19:
This PR documents all `scripts_pokemon_league.*.s` files, with some small side tangents because of everything that happens in `scripts_pokemon_league_hall_of_fame.s` after beating the pokemon league.

I renamed `.*pokemon_league_elevator_to_hall_of_fame.*` to `.*pokemon_league_hallway_to_hall_of_fame.*` because there is no actual elevator in this map.

I renamed `unk_02052C6C.c/h` to `clear_game.c/h` with the `ClearGame_` prefix for functions and the static struct `ClearGameStruct`. I think the file name can be improved, and especially the struct can use a more descriptive name. I still wanted to give the struct a placeholder name to make documenting this file easier, and since this file is now mostly documented, a more descriptive name should definitely be possible (but I'm personally not sure what that might be).

I also added `constants/roamer_states.h` with:
```c
#define ROAMER_STATE_ROAMING  0
#define ROAMER_STATE_CAPTURED 1
#define ROAMER_STATE_DEFEATED 2
#define ROAMER_STATE_RESET    3
```
The one name that could be improved here is `ROAMER_STATE_RESET`. This state is set in the hall of fame if the roamer was previously defeated. The player can then go back to wherever the respective roamer gets triggered to make it start roaming once more, then setting the state back to `ROAMER_STATE_ROAMING`. Maybe something like `ROAMER_STATE_INACTIVE` would be more descriptive?


res/field/scripts/scripts_pokemon_league_elevator_to_aaron_room.s
```diff
@@ -1,12 +1,12 @@
...
 
-    ScriptEntry _000E
+    ScriptEntry PokemonLeagueElevatorToAaronRoom_OnTransition
     ScriptEntry _0029
-    ScriptEntry _0031
+    ScriptEntry PokemonLeagueElevatorToAaronRoom_TriggerPlatformLift
     ScriptEntryEnd
 
-_000E:
+PokemonLeagueElevatorToAaronRoom_OnTransition:
```
lhearachel, line 9: **question**: What's the difference between `OnTransition`, `OnFrame`, etc.? I assume this is terminology borrowed from `pokeemerald`.

res/field/scripts/scripts_pokemon_league_hallway_to_hall_of_fame.s
```diff
@@ -0,0 +1,141 @@
...
+    WaitMovement
+    BufferPlayerName 0
+    Message PokemonLeagueHallwayToHallOfFame_Text_YourLastBattleWasSplendid
+    CloseMessage
+    ApplyMovement POKEMON_LEAGUE_HALLWAY_TO_HALL_OF_FAME_CYNTHIA, PokemonLeagueHallwayToHallOfFame_Movement_ExclamationMark
+    WaitMovement
+    Message PokemonLeagueHallwayToHallOfFame_Text_HelloProfessorRowan
+    BufferPlayerName 0
+    Message PokemonLeagueHallwayToHallOfFame_Text_YouveGrownIntoARealChampion
+    ApplyMovement POKEMON_LEAGUE_HALLWAY_TO_HALL_OF_FAME_CYNTHIA, PokemonLeagueHallwayToHallOfFame_Movement_WalkInPlaceWest
```
lhearachel: **thought**: I wonder if this setup makes sense, or if we should do what `pokeemerald` does and use `LOCALID_CYNTHIA`, etc.

src/clear_game.c
```diff
@@ -0,0 +1,316 @@
...
+    startLocation = FieldOverworldState_GetSpecialLocation(SaveData_GetFieldOverworldState(fieldSystem->saveData));
+    firstRespawnLocation = FieldOverworldState_GetExitLocation(SaveData_GetFieldOverworldState(fieldSystem->saveData));
+
+    clearGameStruct->gameCompleted = SystemFlag_CheckGameCompleted(varsFlags);
+    clearGameStruct->displayData.trainerInfo = SaveData_GetTrainerInfo(fieldSystem->saveData);
+    clearGameStruct->displayData.party = SaveData_GetParty(fieldSystem->saveData);
+    clearGameStruct->displayData.playTime = SaveData_GetPlayTime(fieldSystem->saveData);
+    clearGameStruct->unk_10.gender = TrainerInfo_Gender(SaveData_GetTrainerInfo(fieldSystem->saveData));
+    clearGameStruct->unk_10.gameCompleted = SystemFlag_CheckGameCompleted(varsFlags);
+    clearGameStruct->unk_10.pokedex = SaveData_GetPokedex(fieldSystem->saveData);
```
lhearachel: **thought**: This seems like it's just some basic data about the player. Could we name it and the associated struct? Should the struct's definition also just be part of this source file, or is it used elsewhere in the codebase?

src/clear_game.c
```diff
@@ -0,0 +1,316 @@
...
+{
+    Options *options = SaveData_GetOptions(fieldSystem->saveData);
+
+    clearGameStruct->string = MessageBank_GetNewStringFromNARC(NARC_INDEX_MSGDATA__PL_MSG, TEXT_BANK_COMMON_STRINGS, 15, HEAP_ID_FIELD3);
+
+    FieldMessage_AddWindow(fieldSystem->bgConfig, &clearGameStruct->window, BG_LAYER_MAIN_3);
+    FieldMessage_DrawWindow(&clearGameStruct->window, options);
+
+    clearGameStruct->printerID = FieldMessage_Print(&clearGameStruct->window, clearGameStruct->string, options, 1);
+    clearGameStruct->dial = Window_AddWaitDial(&clearGameStruct->window, 1024 - (18 + 12));
```
lhearachel: **polish**: Replace `1024 - (18 + 12)` with `BASE_TILE_SCROLLING_MESSAGE_BOX` from `include/constants/field_base_tiles.h`.

src/clear_game.c
```diff
@@ -0,0 +1,316 @@
...
+    Window_Init(&clearGameStruct->window);
+    GXLayers_SetBanks(&v0);
+
+    GX_SetDispSelect(GX_DISP_SELECT_MAIN_SUB);
+
+    SetAllGraphicsModes(&v1);
+    Bg_MaskPalette(BG_LAYER_MAIN_3, 0x0);
+    Bg_InitFromTemplate(fieldSystem->bgConfig, BG_LAYER_MAIN_3, &v2, 0);
+    Bg_ClearTilesRange(BG_LAYER_MAIN_3, 0x20, 0, HEAP_ID_FIELD3);
+    Bg_FillTilemapRect(fieldSystem->bgConfig, 3, 0x0, 0, 0, 32, 32, 17);
```
lhearachel: **polish**:

- `3` -> `BG_LAYER_MAIN_3`
- `17` -> `TILEMAP_FILL_VAL_INCLUDES_PALETTE`

src/clear_game.c
```diff
@@ -0,0 +1,316 @@
...
+}
+
+static void sub_02053028(FieldSystem *fieldSystem, ClearGameStruct *clearGameStruct, int saveResult)
+{
+    MessageLoader *msgLoader = MessageLoader_Init(MSG_LOADER_LOAD_ON_DEMAND, NARC_INDEX_MSGDATA__PL_MSG, TEXT_BANK_COMMON_STRINGS, HEAP_ID_FIELD1);
+
+    if (saveResult == SAVE_RESULT_OK) {
+        StringTemplate *strTemplate = StringTemplate_Default(HEAP_ID_FIELD1);
+        StringTemplate_SetPlayerName(strTemplate, 0, SaveData_GetTrainerInfo(fieldSystem->saveData));
+        clearGameStruct->string = MessageUtil_ExpandedString(strTemplate, msgLoader, 16, HEAP_ID_FIELD1);
```
lhearachel: **polish**: `16` is a text bank entry ID; in this instance, it's `pl_msg_00000213_00016` from the `common_strings` bank. We should also update the `id` of the string in the bank.

src/clear_game.c
```diff
@@ -0,0 +1,316 @@
...
+        StringTemplate_Free(strTemplate);
+    } else {
+        clearGameStruct->string = MessageLoader_GetNewString(msgLoader, 18);
```
lhearachel: **polish**: Similar to the above, `18` is a text bank entry ID from `common_strings`.

src/clear_game.c
```diff
@@ -0,0 +1,316 @@
...
+    Bg_ClearTilesRange(BG_LAYER_MAIN_3, 0x20, 0, HEAP_ID_FIELD3);
+    Bg_FillTilemapRect(fieldSystem->bgConfig, 3, 0x0, 0, 0, 32, 32, 17);
+    Bg_CopyTilemapBufferToVRAM(fieldSystem->bgConfig, 3);
+}
+
+static void sub_02052FA8(FieldSystem *fieldSystem, ClearGameStruct *clearGameStruct)
+{
+    Options *options = SaveData_GetOptions(fieldSystem->saveData);
+
+    clearGameStruct->string = MessageBank_GetNewStringFromNARC(NARC_INDEX_MSGDATA__PL_MSG, TEXT_BANK_COMMON_STRINGS, 15, HEAP_ID_FIELD3);
```
lhearachel: **polish**: `15` is also a text bank entry from `common_strings`.

res/field/scripts/scripts_pokemon_league_elevator_to_aaron_room.s
Viperio19: I'm not completely sure about the details either, I just followed the terminology used in the init scripts (which is probably borrowed from `pokeemerald`). Here is an example of an init script:
```
    InitScriptEntry_OnTransition 1
    InitScriptEntry_OnFrameTable InitScriptFrameTable
    InitScriptEntryEnd

InitScriptFrameTable:
    InitScriptGoToIfEqual VAR_MAP_LOCAL_3, 0, 2
    InitScriptFrameTableEnd

    InitScriptEnd
```
Here I would suffix script 1 with `_OnTransition` and 2 with `_OnFrame`.

res/field/scripts/scripts_pokemon_league_hallway_to_hall_of_fame.s
Viperio19: I think this setup with using the already defined `id`s is quite clean, and it is robust for when object events get added, removed, or moved around. If you separately define local ids in the script and you were to remove an object event, you would also have to manually correct the local id (unless you were to define it like `#define LOCALID_CYNTHIA POKEMON_LEAGUE_HALLWAY_TO_HALL_OF_FAME_CYNTHIA` but then what's the point). I know that was already from a modding standpoint, but in general I don't see a reason for adding an additional layer of defining the `id`s.

I think the reason that `pokeemerald` uses that setup is because they don't have `id`s for their object events:
```json
    {
      "graphics_id": "OBJ_EVENT_GFX_WALLACE",
      "x": 9,
      "y": 13,
      "elevation": 3,
      "movement_type": "MOVEMENT_TYPE_FACE_UP",
      "movement_range_x": 1,
      "movement_range_y": 1,
      "trainer_sight_or_berry_tree_id": "0",
      "script": "CaveOfOrigin_B1F_EventScript_Wallace",
      "flag": "FLAG_HIDE_CAVE_OF_ORIGIN_B1F_WALLACE"
    }
```
Since there is no `id` in there, unlike for example `"id": "POKEMON_LEAGUE_HALLWAY_TO_HALL_OF_FAME_CYNTHIA",`, the `id`s need to be defined in the scripts, which is not the case here.

src/clear_game.c
Viperio19: `UnkStruct_0203E274` is used elsewhere so I can't just define the struct here, but namewise I'll name it as follows unless there is a better suggestion:
```diff
typedef struct ClearGameStruct {
    BOOL gameCompleted;
    HallOfFameDisplayData displayData;
-   UnkStruct_0203E274 unk_10;
+   ClearGamePlayerInfo playerInfo;
    Window window;
    String *string;
    void *dial;
    int printerID;
    int delay;
} ClearGameStruct;
```

res/field/scripts/scripts_pokemon_league_hallway_to_hall_of_fame.s
lhearachel: I think that I misphrased here.

Rephrasing: why do the `id` properties of the individual overworlds need a verbose prefix? Why not just use `LOCALID_CYNTHIA` in the events file's `id` property?

src/clear_game.c
lhearachel: Sounds good to me!

lhearachel (approved):

PR #872, opened 2025-12-29
Document twinleaf town scripts

Viperio19:
This PR documents all `scripts_twinleaf_town.*.s` files. There are a few vars and flags that appear in these files that I have not documented yet because they are more relevant in later scripts so I didn't fully look into what they do yet.

Additionally, `tools/json2bin/event.py` can now parse map header constants for the `"hidden_flag"` field for clone object events, like this:
```diff
        {
            "id": "TWINLEAF_TOWN_ARROW_SIGNPOST",
            "clone_id": 3,
            "graphics_id": "OBJ_EVENT_GFX_ARROW_SIGNPOST",
            "movement_type": "MOVEMENT_TYPE_NONE",
            "trainer_type": "TRAINER_TYPE_NONE",
-           "hidden_flag": "FLAG_UNK_0x0156",
+           "hidden_flag": "MAP_HEADER_ROUTE_201",
            "script": 65535,
            ...
            "y": 0
        }
```


res/field/events/events_twinleaf_town.json
```diff
@@ -41,11 +41,11 @@
             "y": 1
         },
         {
-            "id": "TWINLEAF_TOWN_RIVAL",
+            "id": "LOCALID_RIVAL",
```
lhearachel, line 44: **issue**: Use `BARRY` instead of `RIVAL`, similar to changes in #869.

```suggestion
            "id": "LOCALID_BARRY",
```

Viperio19: As I mentioned on discord, I personally prefer `RIVAL` over `BARRY` because he is never referred to as `BARRY`, and there are a ton of instances like `MAP_HEADER_TWINLEAF_TOWN_RIVAL_HOUSE_1F` and `FollowerPartners_Text_Rival_DoYouBelieveIt` where I don't think it makes sense to change `RIVAL` to `BARRY`. If there's better reasons for `BARRY` over `RIVAL` though, let me know and I'll change it.

PR #866, opened 2025-12-24
Generate Contest Effects and use generated values in move data files

CharlesFolz4:
Generate Contest Effects
Generate Contest Ranks
Document contest effect messages & a few associated functions
Consolidate duplicate contest constants:
-  `MAX_APPEAL_HEARTS`  and `MAX_APPEAL_PTS`
-  `CONTEST_EFFECT_PER_APPEAL_PT` and `POINTS_PER_APPEAL_HEART`

Convert move `data.json` files to use generated Contest Effect values instead of numbers for the "effect" field

PR #862, opened 2025-12-22
Add enum for friendship change events

CharlesFolz4:
Add enum for friendship change events

PR #874, opened 2026-01-01
Document link contest records

CharlesFolz4:
Documents the Link Contest Records and the display for those records in the contest registration hall
Screenshot_20251231_212817

Removed duplicate declarations for ScrCmd_0F8 through ScrCmd_116 in unk_02046C7C.c

~~This requires changes from currently unmerged PRs #862 and #866.~~ 
Required PRs have been merged, so this is now ready for review.


res/field/scripts/scripts_unk_0212.s
```diff
@@ -1319,10 +1319,10 @@ _1393:
...
     End
 
-_1399:
+ContestRegistration_LinkContestRecordsDisplay:
     PlayFanfare SEQ_SE_CONFIRM
     LockAll
-    ScrCmd_116
+    ShowLinkContestRecords
     ReleaseAll
     End
```
lhearachel, line 1327: **suggestion**: Unify the naming scheme; change the script label to `ContestRegistration_ShowLinkContestRecords`.

src/link_contest_records.c
```diff
@@ -0,0 +1,45 @@
...
+
+void LinkContestRecords_IncrementSavaData(SaveData *saveData, enum PokemonContestType contestType, int placement)
+{
+    LinkContestRecords *LinkContestRecords = SaveData_SaveTable(saveData, SAVE_TABLE_ENTRY_LINK_CONTEST_RECORDS);
+
+    if (LinkContestRecords->contestPlacement[contestType][placement] < 9999) {
+        LinkContestRecords->contestPlacement[contestType][placement]++;
+    }
+
+    SaveData_SetChecksum(SAVE_TABLE_ENTRY_LINK_CONTEST_RECORDS);
```
lhearachel: **issue**: The variable should be named `linkContestRecords`. (Another instance below in this file)

src/unk_02046C7C.c
```diff
@@ -413,11 +380,11 @@ BOOL ScrCmd_115(ScriptContext *param0)
...
     void **v0 = FieldSystem_GetScriptMemberPtr(param0->fieldSystem, SCRIPT_MANAGER_DATA_PTR);
     u16 *v1 = ScriptContext_GetVarPointer(param0);
-    int v2, v3, v4, v5, v6;
+    int v2, isLinkContest, v4, v5, v6;
 
-    sub_02094BB4(*v0, &v2, &v3, &v4, &v5, &v6);
+    sub_02094BB4(*v0, &v2, &isLinkContest, &v4, &v5, &v6);
 
-    if ((v3 == 1) || (v4 == 1) || (v5 == 1) || (v6 == 1)) {
+    if ((isLinkContest == TRUE) || (v4 == 1) || (v5 == 1) || (v6 == 1)) {
```
lhearachel: **polish**: Unnecessary parens.

src/overlay006/link_contest_record_display.c
```diff
@@ -0,0 +1,236 @@
...
+    Window_DrawStandardFrame(&linkContestRecord->window, TRUE, 1, PLTT_11);
+    Window_FillTilemap(&linkContestRecord->window, 15);
+
+    String *linkContestRecordString = String_Init(100, HEAP_ID_FIELD2);
+    String *message = MessageLoader_GetNewString(linkContestRecord->messageLoader, LinkContestRecord_Text_PlayerName);
+
+    StringTemplate_SetPlayerName(linkContestRecord->stringTemplate, 0, SaveData_GetTrainerInfo(saveData));
+    StringTemplate_Format(linkContestRecord->stringTemplate, linkContestRecordString, message);
+
+    int textX = CalcAlignedTextX(linkContestRecordString, FONT_SYSTEM, TEXT_ALIGN_CENTER, (15 * 8));
```
lhearachel: **question**: Is `90` (15 x 8) an important value here? I'm curious if there's a semantic definition we could use, if it refers to a column of text.

src/overlay006/link_contest_record_display.c
```diff
@@ -0,0 +1,236 @@
...
+
+    Text_AddPrinterWithParams(&linkContestRecord->window, FONT_SYSTEM, linkContestRecordString, textX, (8 * 0), TEXT_SPEED_NO_TRANSFER, NULL);
+    String_Free(message);
+
+    message = MessageLoader_GetNewString(linkContestRecord->messageLoader, LinkContestRecord_Text_LinkContestRecord);
+    textX = CalcAlignedTextX(message, FONT_SYSTEM, TEXT_ALIGN_CENTER, (15 * 8));
+
+    Text_AddPrinterWithParams(&linkContestRecord->window, FONT_SYSTEM, message, textX, (8 * 2), TEXT_SPEED_NO_TRANSFER, NULL);
```
lhearachel: **suggestion**: Replace `8 * 2` with `16`.

res/field/scripts/scripts_unk_0212.s
CharlesFolz4: I think there may be a misunderstanding from an ambiguous naming choice.  The intention with this name was for it to be similar to the other entry scripts, the same way as `ContestRegistration_LinkContestReceptionist` would be, with "Display" as a noun not as a verb, meaning the physical display shown on the map that is interacted with to trigger this script.
Screenshot_20260102_203449
Should "Display" be changed to "Monitor" here instead, to make that more clear that it's an entry point triggered from the bg_event, or should this label match the action of the called command?

src/overlay006/link_contest_record_display.c
CharlesFolz4: I had initially wanted to preserve the 8 *, because that 8 value seemed to be related to text or font cell sizes in some way, but I couldn't track down any relevant constants.  On closer examination, I'm now pretty sure that these values are meant to align the text with the background tilemap grid, but given that it's a featureless white background, I don't think that's particularly meaningful anymore.

res/field/scripts/scripts_unk_0212.s
lhearachel: Ah! Yes, I think that something like `Monitor` or `Screen` would better communicate the intent here. I would generally think that scripts which are used by events should be more noun-ish than verb-ish, if that makes sense. I'm open to other thoughts here, though.

lhearachel (approved):

PR #878, opened 2026-01-02
Document Battle Party Sprites

Morglom:
Most of the pull request is focused on battle_party_sprites but I also cleaned up some stuff I saw in related sections.

### DIFF
--- a/include/overlay013/battle_bag.h
+++ b/include/overlay013/battle_bag.h
@@ -82,7 +82,7 @@ typedef struct BattleBagContext {
     u8 pocketCurrentPages[BATTLE_POCKET_MAX];
 } BattleBagContext;
 
-#define NUM_BATTLE_BAG_SPRITES 6
+#define NUM_BATTLE_BAG_SPRITES BATTLE_POCKET_ITEMS_PER_PAGE
 
 typedef struct BattleBag {
     BattleBagContext *context;


include/overlay013/battle_sub_menu_sprites_defs.h
lhearachel, line 1: **thought**: This file seems nice to have, but these also feel pretty generic. Should they be exposed at a higher level somewhere? (I'm alright with tabling this for now, as long as we address it eventually)

src/overlay013/battle_party_sprites.c
```diff
@@ -0,0 +1,1342 @@
...
+#define SPRITE_MANAGER_CELL_RESOURCE_CAPACITY       6
+#define SPRITE_MANAGER_ANIM_RESOURCE_CAPACITY       6
+#define SPRITE_MANAGER_MULTI_CELL_RESOURCE_CAPACITY 0
+#define SPRITE_MANAGER_MULTI_ANIM_RESOURCE_CAPACITY 0
+#define SPRITE_MANAGER_MAX_SPRITES                  43
+
+#define HELD_ITEMS_CHAR_RESOURCE_ID 45078
+#define HELD_ITEMS_PLTT_RESOURCE_ID 45066
+#define HELD_ITEMS_CELL_RESOURCE_ID 45066
+#define HELD_ITEMS_ANIM_RESOURCE_ID 45066
```
lhearachel, line 34: **thought**: We may want to start collating these resource IDs into a collective enumeration of some sort. Probably out-of-scope for this PR.

lhearachel (approved):
Changes look good to me. 👍  Mostly some leading thoughts below. Feel free to mark as out-of-scope for this PR.

include/overlay013/battle_sub_menu_sprites_defs.h
Morglom: There were other sections that seemed to resemble the setup with "SpriteResourcesAndPriority" such as the "resources" field inside the "SpriteTemplate" struct in "sprite_system.h". So I think there's a good chance it could be repurposed into something more generic.

src/overlay013/battle_party_sprites.c
Morglom: That does sound like a good idea, though I suppose it'd need to account for the same resource being needed in several places.

lhearachel: > I suppose it'd need to account for the same resource being needed in several places.

Could you elaborate? Do you mean resources that overlap ID values? Or something else?

Morglom: I was just thinking aloud about for example how you'd categorize "45065" if it's being used for pokemon and move type and move class sprites. Nothing of concern or anything, again just thinking about it.

lhearachel: Right, it will require some finesse. We can keep a simple solution for now and think about something more robust as future scoping. 😁

PR #879, opened 2026-01-02
Convert multiple `int heapID` to `enum HeapID heapID`

AsparagusEduardo:
There's still 42 instances left of `int heapID`, but those require more tinkering to make them match.

~~Also, since allocating memory is such an universal thing across the repo, I added `include/constants/heap.h` to `include/pch/global_pch.h` and removed the individual calls to it.~~
EDIT: Reverted this.


AsparagusEduardo:
I was able to reduce the instances of `int heapID` to 13.

PR #889, opened 2026-01-09
document underground vendors

scbroede:


include/constants/traps.h
```diff
@@ -35,5 +35,6 @@
 #define TRAP_FIRE          31
 #define TRAP_RADAR_DUMMY   32
 #define TRAP_DIGGER_DRILL  33
+#define TRAP_MAX           34
```
lhearachel: **thought**: Probably worth turning these into `enum`s now?

src/overlay023/trap_prices.c
```diff
@@ -0,0 +1,61 @@
+#include "overlay023/trap_prices.h"
+
+#include <nitro.h>
+#include <string.h>
+
+#include "constants/traps.h"
+
+#include "overlay023/underground_defs.h"
+
+static const SpherePrice sTrapPrices[] = {
```
lhearachel, line 11: **thought**: This is likely out-of-scope, but I wonder if we can set up some guard-rails to make sure adding a trap / treasure / whatever causes a compilation failure if there isn't an entry in these tables.

include/constants/traps.h
scbroede: done. this leaves a bunch of int parameters i should probably change to use the new enums, but i'll take care of that in a future pr

src/overlay023/trap_prices.c
scbroede: good idea for the future, left a todo comment as a reminder

PR #892, opened 2026-01-09
Standarized Barry as Rival

AsparagusEduardo:
Here's the only instances I left of Barry:
image


generated/object_events.txt
```diff
@@ -146,7 +146,7 @@ OBJ_EVENT_GFX_BUCK
 OBJ_EVENT_GFX_MIRA
 OBJ_EVENT_GFX_DUMMY_146
 OBJ_EVENT_GFX_DUMMY_147
-OBJ_EVENT_GFX_BARRY
+OBJ_EVENT_GFX_RIVAL
```
lhearachel: **thought**: I would say these should actually be `BARRY`, because they refer to Barry, the character, graphically.

AsparagusEduardo: > **thought**: I would say these should actually be `BARRY`, because they refer to Barry, the character, graphically.

Makes sense. Will revert

AsparagusEduardo:
Reverted `OBJ_EVENT_GFX_RIVAL` 👍

PR #898, opened 2026-01-15
Document Poffin feeding cutscene

VicSevenT:
Documenting the rest of overlay 79, which contains the cutscene where a poffin is fed to a pokemon. The cutscene follows a pretty simple execution flow, with the biggest complexity coming from a series of tasks used to animate the poffin's and pokemon's movements.


res/graphics/poffin_case/feeding_cutscene/bottom_screen_tiles.png
lhearachel, line 1: **question**: Should this be named with `_tiles` as a suffix (before the extension) to denote that it's a tilemap input? Ditto for the top screen.

src/applications/poffin_case/cutscene.c
```diff
@@ -0,0 +1,1155 @@
...
+
+    NARC_dtor(narc);
+
+    Bg_ScheduleTilemapTransfer(app->bgConfig, BG_LAYER_MAIN_2);
+    Bg_ScheduleTilemapTransfer(app->bgConfig, BG_LAYER_SUB_0);
+}
+
+static void InitWindow(PoffinCutscene *app)
+{
+    static const WindowTemplate winTemplate = {
```
lhearachel, line 470: **polish**: Use designated initializers for these constructions.

src/applications/poffin_case/cutscene.c
```diff
@@ -0,0 +1,1155 @@
...
+    man->poffinSprite = app->poffinSprite->sprite->sprite;
+    man->monSprite = app->monSprite.sprite;
+    man->poffinPreference = app->mon.poffinPreference;
+    man->unused = 24;
+    man->poffinSpriteState.startPos.x = FX32_CONST(128);
+    man->poffinSpriteState.startPos.y = FX32_CONST(192 + 32);
+    man->poffinSpriteState.endPos.x = FX32_CONST(128);
+    man->poffinSpriteState.endPos.y = FX32_CONST(96);
+    man->poffinSpriteState.startSize.x = 0x1000;
+    man->poffinSpriteState.startSize.y = 0x1000;
```
lhearachel: **question**: Are these related to the `MON_SPRITE_*_SIZE` constants at all?

res/graphics/poffin_case/feeding_cutscene/bottom_screen_tiles.png
VicSevenT: Can do

src/applications/poffin_case/cutscene.c
VicSevenT: Not exactly? The `MON_SPRITE` constants define the pokemon sprite's size as it moves towards the screen to receive the poffin, and then moves away after getting it. The poffin sprites values for correspondingly for the poffin, which is only present in the first half of the animation (it only grows, then disappears, and doesn't shrink). The exact values of the size parameters for the two sprites are unrelated AFAIK, they're just set to something that makes the animation look meaningful. I think I initally ignored using constants for the poffin's size because those are less used that the mon sprite's param, but I just threw some in there now to hopefully clarify.

PR #904, opened 2026-01-17
Document route 201 and lake verity scripts, misc documentation

Viperio19:
This PR documents the route 201, verity lakefront, and lake verity (normal and low water) script files. There are also related functions and constants that I documented as well.

### DIFF
--- a/asm/macros/scrcmd.inc
+++ b/asm/macros/scrcmd.inc
@@ -721,7 +721,7 @@
     .short \z
     .endm
 
-    .macro SetObjectFlagPersistent localID, flag
+    .macro SetObjectFlagIsPersistent localID, flag
     .short 108
     .short \localID
     .byte \flag
--- a/res/field/scripts/scripts_route_201.s
+++ b/res/field/scripts/scripts_route_201.s
@@ -777,7 +777,7 @@ Route201_ResetRivalPartnerEast:
 Route201_ResetRivalPartner:
     SetHasPartner
     SetMovementType LOCALID_RIVAL, MOVEMENT_TYPE_FOLLOW_PLAYER
-    SetObjectFlagPersistent LOCALID_RIVAL, TRUE
+    SetObjectFlagIsPersistent LOCALID_RIVAL, TRUE
     Return
 
     .balign 4, 0
@@ -1036,7 +1036,7 @@ _0CF1:
     RemoveObject LOCALID_COUNTERPART
     RemoveObject LOCALID_PROF_ROWAN
     SetVar VAR_FOLLOWER_RIVAL_STATE, 3
-    SetObjectFlagPersistent 254, FALSE
+    SetObjectFlagIsPersistent 254, FALSE
     ClearHasPartner
     SetFlag FLAG_HIDE_ROUTE_201_RIVAL
     SetFlag FLAG_UNK_0x0195
@@ -1256,7 +1256,7 @@ Route201_SetRivalPartner:
     SetStepFlag
     SetHasPartner
     SetMovementType LOCALID_RIVAL, MOVEMENT_TYPE_FOLLOW_PLAYER
-    SetObjectFlagPersistent LOCALID_RIVAL, TRUE
+    SetObjectFlagIsPersistent LOCALID_RIVAL, TRUE
     SetFlag FLAG_HIDE_ROUTE_201_RIVAL
     ReleaseAll
     End
--- a/src/scrcmd.c
+++ b/src/scrcmd.c
@@ -352,7 +352,7 @@ static BOOL ScrCmd_GetPlayerMapPos(ScriptContext *ctx);
 static BOOL ScrCmd_Unused_06A(ScriptContext *ctx);
 static BOOL ScrCmd_GetPlayerDir(ScriptContext *ctx);
 static BOOL ScrCmd_MoveCamera(ScriptContext *ctx);
-static BOOL ScrCmd_SetObjectFlagPersistent(ScriptContext *ctx);
+static BOOL ScrCmd_SetObjectFlagIsPersistent(ScriptContext *ctx);
 static BOOL ScrCmd_SetMovementType(ScriptContext *ctx);
 static BOOL ScrCmd_GetMovementType(ScriptContext *ctx);
 static BOOL ScrCmd_Unused_06E(ScriptContext *ctx);
@@ -876,7 +876,7 @@ const ScrCmdFunc Unk_020EAC58[] = {
     ScrCmd_GetPlayerMapPos,
     ScrCmd_Unused_06A,
     ScrCmd_MoveCamera,
-    ScrCmd_SetObjectFlagPersistent,
+    ScrCmd_SetObjectFlagIsPersistent,
     ScrCmd_SetMovementType,
     ScrCmd_Unused_06E,
     ScrCmd_GiveMoney,
@@ -3344,7 +3344,7 @@ static BOOL ScrCmd_MoveCamera(ScriptContext *ctx)
     return FALSE;
 }
 
-static BOOL ScrCmd_SetObjectFlagPersistent(ScriptContext *ctx)
+static BOOL ScrCmd_SetObjectFlagIsPersistent(ScriptContext *ctx)
 {
     MapObject *mapObj = MapObjMan_LocalMapObjByIndex(ctx->fieldSystem->mapObjMan, ScriptContext_GetVar(ctx));
     u8 flag = ScriptContext_ReadByte(ctx);
--- a/res/field/scripts/scripts_eterna_forest.s
+++ b/res/field/scripts/scripts_eterna_forest.s
@@ -57,7 +57,7 @@ _008E:
     SetVar VAR_PARTNER_TRAINER_ID, TRAINER_CHERYL_ETERNA_FOREST
     SetHasPartner
     SetMovementType 0, 48
-    SetObjectFlagPersistent 0, TRUE
+    SetObjectFlagIsPersistent 0, TRUE
     ReleaseAll
     End
 
@@ -89,7 +89,7 @@ _0108:
     LockAll
     ClearHasPartner
     SetMovementType 0, 15
-    SetObjectFlagPersistent 0, FALSE
+    SetObjectFlagIsPersistent 0, FALSE
     ApplyMovement 0, _0190
     WaitMovement
     Message 4


include/map_object.h
```diff
@@ -72,7 +72,7 @@ void MapObjectMan_SaveAll(FieldSystem *fieldSystem, const MapObjectManager *mapO
 void MapObjectMan_LoadAllObjects(const MapObjectManager *mapObjMan, MapObjectSave *mapObjSave, int size);
 void sub_02062068(const MapObjectManager *mapObjMan, int param1, int param2, const ObjectEvent *objectEvent);
 MapObject *MapObjMan_LocalMapObjByIndex(const MapObjectManager *mapObjMan, int index);
-MapObject *sub_02062570(const MapObjectManager *mapObjMan, int movementType);
+MapObject *MapObjMan_GetLocalMapObjByMovementType(const MapObjectManager *mapObjMan, int movementType);
```
Gudf, line 78: thought: it looks like we have a mix of `MapObjectMan` and `MapObjMan`, we should probably choose one of these and stick to it

src/scrcmd.c
```diff
@@ -3344,21 +3344,21 @@ static BOOL ScrCmd_MoveCamera(ScriptContext *ctx)
...
-    u8 v1 = ScriptContext_ReadByte(ctx);
+    u8 flag = ScriptContext_ReadByte(ctx);
 
-    sub_02062E5C(mapObj, v1);
+    MapObject_SetFlagPersistent(mapObj, flag);
     return FALSE;
 }
 
-static BOOL ScrCmd_06D(ScriptContext *ctx)
+static BOOL ScrCmd_SetMoveCode(ScriptContext *ctx)
```
Gudf: **suggestion**: Since the constants are called `MOVEMENT_TYPE_[...]`, this should be named accordingly. 
I'm not sure what to do with `MapObject_SetMoveCode`, since we probably want to keep `MapObject_SetMovementType` for the straight `MapObject::movementType` getter? But `MapObject_SetMovementType` is also not used outside of `map_object.c` afaict so it's probably safe to stop exposing it, rename it `MapObject_SetMovementTypeInternal` or smth and rename `MapObject_SetMoveCode` to `MapObject_SetMovementType` (or something like `MapObject_SwitchMovementType` if you feel like it's doing too much for a setter)

src/map_object.c
```diff
@@ -1851,12 +1851,12 @@ int MapObject_IsHeightCalculationDisabled(const MapObject *mapObj)
     return FALSE;
 }
 
-void sub_02062E5C(MapObject *mapObj, int param1)
+void MapObject_SetFlagPersistent(MapObject *mapObj, BOOL flag)
```
Gudf: **suggestion**: `SetFlagPersistent` sounds to me like you're setting a flag such that it won't get reset, instead of setting a flag called `persistent`. I would probably name it `SetPersistent`(/`SetIsPersistent`)?

res/field/events/events_lake_verity_low_water.json
```diff
@@ -32,11 +32,11 @@
             "y": 0
         },
         {
-            "id": "LAKE_VERITY_LOW_WATER_STARLY_2",
+            "id": "LOCALID_STARLY_2",
```
Gudf: **suggestion**: Now that the IDs don't have their local ID appended systematically, having a `STARLY_2` and a `STARLY_3` with no `STARLY_1` feels unintuitive. It would probably be best to rename those based on their position? i.e. this would be `LOCALID_STARLY_EAST`, the other one would be `LOCALID_STARLY_SOUTH`

res/field/events/events_lake_verity.json
```diff
@@ -130,11 +130,11 @@
             "y": 0
         },
         {
-            "id": "LAKE_VERITY_GRUNT_M_8",
+            "id": "LOCALID_GRUNT_M_8",
```
Gudf: **suggestion**: Similarly here, there aren't 8 grunts at Lake Verity, this should probably be `LOCALID_GALACTIC_GRUNT_5` (or `LOCALID_GALACTIC_GRUNT_M_3` and rename the other grunts' IDs to also include the gender)

res/field/events/events_lake_verity.json
```diff
@@ -70,11 +70,11 @@
             "y": 0
         },
         {
-            "id": "LAKE_VERITY_POKEBALL_4",
+            "id": "LOCALID_POKEBALL_4",
```
Gudf: **suggestion**: there's only one pokeball at Lake Verity:
```suggestion
            "id": "LOCALID_POKEBALL",
```

asm/macros/scrcmd.inc
```diff
@@ -4966,6 +4966,14 @@
     GoToIf 5, \offset
     .endm
 
+   .macro GoToIfInRange varID, lower, upper, offset
+    CompareVar \varID, \lower
+    GoToIf 1, \offset
```
Gudf, line 5017: **thought**: we should probably have constants for `GoToIf`'s first argument

res/field/scripts/scripts_lake_verity.s
```diff
@@ -1,233 +1,234 @@
...
     GetPlayerGender VAR_MAP_LOCAL_0
-    GoToIfEq VAR_MAP_LOCAL_0, GENDER_MALE, _0061
-    GoToIfEq VAR_MAP_LOCAL_0, GENDER_FEMALE, _0069
+    GoToIfEq VAR_MAP_LOCAL_0, GENDER_MALE, LakeVerity_SetCounterpartGraphicsDawn
+    GoToIfEq VAR_MAP_LOCAL_0, GENDER_FEMALE, LakeVerity_SetCounterpartGraphicsLucas
     End
 
-_0061:
+LakeVerity_SetCounterpartGraphicsDawn:
     SetVar VAR_OBJ_GFX_ID_0, 97
```
Gudf: **question**: Do we have constants for this `97` and Lucas's `0` below?

res/field/scripts/scripts_lake_verity.s
```diff
@@ -1,233 +1,234 @@
...
+    SetObjectEventDir LOCALID_PROF_ROWAN, DIR_WEST
+    SetObjectEventPos LOCALID_COUNTERPART, 50, 39
+    SetObjectEventMovementType LOCALID_COUNTERPART, MOVEMENT_TYPE_LOOK_WEST
+    SetObjectEventDir LOCALID_COUNTERPART, DIR_WEST
     Return
 
-_00C7:
+LakeVerity_OnLoad:
     End
     End
```
Gudf, line 57: **suggestion**: For the sake of semantics, I'd split the second `End` into it's own script:
```suggestion

LakeVerity_Unused:
    End
```

res/field/events/events_lake_verity.json
```diff
@@ -85,9 +85,9 @@
             "y": 0
         },
         {
-            "id": "LAKE_VERITY_PROF_ROWAN_5",
+            "id": "LOCALID_PROF_ROWAN",
             "graphics_id": "OBJ_EVENT_GFX_PROF_ROWAN",
-            "movement_type": "MOVEMENT_TYPE_LOOK_RIGHT",
+            "movement_type": "MOVEMENT_TYPE_LOOK_EAST",
             "trainer_type": "TRAINER_TYPE_NONE",
             "hidden_flag": "FLAG_UNK_0x028C",
```
Gudf: **question**: Can we not document this flag now?

res/field/events/events_lake_verity.json
```diff
@@ -100,9 +100,9 @@
             "y": 0
         },
         {
-            "id": "LAKE_VERITY_COUNTERPART_6",
+            "id": "LOCALID_COUNTERPART",
             "graphics_id": "OBJ_EVENT_GFX_COUNTERPART",
-            "movement_type": "MOVEMENT_TYPE_LOOK_RIGHT",
+            "movement_type": "MOVEMENT_TYPE_LOOK_EAST",
             "trainer_type": "TRAINER_TYPE_NONE",
             "hidden_flag": "FLAG_UNK_0x028D",
```
Gudf: **question**: Can we not document this flag now?

res/field/scripts/scripts_route_201.s
```diff
@@ -580,338 +580,335 @@ _07A8:
...
 
     .balign 4, 0
-_07F0:
+Route201_Movement_ProfRowanFacPlayer:
     WalkOnSpotNormalEast
     EndMovement
 
     .balign 4, 0
-_07F8:
+Route201_Movement_ProfRowanFacePlayer:
```
Gudf: **typo**: Missing the `e` in `Face`. Though that'll make both labels have the same name, I'd rename them based on the player's direction relative to Prof. Rowan:
```suggestion
Route201_Movement_ProfRowanFacePlayerEast:
    WalkOnSpotNormalEast
    EndMovement

    .balign 4, 0
Route201_Movement_ProfRowanFacePlayerSouth:
```

src/scrcmd.c
Viperio19: I definitely agree with `ScrCmd_SetMovementType`. For `MapObject_SetMoveCode`, I think `MapObject_SwitchMovementType` makes more sense.

src/map_object.c
Viperio19: That makes sense, although personally I'd prefer `MapObject_SetFlagIsPersistent` to keep it in line with the other functions that set status flags

res/field/events/events_lake_verity.json
Viperio19: This a slightly different case compared to the starly one, since here some grunts are linked to their trainer constants:
```json
            "id": "LOCALID_GALACTIC_GRUNT_3",
            ...
            "script": "TRAINER_GALACTIC_GRUNT_LAKE_VERITY_3",
```
For example, `LOCALID_GALACTIC_GRUNT_3`'s `id` is 2. `LOCALID_GRUNT_M_8` is instead tied to the `graphicsid` `OBJ_EVENT_GFX_GRUNT_M`, and I'm in a way also using the gender to indicate that it's not a trainer, but this is definitely confusing and I would be interested to hear other potential solutions. I'm against `LOCALID_GALACTIC_GRUNT_5` because how it is done right now would indicate to me that it's the fifth trainer here. Maybe a solution would be to just make the `id`s separate from the trainer constants, so you'd have `LOCALID_GALACTIC_GRUNT_2` with script `TRAINER_GALACTIC_GRUNT_LAKE_VERITY_3`, but at least there's no confusion within the `id`s.

res/field/scripts/scripts_lake_verity.s
Viperio19: Ah yeah I made a separate PR to allow using `OBJ_EVENT_GFX_[...]`, I'll merge main into this branch and then use those constants here.

src/map_object.c
Gudf: Sure

res/field/events/events_lake_verity.json
Gudf: Ah, I see. I'd still get rid of the `8` but otherwise it's probably fine to keep as-is in that case?

Viperio19: Yeah, I'll keep `LOCALID_GALACTIC_GRUNT_[X]` for trainers and if in other scripts there are multiple grunts that aren't trainers I can name them `LOCALID_GRUNT_[M/F]_[X]` or something related to positions like the starly

src/scrcmd.c
```diff
@@ -3344,21 +3344,21 @@ static BOOL ScrCmd_MoveCamera(ScriptContext *ctx)
     return FALSE;
 }
 
-static BOOL ScrCmd_06C(ScriptContext *ctx)
+static BOOL ScrCmd_SetObjectFlagPersistent(ScriptContext *ctx)
```
Gudf: **suggestion**: to be coherent with `MapObject_SetFlagIsPersistent` (and for the same reasons I objected about `SetFlagPersistent` in the first place, really). Other than that I'm good to merge.
```suggestion
static BOOL ScrCmd_SetObjectFlagIsPersistent(ScriptContext *ctx)
```

Viperio19: good call, done!

PR #907, opened 2026-01-18
Fix Dawn/Lucas Trainer Identifiers

ljtpetersen:


scbroede:
these are inconsistent; for jubilife you're using the convention where "lucas/dawn turtwig" is the version where lucas/dawn has a turtwig, but in veilstone "lucas/dawn turtwig" means the version where the player started with turtwig.

i don't really care which version we use but it should be consistent (and the rival battles should also use the same naming scheme)

lhearachel:
> i don't really care which version we use but it should be consistent (and the rival battles should also use the same naming scheme)

I would favor having the marker reference the player's starter, personally (the rival teams should already do this). But I don't mind either way.

ljtpetersen:
> > i don't really care which version we use but it should be consistent (and the rival battles should also use the same naming scheme)
> 
> I would favor having the marker reference the player's starter, personally (the rival teams should already do this). But I don't mind either way.

I have changed the names so they reference the player's starter.

PR #909, opened 2026-01-23
document underground top screen

scbroede:
- eliminating the decl for `UndergroundTopScreenContext` and defining it in the same file as `FieldSystem` is possible if the type of `parent` is changed to `void *` but i'm not sure if it's worth it

- needed to make the `-noSkip` flag in nitrogfx available when converting png to ncgr because the cell file for map markers does some absolutely ridiculous nonsense where it uses oams for future cells ahead of time and then completely overwrites them anyway. will port the change upstream if it's accepted here


src/overlay023/underground_spheres.c
```diff
@@ -559,25 +559,25 @@ void SphereRadar_Exit(void)
...
 }
 
-int SphereRadar_GetZCoordOfBuriedSphere(int param0)
+int SphereRadar_GetZCoordOfBuriedSphere(int radarIndex)
 {
     if (buriedSpheresEnv && buriedSpheresEnv->sphereRadarTimer) {
         // bug: only the first 66 buried spheres can show up on the radar
         int index = buriedSpheresEnv->sphereRadarTimer->timer / 2;
-        index = (index + param0) % 100;
+        index = (index + radarIndex) % 100;
```
ravepossum: question: I'm not sure I understand the bug here. Is that why you didn't use `MAX_BURIED_SPHERES` here or was that just an oversight?

include/overlay023/ov23_02241F74.h
```diff
@@ -74,16 +92,16 @@ void ov23_02242C78(void);
 void ov23_02242CB4(void);
 void ov23_02242D08(void);
 void ov23_02242D44(FieldSystem *fieldSystem);
-BOOL ov23_02242D60(String *param0);
+BOOL CommManUnderground_GetQueuedMessage(String *param0);
```
ravepossum, line 95: polish: Parameter names in declaration should match the ones in the definition. Same with the functions below.

platinum.us/filesys.csv
```diff
@@ -131,7 +131,7 @@ res/prebuilt/data/uground_pltt.resdat,/data/uground_pltt.resdat
 res/prebuilt/data/uground_pltt2.resdat,/data/uground_pltt2.resdat
 res/prebuilt/data/ugroundeffect.naix,/data/ugroundeffect.naix
 res/prebuilt/data/ugroundeffect.narc,/data/ugroundeffect.narc
-res/prebuilt/data/underg_radar.narc,/data/underg_radar.narc
+res/graphics/underground_top_screen/underground_top_screen.narc,/data/underg_radar.narc
```
ravepossum, line 134: praise: Readable NARC names my beloved. We should all be doing this now that we can

res/graphics/underground_top_screen/underground_top_screen.order
```diff
@@ -0,0 +1,7 @@
+background_tiles.NCLR
+background_tiles.NCGR
+background.NSCR
+map_markers.NCLR
+map_markers.NCGR
+map_markers_cell.NCER
+map_markers_anim.NANR
```
ravepossum: nitpick: Missing EOF newline

src/overlay023/underground_top_screen.c
```diff
@@ -0,0 +1,612 @@
...
+    switch (ctx->state) {
+    case TOP_SCREEN_STATE_INIT:
+        UndergroundTopScreen_InitBackgrounds(bgConfig, &ctx->window);
+        UndergroundMap_InitPlayerPositions(ctx->playerCoords, ctx->playerMarkers);
+        MessageQueue_Init(&ctx->messageQueue);
+        ctx->state++;
+        break;
+    case TOP_SCREEN_STATE_INIT_SPRITES:
+        UndergroundMap_InitSpriteResources(ctx);
+        SpriteResourcesHeader_Init(&ctx->resourceData, 1000, 1000, 1000, 1000, 0xffffffff, 0xffffffff, FALSE, 0, ctx->spriteResourceCollection[SPRITE_RESOURCE_CHAR], ctx->spriteResourceCollection[SPRITE_RESOURCE_PLTT], ctx->spriteResourceCollection[SPRITE_RESOURCE_CELL], ctx->spriteResourceCollection[SPRITE_RESOURCE_ANIM], NULL, NULL);
```
ravepossum: polish: `0xffffffff` -> `-1`
suggestion: It also might not hurt to split this invocation across multiple lines.
suggestion: We may want to add constants for the resource IDs.

src/overlay023/underground_top_screen.c
```diff
@@ -0,0 +1,612 @@
...
+        break;
+    case MESSAGE_STATE_WAIT_FOR_PRINTER:
+        if (FieldMessage_FinishedPrinting(*printerID)) {
+            *state = MESSAGE_STATE_WAIT_FOR_NEW_MESSAGE;
+            *timer = 0;
+        }
+        break;
+    case MESSAGE_STATE_WAIT_FOR_NEW_MESSAGE:
+        if (!MessageQueue_IsEmpty(queue)) {
+            Window_Scroll(window, SCROLL_DIRECTION_UP, 2, 0x0);
```
ravepossum: nitpick: `0x0` -> `0`

src/overlay023/underground_top_screen.c
```diff
@@ -0,0 +1,612 @@
...
+    for (int i = 0; i < MAX_QUEUE_SIZE; i++) {
+        queue->queuedLines[i] = String_Init(50 * 2, HEAP_ID_FIELD1);
+    }
+
+    queue->endIndex = 0;
+    queue->currentIndex = 0;
+    queue->tempString = String_Init(50 * 2, HEAP_ID_FIELD1);
+
+    for (int i = 0; i < (int)NELEMS(queue->tempStringLines); i++) {
+        queue->tempStringLines[i] = String_Init(80, HEAP_ID_FIELD1);
```
ravepossum: suggestion: Wouldn't hurt to make a `TEMP_STRING_LINE_SIZE` constant or something since it's at least used here, in the struct, and in `MessageQueue_SplitMessage` below.

src/overlay023/underground_top_screen.c
```diff
@@ -0,0 +1,612 @@
...
+
+            Sprite_SetPosition(sprites[netID], &pos);
+            Sprite_SetDrawFlag(sprites[netID], TRUE);
+        } else {
+            Sprite_SetDrawFlag(sprites[netID], FALSE);
+        }
+    }
+}
+
+static void UndergroundMap_GetRadarAndBaseInfo(OtherMarkerData markerData[])
```
ravepossum: question: Would this make more sense called `Set` rather than `Get`?

src/overlay023/underground_spheres.c
scbroede: the timer caps at 100 (see `SphereRadar_TimerSysTask`), and since it's halved here when calculating the index that means the index can't go above 50 + 15 (the max value of `radarIndex`)

i don't know why i didn't use `MAX_BURIED_SPHERES` here lol, it really should be that

include/overlay023/ov23_02241F74.h
scbroede: yeah, my bad. i noticed that while documenting this file in another branch and it'll be part of that pr to avoid conflicts

src/overlay023/underground_top_screen.c
scbroede: my thought process was that it's getting that data and putting it in `markerData` for future use, though admittedly the name does make it sound like a function that returns that data. thinking `Retrieve` instead of `Get` is probably better (or maybe the parameter should be called `out` instead of `markerData` or something like that)

scbroede:
updated. please squash if it's ready for merge (or if you'd rather let someone more familiar with nitrogfx take a look, that's fine too)

PR #910, opened 2026-01-23
Document distortion world platform and obstacle ghost props

Kuruyia:
This continues documentation of the distortion world overlay by starting documenting the runtime side of ghost props. This focuses on the two big families of ghost props

- platform props, which are platforms the player can jump on (they are different from what's been previously documented as floating platforms in that they do not necessarily rotate the camera and player sprite; only two floating platforms are also part of the ghost prop infrastructure)
- obstacle props, which are the rocks and flowers that appear and disappear to block and unblock passage for the player

This also documents parts of the code that manages the 3D model and animation of ov9-managed map props.


include/map_object_move.h
```diff
@@ -32,7 +32,7 @@ void MapObject_MovePosInDir(MapObject *mapObj, int dir, fx32 distance);
 int MapObject_RecalculateObjectHeight(MapObject *mapObj);
 int MapObject_SetTileBehaviors(MapObject *mapObj);
 void VecFx32_StepDirection(int param0, VecFx32 *vec, fx32 val);
-void sub_02064450(int x, int z, VecFx32 *pos);
+void VecFx32_FromMapObjectPos(int x, int z, VecFx32 *pos);
```
ravepossum: question: Would the last parameter be clearer as something like `outVec`? Just to differentiate from the map pos and to make it clearer it's assigned to. And would the function name be clearer as something like `VecFx32_SetPosFromMapCoords()`?

src/overlay009/ov9_02249960.c
```diff
@@ -92,6 +93,80 @@
...
+#define GHOST_PROP_RENDERER_COUNT 34
+#define GHOST_PROP_OPACITY_MIN    0
+#define GHOST_PROP_OPACITY_MAX    31
+
+#define PLATFORM_PROP_ANIM_TIMING_COUNT 8
+#define PLATFORM_PROP_ANIM_DELTA        0x800
+
+#define OBSTACLE_PROP_ANIM_DELTA 2
+
+enum FloatingPlatformKind {
```
ravepossum, line 106: question: Was there a particular reason for using `Kind` throughout here? This is kind of a minor quibble, but we tend to use `Type`. Additionally, do we need to use `KIND`/`TYPE` in the constants themselves? I'm not sure `FLOATING_PLATFORM_KIND_FLOOR` is more descriptive than `FLOATING_PLATFORM_FLOOR`.

Edit: I see now that these enums were just moved, don't consider this blocking at all, but I'll leave in my 2 cents lol.

src/overlay009/ov9_02249960.c
```diff
@@ -5364,285 +5360,250 @@ static const DistWorldMapConnections *GetConnectionsForMap(int mapHeaderID)
...
-    NNS_G3dMdlUseMdlAlpha(v2->unk_00[v1].unk_04.g3DModel);
-    NNS_G3dMdlSetMdlAlphaAll(v2->unk_00[v1].unk_04.g3DModel, param2);
+    NNS_G3dMdlUseMdlAlpha(renderBuffs->models[propKind2].model.g3DModel);
+    NNS_G3dMdlSetMdlAlphaAll(renderBuffs->models[propKind2].model.g3DModel, opacity);
 
-    if (param3 != -1) {
-        param3 &= 0x3f;
-        NNS_G3dMdlSetMdlPolygonIDAll(v2->unk_00[v1].unk_04.g3DModel, param3);
+    if (polygonID != -1) {
+        polygonID &= 0x3f;
```
ravepossum: question: Could we use a constant to make the significance of this mask clearer?

src/overlay009/ov9_02249960.c
```diff
@@ -9600,10 +9557,10 @@ BOOL ov9_022511A0(FieldSystem *fieldSystem, int param1, int param2, int param3)
     return 0;
 }
 
-static void ov9_022511E0(u16 param0)
+static void PlaySoundEffect(u16 seqID)
```
ravepossum: suggestion: It'd be more verbose, but it may help to give this a more descriptive name. It would make it clearer why this function would be used instead of the standard `Sound_PlayEffect()` function. Maybe something like `PlaySoundIfNotActive` or just `TryPlaySoundEffect()`?

src/overlay009/ov9_02249960.c
```diff
@@ -351,43 +426,43 @@ typedef struct {
...
-    UnkStruct_ov9_0224D744_sub1 unk_00[25];
-    UnkStruct_ov9_0224D744_sub2 unk_258[5];
-    UnkStruct_ov9_0224D928 unk_280[34];
-} UnkStruct_ov9_0224D744;
-
-typedef struct {
+typedef struct DistWorldPropAnimInfo {
+    u16 propKind;
+    u16 animKind;
+    BOOL hasNoAnimation;
```
ravepossum: thought: Would this be just as clear as something like `isStatic`? (although I guess that could be confused with the `static` keyword) It would be nice to avoid some of the double negatives elsewhere in the file (`!HasNoAnimation`).

ravepossum (commented):
Your continued crusade on these monstrous distortion world files is very appreciated! Looks great, just a few thoughts here.

src/overlay009/ov9_02249960.c
Kuruyia: I agree that the name was confusing, `isStatic` sounds good to me.

src/overlay009/ov9_02249960.c
Kuruyia: Went with `PlaySoundIfNotActive`.

src/overlay009/ov9_02249960.c
Kuruyia: Added a `NNS_G3D_POLYGON_ID_MASK` define to our `constants/graphics.h` file.

include/map_object_move.h
Kuruyia: Went with both suggestions.

src/overlay009/ov9_02249960.c
Kuruyia: I can add a second commit to this PR to take care of this. Do we want to drop the `_KIND` part for all the `SomethingKind` enums?

Also, about the `Kind` -> `Type`, no problem for me to do that change. There was no specific reasons for using the first one over the other. Although I've used `Kind` in previous PRs, so we'll want to get those updated as well.

Kuruyia:
> Your continued crusade on these monstrous distortion world files is very appreciated! Looks great, just a few thoughts here.

More incoming! Finally gathered the courage to tackle a good chunk of this overlay.

I tried to split the work done into small-ish commits; hopefully they won't be too annoying for ya to review.

src/overlay009/ov9_02249960.c
ravepossum: Eh, it's not that big of a deal either way. We can re-eval on the follow-up if it's that impactful. I'll go ahead and get this merged.

PR #912, opened 2026-01-24
Document `evolution.c`

Viperio19:
I was so close to for once making a PR where I didn't get sidetracked and I just stayed under 10 or so files changed. Then I renamed `UnkStruct_02099F80` to `GXBanks`. For the rest, this PR just documents `evolution.c` which mainly revolves around the big state machine `Evolution_Main`.


include/evolution.h
```diff
@@ -20,55 +20,55 @@
...
+    u8 moveSlot;
+    u8 unused;
+    u8 animationState;
+    u8 attributeDelta;
+    u8 left;
+    u8 top;
+    u8 right;
+    u8 bottom;
+    int method;
     int unk_7C;
```
ravepossum: thought: It looks to me like this is related to the evolution method of the Pokemon evolving.

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
+static void Evolution_FreeBgs(BgConfig *bgConfig);
+static void Evolution_CreateSprites(EvolutionData *evolutionData);
+static void Evolution_VBlankCallback(void *data);
+static u8 Evolution_PrintString(EvolutionData *evolutionData, int entryID);
+static BOOL Evolution_TextPrinterCallback(TextPrinterTemplate *template, u16 param1);
+static void Evolution_NewSummaryScreenApplicationManager(EvolutionData *evolutionData);
+
+static const WindowTemplate sYesNowWindowTemplate = {
+    .bgLayer = BG_LAYER_MAIN_2,
+    .tilemapLeft = 0x17,
```
ravepossum: polish: We've been trying to convert hex numbers to decimal numbers unless it aids readability to keep them in hex. There's more of these sprinkled throughout the file.

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
-    sub_0207C498(v0);
-    PaletteData_StartFade(v0->unk_14, 0x1 | 0x2 | 0x4 | 0x8, 0xffff, 1, 16, 0, 0x0);
-    PokemonSpriteManager_StartFadeAll(v0->unk_18, 16, 0, 0, 0x0);
-
-    v0->unk_58 = sub_0201567C(v0->unk_14, 0, 0xb, heapID);
-
-    sub_02015738(v0->unk_58, 1);
-    SysTask_Start(sub_0207B0A0, v0, 0);
+    Evolution_InitGraphics(evolutionData, evolutionData->bgConfig);
+    Window_Add(evolutionData->bgConfig, evolutionData->window, 1, 2, 0x13, 27, 4, PLTT_11, (18 + 12) + 1);
```
ravepossum: polish: `1` -> `BG_LAYER_MAIN_1`.

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
+    Window_FillTilemap(evolutionData->window, 0xff);
```
ravepossum: nitpick: not remotely blocking but we've also been trying to standardize hex characters to be uppercase.

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
+    evolutionData->monSummary->showContest = showContest;
+    evolutionData->pokedex = pokedex;
+    evolutionData->bag = bag;
+    evolutionData->records = records;
+    evolutionData->poketch = poketch;
+    evolutionData->method = method;
+    evolutionData->unk_7C = param10;
+
+    Evolution_CreateSprites(evolutionData);
+    PaletteData_StartFade(evolutionData->paletteData, 0x1 | 0x2 | 0x4 | 0x8, 0xffff, 1, 16, 0, 0x0);
```
ravepossum: polish: The buffer constants here could be replaced with `PLTTBUF_` constants from `palette.h`. Same with other calls to this function.

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
-        if ((param0->unk_70 & 1) == 0) {
-            PokemonSprite_AddAttribute(param0->unk_1C[0], MON_SPRITE_SCALE_X, param0->unk_71 * -1);
-            PokemonSprite_AddAttribute(param0->unk_1C[0], MON_SPRITE_SCALE_Y, param0->unk_71 * -1);
-            PokemonSprite_AddAttribute(param0->unk_1C[1], MON_SPRITE_SCALE_X, param0->unk_71);
-            PokemonSprite_AddAttribute(param0->unk_1C[1], MON_SPRITE_SCALE_Y, param0->unk_71);
-
-            if (PokemonSprite_GetAttribute(param0->unk_1C[0], MON_SPRITE_SCALE_X) == 0) {
-                param0->unk_70 ^= 1;
+    if (evolutionData->animationState) {
+        if ((evolutionData->animationState & 1) == 0) {
```
ravepossum: suggestion: `animationState` could use some constants or maybe a more descriptive name to aid readability. It's not immediately clear what the different states are or what it indicates.

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
+                    evolutionData->attributeDelta *= 2;
                 }
             }
         }
     }
 
-    if ((param0->unk_7C & 0x1) && (param0->unk_64 == 8) && (gSystem.pressedKeys & PAD_BUTTON_B)) {
-        PaletteData_StartFade(param0->unk_14, 0x1 | 0x2 | 0x4 | 0x8, 0xc00 ^ 0xffff, 0, 0, 16, 0x7fff);
-        param0->unk_64 = 41;
+    if ((evolutionData->unk_7C & 0x1) && evolutionData->state == 8 && gSystem.pressedKeys & PAD_BUTTON_B) {
```
ravepossum: polish: `8` -> `EVOLUTION_STATE_ANIMATION_ALTERNATE_POKEMON`
polish `gSystem.pressedKeys & PAD_BUTTON_B` -> `JOY_NEW(PAD_BUTTON_B)`

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
+            evolutionData->delay = 40;
+            evolutionData->state++;
         }
         break;
-    case 7:
-        if (param0->unk_73 < 0x28) {
-            param0->unk_73 += 2;
-            param0->unk_75 -= 2;
+    case EVOLUTION_STATE_CLAMP_IN:
+        if (evolutionData->top < 0x28) {
```
ravepossum: thought: I find the `top`/`bottom`/etc names here a bit ambiguous. Top of what?

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
+        if (sub_0207C8E0(evolutionData->unk_30) == FALSE) {
+            sub_0207C8C4(evolutionData->unk_30, 3);
+            sub_0207C8C4(evolutionData->unk_30, 4);
+            sub_0207C8C4(evolutionData->unk_30, 5);
+            sub_0207C8C4(evolutionData->unk_30, 6);
+            sub_0207C8C4(evolutionData->unk_30, 10);
+            PaletteData_StartFade(evolutionData->paletteData, 0x1 | 0x2 | 0x4 | 0x8, 0xc00 ^ 0xffff, 2, 0, 16, 0x7fff);
+            PokemonSprite_SetAttribute(evolutionData->monSprites[0], MON_SPRITE_SCALE_X, 0x0);
+            PokemonSprite_SetAttribute(evolutionData->monSprites[0], MON_SPRITE_SCALE_Y, 0x0);
+            PokemonSprite_SetAttribute(evolutionData->monSprites[1], MON_SPRITE_SCALE_X, 0x100);
```
ravepossum: polish: `0x100` -> `MON_AFFINE_SCALE(1)` (same on other invocations)

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
 
-                MI_CpuClearFast(&v4, sizeof(BallCapsule));
+            MI_CpuClearFast(&ballCapsule, sizeof(BallCapsule));
 
-                Pokemon_SetValue(shedinja, MON_DATA_BALL_CAPSULE, (BallCapsule *)&v4);
-                Pokemon_CalcAbility(shedinja);
+            Pokemon_SetValue(shedinja, MON_DATA_BALL_CAPSULE, (BallCapsule *)&ballCapsule);
+            Pokemon_CalcAbility(shedinja);
 
-                i = Pokemon_GetGender(shedinja);
```
ravepossum, line 733: ...GF, why?

src/evolution.c
```diff
@@ -56,933 +55,946 @@
...
+    Bg_ClearTilemap(bgConfig, BG_LAYER_MAIN_1);
+    Bg_InitFromTemplate(bgConfig, BG_LAYER_MAIN_2, &bgTemplates[1], BG_TYPE_STATIC);
+    Bg_ClearTilemap(bgConfig, BG_LAYER_MAIN_2);
+    Bg_InitFromTemplate(bgConfig, BG_LAYER_MAIN_3, &bgTemplates[2], BG_TYPE_STATIC);
+    Bg_ClearTilemap(bgConfig, BG_LAYER_MAIN_3);
+
+    G2_SetBG0Priority(0x1);
+    GXLayers_EngineAToggleLayers(GX_PLANEMASK_BG0, 1);
+
+    BgTemplate BgTemplate[] = {
```
ravepossum: polish: `BgTemplate` -> `bgTemplate`

include/evolution.h
Viperio19: `unk_7C` is related to from where the evolution is called. I could maybe call it something like `flags` and add constants for it? I'm open to other ideas as well. Here is a small overview of the effects of `unk_7C` and what `unk_7C` is set to in different calls to `Evolution_Begin`:

**effects**
`evolutionData->unk_7C & 0x1`: can't cancel evolution
`evolutionData->unk_7C & 0x2`: start evolution with separate "what?" and "pokemon is evolving" messages instead of one "what? pokemon is evolving" message 

**values**
in `StartMenu_EvolveInit` in `start_menu.c`: 
```c
UnkStruct_0203C7B8 *v2 = menu->taskData;
...
    if (v2->unk_01 == 0) {
        v5 = Evolution_Begin(..., 0x1, ...);
    } else {
        v5 = Evolution_Begin(..., NULL, ...);
    }
```
in `sub_0203DBF0` in `unk_0203D1B8.c`:
```c
v4 = Pokemon_GetEvolutionTargetSpecies(..., EVO_CLASS_BY_TRADE, ...)
Evolution_Begin(..., v4, 0x4, ...)
```
in `Battle_Main` in `ov16_0223B140.c`:
```c
Evolution_Begin(..., 0x1 | 0x2, ...)
```
in `GTSApplication_Trade_Main` in `overlay094/screens/trade.c`:
```c
Evolution_Begin(..., 0x4, ...)
```

src/evolution.c
Viperio19: it's the corners of the screen:
`G2_SetWnd0Position(evolutionData->left, evolutionData->top, evolutionData->right, evolutionData->bottom);`
would `screenLeft` etc be clearer?

src/evolution.c
Viperio19: I think constants would still be a bit confusing here because the only instance where it would be descriptive is a negation; you'd have something like `if ((evolutionData->animationState & SHRINK_CURRENT_AND_GROW_EVOLUTION) == 0)` for when the current pokemon grows and the evolution shrinks. Something like `shrinkAndGrow(Animation)State` could work (maybe with better word choices), or maybe some short comments could clarify on the states here

Viperio19:
> A small handful of thoughts on names and magic numbers.
> 
> For the future, making big rename changes like the `GXBanks` one is a lot easier to review if you split it into its own commit (if not its own PR). While it's easy to scroll past renames, it also makes it easy to miss needles of relevant changes in the haystack of renames.

The comment about renaming to `GXBanks` makes sense, I can still split it up into different commits or PRs if that would be better

Viperio19:
I pushed a new commit with most of the requested changes, that just leaves the feedback that I had questions about

include/evolution.h
ravepossum: `flags` and an associated enum that explains each flag seems sensible to me. If we're not 100% confident in the meanings though, I'm totally fine with leaving as-is for now or doing it in a follow-up PR.

src/evolution.c
ravepossum: I like `shrinkAndGrowState` or `shrinkGrowState`. I think the extra specificity as opposed to `animation` helps a lot there.

src/evolution.c
ravepossum: Yep, something like screen or window there sounds good to me.

PR #913, opened 2026-01-26
Document spawn locations and fly/warp system

KalaayPT:
- Renamed `unk_0203A7D8.c/h` to `spawn_locations.c/h`
- Renamed related functions and documented struct fields/name for table entries

### DIFF
--- a/include/spawn_locations.h
+++ b/include/spawn_locations.h
@@ -7,9 +7,9 @@
 
 int FieldOverworldState_GetDefaultWarpID(void);
 void Location_InitFly(int flyDestination, Location *location);
-void Location_InitWhiteOut(int whiteOutDestination, Location *location);
-int GetWhiteOutWarpIdByMap(int param0);
-int GetFlyWarpIdByMap(int param0);
+void Location_InitBlackOut(int blackOutDestination, Location *location);
+int GetMapBlackOutWarpId(int mapId);
+int GetMapFlyWarpId(int mapId);
 int GetSpawnIdByMapAndCoords(int mapID, int param1, int param2);
 void TryUnlockFlyLocationByMap(FieldSystem *fieldSystem, int param1);
 BOOL CheckFlyLocationUnlocked(FieldSystem *fieldSystem, int param1);
--- a/src/spawn_locations.c
+++ b/src/spawn_locations.c
@@ -13,13 +13,13 @@
 #include "vars_flags.h"
 
 typedef struct SpawnLocation {
-    u16 whiteOutMapId;
-    u16 whiteOutX;
-    u16 whiteOutZ;
+    u16 blackOutMapId;
+    u16 blackOutX;
+    u16 blackOutZ;
     u16 flyMapId;
     u16 flyX;
     u16 flyZ;
-    u8 isTeleportPos;
+    u8 isWarpPos;
     u8 unlockOnMapEntry;
     u16 firstArrival;
 } SpawnLocation;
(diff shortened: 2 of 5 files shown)


src/spawn_locations.c
```diff
@@ -85,33 +85,33 @@ void Location_InitWhiteOut(int whiteOutDestination, Location *location)
     location->faceDirection = FACE_UP;
 }
 
-int sub_0203A858(int param0)
+int GetWhiteOutWarpIdByMap(int param0)
```
ravepossum: suggestion: We should name the parameter here and in the declaration now that we know what the function does. Same with the local var. I assume it'd be appropriate as something like `mapId` and `i` respectively.

src/spawn_locations.c
```diff
@@ -85,33 +85,33 @@ void Location_InitWhiteOut(int whiteOutDestination, Location *location)
...
 {
     int v0;
 
     for (v0 = 0; v0 < NELEMS(sSpawnLocations); v0++) {
-        if ((sSpawnLocations[v0].whiteOutMapId == param0) && sSpawnLocations[v0].unk_0C) {
+        if ((sSpawnLocations[v0].whiteOutMapId == param0) && sSpawnLocations[v0].isTeleportPos) {
             return v0 + 1;
         }
     }
```
ravepossum: suggestion:
```suggestion
     for (int v0 = 0; v0 < NELEMS(sSpawnLocations); v0++) {
        if (sSpawnLocations[v0].whiteOutMapId == param0 && sSpawnLocations[v0].isTeleportPos) {
            return v0 + 1;
        }
    }
```

We've been cleaning up unnecessary declarations and parentheses as we document code. `GetFlyWarpIdByMap` is similar.

include/spawn_locations.h
```diff
@@ -0,0 +1,17 @@
...
+#define POKEPLATINUM_SPAWN_LOCATIONS_H
+
+#include "field/field_system_decl.h"
+
+#include "location.h"
+
+int FieldOverworldState_GetDefaultWarpID(void);
+void Location_InitFly(int flyDestination, Location *location);
+void Location_InitWhiteOut(int whiteOutDestination, Location *location);
+int GetWhiteOutWarpIdByMap(int param0);
```
ravepossum: suggestion: The usage of `ByMap` here and in the functions below reads a little odd to me. Nothing wrong with it exactly, but I feel like I would opt for something like `GetMapWhiteOutWarpId`. Or maybe just use `For` instead of `By`.

src/spawn_locations.c
```diff
@@ -12,19 +12,19 @@
...
+typedef struct SpawnLocation {
     u16 whiteOutMapId;
     u16 whiteOutX;
     u16 whiteOutZ;
     u16 flyMapId;
     u16 flyX;
     u16 flyZ;
-    u8 unk_0C;
-    u8 unk_0D;
+    u8 isTeleportPos;
```
ravepossum: thought: I'm a little iffy on this name because it could be confused with Teleport as a field move. That being said, I'm not super solid on my suggestions either. Maybe something like `isWhiteOutAndFlySpot` or `allowFlyAndWhiteOut`? Admittedly, a lot clunkier of a name.

src/spawn_locations.c
```diff
@@ -12,19 +12,19 @@
...
+    u8 unlockOnMapEntry;
     u16 firstArrival;
-} UnkStruct_020E97B4;
+} SpawnLocation;
 
-static const UnkStruct_020E97B4 sSpawnLocations[] = {
+static const SpawnLocation sSpawnLocations[] = {
```
ravepossum, line 27: note: I don't know that this needs to be handled by this PR, but at some point for legibility we should format this table to use the struct member names and `TRUE`/`FALSE` for the values of boolean-esque members.

ravepossum (commented):
Thanks for the contribution! A couple thoughts on names and bits of polish.

lhearachel:
Additional note: Platinum and Diamond refer to this as a "Black Out", not a "White Out".

include/spawn_locations.h
KalaayPT: probably GetMapBlackOutWarpId then, given lheas comment

src/spawn_locations.c
KalaayPT: what about isWarpSpot or isWarpPos then WhiteOutAndFly is clunkier, yes, and it also doesnt consider teleport, which also reads from this table iirc

ravepossum: I think that's the best out of the options we've mentioned so far. Sounds good to me.

KalaayPT:
resolved the discussed issues

PR #914, opened 2026-01-27
document underground manager

scbroede:
went with underground manager instead of the previously used "CommManUnderground" because that seemed like a better name for `ov23_022499E4` in the future. there are also a few things i left undocumented for now because what they do will hopefully become clearer when that file is documented.

### DIFF
--- a/include/overlay023/ov23_0223E140.h
+++ b/include/overlay023/ov23_0223E140.h
@@ -47,7 +47,7 @@ void ov23_0224160C(void);
 BOOL Mining_IsPlayerMining(int netID);
 void ov23_02241648(int param0);
 BOOL Mining_IsMiningGameTaskActive(void);
-void Mining_SetPlayerMiningStatus(int param0, int param1);
+void Mining_SetPlayerMiningStatus(int netID, int isMining);
 
 enum MiningObjectIDs {
     MINING_TREASURE_NONE = 0,
--- a/include/overlay023/underground_manager.h
+++ b/include/overlay023/underground_manager.h
@@ -58,8 +58,8 @@ enum MapMarkerAnim {
     MARKER_ANIM_MINING_SPOT,
 };
 
-#define BIT_TRAPS_FULL             1 << 4
-#define BIT_BURIED_SPHERE_IN_FRONT 1 << 5
+#define FLAG_TRAPS_FULL             (1 << 4)
+#define FLAG_BURIED_SPHERE_IN_FRONT (1 << 5)
 
 #define MAX_RADAR_BLIPS 16
 
@@ -78,12 +78,12 @@ BOOL UndergroundMan_CheckForTouchInput(void);
 void UndergroundMan_ProcessTouchInput(int netID, int unused1, void *data, void *unused3);
 void UndergroundMan_ProcessTouchRadarTrapResults(int unused0, int size, void *data, void *unused3);
 void UndergroundMan_ProcessTouchRadarMiningSpotResults(int unused0, int size, void *data, void *unused3);
-int UndergroundMan_CalculateCoordinatesIndexGet(Coordinates *coordinates);
-int UndergroundMan_CalculateCoordinatesIndexInsert(Coordinates *coordinates);
-void UndergroundMan_InitCoordinatesOrderingState(int orderedArrayLength, CoordinatesGetter coordinatesGetter);
+int UndergroundMan_CalcCoordsIndexGet(Coordinates *coordinates);
+int UndergroundMan_CalcCoordsIndexInsert(Coordinates *coordinates);
+void UndergroundMan_InitCoordsOrderingState(int orderedArrayLength, CoordinatesGetter coordinatesGetter);
 void UndergroundMan_OpenMenu(void);
 void UndergroundMan_ProgressInteractCooldown(void);
-void UndergroundMan_ProcessInteract(u8 bits);
+void UndergroundMan_ProcessInteract(u8 flags);
 int CommPacketSizeOf_InteractEvent(void);
 void UndergroundMan_ProcessInteractEvent(int netID, int unused1, void *data, void *unused3);
 void UndergroundMan_Process(void);
--- a/res/text/underground_common.json
+++ b/res/text/underground_common.json
@@ -552,7 +552,7 @@
       "en_US": "{STRVAR_1 3, 0, 0} disengaged the Trap!"
     },
     {
-      "id": "UndergroundCommon_PlayerHasEntered",
+      "id": "UndergroundCommon_Text_PlayerHasEntered",
       "en_US": "{STRVAR_1 3, 1, 0} has entered."
     },
     {
--- a/src/comm_player_manager.c
+++ b/src/comm_player_manager.c
@@ -129,7 +129,7 @@ void CommPlayerMan_Disable(void)
     }
 
     if (sCommPlayerManager->isUnderground) {
-        sCommPlayerManager->isResetting = TRUE;
+        sCommPlayerManager->isDisabled = TRUE;
     }
 }
 
@@ -140,7 +140,7 @@ void CommPlayerMan_Restart(void)
     }
 
     sCommPlayerManager->sendAllPos = TRUE;
-    sCommPlayerManager->isResetting = FALSE;
+    sCommPlayerManager->isDisabled = FALSE;
 
     CommPlayer_InitPersonal();
     CommPlayer_SendPos(TRUE);
@@ -382,7 +382,7 @@ static void CommPlayer_Add(u8 netId)
         return;
     }
 
-    if (sCommPlayerManager->isResetting) {
+    if (sCommPlayerManager->isDisabled) {
         return;
     }
 
@@ -415,7 +415,7 @@ static void CommPlayer_Add(u8 netId)
         }
 
         if (sCommPlayerManager->isUnderground && !sCommPlayerManager->isActive[netId]) {
-            if (!sCommPlayerManager->isResetting) {
+            if (!sCommPlayerManager->isDisabled) {
                 ov5_021F5634(sCommPlayerManager->fieldSystem, sCommPlayerManager->playerLocation[netId].x, 0, sCommPlayerManager->playerLocation[netId].z);
             }
 
@@ -864,7 +864,7 @@ void CommPlayer_RecvDelete(int unused0, int unused1, void *src, void *unused2)
     if (sCommPlayerManager->isUnderground) {
         UndergroundMan_SetPlayerLeft(netId);
 
-        if (!sCommPlayerManager->isResetting) {
+        if (!sCommPlayerManager->isDisabled) {
             ov5_021F5634(sCommPlayerManager->fieldSystem, sCommPlayerManager->playerLocation[netId].x, 0, sCommPlayerManager->playerLocation[netId].z);
         }
     }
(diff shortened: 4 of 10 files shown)


include/overlay023/underground_manager.h
```diff
@@ -0,0 +1,126 @@
...
+    MARKER_ANIM_GRAY_SHINE_UNUSED,
+    MARKER_ANIM_YELLOW_SHINE_UNUSED,
+    MARKER_ANIM_LIGHT_YELLOW_SHINE_UNUSED,
+    MARKER_ANIM_SPHERE,
+    MARKER_ANIM_GREEN_SHINE_UNUSED,
+    MARKER_ANIM_TRAP,
+    MARKER_ANIM_MINING_SPOT,
+};
+
+#define BIT_TRAPS_FULL             1 << 4
```
ravepossum: suggestion: `(1 << 4)` and same below to avoid operator precedence issues

include/unk_02030EE0.h
```diff
@@ -39,7 +39,7 @@ BOOL sub_0203208C(BOOL param0);
 BOOL sub_020320C4(void);
 void sub_020320E8(void);
 void sub_020320FC(UnkFuncPtr_020320FC param0);
-void sub_02032110(UnkFuncPtr_02032110 param0);
+void sub_SetDisconnectCallback(UnkFuncPtr_02032110 param0);
```
ravepossum, line 42: question: Was leaving `sub_` on here intentional? Also, parameter name should match definition.

res/text/underground_common.json
```diff
@@ -552,7 +552,7 @@
       "en_US": "{STRVAR_1 3, 0, 0} disengaged the Trap!"
     },
     {
-      "id": "pl_msg_00000634_00091",
+      "id": "UndergroundCommon_PlayerHasEntered",
```
ravepossum: question: Is there a reason the `Text` infix isn't on this entry?

include/overlay023/ov23_0223E140.h
```diff
@@ -46,8 +46,8 @@ BOOL ov23_022415B8(String *param0);
 void ov23_0224160C(void);
 BOOL Mining_IsPlayerMining(int netID);
 void ov23_02241648(int param0);
-BOOL ov23_02241670(void);
-void ov23_02241690(int param0, int param1);
+BOOL Mining_IsMiningGameTaskActive(void);
+void Mining_SetPlayerMiningStatus(int param0, int param1);
```
ravepossum: polish: Declaration param names should match the definition.

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+#include "res/graphics/trap_effects/trap_effects.naix.h"
+#include "res/text/bank/underground_common.h"
+
+#define TOUCH_RADAR_RADIUS 6
+
+#define MAX_TOUCH_RADAR_BLIPS 8
+
+#define MAX_NORMAL_RADAR_BLIPS 8
+
+typedef BOOL (*TouchRadarItemCheckFunc)(int x, int z);
```
ravepossum, line 65: suggestion: In both this typedef and in some of the struct members below, does `touch` in `touchRader` add enough to justify the length it adds? I feel like `radar` might be sufficient to convey the meaning.

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+    MI_CpuCopy8(&sphereResults[1], sUndergroundMan->touchRadarBuriedSphereResults, sUndergroundMan->touchRadarBuriedSphereResultCount);
+    UndergroundTraps_StartTouchRadarTask(sUndergroundMan->fieldSystem, sUndergroundMan->touchedTileCoordinates.x, sUndergroundMan->touchedTileCoordinates.z, sUndergroundMan->touchCoordinates.x, sUndergroundMan->touchCoordinates.y, sUndergroundMan->touchRadarTrapResults, sUndergroundMan->touchRadarTrapResultCount, sUndergroundMan->touchRadarMiningSpotResults, sUndergroundMan->touchRadarMiningSpotResultCount, sUndergroundMan->touchRadarBuriedSphereResults, sUndergroundMan->touchRadarBuriedSphereResultCount);
+}
+
+static int UndergroundMan_GetOrderedCoordinatesValue(Coordinates *coordinates)
+{
+    int x = 0, z = 0;
+
+    if (coordinates == NULL) {
+        return 30 * 30 * 32 * 32;
```
ravepossum: question: Do we know what the significance of this value is? Could we use some constants here and in the return value below?

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+
+    return (z * 30 * 32) + x;
+}
+
+int UndergroundMan_CalculateCoordinatesIndexGet(Coordinates *coordinates)
+{
+    int index = 0;
+    int max = sUndergroundMan->orderedArrayLength - 1;
+    int orderedValue = UndergroundMan_GetOrderedCoordinatesValue(coordinates);
+    Coordinates _;
```
ravepossum, line 498: suggestion: `_` -> `outCoords` maybe?

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+
+    return -1;
+}
+
+int UndergroundMan_CalculateCoordinatesIndexInsert(Coordinates *coordinates)
+{
+    int index = 0;
+    int max = sUndergroundMan->orderedArrayLength - 2;
+    int orderedValue = UndergroundMan_GetOrderedCoordinatesValue(coordinates);
+    Coordinates _;
```
ravepossum, line 523: suggestion: same as above for this similar local var

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+    }
+
+    if (UndergroundMan_GetOrderedCoordinatesValue(getCoordinates(&_, index)) == orderedValue) {
+        return index;
+    }
+
+    return -1;
+}
+
+int UndergroundMan_CalculateCoordinatesIndexInsert(Coordinates *coordinates)
```
ravepossum: thought: I'm not sure I understand the difference between these two functions from the names alone. Is it just a functional difference in how they're used?

It also might not hurt to abbreviate some things for conciseness here. `Calc` and `Coords` are both a lot shorter and I don't think they discard much meaning.

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+    if (sUndergroundMan->interactCooldown != 0 && sUndergroundMan->touchCooldown != 0) {
+        return;
+    }
+
+    if (CommPlayer_GetMovementTimer(CommSys_CurNetId()) != 0) {
+        return;
+    }
+
+    if (Underground_GetTrapCount(underground) == MAX_TRAP_SLOTS) {
+        bits |= BIT_TRAPS_FULL;
```
ravepossum: thought: I feel like we tend to use `flags` or `flagBits` to indicate this sort of thing. `bits` feels too ambiguous.

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+
+        return;
+    }
+
+    if (event->bits & BIT_BURIED_SPHERE_IN_FRONT) {
+        if (UndergroundPlayer_BuriedObjectHeldFlagCheck(netID)) {
+            return;
+        }
+
+        if (event->pos == (coordinates.x & 0xF) * 16 + (coordinates.z & 0xF)) {
```
ravepossum: thought: I've only seen the `(foo.x & 0xF) * 16 + (foo.z & 0xF)` pattern a couple times now but it might not hurt to make a macro, especially if it appears more. `COORDS_TO_EVENT_POS` or something along those lines.

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+{
+    if (!sUndergroundMan) {
+        return 0;
+    }
+
+    return SecretBases_GetBaseZCoordinate();
+}
+
+// also returns true for out of bounds coordinates in general
+BOOL UndergroundMan_AreCoordinatesInSecretBase(int x, int z)
```
ravepossum, line 895: question: Is this the best name for this function? The contents don't really seem to reflect the name to me.

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+{
+    sUndergroundMan->activeRadar = RADAR_TRAP;
+}
+
+void UndergroundMan_SetTreasureRadarActive(void)
+{
+    sUndergroundMan->activeRadar = RADAR_TREASURE;
+}
+
+void UndergroundMan_SetLinkTalkedTo(int netID, int targetNetID)
```
ravepossum, line 1011: question: Why `Link` instead of `Player` here?

src/overlay023/underground_manager.c
```diff
@@ -0,0 +1,1241 @@
...
+
+void UndergroundMan_StoreCursorAndListPos(u16 menuKey, u16 cursorPos, u16 listPos)
+{
+    int i;
+
+    if (sUndergroundMan->storedPositionKey == UNDERGROUND_STORED_POS_NONE) {
+        return;
+    }
+
+    for (i = 0; i < 20; i++) {
```
ravepossum: suggestion: Seeing that this is used a few times, making a `MAX_STORED_POSITIONS` constant might be a good idea.

src/comm_player_manager.c
ravepossum: thought: I don't really have the full context on what this struct member does, but from a quick glance, with this function renamed, maybe `isDisabled` would be more appropriate?

include/unk_02030EE0.h
scbroede: yes, that's normally what i do when naming a function in an otherwise-undocumented file

res/text/underground_common.json
scbroede: there isn't, that was a mistake

src/overlay023/underground_manager.c
scbroede: the annoying thing is that there are two things that use the term "radar": the feature where you tap the touch screen, and the secret base pc feature that shows things on the top screen. i use "touch radar" to specify the former

src/overlay023/underground_manager.c
scbroede: i feel like they *should* be the dimensions of the underground map in tiles, but these are exactly twice as big for some reason

src/overlay023/underground_manager.c
scbroede: i went with `_` because it only exists because `Underground_GetOrderedCoordinatesValue` requires a `Coordinates *` parameter and the value that gets put there is never used for anything.

src/overlay023/underground_manager.c
scbroede: basically these convert a set of coordinates to an index in the array of buried spheres/traps, and the way that index is calculated is different when inserting a new buried sphere/trap or getting an existing one. not a huge fan of the names but couldn't think of anything better at the time

agreed w/r/t abbreviating

src/overlay023/underground_manager.c
scbroede: yes, while having an x or z coordinate over the max would also return true, that's not actually possible without glitches/cheating, and in practice whenever this function is called it's in the context of wanting to know if the player is currently in a secret base (the bases are on the same map as the rest of the underground along the western and northern edges).

really this probably could have just been a check for z < 64 since afaik all the secret bases that are actually accessible meet that criterion

src/overlay023/underground_manager.c
scbroede: that's our standard term for "connected player" iirc

scbroede:
updated

PR #916, opened 2026-01-27
decompile ov61_0222DCDC

scbroede:


ravepossum (approved):
Looks good! One tiiiny nitpick - `lib/gds/include/gds_internal.h` and `lib/gds/meson.build` are both missing EOF newlines. I know you have a follow-up cooking though - can you fix those in that one? I'll go ahead and merge this.

PR #917, opened 2026-01-28
Document `battle_strings.json`

Viperio19:
Almost all messages in `battle_strings.json` now have names formatted as `BattleStrings_Text_[Description]`. There are 45 messages left with `BattleStrings_Text_[Index]` because I wasn't fully certain about what the `STRVAR`s meant there. This PR is limited to only renaming the messages in `battle_strings.json`, all changes in other files are because they already referred to the original `pl_msg_00000368_[Index]` names.


res/text/battle_strings.json
```diff
@@ -2,866 +2,866 @@
...
-      "id": "pl_msg_00000368_00005",
+      "id": "BattleStrings_Text_PokemonIsTryingToLearnMove",
       "en_US": [
         "{STRVAR_1 1, 0, 0} is trying to\n",
         "learn {STRVAR_1 6, 1, 0}.\r"
       ]
     },
     {
-      "id": "pl_msg_00000368_00006",
+      "id": "BattleStrings_Text_ButPokemonCantLearnMoreThanFourMoves",
```
lhearachel, line 44: **suggestion**: 

```suggestion
      "id": "BattleStrings_Text_CantLearnMoreThanFourMoves",
```

res/text/battle_strings.json
```diff
@@ -2,866 +2,866 @@
...
-      "id": "pl_msg_00000368_00020",
+      "id": "BattleStrings_Text_TheFoesPokemonAvoidedDamageByUsingAbility",
       "en_US": [
         "The foe’s {STRVAR_1 1, 0, 0} avoided\n",
         "damage by using {STRVAR_1 5, 1, 0}!"
       ]
     },
     {
-      "id": "pl_msg_00000368_00021",
+      "id": "BattleStrings_Text_PokemonMakesGrondMovesMissByUsingAbility",
```
lhearachel: **typo**:


```suggestion
      "id": "BattleStrings_Text_PokemonMakesGroundMovesMissByUsingAbility",
```

res/text/battle_strings.json
```diff
@@ -2,866 +2,866 @@
...
       "en_US": [
         "{STRVAR_1 1, 0, 0} makes Ground moves\n",
         "miss by using {STRVAR_1 5, 1, 0}!"
       ]
     },
     {
-      "id": "pl_msg_00000368_00022",
+      "id": "BattleStrings_Text_TheWildPokemonMakesGrondMovesMissByUsingAbility",
```
lhearachel: **typo**:

```suggestion
      "id": "BattleStrings_Text_TheWildPokemonMakesGroundMovesMissByUsingAbility",
```

res/text/battle_strings.json
```diff
@@ -2,866 +2,866 @@
...
       "en_US": [
         "The wild {STRVAR_1 1, 0, 0} makes Ground\n",
         "moves miss by using {STRVAR_1 5, 1, 0}!"
       ]
     },
     {
-      "id": "pl_msg_00000368_00023",
+      "id": "BattleStrings_Text_TheFoesPokemonMakesGrondMovesMissByUsingAbility",
```
lhearachel: **typo**:


```suggestion
      "id": "BattleStrings_Text_TheFoesPokemonMakesGroundMovesMissByUsingAbility",
```

res/text/battle_strings.json
```diff
@@ -2,866 +2,866 @@
...
-      "id": "pl_msg_00000368_00040",
+      "id": "BattleStrings_Text_TheWildPokemonPreventsEscapeWithAbility",
       "en_US": [
         "The wild {STRVAR_1 1, 0, 0} prevents\n",
         "escape with {STRVAR_1 5, 1, 0}!\r"
       ]
     },
     {
-      "id": "pl_msg_00000368_00041",
+      "id": "BattleStrings_Text_ThFoesPokemonPreventsEscapeWithAbility",
```
lhearachel: **typo**:


```suggestion
      "id": "BattleStrings_Text_TheFoesPokemonPreventsEscapeWithAbility",
```

res/text/battle_strings.json
```diff
@@ -2,866 +2,866 @@
...
+      "id": "BattleStrings_Text_TheFoesPokemonsAbilityFrozePokemonSolid",
       "en_US": [
         "The foe’s {STRVAR_1 1, 0, 0}’s\n",
         "{STRVAR_1 5, 1, 0} froze {STRVAR_1 1, 2, 0}\f",
         "solid!"
       ]
     },
     {
-      "id": "pl_msg_00000368_00110",
+      "id": "BattleStrings_Text_TheFoesPokemonsAbilityFrozeTheFoesPokemonSolid",
```
lhearachel: **thought**: After reading so many of these, I wonder if there's a better notation that we could devise. Maybe something like:

```
BattleStrings_Text_AbilityFreeze_AllyToAlly
BattleStrings_Text_AbilityFreeze_AllyToWild
BattleStrings_Text_AbilityFreeze_AllyToFoe
BattleStrings_Text_AbilityFreeze_WildToAlly
BattleStrings_Text_AbilityFreeze_WildToWild
BattleStrings_Text_AbilityFreeze_FoeToAlly
BattleStrings_Text_AbilityFreeze_FoeToFoe
```

I'm curious what others think here, too.

res/text/battle_strings.json
```diff
@@ -888,2669 +888,2669 @@
...
       ]
     },
     {
-      "id": "pl_msg_00000368_00148",
+      "id": "BattleStrings_Text_00148",
       "en_US": [
         "The effect of your team’s\n",
         "{STRVAR_1 6, 0, 0} wore off!"
       ]
     },
```
lhearachel, line 1022: **thought**: Looks like it's just semantically identical to `00288`, and thus a leftover. I don't see any instances in the code where `148` is referenced as a string ID for the battle code.

Viperio19:
> One thought that springs to mind is if we want to use IDs that are direct translations of the text or instead "adapt" them to the specific circumstances. e.g., do we want `pl_msg_00000368_00192` to be `BattleStrings_Text_MoveRaisedYourTeamsDefenseSlightly` or something like `BattleStrings_Text_Reflect_AllyDoubles`?

Yeah that's probably the main question here. Translations are nice for extendibility in my opinion, like adding new moves with similar or the same effects, but ofc that's just for potential future modding and not much about the repo itself atm. Besides that point I don't have a strong preference between the two options.

ravepossum:
> One thought that springs to mind is if we want to use IDs that are direct translations of the text or instead "adapt" them to the specific circumstances. e.g., do we want pl_msg_00000368_00192 to be BattleStrings_Text_MoveRaisedYourTeamsDefenseSlightly or something like BattleStrings_Text_Reflect_AllyDoubles?

I personally prefer direction translations because there could be scenarios where the same string is used in multiple situations (I think I came across this in the summary screen).

res/text/battle_strings.json
Viperio19: I think this depends on the overall route between translation and description. I could see a hybrid though:
```
BattleStrings_Text_PokemonsAbilityFrozePokemonSolid_AllyToAlly
...
BattleStrings_Text_PokemonsAbilityFrozePokemonSolid_FoeToFoe
```
This keeps the translation bit while also giving a much better overview instead of all these messy bunches of variants of the same message

lhearachel: Sounds good to me. 😁

Viperio19: I also want to suggest doing this for variants with a single pokemon:
```diff
- BattleStrings_Text_PokemonProtectedItself
- BattleStrings_Text_TheWildPokemonProtectedItself
- BattleStrings_Text_TheFoesPokemonProtectedItself
+ BattleStrings_Text_PokemonProtectedItself_Ally
+ BattleStrings_Text_PokemonProtectedItself_Wild
+ BattleStrings_Text_PokemonProtectedItself_Foe
```
This is not as necessary, but it's still worth it as a small improvement in my opinion

PR #980, opened 2026-02-27
Unpack pl_batt_obj

lhearachel:
This archive contains a mish-mash of stuff used for sprite-work in battles, mostly related to the interface and scenery:

- the terrain blobs that the battlers stand on, which show on either side of the screen,
- the sprites used for thrown Poke Balls during capture attempts,
- the bottom-screen interface,
- the base healthboxes, and
- the icons used for both move and contest types (which are also used by the summary screen).

There are also some duplicate copies of trainer back-sprites that appear to be unused in the retail ROM and a handful of debug-related files.

Graphics files are split out into a separate commit for ease of reviewing other changes.

I'm not super happy with how the `terrain` files are built, but Meson does not seem to support generating filenames from template-strings.

I took the liberty of decorating references to the member files that are actually used where I could find them, at least for anything outside of the input interface.


res/graphics/battle/ball_throws/bait.png
scbroede, line 1: this is the bait you throw in the great marsh, not a rock

res/graphics/battle/misc/unk_117.png
scbroede, line 1: this appears to be an unused healthbox. imo it looks best with width 8, though given that it's unused it probably doesn't matter

res/graphics/battle/misc/unk_120.png
scbroede, line 1: this is the gen 3 enemy healthbox. you could pass in -palindex 1 (or 2? i forget whether it's 0-indexed or 1-indexed) to give it its proper colouring if you want

res/graphics/battle/misc/unk_121_124.png
scbroede, line 1: do we want to just call this a debug image or something so we're not left with an `unk`?

res/graphics/battle/misc/unk_117.png
lhearachel: It's especially odd because the cell application doesn't seem to produce matches. Might be stripping tiles out or something? I recall we talked about this in Discord.

I don't think it's worth resizing. It's unused, and a debug-leftover.

res/graphics/battle/misc/unk_120.png
lhearachel: I don't think it's particularly valuable; the bad palette index is arguably a positive signal that this is not a graphic referenced in the base game.

res/graphics/battle/misc/unk_117.png
scbroede: >It's especially odd because the cell application doesn't seem to produce matches. Might be stripping tiles out or something?

probably because it overlaps itself; it has the white background as one image and then places the actual details of the healthbox on top of that

scbroede (approved):

PR #987, opened 2026-03-03
document underground map transitions

scbroede:
~~99% done, parking this here as a draft until the resdat stuff is merged, at which point i'll finish it~~


src/field_map_change.c
```diff
@@ -1072,246 +1086,231 @@ void FieldSystem_StartMapChangeWarpTask(FieldSystem *fieldSystem, int param1, in
...
+            ctx->state++;
         }
         break;
-    case 7:
-        Heap_Free(mapChangeUndergroundData);
-        return 1;
+    case EXIT_UNDERGROUND_END:
+        Heap_Free(ctx);
+        return TRUE;
         break;
```
lhearachel: **polish**: Unnecessary `break`

lhearachel (approved):
One nitpick comment

src/field_map_change.c
scbroede: fixed

PR #1026, opened 2026-03-22
rom: build the trainer AI script with arm-none-eabi-gcc

lhearachel:
## Description

The Trainer AI script is currently built with `mwasmarm` (hereafter called `mwas`), which implements a different assembler dialect than `arm-none-eabi-gcc` (hereafter called `gas`). This creates a few pitfalls:

1. The implementation of [`.equ` in `gas`](https://sourceware.org/binutils/docs/as/Equ.html) allows for setting the definition of a symbol multiple times; this is not the case in `mwas`, which will throw an error like the [`.equiv` directive in `gas`](https://sourceware.org/binutils/docs/as/Equiv.html).
2. `gas` does not recognize `;` as a comment-prefix.
3. The [`.macro` directive in `gas`](https://sourceware.org/binutils/docs/as/Macro.html) supports the specification of required parameters, variadic arguments, and default values for unspecified parameters.
4. `mwas` supports far fewer preprocessor features and options. In particular, `gas`, as a `gcc`-derivative, will provide [standard pre-defined macros](https://gcc.gnu.org/onlinedocs/cpp/Standard-Predefined-Macros.html) on invocation.

It makes sense to unify how we compile bytecode scripts, so that the behavior is consistent throughout the project.

## Other Approaches

One way that I tried to do this was specifying `arm-none-eabi-gcc` as the `nasm` compiler in the Meson cross-files. That didn't seem to work, as `arm-none-eabi-gcc` is not directly recognized by Meson as a known compiler. I'm sure that there's a way to do this, but I didn't try much harder past the first hour or two of debugging.

## Clean-up Changes

- `asm/macros/aicmd.inc` - All macro parameters have been marked as required.
- `src/battle/trainer_ai/script.s` - All comment-strings have been updated to use `//` rather than `;` as a prefix.

### DIFF
--- a/asm/macros/aicmd.inc
+++ b/asm/macros/aicmd.inc
@@ -1,185 +1,185 @@
     .equ FALSE, 0
     .equ TRUE, 1
 
-    .macro IfRandomLessThan i, jump
+    .macro IfRandomLessThan i:req, jump:req
     .long 0
     .long \i
     .long (\jump-.) / 4 - 1
     .endm
 
-    .macro IfRandomGreaterThan i, jump
+    .macro IfRandomGreaterThan i:req, jump:req
     .long 1
     .long \i
     .long (\jump-.) / 4 - 1
     .endm
 
-    .macro IfRandomEqualTo i, jump
+    .macro IfRandomEqualTo i:req, jump:req
     .long 2
     .long \i
     .long (\jump-.) / 4 - 1
     .endm
 
-    .macro IfRandomNotEqualTo i, jump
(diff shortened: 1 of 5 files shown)


ravepossum (approved):
LGTM. This is a nice change to have! I'll let it sit for a bit longer and then merge if no objections.

lhearachel:
Linking here so that it's not lost: the use of `gas` as a language is a known feature-request on the Meson side. https://github.com/mesonbuild/meson/issues/13940

PR #1041, opened 2026-03-28
Remove Species Graphics Script

h2o-DS:
Moves the species front and back sprites narc generation from a python script to meson generators. The order file is an additional output to `speciesproc`.


res/pokemon/meson.build
```diff
@@ -244,20 +257,54 @@ pl_poke_icon_narc = custom_target('pl_poke_icon.narc',
...
-    depends: [ py_consts_generators ],
+    output: [
+        'pl_pokegra.narc',
+        'pl_pokegra.naix',
+    ],
+    input: [
+        pokegra_ncgr,
+        pokegra_nclr,
+        pokegra_none_nclr,
+        pokegra_empty,
```
lhearachel: **suggestion**: Move the `generator.process` calls inline to `input`, which both makes it clear that the outputs are tied to this archive-pack *and* removes an entry in Meson's variable-store.

tools/dataproc/src/common.h
```diff
@@ -77,6 +77,13 @@ struct header_template {
     char       *footer;
 };
 
+// Template-struct for an order file to be written. Specify the output filename
```
lhearachel: **polish**: Put a period (`.`) at the end of this comment.

tools/dataproc/src/speciesproc.c
```diff
@@ -111,6 +112,10 @@ static header_template_t headers[] = {
     { .out_filename = NULL,                          },
 };
 
+static order_template_t orders[] = {
+    { .out_filename = "pl_pokegra.order"             },
+};
```
lhearachel, line 118: **question**: Does this need a `NULL` entry as the last element?

tools/dataproc/src/common.h
```diff
@@ -77,6 +77,13 @@ struct header_template {
...
+typedef struct order_template order_template_t;
+struct order_template {
+    const char *out_filename;
+    FILE       *out_file;
+};
```
lhearachel, line 85: **question**: Couldn't this just be a `header_template_t` without a `header` and `footer`? e.g., just use a file in `tools/dataproc/data` that only has the magic comment-string. I think that might require some refactoring to support a generic template name, though, rather than one that maps to the expected output filename.

h2o-DS: My thought was to be explicit about the intention of the file and skip the `header_comment`

lhearachel: That's fair! Something we might consider, then, is making the `common_init` API more extensible by passing a meta-struct (or a pointer to one) that contains all of these templates.

h2o-DS:
Addressed comments

PR #1042, opened 2026-03-28
Document route 212 and pokemon mansion scripts

Viperio19:
This PR documents the event, text and script files for route 212 and the pokemon mansion.


src/scrcmd.c
```diff
@@ -3020,12 +3020,12 @@ static BOOL ScrCmd_DrawPokemonPreviewFromPartySlot(ScriptContext *ctx)
     return FALSE;
 }
 
-static BOOL ScrCmd_209(ScriptContext *ctx)
+static BOOL ScrCmd_RemovePokemonPreview(ScriptContext *ctx)
 {
     void **v1 = FieldSystem_GetScriptMemberPtr(ctx->fieldSystem, SCRIPT_MANAGER_DATA_PTR);
```
ravepossum: question: Any reason not to name this?

src/scrcmd.c
```diff
@@ -3020,12 +3020,12 @@ static BOOL ScrCmd_DrawPokemonPreviewFromPartySlot(ScriptContext *ctx)
...
 
-    u8 *v0 = *v1;
-    *v0 = 1;
+    u8 *previewState = *v1;
+    *previewState = 1;
```
ravepossum, line 3028: suggestion: We should make an `enum` for the states to make this much clearer, but I'm fine with considering that out of scope for this PR.

ravepossum (commented):
Couple thoughts on the script command, rest looks good!

src/scrcmd.c
Viperio19: No reason besides that I'm not sure what exactly to name it

src/scrcmd.c
Viperio19: Yeah that would help, but also sounds out of scope

src/scrcmd.c
ravepossum: It's just the state, same as the var below right? Except one level of indirection higher. The script data pointer is just a generic way to store data between command calls. `dataPtr` or something seems fine to me.

Viperio19: done!

ravepossum (approved):

PR #1045, opened 2026-03-29
Start documenting some undocumented player data

ItsLeaze:


include/overlay066/ov66_0222DDF0.h
```diff
@@ -67,21 +67,21 @@ UnkStruct_ov66_0222E71C *ov66_0222E3BC(UnkStruct_ov66_0222DFF8 *param0);
...
-void ov66_0222E56C(UnkStruct_ov66_0222DFF8 *param0, int param1, int param2);
+void ov66_0222E56C(UnkStruct_ov66_0222DFF8 *param0, int type1, int type2);
 void ov66_0222E5D8(UnkStruct_ov66_0222DFF8 *param0, u32 param1, u32 param2);
 void ov66_0222E640(const UnkStruct_ov66_0222E71C *param0, TrainerInfo *param1, enum HeapID heapID);
 u32 ov66_0222E71C(const UnkStruct_ov66_0222E71C *param0);
 int ov66_0222E728(const UnkStruct_ov66_0222E71C *param0, u32 param1);
 s32 ov66_0222E760(const UnkStruct_ov66_0222E71C *param0, u32 param1);
 s32 ov66_0222E798(const UnkStruct_ov66_0222E71C *param0);
-u32 ov66_0222E79C(const UnkStruct_ov66_0222E71C *param0);
+u32 GetTrainerID(const UnkStruct_ov66_0222E71C *param0);
```
ravepossum: issue: We follow the general convention of `Module_Function()` when naming non-`static` functions to minimize namespace issues and mimic the pseudo-OOP structure of the game. Since that hasn't been identified yet here, you'll need to identify that. This would go for all the non-static functions here.

We have a list of descriptions of each overlay [here](gist.github.com/lhearachel/7983d1e594ff21300941b94bc2fbbb62) which indicates that overlay66 is for the Wi-Fi Plaza orchestrator. There's a handful of files in this overlay though, so it's one specific part of the plaza orchestrator. If you're having trouble identifying or naming the module, feel free to ask in the discord.

include/overlay066/ov66_0222DDF0.h
```diff
@@ -67,21 +67,21 @@ UnkStruct_ov66_0222E71C *ov66_0222E3BC(UnkStruct_ov66_0222DFF8 *param0);
...
 int ov66_0222E7A0(const UnkStruct_ov66_0222E71C *param0);
 UnkStruct_ov66_02231428 ov66_0222E7C4(const UnkStruct_ov66_0222E71C *param0);
-u32 ov66_0222E7C8(const UnkStruct_ov66_0222E71C *param0);
-u32 ov66_0222E80C(const UnkStruct_ov66_0222E71C *param0);
+u32 GetGender(const UnkStruct_ov66_0222E71C *param0);
```
ravepossum: suggestion: It may also be a good idea to identify what exactly `UnkStruct_ov66_0222E71C` is before naming these functions. I don't want to expand the scope of the PR (I think starting small is good), but we tend to hold off on naming functions until we're certain we know what they do. When reviewing, I apply a lot more scrutiny to a named function than one left unnamed. Leaving a function (or anything really) unnamed that you're not sure about is perfectly valid.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -65,27 +65,27 @@
...
-    u32 unk_04;
-    u16 unk_08[8];
-    UnkStruct_ov66_02231428 unk_18;
-    UnkStruct_ov66_02231428 unk_1C;
-    u16 unk_20[6];
-    u8 unk_2C[6];
-    u8 unk_32[6];
-    u8 unk_38;
+    u32 trainerID;
+    u16 trainerName[8];
```
ravepossum: polish: `8` here is `TRAINER_NAME_LEN + 1`.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -65,27 +65,27 @@
...
+    u8 isMainStoryCleared;
     u8 unk_41;
-    u8 unk_42;
+    u8 gameVersion;
     u8 unk_43;
     s64 unk_44;
     u8 unk_4C[12];
     s32 unk_58[12];
-    u16 unk_88[2];
+    u16 types[2];
```
ravepossum: suggestion: I'm a little hesitant to name this without having more context on its usage. I think `types` is pretty vague.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -372,23 +372,23 @@ static const s32 Unk_ov66_0225892C[9] = {
...
-    { 0x46, 0x0 },
-    { 0x6, 0x1 },
-    { 0x7, 0x1 },
-    { 0xD, 0x1 },
-    { 0xE, 0x1 },
-    { 0x23, 0x1 },
-    { 0x25, 0x1 },
-    { 0x2A, 0x1 },
-    { 0x3F, 0x1 }
+static const TrainerAppearanceAndGender AppearanceGenderCombinations[16] = {
```
ravepossum: suggestion: These constants look like graphics ID values to me - take a look the generated constants in `build/generated/object_events_gfx.h`. We should use those constants for them and also take that into account in how the related functionality here is named. For instance, I think a more appropriate name for the corresponding struct member may be `graphicsID`.

I think it would also be better to be more specific with the structure name - maybe something like `WifiPlazaTrainerAppearance`. If we did that, then I think a more appropriate name for the array would be something like `sWifiPlazaTrainerAppearances`. In general, when I document things, I tend to err on the side of specificity to eliminate ambiguity and namespace conflicts. Especially since there's `TrainerAppearance`s elsewhere in the codebase like `appearance.c`.

issue: `static` variables should be prefixed with `s` and be camelCase.

polish: `16` here could be `APPEARANCES_COUNT * 2`. Now that we know it's used elsewhere, we could lift that constant out of `appearance.c` into `appearance.h` and include it here.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -926,30 +926,30 @@ void ov66_0222E528(UnkStruct_ov66_0222DFF8 *param0, u32 param1)
     ov66_0222F964(param0);
 }
 
-void ov66_0222E56C(UnkStruct_ov66_0222DFF8 *param0, int param1, int param2)
+void ov66_0222E56C(UnkStruct_ov66_0222DFF8 *param0, int type1, int type2)
```
ravepossum, line 910: suggestion: While the constants here are definitely suggestive of types, do we know enough about what this function does to say with certainty that they're types? It's also possible we could give more descriptive names than `type1` and `type2` if we know what they're used for.

It looks like this function is called in overlay70, which we have described as the Wi-Fi Plaza 3D Map. I know that the plaza has type-based room but only a handful of types are included. I'm not sure what this is for without digging into it more. Like I mentioned else, feel free to give a shout in discord if you're unsure and someone'll likely be happy to give a hand.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1911,17 +1911,17 @@ void ov66_0222F198(UnkStruct_ov66_0222DFF8 *param0)
...
+static BOOL ov66_0222F1B4(const u16 *trainerName, u32 param1)
 {
-    int v0;
+    int idx;
 
-    for (v0 = 0; v0 < param1; v0++) {
-        if (param0[v0] == 0xffff) {
-            return 1;
+    for (idx = 0; idx < param1; idx++) {
+        if (trainerName[idx] == 0xffff) {
```
ravepossum: polish: `0xffff` looks to be `CHAR_EOS` in this context.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -978,54 +978,54 @@ void ov66_0222E5D8(UnkStruct_ov66_0222DFF8 *param0, u32 param1, u32 param2)
...
-    v1 = 0;
+    isTrainerNameValid = ov66_0222F1B4(param0->trainerName, TRAINER_NAME_LEN + 1);
+    isTrainerNameInvalid = FALSE;
 
-    if (v0 == 1) {
-        TrainerInfo_SetName(param1, param0->unk_08);
+    if (isTrainerNameValid == TRUE) {
+        TrainerInfo_SetName(trainerInfo, param0->trainerName);
 
         {
```
ravepossum: polish: These unnecessary scope braces can be removed. We like to clean stuff like this up while we document.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -978,54 +978,54 @@ void ov66_0222E5D8(UnkStruct_ov66_0222DFF8 *param0, u32 param1, u32 param2)
...
-        MessageLoader *v4;
-        String *v5;
+    if (isTrainerNameInvalid) {
+        MessageLoader *messageLoader;
+        String *fallbackName;
 
-        v4 = MessageLoader_Init(MSG_LOADER_LOAD_ON_DEMAND, NARC_INDEX_MSGDATA__PL_MSG, TEXT_BANK_UNK_0673, heapID);
-        v5 = MessageLoader_GetNewString(v4, 64);
+        messageLoader = MessageLoader_Init(MSG_LOADER_LOAD_ON_DEMAND, NARC_INDEX_MSGDATA__PL_MSG, TEXT_BANK_UNK_0673, heapID);
+        fallbackName = MessageLoader_GetNewString(messageLoader, 64);
```
ravepossum: polish `64` here appears to be `pl_msg_00000673_00064`. You could also give it a name while you're at it, if you've identified what the text bank should be called. Leaving it as the undocumented constant would be totally fine though.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -978,54 +978,54 @@ void ov66_0222E5D8(UnkStruct_ov66_0222DFF8 *param0, u32 param1, u32 param2)
     ov66_0222F964(param0);
 }
 
-void ov66_0222E640(const UnkStruct_ov66_0222E71C *param0, TrainerInfo *param1, enum HeapID heapID)
+void ov66_0222E640(const UnkStruct_ov66_0222E71C *param0, TrainerInfo *trainerInfo, enum HeapID heapID)
 {
-    BOOL v0;
-    BOOL v1;
+    BOOL isTrainerNameValid;
```
ravepossum: polish: You should be able to change these lines so the initialization and declaration of the vars occurs at the same time. This is something we've been trying to clean up as we go as long as it still matches.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1086,21 +1086,21 @@ s32 ov66_0222E798(const UnkStruct_ov66_0222E71C *param0)
...
+    u32 trainerAppearance;
+    int idx;
 
-    v0 = ov66_0222E858(param0);
+    trainerAppearance = GetTrainerAppearance(param0);
 
-    for (v1 = 0; v1 < 16; v1++) {
-        if (Unk_ov66_02258974[v1].unk_00 == v0) {
-            return v1;
+    for (idx = 0; idx < 16; idx++) {
```
ravepossum: suggestion: Same as in the array declaration, the `16` here looks like `APPEARANCES_COUNT * 2`.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1086,21 +1086,21 @@ s32 ov66_0222E798(const UnkStruct_ov66_0222E71C *param0)
...
 {
-    return param0->unk_04;
+    return param0->trainerID;
 }
 
 int ov66_0222E7A0(const UnkStruct_ov66_0222E71C *param0)
 {
-    u32 v0;
-    int v1;
+    u32 trainerAppearance;
```
ravepossum: polish: This looks like another opportunity to declare the vars as they're initialized.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1112,32 +1112,32 @@ UnkStruct_ov66_02231428 ov66_0222E7C4(const UnkStruct_ov66_0222E71C *param0)
...
-    if (param0->unk_38 >= 2) {
-        v2 = ov66_02230C2C(param0->unk_3A);
+    if (param0->trainerGender >= GENDER_NONE) {
+        trainerAppearance = GetAppearanceFromIdx(param0->trainerAppearanceIdx);
 
-        if (v2 != 0xffff) {
-            for (v1 = 0; v1 < NELEMS(Unk_ov66_02258974); v1++) {
-                if (Unk_ov66_02258974[v1].unk_00 == param0->unk_3A) {
-                    v0 = Unk_ov66_02258974[v1].unk_00;
+        if (trainerAppearance != 0xffff) {
```
ravepossum: note: Just for your knowledge, the 0xFFFF is a constant that refers to `OBJ_EVENT_GFX_NONE`. I have that currently documented in an unfinished PR.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1112,32 +1112,32 @@ UnkStruct_ov66_02231428 ov66_0222E7C4(const UnkStruct_ov66_0222E71C *param0)
...
-    int v1;
-    u32 v2;
+    u32 gender;
+    int idx;
+    u32 trainerAppearance;
 
-    if (param0->unk_38 >= 2) {
-        v2 = ov66_02230C2C(param0->unk_3A);
+    if (param0->trainerGender >= GENDER_NONE) {
+        trainerAppearance = GetAppearanceFromIdx(param0->trainerAppearanceIdx);
```
ravepossum: suggestion: As I mentioned in an earlier comment, this looks more like a `graphicsID`.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -3686,21 +3686,21 @@ static BOOL ov66_02230C04(const UnkStruct_ov66_02230A6C *param0)
...
+        if (AppearanceGenderCombinations[idx].trainerAppearance == param0) {
+            return idx;
         }
     }
 
     return 16;
 }
 
-static u32 ov66_02230C2C(u32 param0)
+static u32 GetAppearanceFromIdx(u32 trainerAppearanceIdx)
```
ravepossum: suggestion: Mentioned this in a couple places already so apologies if this is getting repetitive, but what you've referred to as `Appearance` seems to be more appropriate as `GraphicsID`. Building from that, I think what you've referred to as the `appearanceIdx` makes more sense as the `appearance`.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -3686,21 +3686,21 @@ static BOOL ov66_02230C04(const UnkStruct_ov66_02230A6C *param0)
...
 {
-    if (param0 < 16) {
-        return Unk_ov66_02258974[param0].unk_00;
+    if (trainerAppearanceIdx < 16) {
```
ravepossum: polish: `16` -> `APPEARANCES_COUNT * 2`

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1168,20 +1168,20 @@ BOOL ov66_0222E824(const UnkStruct_ov66_0222E71C *param0)
...
 
-    if (v0 != 0xffff) {
-        return v0;
+    if (trainerAppearance != 0xffff) {
+        return trainerAppearance;
     }
 
-    if (ov66_0222E7C8(param0) == 0) {
+    if (GetGender(param0) == GENDER_MALE) {
         return 0x3;
```
ravepossum: polish: This looks like `OBJ_EVENT_GFX_SCHOOL_KID_M`.

src/overlay066/ov66_0222DDF0.c
ravepossum: polish: This looks like `OBJ_EVENT_GFX_LASS`.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1190,32 +1190,32 @@ u32 ov66_0222E858(const UnkStruct_ov66_0222E71C *param0)
...
     }
 
-    return v0;
+    return country;
 }
 
 u32 ov66_0222E8A4(const UnkStruct_ov66_0222E71C *param0)
 {
-    u32 v0;
+    u32 region;
```
ravepossum: question: Do we know enough here to know that `region` makes sense for this variable (and the corresponding struct member)?

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1244,15 +1244,15 @@ BOOL ov66_0222E8D8(const UnkStruct_ov66_0222E71C *param0)
     return 1;
 }
 
-int ov66_0222E8E8(const UnkStruct_ov66_0222E71C *param0, u32 param1)
+int ov66_0222E8E8(const UnkStruct_ov66_0222E71C *param0, u32 index)
```
ravepossum: note: Similar thoughts on using `types` and `index` here - it could potentially be more descriptive in context once we know the context. I don't know enough about the Wi-Fi Plaza to recognize it off-hand.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1546,22 +1546,22 @@ BOOL ov66_0222EC70(UnkStruct_ov66_0222DFF8 *param0, u32 param1)
 
 void ov66_0222ECD4(UnkStruct_ov66_0222DFF8 *param0, u16 param1)
 {
-    BOOL v0 = 0;
+    BOOL isTrainerNameValid = 0;
```
ravepossum: polish: `0` -> `FALSE`. `1` -> `TRUE` for the similar assignments below.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -1911,17 +1911,17 @@ void ov66_0222F198(UnkStruct_ov66_0222DFF8 *param0)
     }
 }
 
-static BOOL ov66_0222F1B4(const u16 *param0, u32 param1)
+static BOOL ov66_0222F1B4(const u16 *trainerName, u32 param1)
 {
-    int v0;
+    int idx;
```
ravepossum: polish: I bet this could be declared as it's initialized in the loop.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -2275,63 +2275,63 @@ static void ov66_0222F768(UnkStruct_ov66_0222DFF8 *param0)
...
-    TrainerInfo *v0;
-    Party *v1;
+    TrainerInfo *trainerInfo;
+    Party *party;
     WiFiHistory *wiFiHistory;
-    Pokedex *v3;
+    Pokedex *pokedex;
     GameTime *v4;
 
     {
```
ravepossum: polish: There's a handful of unnecessary scope blocks we could remove here.

src/overlay066/ov66_0222DDF0.c
```diff
@@ -2275,63 +2275,63 @@ static void ov66_0222F768(UnkStruct_ov66_0222DFF8 *param0)
 
 static void ov66_0222F7C8(UnkStruct_ov66_0222F6C4 *param0, SaveData *saveData, enum HeapID heapID)
 {
-    TrainerInfo *v0;
-    Party *v1;
+    TrainerInfo *trainerInfo;
```
ravepossum, line 2235: polish: I bet a lot of these could be declared as they're initialized as well, but don't stress trying to get every single one if it seems like they don't all match.

ravepossum (commented):
Welcome, and thank you for the contribution! Don't hesitate to ask for clarification or help with any of the feedback here.

lhearachel:
I'm taking over cleanup of this PR given that it is stale and the contributor is unresponsive.

include/overlay066/ov66_0222DDF0.h
```diff
@@ -14,10 +16,37 @@
...
+    u8 isMainStoryCleared;
+    u8 unk_41;
+    u8 gameVersion;
+    u8 unk_43;
+    s64 startTimestamp;
+    u8 unk_4C[12];
+    s32 unk_58[12];
+    u16 unk_88[2];
+    UnkStruct_ov66_0222E908 unk_8C;
+} WiFiPlazaProfile;
```
scbroede, line 48: imo this struct should live in the library instead if possible. feels a bit wrong for a library to be including stuff from the game code

scbroede (approved):
just one nitpick, which you can feel free to disregard if you want

include/overlay066/ov66_0222DDF0.h
lhearachel: It would still need to include constants that are related to game-code.

lhearachel: Carry-over from Discord: we're judging this move as out-of-scope for now.

PR #1048, opened 2026-03-29
Document routes 219-221, pal park and fuego ironworks scripts

Viperio19:
This PR documents the event, script and text files for routes 219-211 and fuego ironworks, and it improves some documentation for pal park.


res/field/scripts/scripts_fuego_ironworks_building.s
```diff
@@ -1,122 +1,123 @@
...
+FuegoIronworksBuilding_AskTrade10StarPieces:
+    Message FuegoIronworksBuilding_Text_WantToTrade10StarPieces
     InitGlobalTextMenu 31, 11, 0, VAR_0x8001
     SetMenuXOriginToRight
-    AddMenuEntry 0x116, 0
-    AddMenuEntry 0x117, 1
-    AddMenuEntry 11, 2
+    AddMenuEntry MenuEntries_Text_StarPieces_1, 0
+    AddMenuEntry MenuEntries_Text_StarPieces_10, 1
+    AddMenuEntry MenuEntries_Text_Battle_Exit, 2
```
lhearachel: **thought**: Is this entry named too specifically?

lhearachel (commented):
Looks good! Just one thought-provocation below.

res/field/scripts/scripts_fuego_ironworks_building.s
Viperio19: Yeah, `MenuEntries_Text_Battle_Exit` was already named this because of its usage in `scripts_pokemon_center_2f_attendants.s`. There is also a general `MenuEntries_Text_Exit` but that's used for both starting communications (join group, become leader, exit) and in the move tutor screen. It might be best to just rename `MenuEntries_Text_Battle_Exit` to `MenuEntries_Text_Exit2`.

lhearachel: I agree with the latter name.

Viperio19: renamed it!

lhearachel (approved):

PR #1056, opened 2026-04-02
Improve some documentation of the trainer AI

lhearachel:


ravepossum (approved):
This is arguably out of scope of this PR, but if `FlagMoveDamageScore` is going to use the move damage constants instead of `TRUE`/`FALSE`, does that mean we should also update it to use those constants in `AICmd_IfBattlerDealsMoreDamage` and `TrainerAI_CalcAllDamage`? I did get a bit confused reading through that change and those functions while reviewing this (as someone who has never looked at the AI code before lol).

Otherwise LGTM!

lhearachel:
> This is arguably out of scope of this PR, but if `FlagMoveDamageScore` is going to use the move damage constants instead of `TRUE`/`FALSE`, does that mean we should also update it to use those constants in `AICmd_IfBattlerDealsMoreDamage` and `TrainerAI_CalcAllDamage`? I did get a bit confused reading through that change and those functions while reviewing this (as someone who has never looked at the AI code before lol).
> 
> Otherwise LGTM!

I think that there's certainly an argument to do so, but we would need to come up with a type that is sufficiently descriptive to serve as an alias. The thought behind the AI using those aliasing-constants is that a user is more likely to modify the AI script than the C code. Maybe that's misguided.

In any case, it seems like the answer is unclear, so I think we want to table this for the sake of this PR.

PR #1057, opened 2026-04-03
rom: Macro-ify the command-tables for scripting VMs

lhearachel:
This is adapting the practice used by `scrcmd` to the other scripting VMs that are currently documented.

Movement actions unfortunately cannot be declared using this same technique due to the use of the enum in C code.


ravepossum (approved):
LGTM, will merge on conflict resolution

PR #1063, opened 2026-04-05
Build species text banks and Pokedex data in speciesproc

lhearachel:
## Overview of Changes

1. Bump the required Meson version to 1.10 for the `array.slice()` function.
2. Implement a writer API for `dataproc` tools to produce formatted output files.
3. Use the writer API to produce text banks for species data: heights, weights, Pokedex categories, etc.
4. Update the structure in all Species data files to display height and weight in Imperial units (inches and pounds).
5. Unify all Giratina-based data into the base `data.json` file. Pokedex data related to the Origin forme is stored in a sub-member `pokedex_data.origin_forme`.
6. Provide functionality to read either Imperial or Metric units and perform the appropriate conversions in `speciesproc`.
7. Provide a script for users to flip between Imperial and Metric units in Species data files, at their discretion.
8. Output Pokedex data as part of `speciesproc` (rather than updating `tools/scripts/make_pokedex_data.py` to support the updated keys for height and weight).
9. Delete unnecessary `pokedex_data` members in alternate-form `data.json` files.

> [!IMPORTANT]
> Maintainers will need to remove their existing `build` directory after pulling these changes.


lhearachel:
Build fails due to a bunch of compiler args that will be fixed in a separate PR.

tools/dataproc/src/speciesproc.c
```diff
@@ -491,6 +699,120 @@ static enum Species proc_offspring(datafile_t *df, size_t i) {
...
+}
+
+static u32 get_weight_hectograms(datanode_t dexdata, enum Species species) {
+    bool   metric = dp_hasmemb(dexdata, "weight_kilograms");
+    double weight = metric
+                  ? dp_float(dp_objmemb(dexdata, "weight_kilograms"))
+                  : LB_TO_KG(dp_float(dp_objmemb(dexdata, "weight_pounds")) - species_weight_rounding_override(species));
+    double hectog = round(weight * 10);
+    return (u32)hectog;
+}
```
Kuruyia, line 761: **question**: Should an error be thrown if both (`height_meters` and `height_inches`) or (`weight_kilograms` and `weight_pounds`) are defined?

If this (admittedly, unlikely) case presents itself, the user may be left wondering why their modification isn't being taken into account if they modify the field with the least precedence.

lhearachel: Good thought! I would call this a warning, and we would use the metric values before any imperial values.

lhearachel: I've implemented this, but as an `error`, and upgraded the other `warning`s already present in `speciesproc` to `error`s. This is because Ninja buffers `stderr` into `stdout`, and all output gets suppressed if the command succeeds.

~~I'll need to do some more research into how this gets bypassed for compiler warnings.~~

Turns out it's not Ninja, but instead is Meson's subprocess wrapper for custom process environments. I've filed a patch to emit the `stderr` content: https://github.com/mesonbuild/meson/pull/15683

If that patch is not accepted, then I'll ascertain how to patch the copy of Meson that we clone as a subproject.

tools/dataproc/src/speciesproc.c
```diff
@@ -660,16 +1000,224 @@ static void emit_iconpalettes(datafile_t *df, size_t i, const char *species) {
...
+        p[0] = 0xE2;
+        p[1] = 0x80;
+        p[2] = 0x89;
+        p   += 3;
+    }
+
+    memcpy(p, tmp, len);
+}
+
+static void emit_heights( SpeciesDexIndex *index, enum Species species, const char *species_name, char *buf) {
```
Kuruyia: **polish**: Remove extra space.


```suggestion
static void emit_heights(SpeciesDexIndex *index, enum Species species, const char *species_name, char *buf) {
```

tools/dataproc/src/speciesproc.c
```diff
@@ -678,13 +1226,274 @@ static void pack(Container *species, size_t i) {
...
+    return 0;
+}
+
+static int sort_by_height_asc(const void *l, const void *r) {
+    const SpeciesDexIndex *lhs = l;
+    const SpeciesDexIndex *rhs = r;
+    if (lhs->height < rhs->height) return -1;
+    if (lhs->height > rhs->height) return  1;
+    return 0;
+}
```
Kuruyia, line 1370: **issue**: Sorting order is ill-defined when multiple species share the same weight/height. This fails to build a matching ROM on aarch64 macOS.

**suggestion**: Fallback to sorting on the species index when the case arises:


```suggestion
static int sort_by_species_asc(const void *l, const void *r) {
    const SpeciesDexIndex *lhs = l;
    const SpeciesDexIndex *rhs = r;
    if (lhs->species < rhs->species) return -1;
    if (lhs->species > rhs->species) return  1;
    return 0;
}

static int sort_by_weight_desc(const void *l, const void *r) {
    const SpeciesDexIndex *lhs = l;
    const SpeciesDexIndex *rhs = r;
    if (lhs->weight > rhs->weight) return -1;
    if (lhs->weight < rhs->weight) return  1;
    return sort_by_species_asc(l, r);
}

static int sort_by_weight_asc(const void *l, const void *r) {
    const SpeciesDexIndex *lhs = l;
    const SpeciesDexIndex *rhs = r;
    if (lhs->weight < rhs->weight) return -1;
    if (lhs->weight > rhs->weight) return  1;
    return sort_by_species_asc(l, r);
}

static int sort_by_height_desc(const void *l, const void *r) {
    const SpeciesDexIndex *lhs = l;
    const SpeciesDexIndex *rhs = r;
    if (lhs->height > rhs->height) return -1;
    if (lhs->height < rhs->height) return  1;
    return sort_by_species_asc(l, r);
}

static int sort_by_height_asc(const void *l, const void *r) {
    const SpeciesDexIndex *lhs = l;
    const SpeciesDexIndex *rhs = r;
    if (lhs->height < rhs->height) return -1;
    if (lhs->height > rhs->height) return  1;
    return sort_by_species_asc(l, r);
}
```

tools/dataproc/meson.build
```diff
@@ -1,11 +1,14 @@
 dataproc_cflags = [
     '-std=gnu17',
-    '-O3',
+    '-O0',
+    '-g3',
```
Kuruyia: **question**: Is this intentional?

tools/scripts/convert_species_units.py
Kuruyia, line 1: **question**: Do we also want a warning/error here if e.g. converting to metric but both `height_inches` and `height` are defined?

Kuruyia (changes requested):

tools/dataproc/meson.build
lhearachel: Nope, a leftover from my debugging sessions.

tools/scripts/convert_species_units.py
lhearachel: I think that's unnecessary; `height` and `weight` are now legacy fields, and they aren't used by the processing code at all. It's a bit different, in my mind, than having two fields that are usable and informing a user that we're preferring one over the other.

There's an argument, I suppose, that a user might modify `height` instead of `height_inches` and be confused when the update isn't reflected in-game. I think that's not worth handling, personally, because we are still in a "expect breaking change" state for modders.

tools/dataproc/src/speciesproc.c
lhearachel: Good catch. I keep getting bitten by LLVM using quick-sort and GNU using merge-sort, it seems.

tools/scripts/convert_species_units.py
Kuruyia: I also think this is unnecessary given how I imagine this script will be used. Let's keep it as it is then :)

tools/dataproc/src/speciesproc.c
lhearachel: I've applied this change; curiously, applying it to the name-ordering causes a mismatch, but I think we we are not concerned about that.

Kuruyia (approved):
No more comments from me. Really cool to see dataproc in action!

PR #1068, opened 2026-04-07
Document snowpoint city scripts

Viperio19:
This PR documents the event, script and text files for snowpoint city and snowpoint temple.


res/field/events/events_snowpoint_city_west_house.json
```diff
@@ -17,7 +17,7 @@
             "y": 0
         },
         {
-            "id": "SNOWPOINT_CITY_WEST_HOUSE_BEAUTY_1",
+            "id": "LOCALID_BEAUTY",
```
lhearachel: **suggestion**: `LOCALID_MINDY`

res/field/scripts/scripts_snowpoint_city_west_house.s
```diff
@@ -2,77 +2,77 @@
 #include "res/text/bank/snowpoint_city_west_house.h"
 
 
-    ScriptEntry _000A
-    ScriptEntry _0033
+    ScriptEntry SnowpointCityWestHouse_AceTrainerM
+    ScriptEntry SnowpointCityWestHouse_Beauty
```
lhearachel: **suggestion**: `SnowpointCityWestHouse_Mindy`

res/field/events/events_snowpoint_city_west_house.json
Viperio19: So far my standard has been that I name NPCs if they are referred to by their name at any point, but that's not the case with Mindy from what I can tell, I think her name is only used for the OT of the traded Haunter. Would we prefer the standard "name any NPC that has an actual name in any shape or form"?

lhearachel: For the trade NPCs, I think their names are notable enough that we should name the local IDs. I only noticed it here because Mindy is so iconic/infamous.

Viperio19: Makes sense to me, especially if people are searching for the name. I'll push a commit for it in a second

lhearachel (approved):

PR #1069, opened 2026-04-07
Standardize NPC/Npc to NPC

Viperio19:


src/overlay006/npc_trade.c
```diff
@@ -21,16 +21,16 @@
 #include "unk_020559DC.h"
 #include "unk_02092494.h"
 
-static inline String *NpcTrade_GetOtName(enum HeapID heapID, u32 npcTradeID);
-static String *NpcTrade_GetNickname(enum HeapID heapID, u32 npcTradeID);
-static void NpcTrade_CreateMon(Pokemon *mon, NpcTradeMon *npcTrade, u32 level, u32 npcTradeID, enum HeapID heapID, u32 mapID);
+static inline String *NPCTrade_GetOtName(enum HeapID heapID, u32 npcTradeID);
```
scbroede: while you're changing this function, imo it should be `OTName` over `OtName`

Viperio19: makes sense, I just pushed a commit for that

scbroede (approved):

PR #1072, opened 2026-04-08
Document the Ranger link

Gudf:
I unfortunately don't have it in me to finish documenting the ranger communication states at this time, since those would probably require documenting a bunch of stuff from `ov97_02232DC8`.


src/main_menu/main_menu.c
```diff
@@ -1293,7 +1294,6 @@ static BOOL MainMenu_Main(ApplicationManager *appMan, int *state)
 
 extern const ApplicationManagerTemplate gMysteryGiftAppTemplate;
 extern const ApplicationManagerTemplate gGBAMigratorAppTemplate;
-extern const ApplicationManagerTemplate gRangerLinkAppTemplate;
```
lhearachel, line 1296: **question**: Are we planning to do this with all of the templates? If so, would it make sense to do that here and extract them all to a single header, e.g., `include/main_menu/subapps.h`?

src/main_menu/ranger_link.c
```diff
@@ -0,0 +1,1148 @@
...
+
+#include "res/graphics/main_menu/main_menu_graphics.naix"
+#include "res/text/bank/main_menu_alerts.h"
+#include "res/text/bank/mystery_gift_menu.h"
+
+FS_EXTERN_OVERLAY(game_opening);
+
+#define BASE_TILE_STANDARD_WINDOW_FRAME 1
+#define BASE_TILE_MESSAGE_BOX_FRAME     (BASE_TILE_STANDARD_WINDOW_FRAME + STANDARD_WINDOW_TILE_COUNT)
+#define WINDOW_CONTENT_TILES_START      (BASE_TILE_MESSAGE_BOX_FRAME + SCROLLING_MESSAGE_BOX_TILE_COUNT)
```
lhearachel, line 57: **question**: Is there danger in re-using preprocessor tokens here that are defined by `include/constants/field_base_tiles.h`? Probably not.

src/main_menu/ranger_link.c
```diff
@@ -0,0 +1,1148 @@
...
+    }
+
+    trainerID = TrainerInfo_ID_LowHalf(param0->trainerInfo);
+    TrainerIDToRangerString(sRangerCharsTrainerID, trainerID);
+    sRangerCharsTrainerID[5] = 0;
+}
+
+static void InitAndLoadGraphics(RangerLinkManager *param0)
+{
+    int frameType;
```
lhearachel: **polish**: Declare inline if matching is preserved.

lhearachel (commented):
This generally looks great! Thanks for digging back into the main menu. A few bits of feedback, mostly questions and a bit of polish.

src/main_menu/main_menu.c
Gudf: I'd say there isn't much of a point to putting them in a single header, it's not like anyone will be reusing them. I mostly separated this out for the clarity of "this is where it's defined" but ultimately it's probably not that useful, it's not like these show up in any other file than this one and the one they're defined in anyway...

src/main_menu/ranger_link.c
Gudf: I don't think so, they're used in incompatible overlays so there really shouldn't be any cross-reference

src/main_menu/ranger_link.c
```diff
@@ -0,0 +1,1148 @@
...
+{
+    dest[0] = id / 10000 + 48;
+    id = id % 10000;
+    dest[1] = id / 1000 + 48;
+    id = id % 1000;
+    dest[2] = id / 100 + 48;
+    id = id % 100;
+    dest[3] = id / 10 + 48;
+    id = id % 10;
+    dest[4] = id + 48;
```
Gudf, line 482: **thought**: These `63` and `48` match with the ASCII character codes for the characters they represent (respectively `'?'` and `'0'`), would it make sense to use char literals here instead of the raw number?
Worth noting that this doesn't apply to every character code from the Ranger games.

Gudf: Ended up deciding against that and simply having a few comments to say which character they represent.

lhearachel (approved):

PR #1074, opened 2026-04-08
document overlay 1

scbroede:
very important


src/dummy/dummy.c
```diff
@@ -0,0 +1,13 @@
...
+void Dummy(void)
+{
+    static int dummy = 0;
+}
+
+static void NitroStaticInit(void)
+{
+    static int dummy = 0;
+    dummy++;
+}
```
lhearachel, line 13: **question**: Is the `Dummy` function needed even if you extract `static int dummy = 0;` to outside the function?

scbroede: apparently. i wasn't able to get the bss ordering right any way other than leaving it inside a function

lhearachel (approved):

PR #1075, opened 2026-04-09
Document and rename some TV-related structs

VicSevenT:
This started out as normal documentation of what would become the `TVEpisode` struct. As I was doing this though, I quickly found myself repeatedly confused by all the structs, particularly the existing `TVEpisodeSegment` and `TVProgramSegment` and some others that I identified in this PR. I've opted to rename the former two to try and make their meanings clearer. I did find a comment about the reason behind these names (https://github.com/pret/pokeplatinum/pull/724#discussion_r2423195921), but I'm not satisfied by that especially in light of some of the additional context the newly identified structs in this PR provide.

The existing structs have been renamed:
* `TVProgramSegment` -> `TVSegment`: I considered `TVSegmentFuncs` for this given its contents, but based on how it is used, this sort of serves as the "public" interface for playing a segment's message, so I think the name is fine.
* `TVEpisodeSegment` -> `TVSegmentData`. I know that using `Data` in a struct name isn't preferred, but this struct is ultimately just a massive union of all the different data structs for the different segments. I think this is simpler than trying to explain the difference between `EpisodeSegment` and `ProgramSegment`.

The newly identified structs are:
* `TVSegmentInstance`: A small wrapper around a `TVSegmentData` that stores a segment ID, creation timestamp, and a counter that increments whenever the instance airs.
* `TVEpisode`: This is the main new struct in this PR. I'm not 100% confident in this name for the same reasons as above, that I don't know if it's meaning is clear enough (see below). This is a wrapper around `TVSegmentInstance` that includes additional data such as the player's name, gender, and game language. This it the top level struct used to generate the final message that actually shows when the player interacts with a TV.
* `TVWifiEpisode`: Very similar to `TVEpisode` but comes is built from a different trainer's data when interacted with in the Union room.

Overall this is the hierarchy of terms I'm using related to the TV programming. See [bulbapedia](https://bulbapedia.bulbagarden.net/wiki/TV_programming_in_Sinnoh) for this comparison
* A **Program** is one of the nine main groups of segments (1.x headers in bulbapedia). All segments within a group share the same intro/outro.
* A **Segment** is a specific show within a program (1.x.x headers in bulbapedia). These are templates that store the core structure of what is played on the TV but lack the custom values needed to fully build the message
* An **Episode** is a segment + all the specific info needed to fully play it. This defines the player's general information such as name and gender as well as segment-specific info, such as the player's response to an interview question.

The organization of some of the structs and header files is still a mess, but I didn't want to make this PR any bigger with all of the renaming that I did.


include/tv_segment.h
```diff
@@ -0,0 +1,80 @@
...
+#include "pokemon.h"
+#include "savedata.h"
+#include "string_template.h"
+#include "trainer_info.h"
+
+int TVSegment_LoadMessage(int programTypeID, FieldSystem *fieldSystem, StringTemplate *template, TVEpisode *episode, u16 *bankDestVar);
+BOOL TVSegment_IsEligible(int programTypeID, FieldSystem *fieldSystem, TVEpisode *episode);
+void TVBroadcast_SetContestHallShowInfo(TVBroadcast *broadcast, Pokemon *mon, enum PokemonContestType contestType, enum PokemonContestRank contestRank, int contestPlacement);
+void sub_0206CF48(TVBroadcast *broadcast, Pokemon *param1, enum HeapID heapID);
+void FieldSystem_SaveTVSegment_PlantingAndWateringShow(FieldSystem *fieldSystem, u16 berryItemID, u8 yieldRating, u16 yieldAmount);
```
lhearachel: **suggestion**: Group the header more logically by separating this and the other `Save` routines together in a block.

I would probably also polish it by splitting block-groups with an empty line.

lhearachel (commented):
This looks great! Just one comment, more a suggestion for polish than anything.

I'm not a huge fan of keeping `Data` suffixes for struct type-names, but I agree with you here that it's just a big union. Not really worth futzing over too much in this case.

include/tv_segment.h
VicSevenT: Reorganized it a little. Can probably do more once the remaining functions are identified

lhearachel (approved):

PR #1076, opened 2026-04-10
Document villa furniture

CharlesFolz4:
Document villa furniture


include/struct_defs/villa_persisted_feature.h
Kuruyia, line 1: **suggestion**: I think this can be merged into `villa_furniture.h`.

include/struct_defs/villa_persisted_feature.h
```diff
@@ -0,0 +1,8 @@
+#ifndef POKEPLATINUM_STRUCT_VILLA_PERSISTED_FEATURE_H
+#define POKEPLATINUM_STRUCT_VILLA_PERSISTED_FEATURE_H
+
+typedef struct VillaPersistedFeature {
+    int persistedFeature;
+} VillaPersistedFeature;
```
Kuruyia: **suggestion**: Rename to `VillaPersistedData`

I initially wanted to have the `PersistedMapFeatures_GetBuffer`-related structs all end in `PersistedData`, but I know some of them already fell through the cracks.

```suggestion
typedef struct VillaPersistedData {
    int persistedFeature;
} VillaPersistedData;
```

include/map_object.h
```diff
@@ -25,7 +25,8 @@
 #include "narc.h"
 #include "sys_task_manager.h"
 
-#define MAP_OBJECT_COORD_TO_FX32(coord) ((coord << 4) * FX32_ONE) + (MAP_OBJECT_TILE_SIZE >> 1)
+#define MAP_OBJECT_COORD_TO_FX32(coord)      ((coord << 4) * FX32_ONE) + (MAP_OBJECT_TILE_SIZE >> 1)
+#define ALSO_MAP_OBJECT_COORD_TO_FX32(coord) ((coord << 4) * FX32_ONE)
```
Kuruyia: **question**: Shouldn't `MAP_OBJECT_COORD_TO_FX32` be `MAP_OBJECT_COORD_TO_FX32_CENTERED`?


```suggestion
#define MAP_OBJECT_COORD_TO_FX32(coord)          ((coord << 4) * FX32_ONE)
#define MAP_OBJECT_COORD_TO_FX32_CENTERED(coord) ((coord << 4) * FX32_ONE) + (MAP_OBJECT_TILE_SIZE >> 1)
```

include/struct_defs/villa_persisted_feature.h
```diff
@@ -0,0 +1,8 @@
+#ifndef POKEPLATINUM_STRUCT_VILLA_PERSISTED_FEATURE_H
+#define POKEPLATINUM_STRUCT_VILLA_PERSISTED_FEATURE_H
+
+typedef struct VillaPersistedFeature {
+    int persistedFeature;
```
Kuruyia: **question**: Is this actually referenced anywhere? If not, I think you can rename it to `dummy`.

```suggestion
    int dummy;
```

Kuruyia (changes requested):
Only minor stuff here, this looks very clean!

src/overlay005/villa_furniture.c
```diff
@@ -0,0 +1,544 @@
...
+    if (furnitureCollision->isCollisionEnabled == TRUE && FieldSystem_OwnsVillaFurniture(fieldSystem, furnitureCollision->furniture) == TRUE) {
+        if (tileZ >= collisionBounds->top && tileZ <= collisionBounds->bottom && tileX >= collisionBounds->left && tileX <= collisionBounds->right) {
+            return TRUE;
+        }
+    }
+
+    return FALSE;
+}
+
+static const VillaFurnitureCollision sVillaFurnitureCollision[FURNITURE_SLOTS] = {
```
Kuruyia: **suggestion**: Seeing that it also holds the position of the furniture, the name can be simplified.

```suggestion
static const VillaFurniture sVillaFurnitures[FURNITURE_SLOTS] = {
```

include/map_object.h
Kuruyia: Additionally, we can improve those macros a bit:

```c
#define MAP_OBJECT_COORD_TO_FX32(coord)          (coord * MAP_OBJECT_TILE_SIZE)
#define MAP_OBJECT_COORD_TO_FX32_CENTERED(coord) (MAP_OBJECT_COORD_TO_FX32(coord) + MAP_OBJECT_TILE_SIZE / 2)
```

include/struct_defs/villa_persisted_feature.h
CharlesFolz4: Updated for consistency, but I disagree with this naming convention on grounds that ending any struct name with "-Data" is inherently redundant, outside of very specific circumstances that I don't think applies here.

src/overlay005/villa_furniture.c
CharlesFolz4: `VillaFurniture` is already defined as the name of the enum, the compiler doesn't like it when that symbol gets redefined as a struct here.

include/struct_defs/villa_persisted_feature.h
Kuruyia: The rationale behind suffixing this family of structs with `Data` is that their content are directly written to the save data. So this is a pretty generic system. 

I'm not really attached to that naming convention tho'. We can discuss about that for another PR. As long as the naming stays consistent :)

src/overlay005/villa_furniture.c
Kuruyia: Would it make sense to rename this enum to `VillaFurnitureType`?

src/overlay005/villa_furniture.c
```diff
@@ -0,0 +1,542 @@
...
+typedef struct VillaFurnitureProp {
+    u16 isLoaded;
+    u16 propIdx;
+    u32 modelID;
+    MapProp *mapProp;
+} VillaFurnitureProp;
+
+typedef struct VillaDynamicMapData {
+    FieldSystem *fieldSystem;
+    VillaPersistedData *persistedFeature;
```
Kuruyia: **polish**: Consistency with the rename that was done.

```suggestion
    VillaPersistedData *persistedData;
```

src/overlay005/villa_furniture.c
```diff
@@ -0,0 +1,542 @@
...
+static BOOL CheckFurnitureCollision(int tileX, int tileZ, const VillaFurnitureCollision *furnitureCollision, FieldSystem *fieldSystem);
+
+static const VillaFurnitureCollision sVillaFurnitures[FURNITURE_SLOTS];
+static const u32 sVillaFurnitureModelIDs[VILLA_FURNITURE_MAX];
+static const u32 sVillaFurnitureScriptIDs[VILLA_FURNITURE_MAX];
+
+void Villa_DynamicMapFeaturesInit(FieldSystem *fieldSystem)
+{
+    PersistedMapFeatures *persistedFeatures = MiscSaveBlock_GetPersistedMapFeatures(FieldSystem_GetSaveData(fieldSystem));
+    VillaPersistedData *villaPersistedFeatures = PersistedMapFeatures_GetBuffer(persistedFeatures, DYNAMIC_MAP_FEATURES_VILLA);
```
Kuruyia: **polish**: Consistency with the rename that was done.


```suggestion
    VillaPersistedData *persistedData = PersistedMapFeatures_GetBuffer(persistedFeatures, DYNAMIC_MAP_FEATURES_VILLA);
```

src/overlay005/villa_furniture.c
CharlesFolz4: This whole thing comes together pretty nicely when the enum is VillaFurnitureID.

src/overlay005/villa_furniture.c
```diff
@@ -0,0 +1,542 @@
...
+
+typedef struct FurnitureCollisionBoundingBox {
+    s16 left;
+    s16 top;
+    s16 right;
+    s16 bottom;
+} FurnitureCollisionBoundingBox;
+
+typedef struct VillaFurniture {
+    u16 furnitureID;
```
Kuruyia: **nitpick**: It can be even nicer :)

```suggestion
    u16 id;
```

CharlesFolz4: Given that the index of `sVillaFurnitures` doesn't always match up with the furnitureIDs, I think it's better to keep this field explicit to avoid any potential confusion.

Kuruyia: I think the initial `Type` suggestion would be more appropriate here then. Also because I see that in the mentioned array there are multiple entries with the same `furnitureID`.

CharlesFolz4: You're right.  Updated.

Kuruyia (approved):

PR #1077, opened 2026-04-11
tools: Implement dexproc for generating regional-national Pokedex mappings

lhearachel:


res/pokemon/meson.build
```diff
@@ -244,6 +244,25 @@ species_data_outputs = species_data.to_list()
 h_headers += species_data_outputs.slice(9, 15)
 species_text_banks = species_data_outputs.slice(15, species_data_outputs.length())
 
+pokedex_listings = custom_target('pokedex_listings',
+    output: [
+        'pl_pokezukan.narc',
+        'shinzukan.narc',
```
scbroede: while we're at it do we want to give these narcs better names?

lhearachel: Probably! I just didn't. I'm thinking `national_to_regional.narc` and `regional_to_national.narc`? Is that sufficiently descriptive?

We probably should also come up with descriptive names for the `speciesproc` NARCs.

Tai-disc: Do you want someone to translate them to English? Seems to me that "zukan" comes from "図鑑" meaning "picture / illustrated reference book" (according to jisho: https://jisho.org/search/zukan).

scbroede: works for me

lhearachel: > Do you want someone to translate them to English?

Appreciated, though we know that `zukan` generally refers to the Pokedex.

scbroede (approved):

PR #1079, opened 2026-04-11
document yes-no touch menu

scbroede:
image


src/yes_no_touch_menu.c
```diff
@@ -0,0 +1,414 @@
...
+
+    if (input == YES_NO_TOUCH_MENU_NOTHING_CHOSEN) {
+        if (menu->buttonState == TOUCH_BUTTON_PRESSED) {
+            if (menu->pressedButton == BUTTON_YES) {
+                return YES_NO_TOUCH_MENU_YES_INSTANT;
+            } else {
+                return YES_NO_TOUCH_MENU_NO_INSTANT;
+            }
+        }
+    }
```
lhearachel: **polish**:

```suggestion
    if (input == YES_NO_TOUCH_MENU_NOTHING_CHOSEN
        && menu->buttonState == TOUCH_BUTTON_PRESSED) {
        if (menu->pressedButton == BUTTON_YES) {
            return YES_NO_TOUCH_MENU_YES_INSTANT;
        } else {
            return YES_NO_TOUCH_MENU_NO_INSTANT;
        }
    }
```

Or, if it matches:

```suggestion
    if (input == YES_NO_TOUCH_MENU_NOTHING_CHOSEN
        && menu->buttonState == TOUCH_BUTTON_PRESSED) {
        return menu->pressedButton == BUTTON_YES
            ? YES_NO_TOUCH_MENU_YES_INSTANT
            : YES_NO_TOUCH_MENU_NO_INSTANT;
    }
```

src/yes_no_touch_menu.c
```diff
@@ -0,0 +1,414 @@
...
+
+static void YesNoTouchMenu_LoadPltt(YesNoTouchMenu *menu, const YesNoTouchMenuParams *params)
+{
+    u32 screen;
+
+    if (menu->bgLayer < BG_LAYER_SUB_0) {
+        screen = SCREEN_MAIN;
+    } else {
+        screen = SCREEN_SUB;
+    }
```
lhearachel: **suggestion**:

```suggestion
    u32 screen = menu->bgLayer < BG_LAYER_SUB_0 ? SCREEN_MAIN : SCREEN_SUB;
```

src/yes_no_touch_menu.c
```diff
@@ -0,0 +1,414 @@
...
+        tilemap[i] += baseTile;
+    }
+}
+
+static void YesNoTouchMenu_ApplyPaletteToTiles(const NNSG2dScreenData *screenData, int palette)
+{
+    int i;
+    int size = screenData->szByte / sizeof(u16);
+
+    u16 *tilemap = (u16 *)screenData->rawData;
```
lhearachel, line 392: **thought**: (out-of-scope for this PR) I wonder if we can / should use `void *` on the struct in NitroSystem, which would allow us to get rid of casts like this one.

scbroede: just saw that it's currently defined as `u32 rawData[1]` and yeah, i think this would be an improvement over that

lhearachel (approved):

PR #1085, opened 2026-04-12
Rename "unk_0202EEC0" to "pal_park_transfers"

Tai-disc:
I renamed each instance of `unk_0202EEC0` to `pal_park_transfers` and named the variable `param2` in `unk_0202EEC0.h` to `slot`.


lhearachel (approved):
Thanks for the contribution! I have no qualms here. 🙂

PR #1087, opened 2026-04-13
Refactored `unk_0202C858` to `wifi_history_save_data`

Tai-disc:
I refactored each instance of `unk_0202C858` to `wifi_history_save_data`. Also, I updated a few parameters in the `wifi_history_save_data.h` according to its c file (+ removed an unused variable).


Tai-disc:
I also refactored some other function and variables. I would have made a separate pull request for those, but I didn't know that my latest push would be included here as well. Sorry for that ^^::

lhearachel (approved):

PR #1088, opened 2026-04-13
Document distortion world B4F/B5F waterfall

Kuruyia:
This continues documentation of the distortion world overlay by documenting the waterfall between B4F and B5F, and the surf mount field effect renderer that is used when surfing in the distortion world.

Please check individual commits for more details.


src/overlay005/dist_world_surf_mount_renderer.c
```diff
@@ -0,0 +1,379 @@
...
+
+        OverworldAnimManager_SetPosition(animMan, &pos);
+    }
+}
+
+static void DistWorldSurfMountRenderer_AnimRender(OverworldAnimManager *animMan, void *context)
+{
+    DistWorldSurfMountRenderer *renderer = context;
+
+    if ((renderer->flags & DIST_WORLD_SURF_MOUNT_RENDERER_FLAG_MASK_NO_RENDER) == 0) {
```
lhearachel: **question**: Does this match as `!` instead of `== 0`?

lhearachel (commented):
Looks great! Just one question to maybe polish it up a bit.

src/overlay005/dist_world_surf_mount_renderer.c
Kuruyia: All but one match. I pushed the change regardless, do we keep it?

lhearachel: Yep, I like keeping them, even if it looks inconsistent.

lhearachel (approved):

PR #1090, opened 2026-04-15
Rename overworld anim manager ID field

Kuruyia:
This renames the `id` field of the overworld animation manager to `userInt`.

This field was initially documented as containing some sort of caller-defined identifier based on how it was being used in the HM cut-in file, but it's become more clear that the entire meaning of this field is caller-defined (see its use by the surf mount renderers as a boolean).

This also updates the surf mount renderer to correctly identify a flag that's used to control whether the surf mount position is synchronized with the player position on tick.

The same changes were ported to the distortion world surf mount renderer where applicable.


ravepossum (approved):
Looks good!

PR #1091, opened 2026-04-17
docs: Add persistent wineserver instructions for macOS

Kuruyia:
This adds instructions to start and stop a persistent Wine server for macOS users, to help with the slow build times.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -187,6 +187,26 @@ GNU `coreutils` installed to run the build scripts:
 
 6. [Download the repository](#2-downloading-the-repository).
 
+> [!TIP]
+> You can run a persistent Wine server in the background to speed up builds.
+>
+> To do this, open the Wine Stable application from your Applications folder.
+> In the terminal window that opens, run the following command:
+>
+> ```zsh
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -187,6 +187,26 @@ GNU `coreutils` installed to run the build scripts:
 
 6. [Download the repository](#2-downloading-the-repository).
 
+> [!NOTE]
```
lhearachel: **nitpick**: I would use a `[!TIP]` here instead of a `[!NOTE]`, based on [the original `community` discussion post](https://github.com/orgs/community/discussions/16925).

> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

lhearachel (commented):
Looks great, just a single nitpick.

INSTALL.md
Kuruyia: We success-maxing with this one

lhearachel: *success-maxxing

lhearachel (approved):

PR #1092, opened 2026-04-18
Document distortion world scripts

Viperio19:
This PR documents the script and text files for the distortion world. All the events are handled in `ov9_02249960.c`. This is also where the object event ids are defined as enums. From what I know and tested, I cannot simply import those enums and use them for the scripts. Because of that, I have currently hardcoded the local ids in each script file. The one proposed solution is replacing the `MapObjectEvent...LocalID` enums in `ov9_02249960.c` with `#define`d constants so these can be used in both `ov9_02249960.c` and in the script files. This works, but the downside is that it's ugly, especially for maps with a lot of object_events:
```
#define DIST_WORLD_B6F_LOCALID_MESPRIT_BOULDER_OUTSIDE    DIST_WORLD_BASE_LOCALID
#define DIST_WORLD_B6F_LOCALID_AZELF_BOULDER_OUTSIDE      DIST_WORLD_B6F_LOCALID_MESPRIT_BOULDER_OUTSIDE + 1
#define DIST_WORLD_B6F_LOCALID_UXIE_BOULDER_OUTSIDE       DIST_WORLD_B6F_LOCALID_AZELF_BOULDER_OUTSIDE + 1
#define DIST_WORLD_B6F_LOCALID_MESPRIT                    DIST_WORLD_B6F_LOCALID_UXIE_BOULDER_OUTSIDE + 1
#define DIST_WORLD_B6F_LOCALID_UXIE                       DIST_WORLD_B6F_LOCALID_MESPRIT + 1
...
```
I have a stash for this solution that I could apply, but I'm interested to hear if there are potantially better solutions.

### DIFF
--- a/res/field/scripts/scripts_distortion_world_1f.s
+++ b/res/field/scripts/scripts_distortion_world_1f.s
@@ -1,6 +1,7 @@
 #include "macros/scrcmd.inc"
 #include "res/text/bank/distortion_world_1f.h"
 
+// NOTE: These constants must mirror those in ov9_02249960.c
 #define LOCALID_CYNTHIA 128
 
     ScriptEntry DistortionWorld1F_OnTransition
--- a/res/field/scripts/scripts_distortion_world_1f.s
+++ b/res/field/scripts/scripts_distortion_world_1f.s
@@ -1,31 +1,32 @@
 #include "macros/scrcmd.inc"
 #include "res/text/bank/distortion_world_1f.h"
 
+#define LOCALID_CYNTHIA 128
 
-    ScriptEntry _0016
-    ScriptEntry _001A
-    ScriptEntry _0070
-    ScriptEntry _011D
-    ScriptEntry _012A
+    ScriptEntry DistortionWorld1F_OnTransition
+    ScriptEntry DistortionWorld1F_Portal
+    ScriptEntry DistortionWorld1F_OnFrameFirstEntry
+    ScriptEntry DistortionWorld1F_TriggerCynthiaElevator
+    ScriptEntry DistortionWorld1F_CynthiaElevator
     ScriptEntryEnd
 
-_0016:
+DistortionWorld1F_OnTransition:
     InitPersistedMapFeaturesForDistortionWorld
     End
 
-_001A:
+DistortionWorld1F_Portal:
     PlaySE SEQ_SE_CONFIRM
     LockAll
-    Message 7
+    Message DistortionWorld1F_Text_ReturnToSpearPillar
     ShowYesNoMenu VAR_RESULT
-    GoToIfEq VAR_RESULT, MENU_YES, _003A
+    GoToIfEq VAR_RESULT, MENU_YES, DistortionWorld1F_ReturnToSpearPillar
     CloseMessage
     ReleaseAll
     End
 
-_003A:
+DistortionWorld1F_ReturnToSpearPillar:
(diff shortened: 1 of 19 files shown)


lhearachel (commented):
The changes as-is look fine to me. Some notes from your PR description:

> From what I know and tested, I cannot simply import those enums and use them for the scripts.

That's correct. Traditional assemblers do not have any syntax for `enum`.

> I have currently hardcoded the local ids in each script file. The one proposed solution is replacing the `MapObjectEvent...LocalID` enums in `ov9_02249960.c` with `#define`d constants so these can be used in both `ov9_02249960.c` and in the script files. This works, but the downside is that it's ugly, especially for maps with a lot of object_events.

I think that the hard-coded solution is fine, for now, but I'd like to hear @Kuruyia's thoughts on it. I personally don't mind ugly glue-code when we have a theoretical path-forward in mind, and a hypothetical `enumproc` as a preprocessing layer is on my roadmap.

If we decide to keep this glue-code in place, do we want to have a `TODO` to revisit it?

Kuruyia:
I'm fine with the current implementation for now. We'll definitely need a way to sync those constants between ov9 and the scripts in the future tho'.

I'd be in favor of leaving a big TODO for this, or even open a GitHub issue, to not get this lost in limbo.

lhearachel:
> I'd be in favor of leaving a big TODO for this, or even open a GitHub issue, to not get this lost in limbo.

We could leave `NOTE` comments on top of each block of `#define`s and the parent `enum`(s) that leave hints, in case anyone is trying to edit these at the moment.

Viperio19:
> We could leave `NOTE` comments on top of each block of `#define`s and the parent `enum`(s) that leave hints, in case anyone is trying to edit these at the moment.

Sounds good, I'd suggest something like below. Would that be sufficient?
```
// NOTE: These ids are defined as enums in ov9_02249960.c but there is currently no way to include enums
#define LOCALID_.... 128
...
```
```c
// NOTE: These ids are redefined locally in scripts_distortion_world_[floor].s because there is currently no way to include enums
enum MapObjectEvent1FLocalID {
    MAP_OBJECT_1F_CYNTHIA_PORTAL = MAP_OBJECT_BASE_LOCAL_ID,
    MAP_OBJECT_1F_CYNTHIA_ELEVATOR,
};
```

lhearachel:
I think this would be preferable:

```c
// NOTE: These constants must mirror those in [filename here]
```

Viperio19:
> I think this would be preferable:
> 
> ```c
> // NOTE: These constants must mirror those in [filename here]
> ```

Done!

res/field/scripts/scripts_distortion_world_1f.s
```diff
@@ -1,31 +1,33 @@
 #include "macros/scrcmd.inc"
 #include "res/text/bank/distortion_world_1f.h"
 
-
-    ScriptEntry _0016
-    ScriptEntry _001A
-    ScriptEntry _0070
-    ScriptEntry _011D
-    ScriptEntry _012A
+@ NOTE: These constants must mirror those in ov9_02249960.c
```
lhearachel: **polish**: Use `gcc`-style comments (with `//`, not `@`).

Viperio19: fixed

lhearachel (approved):

PR #1094, opened 2026-04-19
document trainer card signing

scbroede:
couldn't use the cell for the done button image because it overlaps itself. might revisit if we come up with a good solution for that


res/graphics/signature_interface/done_button.png
lhearachel, line 1: **question**: Is there no way to clean up this PNG?

src/applications/signature.c
```diff
@@ -0,0 +1,1137 @@
...
+
+    PlttTransfer_Init(20, HEAP_ID_DRAWING);
+    CharTransfer_ClearBuffers();
+    PlttTransfer_Clear();
+}
+
+static void SignatureApp_LoadButtonSpriteResources(SignatureApp *signatureApp, NARC *narc)
+{
+    NNS_G2dInitOamManagerModule();
+    RenderOam_Init(0, 126, 0, 32, 0, 126, 0, 32, HEAP_ID_DRAWING);
```
lhearachel: **thought**: Super odd that they used `126` here instead of the standard `128`. I wonder if that's a bug.

---

**suggestion**: In a similar vein, you could use constants here like `SetupOAM` in `src/choose_starter/choose_starter_app.c` does.

---

**thought**: Out-of-scope for this PR, but we should absolutely standardize this function (and others like it) as `RenderOAM` rather than `RenderOam`.

src/applications/signature.c
```diff
@@ -0,0 +1,1137 @@
...
+}
+
+static void SignatureApp_AnimateButtonsUnused(Sprite **buttonSprites, int pressedButtonIndex)
+{
+    for (int i = 0; i < 1; i++) {
+        if (i == pressedButtonIndex) {
+            Sprite_SetAnim(buttonSprites[i], sDoneButtonSpriteData[i].animID + 1);
+        } else {
+            Sprite_SetAnim(buttonSprites[i], sDoneButtonSpriteData[i].animID);
+        }
```
lhearachel, line 702: **question**: I highly doubt it, but does this optimize as:


```suggestion
        Sprite_SetAnim(buttonSprites[i], sDoneButtonSpriteData[i].animID + (i == pressedButtonIndex));
```

res/graphics/signature_interface/done_button.png
scbroede: not really. it has 178 tiles, so any width higher than 2 tiles and less than 89 doesn't divide evenly, which results in empty tiles padding out the rest of the bottom row that prevent matching

src/applications/signature.c
scbroede: it does not

lhearachel (approved):

PR #1100, opened 2026-04-20
Name unnamed struct files with named structs

Viperio19:
This PR mainly names all unnamed files in `struct_decls/` and `struct_defs/` for named structs. It also removes some unnecessary structs (`UnkStruct_0209BDF8_sub1` which only contained one `int` field) and `struct_defs/` files (`struct_020147B8` which was only included in `lib/spl/include/spl_resource.h`, so the contents are moved there, and some files without any contents at all).


lhearachel (changes requested):
I would rather we make an earnest attempt to deconstruct these files than introduce another layer of changes to name files that we don't want to be around long-term.

Viperio19:
> I would rather we make an earnest attempt to deconstruct these files than introduce another layer of changes to name files that we don't want to be around long-term.

Is the goal to eventually get rid of `struct_decls/` and `struct_defs/` entirely? If that's the case I can agree with that, but either way I believe that this PR just introduces some changes that should have been made when naming the relevant structs in the first place. I also think that it's helpful to be able to see that a file includes the `MapObject` struct decl for example, so if `struct_decls/` and `struct_defs/` aren't on someone's TODO list, I don't see an issue with some improvements to the current names.

lhearachel:
> Is the goal to eventually get rid of `struct_decls/` and `struct_defs/` entirely?

Yes, as much as is possible.

> Either way I believe that this PR just introduces some changes that should have been made when naming the relevant structs in the first place.

I don't disagree with the notion, but I would like to push for more proper merging of structs into associated headers, if that makes sense. Just renaming the `decl` / `def` file is a stop-gap solution that is fine for an accompanying documentation PR or set of PRs that are in-progress, but this is a standalone.

Viperio19:
> I don't disagree with the notion, but I would like to push for more proper merging of structs into associated headers, if that makes sense. Just renaming the `decl` / `def` file is a stop-gap solution that is fine for an accompanying documentation PR or set of PRs that are in-progress, but this is a standalone.

Fair, that makes sense to me.

lhearachel:
> Fair, that makes sense to me.

To clarify, I'm still happy to review this one, if you'd like to push for it (and it sounds like you do). This is mostly a nudge for future PRs. 😅

Viperio19:
> To clarify, I'm still happy to review this one, if you'd like to push for it (and it sounds like you do). This is mostly a nudge for future PRs. 😅

I always push for fully understanding the reasoning behind something😅. Reviewing this and then not having any future PRs for just renaming `decl` / `def` files sounds totally fine to me!

lhearachel (approved):

PR #1101, opened 2026-04-22
Document victory road and routes 223 and 224 scripts

Viperio19:
This PR documents the event, script and text files for victory road and routes 223 and 224 (route 223 actually only has an event file).


include/constants/sound_volume.h
```diff
@@ -0,0 +1,7 @@
+#ifndef POKEPLATINUM_CONSTANTS_SOUND_VOLUME_H
+#define POKEPLATINUM_CONSTANTS_SOUND_VOLUME_H
+
+#define SOUND_VOLUME_MIN 0
+#define SOUND_VOLUME_MAX 127
```
lhearachel, line 5: **suggestion**: Leave a comment that these are separate (for now) because `include/constants/sound.h` has an `enum` and thus cannot be used in scripts.

include/applications/naming_screen.h
```diff
@@ -13,7 +13,7 @@ enum NamingScreenType {
     NAMING_SCREEN_TYPE_RIVAL,
     NAMING_SCREEN_TYPE_UNK4,
     NAMING_SCREEN_TYPE_GROUP,
-    NAMING_SCREEN_TYPE_UNK6,
+    NAMING_SCREEN_TYPE_TABLET,
```
lhearachel: **thought**: I wonder how descriptive this is. I didn't know what this was for on first-glance until remembering the big white rock for unlocking Seabreak Path.

lhearachel (commented):
Congrats on finishing the main-line game! 🎉 These look good, just a couple notes for polish.

include/constants/sound_volume.h
Viperio19: There were constants for it in `include/sound.h`, but I moved those to this separate file to include it. I can't find the `enum` that you are referring to, `include/constants/sound.h` just contains `PokemonCryMod`.

include/applications/naming_screen.h
Viperio19: That makes sense, this name was mainly just based on `MiscSaveBlock_SetTabletName`. I like scbroede's suggestion of `NAMING_SCREEN_TYPE_SEABREAK_TABLET`, although it might be slightly misleading since it's not on Seabreak path itself, so `NAMING_SCREEN_TYPE_ROUTE_224_TABLET` would be more accurate (but slightly less descriptive again).

Viperio19: Renamed it to `NAMING_SCREEN_TYPE_SHAYMIN_TABLET `

lhearachel (approved):

PR #1108, opened 2026-04-24
Refactor `struct_ov61_0222AFCC.h` and `struct_02030A80.h`.

Tai-disc:
- Documented member variables for `struct_ov61_0222AFCC.h` and `struct_02030A80.h`.
- Refactored variables accordingly in `ov61_0222AE60.c` and `unk_02030A80.c`.
- Renamed variables in `ov5_021D1A94.c`.


src/unk_02030A80.c
```diff
@@ -83,60 +84,60 @@ String *sub_02030B94(const UnkStruct_02030A80 *param0, enum HeapID heapID)
 
 u32 sub_02030BAC(const UnkStruct_02030A80 *param0)
 {
-    if ((param0->unk_14 != 0) && (param0->unk_14 != 1)) {
+    if ((param0->gender != GENDER_MALE) && (param0->gender != GENDER_FEMALE)) {
         return 0;
```
lhearachel: **polish**: Return `GENDER_MALE` (which is 0) here as the default.

src/unk_02030A80.c
```diff
@@ -83,60 +84,60 @@ String *sub_02030B94(const UnkStruct_02030A80 *param0, enum HeapID heapID)
...
 
-    return param0->unk_14;
+    return param0->gender;
 }
 
 int sub_02030BBC(const UnkStruct_02030A80 *param0)
 {
-    if (param0->unk_1C >= 495) {
+    if (param0->species >= MAX_SPECIES) {
         return 0;
```
lhearachel: **polish**: `return SPECIES_NONE;`

src/unk_02030A80.c
```diff
@@ -83,60 +84,60 @@ String *sub_02030B94(const UnkStruct_02030A80 *param0, enum HeapID heapID)
...
 
-    return param0->unk_1C;
+    return param0->species;
 }
 
 int sub_02030BCC(const UnkStruct_02030A80 *param0)
 {
-    if (param0->unk_1C >= 495) {
+    if (param0->species >= MAX_SPECIES) {
         return 0;
```
lhearachel: **polish**: `return SPECIES_NONE;`

src/unk_02030A80.c
```diff
@@ -83,60 +84,60 @@ String *sub_02030B94(const UnkStruct_02030A80 *param0, enum HeapID heapID)
...
-    return Pokemon_SanitizeFormId(param0->unk_1C, param0->unk_1B_1);
+    return Pokemon_SanitizeFormId(param0->species, param0->form);
 }
 
 int sub_02030BEC(const UnkStruct_02030A80 *param0)
 {
-    if (param0->unk_1B_0 > 1) {
+    if (param0->isEgg > 1) {
         return 1;
     }
```
lhearachel, line 118: **thought**: Isn't this impossible, since it's a single-bit? What happens if you just remove this branch...?

src/unk_02030A80.c
```diff
@@ -83,60 +84,60 @@ String *sub_02030B94(const UnkStruct_02030A80 *param0, enum HeapID heapID)
...
-    return param0->unk_1B_0;
+    return param0->isEgg;
 }
 
 int sub_02030BFC(const UnkStruct_02030A80 *param0)
 {
-    if (param0->unk_17 >= 234) {
+    if (param0->country >= 234) {
         return 0;
     }
```
lhearachel, line 128: **thought**: We may want to output an additional constant from `msgenc` that lists the number of entries in the bank. That's out-of-scope for this PR, but can you leave a comment here noting that `234` is the count of entries in the `country_names` text bank?

---

**polish**: Related to the above, return `Country_Text_None` here instead of a bare `0`. You'll also need to `#include "res/text/bank/country_names.h"`.

src/unk_02030A80.c
```diff
@@ -176,18 +177,18 @@ String *sub_02030C28(const UnkStruct_02030A80 *param0, Sentence *param1, enum He
...
 
     return 1;
 }
 
 int sub_02030CDC(const UnkStruct_02030A80 *param0)
 {
-    if (param0->unk_16 > 15) {
+    if (param0->appearance > 15) {
         return 0;
     }
```
lhearachel: **question**: Does this match if you do `>= 16` instead of `> 15`? That matters for my suggestion here.

---

**polish**: Return `TRAINER_APPEARANCE_SCHOOL_KID_M` instead of `0` (from `#include "appearance.h"`).

Tai-disc: Unfortunately, it doesn't match. Is it still ok if I return `TRAINER_APPEARANCE_SCHOOL_KID_M`?

src/unk_02030A80.c
Tai-disc: You're right! However, only the checksum for the filesystem passes if I remove it. :(

lhearachel: Dang, I was hoping the compiler would optimize this out. Guess it doesn't see that it's impossible.

src/unk_02030A80.c
lhearachel: Yes, please do. Let's also replace the 15 with `TRAINER_APPEARANCE_LADY`.

Tai-disc: Done! Thank you for the feedback, Ihearachel! 😊

lhearachel (approved):

PR #1110, opened 2026-04-25
tools: Implement a preprocessor for C-enums in bytecode scripts

lhearachel:
This is the successor to how we generate constants from text files via `metang`. After this PR is accepted, no new entries should be added to `generated` unless they are absolutely required for a Python script. There is some friction there, which I see as dissuading the use of Python scripts in general for data processing. We may also want to write some Python-bindings for `libenum` so that such scripts can use that library, but that feels out-of-scope here.

Any non-`generated` constants used by `dataproc` after this library is merged should follow the paradigm set by `resdatproc`'s import of `constants/narc.h`. This will load the `enum`'s members into a lookup table dynamically.

C-`enum`s should *just* be able to be included in bytecode scripts with this tool and work as-is. The library includes an expression parser and evaluator, which is primarily for bitmask-type `enum`s, but implements a large subset of C's rules for integer-constant expressions.


tools/enumproc/libenum.c
```diff
@@ -0,0 +1,679 @@
...
+    return str;
+}
+
+static inline bool is_alpha(char c) {
+    return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z');
+}
+
+static inline bool is_digit(char c) {
+    return (c >= '0' && c <= '9');
+}
```
Kuruyia: **question**: Why not use `<ctype.h>`?

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,679 @@
...
+
+static inline bool is_ident(char c) {
+    return is_alpha(c) || is_digit(c) || c == '_';
+}
+
+/* ========================== EXPRESSION EVALUATOR ========================== */
+
+typedef struct stack stack_t;
```
Kuruyia: **issue**: Another day, another issue with macOS. `stack_t` is defined by the macOS SDK.

```
FAILED: [code=1] tools/enumproc/liblibenum.a.p/libenum.c.o
gcc-14 -Itools/enumproc/liblibenum.a.p -Itools/enumproc -I../tools/enumproc -fdiagnostics-color=always -std=gnu17 -O3 -Wall -Wextra -Wpedantic -Wconversion -Wno-sign-conversion -Werror -MD -MQ tools/enumproc/liblibenum.a.p/libenum.c.o -MF tools/enumproc/liblibenum.a.p/libenum.c.o.d -o tools/enumproc/liblibenum.a.p/libenum.c.o -c ../tools/enumproc/libenum.c
../tools/enumproc/libenum.c:50:22: error: conflicting types for 'stack_t'; have 'struct stack'
   50 | typedef struct stack stack_t;
      |                      ^~~~~~~
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/signal.h:150,
                 from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:109,
                 from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:70,
                 from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58,
                 from ../tools/enumproc/libenum.c:6:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/_types/_sigaltstack.h:48:33: note: previous declaration of 'stack_t' with type 'stack_t'
   48 | typedef _STRUCT_SIGALTSTACK     stack_t; /* [???] signal stack */
      |
```

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,679 @@
...
+        if (str[0] == '/' && str[1] == '/') {
+            str += 2;
+            while (*str && *str != '\r' && *str != '\n') str++;
+            continue;
+        }
+
+        if (str[0] == '/' && str[1] == '*') {
+            str += 2;
+            while (*str && *str != '*' && str[1] && str[1] != '/') str++;
+            str += 2;
```
Kuruyia: **issue**: If we stopped because of a `\0` character, this goes past the buffer.

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,679 @@
...
+        case OP_BNT: res = ~res;                 break;      \
+                                                             \
+        default: abort();                                    \
+        }                                                    \
+                                                             \
+        push(stk, nums, res);                                \
+    } while (0)
+
+// ref: https://en.wikipedia.org/wiki/Shunting_yard_algorithm
+long eval(char *expr, char **endptr, char *terminators, enum_t *scope) {
```
Kuruyia: **issue**: When testing this, I got 6 as a result for `1 + 2 * 2`. I couldn't really pinpoint the problem to a specific place.

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,679 @@
...
+
+            n = false;
+            u = false;
+            p++;
+            break;
+
+        default: goto early_exit;
+        }
+
+        p = skip(p);
```
Kuruyia: **issue**: If the `p` pointer contains a newline (which happens if the line unexpectedly ends with no terminator or a left parenthesis), then `skip` will start reading the next line.

For example:
```c
enum {
    TEST_1 = 1)
    TEST_2,
    TEST_3
};
```

Gives:
```
#define TEST_1 1
#define EST_2 2
#define TEST_3 3
```

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,679 @@
...
+
+        setup_member();
+
+        switch (*p) {
+        case '}': state       = STATE_DONE; // fall-through
+        case ',': curr->value = next++;     break;
+
+        case '=': {
+            if (flags & LIBENUM_F_EVAL) {
+                curr->value = eval(skip(p + 1), &p, ",}", p_scope);
```
Kuruyia: **issue**: The `p` pointer returned here points to the character after the expression, which will be skipped by the `p++` operation afterwards.

This allows such enums:
```c
enum {
    TEST_1 = 1?
    TEST_2,
    TEST_3
};
```

tools/enumproc/enumproc.c
```diff
@@ -0,0 +1,104 @@
...
+    char *content = argc == 1 ? slurp_stdin() : slurp_file(argv[1]);
+    char *endptr  = NULL;
+    char *p       = content;
+
+    while (*p) {
+        while (isws(*p)) { fputc(*p, stdout); p++; }
+
+        if (strncmp(p, "enum", 4) == 0) {
+            enum_t parsed = enum_map_one(p, &endptr, LIBENUM_F_EVAL, NULL, NULL);
+            if (*endptr != ';') goto put_line; // probably not a C enum
```
Kuruyia: **suggestion**: Exit early and don't print anything if parsing failed.

```suggestion
            enum_t parsed = enum_map_one(p, &endptr, LIBENUM_F_EVAL, NULL, NULL);
            if (parsed.size == 0) exit(EXIT_FAILURE);
            if (*endptr != ';') goto put_line; // probably not a C enum
```

Kuruyia (changes requested):
RIP metang

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,679 @@
...
+
+            state = STATE_DONE;
+            p     = skip(p + 1);
+            break;
+        }
+
+        member_t *curr;
+        char     *e;
+
+        setup_member();
```
Kuruyia: **question**: Do we want to make sure that the enum variant we're currently handling wasn't previously defined?

tools/enumproc/libenum.c
lhearachel: Locales; I prefer not to deal with them where possible, and these functions are trivial to define.

tools/enumproc/libenum.c
lhearachel: Finally punished for my use of `_t`. :sweat_smile:

tools/enumproc/enumproc.c
lhearachel: We explicitly do *not* want to do that because of how we use this tool: it is a mid-preproc translation utility for ASM scripts.

Exiting here would terminate the build in the places that we define enumerations for the script-command tables (because those emit `enum SCRCMD_...` when invoked by the assembler). We could change that behavior, of course.

tools/enumproc/libenum.c
lhearachel: I think that's a valuable control-flag. :slightly_smiling_face:

tools/enumproc/enumproc.c
Kuruyia: Ahh gotcha. I originally asked this question because giving the tool a malformed enum results in an abruptly truncated output.

Consider the following input:
```c
// Some stuff

enum {
    TEST_1 = 1,
    TEST_2 =
};

// Some other stuff
```

The tool outputs:
```
// Some stuff

enum 
```

I believe this is because we put a `\0` character after the `enum` keyword/enum identifier.

Should `enumproc` consider the state of the parser to determine whether to continue or fail? e.g. we expect to fail during `STATE_DECL` because of the `enum SCRCMD_...` case you talked about, but not during `STATE_READ` because we consider at that point that we're parsing a valid C enum.

At the very least, I think it is worth including such information (or an equivalent; I understand we'd want to keep the parser state as internal machinery) in `enum_t` so that the caller (`enumproc` or another) can determine whether it wants to fail or not.

tools/enumproc/libenum.c
Kuruyia: Fair enough!

tools/enumproc/libenum.c
lhearachel: This actually uncovered a bigger bug: the logic for detecting the tail of the comment is wrong. Here's the fixed version:

```c
        if (str[0] == '/' && str[1] == '*') {
            const char *start = str;

            str += 2;
            while (*str && str[1] && !(*str == '*' && str[1] == '/')) str++;
            if    (!*str || !str[1]) { str = start; break; } // unterminated
            str += 2;
            continue;
        }
```

tools/enumproc/libenum.c
lhearachel: The logic for `precedes` was inverted. I updated the code at some point, but not the macro. 😓

Kuruyia: Ah! Glad it wasn't an annoying bug to fix at least.

tools/enumproc/enumproc.c
lhearachel: Yeah, I think it's a good idea for the `enum_t` structure to include some additional information about the ultimate result-state. I'll add that in.

lhearachel: Ah, I actually considered this, I just never implemented it! If `result.members == NULL`, then one of the following must be true:

1. The declaration is not valid C,
2. Some member in the definition is invalid,
3. Some syntax in the definition (either expression-syntax or an unexpected token) is invalid, or
4. The `enum` has no members (which is invalid in standard C).

We will want to remove our assembler-macros as false-positives on case (1). I believe the easiest way for us to do that is to change our macros in the assembler to have prefixes so that they do not get sent to the library at all.

Implementing these detections as an error-case inside of `enumproc` (the tool) will require some additional sensitivity to the input stream so that we can report the error intelligently. The preprocessor output from `arm-none-eabi-gcc`, luckily, includes the source filename, which we can keep for ourselves for tracking.

lhearachel: I've implemented some proper error-handling. Notably, I did not include duplicate member-detection; we can leave that to the assembler.

tools/enumproc/libenum.c
Kuruyia: There's still something broken in here unfortunately: `2 + 3 * 10 / 10` evaluates to 3 instead of the expected 5.

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,725 @@
...
+        switch (*p) {
+        case '}': state       = STATE_DONE; // fall-through
+        case ',': curr->value = next++;     break;
+
+        case '=': {
+            if (flags & LIBENUM_F_EVAL) {
+                curr->value = eval(skip(p + 1), &p, ",}", p_scope);
+                next        = curr->value + 1;
+
+                if (*p != '}' && *p != ',') return_error(LIBENUM_R_TOKEN);
```
Kuruyia: **issue**: If `*p == '}'`, then we need to finish parsing this enum. Current behavior treats `}` and `,` the same and continues parsing.

For example:
```c
enum {
    TEST_1}
    TEST_2,
    TEST_3
};
```

Outputs:
```
#define TEST_1 1
#define TEST_2 2
#define TEST_3 3
```

I think the bigger issue here is an architectural one. The terminator token should be consumed by the two cases we already have defined above.

**suggestion**: Alternatively, I think such an implementation could make sense, although the non-`LIBENUM_F_EVAL` branch is still buggy:

```c
// ...
setup_member();

if (*p == '=') {
    if (flags & LIBENUM_F_EVAL) {
        curr->value = eval(skip(p + 1), &p, ",}", p_scope);
        next        = curr->value + 1;
    } else {
        p          = skip(p + 1);
        curr->expr = p;
        while (*p && *p != ',' && *p != '}') p++;
        *p = '\0';
        p++;
    }
} else {
    curr->value = next++;
}

switch (*p) {
case '}': state = STATE_DONE; // fall-through
case ',': break;
default:  return_error(LIBENUM_R_TOKEN);
}
// ...
```

tools/enumproc/libenum.h
```diff
@@ -0,0 +1,111 @@
...
+    LIBENUM_R_MEMB_EXISTS,
+    LIBENUM_R_ENUM_EXISTS,
+};
+
+typedef struct member_t member_t;
+struct member_t {
+    union {
+        long  value;
+        char *expr;
+    };
```
Kuruyia: **question**: If the `LIBENUM_F_EVAL` flag is not set, how do we know whether `value` or `expr` is the valid one?

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,725 @@
...
+            } else {
+                p          = skip(p + 1);
+                curr->expr = p;
+                while (*p && *p != ',' && *p != '}') p++;
+                *p = '\0';
```
Kuruyia: **issue**: If `*p == '}'`, we just erased it and moved on.

tools/enumproc/enumproc.c
```diff
@@ -0,0 +1,235 @@
...
+        .len  = 0,
+    };
+
+    size_t line = 1;
+    size_t col  = 1;
+    char  *s    = content.data;
+    char  *p    = content.data;
+    char  *e    = content.data + content.len;
+
+    // Find the line and column number for the error.
```
Kuruyia, line 180: **suggestion**: If we also want to be able to parse non-preprocessed files, then we cannot rely on those directives being present.

Kuruyia (changes requested):

tools/enumproc/enumproc.c
lhearachel: The CLI tool is specifically for preprocessed files.

tools/enumproc/libenum.h
lhearachel: `flags` on the parent enum contains the flags that were passed by the caller. So it's up to the caller to recall which they requested.

Kuruyia: I meant, given such an input (with the `LIBENUM_F_EVAL` flag not set):

```c
enum {
    EXPLICIT = 3,
    IMPLICIT,
};
```

I'd expect the result to have two members. The first one is clear to me: `{ .symbol = &"EXPLICIT", .expr = &"3" }`.

The second one, I am less sure: `{ .symbol = &"IMPLICIT", ??? }`. 

My current understanding is that:
- It will most definitely not have a valid `expr`, as there was no expression on that particular member. 
- It will have a `value` (with the current implementation), but it is invalid as well (since we didn't evaluate `EXPLICIT`'s expression, we don't know that `IMPLICIT` should be 4 here).

In my original question, I asked how to know which one is usable. But in this situation, neither really is. How does the caller know about that?

lhearachel: The intent here is that *not* setting the flag means the caller wants to either use their own expression parser and doesn't want the returned `enum` to be sorted. The latter bit is stated in the function documentation in `libenum.h`:

```c
// When `flags` is given with `LIBENUM_F_EVAL`, integer constant expressions
// will be evaluated, and each member of the output mapping shall use `value`.
// Otherwise, the expression will be stored as-is in `expr`.
//
// When `flags` is given with `LIBENUM_F_SORT`, the resulting members will be
// sorted in-order according to their symbol-names. Note that without the use of
// of `LIBENUM_F_EVAL`, this will destroy the successor-context of members; the
// value of any given member will thus be indeterminate.
```

I will add a clarification for the former. In your specific example, the `expr` property is what is set, and `IMPLICIT` will (or should) have a `value` set by the usual incrementing rules.

lhearachel: It mostly just gets a bit fuzzy without specifying `LIBENUM_F_EVAL`, due to the nature of C enums. Like I mentioned, the *intent* is that that flag should not be used *without* `LIBENUM_F_SORT`. But I don't think that I want to codify tying them together, because it reduces flexibility. I can imagine a use-case where `LIBENUM_F_EVAL` is desired, but `LIBENUM_F_SORT` is not, e.g., a small `enum` where the overhead of `qsort` just to use `bsearch` is higher than a linear scan.

Kuruyia: Yeah no problem, I understand how you see this to be used without `LIBENUM_F_EVAL`. In that case, I just think that it should be explicit in the documentation that, when not specifying this flag: 
- the contents of `.value` is always considered invalid
- the contents of `.expr` is considered valid if the pointer is non-NULL

Also, since this is a union, I believe that `.expr` is indeed initialized to `NULL` thanks to `calloc`, but then set to some garbage via `.value` when there is no expression.

tools/enumproc/libenum.c
lhearachel: This is resolved by actually updating the reference to `o2` after popping from the stack in `handle_binary`:

```c
while (stk.n_opts > 0
    && (o2 = top(stk, opts)) != OP_LPA
    && (precedes(o1, o2) || (congruent(o1, o2) && is_left(o1)))) {
    o2 = pop(stk, opts);
    evalbinary(o2);
}
```

tools/enumproc/libenum.h
lhearachel: I decided to turn this from a `union` to a `struct` to support the use-case of `.expr` being invalid if the pointer is non-`NULL`, because it would *not* be `NULL` in this case; it occupies the same memory as `.value`.

tools/enumproc/libenum.c
lhearachel: I settled on this:

```c
member_t *curr;
char     *e, *endp;

setup_member();
if (*p == '=') {
    if (flags & LIBENUM_F_EVAL) {
        curr->value.eval = eval(skip(p + 1), &endp, ",}", p_scope);
        next             = curr->value.eval + 1;
    } else {
        p                = skip(p + 1);
        curr->value.expr = p;
        curr->value.eval = next++; // Store both to keep canonical order
        while (*p && *p != ',' && *p != '}') p++;
        endp = p;
    }
}
else curr->value.eval = next++;

if (*p == ',') p     = skip(p + 1);
if (*p == '}') state = STATE_DONE;

// This check occurs here because we need `next` to be updated.
if (!filter || strncmp(curr->symbol, filter, s_filter) == 0) {
    close_member();
}
```

`endp` is cleared as part of `close_member`. Similar code is used in the `cpp` variant. Note that the `}` branch is always evaluated after checking for `,`, which removes the need for a second check on the next iteration (which no longer exists).

tools/enumproc/libenum.c
lhearachel: See above.

lhearachel:
~~I'll return to the build-error tomorrow. The bug is that `#define`-sets are skipping over the newline-terminator while processing the integer-expression.~~

Oops, found it.

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,747 @@
...
+            if    (!str[0] || !str[1]) { str = start; break; } // unterminated
+            str += 2;
+            continue;
+        }
+
+        break;
+    }
+
+    return (char *)str;
+}
```
Kuruyia: **polish**: It's probably better to share a common implementation.

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,747 @@
...
+        setup_member();
+        if (*p == '=') {
+            if (flags & LIBENUM_F_EVAL) {
+                curr->eval = eval(skip(p + 1), &endp, ",}", p_scope);
+                next       = curr->eval + 1;
+                p          = endp;
+            } else {
+                p          = skip(p + 1);
+                curr->expr = p;
+                curr->eval = next++; // Store both to keep canonical order
```
Kuruyia: **issue**: Because we returned to a `union`, the `curr->eval` assignment immediately overwrites the `curr->expr` assignment done previously.

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,747 @@
...
+                endp = p;
+            }
+        }
+        else if (flags & LIBENUM_F_EVAL) {
+            curr->eval = next++;
+        }
+        else curr->expr = NULL;
+
+        if (*p == ',') p     = skip(p + 1);
+        if (*p == '}') state = STATE_DONE;
```
Kuruyia: **issue**: This character is not consumed, so it appears in the final output.

When given:
```c
enum {
    TEST_1
};

// Some other stuff
```

The output is:
```
#define TEST_1 0
};

// Some other stuff
```

Or, when given (note the missing leading comma):
```c
enum {
    TEST_1 = 3
};

// Some other stuff
```

The output is:
```
#define TEST_1 3
```

It looks like a simple `++p` after the while loop fixes those issues.

Kuruyia: **issue**: We're not handling an unexpected character at that point.

When given:
```c
enum {
    hello how are you,
};
```

The output is:
```
#define hello 0
#define how 1
#define are 2
#define you 3
```

Kuruyia (changes requested):

lhearachel:
I've replaced the underlying library with a split implementation:

1. `libenum` now makes use of a lex-and-parse pipeline for translating `enum`s into member-mappings, but performs no parsing of any underlying expression-assignments, no evaluation of successor values, nor any sorting of the member-mapping; all of these tasks are now the responsibility of the client.
2. Expression evaluation is separated into `libexpr`.
3. `dp_include` in `tools/dataproc/src/common.c` handles all of the glue code to create the necessary lookup table.

> [!WARNING]
> There is an issue with the current build: our `postconf` script turns the `enumproc` executable used during script-processing into an order-only dependency. I am currently evaluating how best to resolve this.

tools/enumproc/libenum.h
```diff
@@ -0,0 +1,120 @@
...
+ * semicolon which terminates the enum definition, if one is present. If the
+ * terminating semicolon is missing, then it is treated as an error.
+ *
+ * Members within the returned mapping are guaranteed to be uniquely named
+ * within the enum's scope. If parsing fails for any reason, then the returned
+ * mapping's `errc` value will be set to an associated error code. Additionally,
+ * if `endptr` is not `NULL`, the address of the first invalid character for the
+ * last-processed token will be stored in `*endptr`.
+ */
+enum_seq_t libenum_load(const char *str, size_t size, char **endptr);
```
Kuruyia, line 80: **question**: Do we want to handle backslash-newline sequences? For example:
```c
enum {
    SOMETHING = \
        3 + 1, // imagine some very long expression instead...
    SOMETHING_ELSE
};
```

I know those are typically handled by the preprocessor. I do not know whether this function expect `str` to have been preprocessed.

tools/dataproc/meson.build
```diff
@@ -1,6 +1,7 @@
 dataproc_cflags = [
     '-std=gnu17',
-    '-O3',
+    '-O0',
+    '-g',
```
Kuruyia: **suggestion**: Revert back to release build flags.

tools/enumproc/meson.build
Kuruyia, line 1: **suggestion**: Ditto about build flags for the `libenum_dep` and `libexpr_dep` targets.

tools/enumproc/libenum.c
```diff
@@ -0,0 +1,509 @@
...
+            if ((expect & EXPECT_EQUALS) == 0) return_error(LIBENUM_E_TOKEN);
+
+            lexer.p = skip(lexer.p, is_wspace);
+            lexer.s = lexer.p;
+
+            for (
+                lexer.p = skip(lexer.p, is_wspace), lexer.s = lexer.p;
+                lexer.p < lexer.e && *lexer.p && *lexer.p != ',' && *lexer.p != '}';
+                lexer.p++
+            );
```
Kuruyia, line 144: **suggestion**: Also stop at newlines to handle such syntax errors:

```c
enum {
    TEST_1 = 1)
    TEST_2,
    TEST_3
};
```

```c
enum {
    TEST_1 = 1
    TEST_2,
    TEST_3
};
```

```suggestion
            for (
                lexer.p = skip(lexer.p, is_wspace), lexer.s = lexer.p;
                lexer.p < lexer.e && *lexer.p && *lexer.p != ',' && *lexer.p != '}' && !is_vspace(*lexer.p);
                lexer.p++
            );
```

Note: the following is apparently valid C (both gcc and clang compiles this), and is handled by the suggested fix:

```c
enum {
    TEST_1 = 1
    ,TEST_2,
    TEST_3
};
```

Kuruyia (changes requested):

tools/dataproc/meson.build
lhearachel: Oop

tools/enumproc/libenum.h
lhearachel: I think that's a valid thing to consider for the future, but let's recall the purpose here: we're mostly looking to parse simple enums. I think that it's fine for us to insist that we only handle a subset of C's possible syntax, and that expressions should *not* include escaped-newlines. This is already noted in `loadcpp` as something that we don't handle:

> Functional macros and macros without substitution-text are skipped. Macros which escape new-lines in their substitution-text are not supported and are treated as if those escapes do not exist.

Kuruyia: Completely fair :)

tools/enumproc/libenum.c
lhearachel: `libexpr` is already built to handle whitespace in the expression-text. `libenum` should accept both of these listed cases, IMO, and defer the syntax-error to `libexpr`.

lhearachel: Discussed on Discord: we decided to defer such cases as errors to `libexpr`. Terminating on newlines would lead to the following being a "missing comma" error in `libenum` rather than an "undefined symbol" error in `libexpr`:

```c
enum {
    TEST_1 = 1+
    TEST_2,
    TEST_3
};
```

Kuruyia (approved):
Goated! I've got an upcoming PR that's eager to try this new stuff out.

lhearachel:
The most recent force-push should fix the build issue.

PR #1117, opened 2026-04-26
Document stark mountain scripts

Viperio19:
This PR documents the event, script and text files for stark mountain.


generated/vars_flags.txt
```diff
@@ -213,13 +213,13 @@ FLAG_PAL_PARK_TALKED_TO_RECEPTIONIST
 FLAG_UNK_0x00D4
 FLAG_RECEIVED_FIGHT_AREA_MART_SCOPE_LENS
 FLAG_ARRESTED_CHARON_STARK_MOUNTAIN
-FLAG_UNK_0x00D7
+FLAG_BATTLED_STARK_MOUNTAIN_ROOM_3_HEATRAN
```
lhearachel, line 216: **question**: As opposed to Room 1 Heatran...?

Viperio19: I mostly default to including full location names just to make it less arbitrary when to include it and when not, but yeah it's not really necessary here (especially since there's also just `FLAG_CAUGHT_HEATRAN`)

lhearachel: That's fair, it just caught me off-guard.

lhearachel (approved):

PR #1124, opened 2026-04-30
Document unused battle park scripts

Viperio19:
This PR documents the event, script and text files for the battle park and marks them as unused (because they are unused). This also includes documenting some BP related script commands, I even documented a few that aren't used (not in these scripts but also not in general) since this it's a nice group of script commands to document and the PR is relatively small anyways.

I also renamed the `MART_FRONTIER_ID_EXCHANGE_SERVICE_CORNER_[UP/DOWN]` constants to `MART_FRONTIER_ID_EXCHANGE_SERVICE_CORNER_[RIGHT/LEFT]` to reflect their usage in `ScrCmd_PokeMartFrontier`. Although in the battle park exchange service corner the maids are oriented north/south, they are oriented right/left in the battle frontier and that is instance where `PokeMartFrontier` is called and actually used.


src/unk_020494DC.c
```diff
@@ -413,119 +414,118 @@ BOOL ScrCmd_UpdateBPDisplay(ScriptContext *ctx)
...
     };
 
-    if (v1 == 1) {
-        v0 = 26;
+    if (martID == 1) {
+        startID = 26;
     } else {
-        v0 = 0;
+        startID = 0;
     }
```
lhearachel, line 528: **thought**: Hm, this seems like a potential foot-gun. The intent here seems to be to have TMs separated from other items. I wonder if there's a way we can guarantee that. 🤔

Viperio19: Yeah, `ScrCmd_PokeMartFrontier` does fully separate the TMs, and that's probably why `ScrCmd_PokeMartFrontier` is used for the actually used exchange service corner in the battle frontier while this `ScrCmd_GetExchangeServiceCornerItemAndCost` is only used in the unused battle park exchange service corner.

lhearachel: Good point, since it's unused, we probably don't care. We should probably (at minimum) leave a comment noting to update both if one of them is touched, though.

Viperio19: I don't see any reason why these two implementations should necessary always have the same shop contents, it's not like we have anything else to do with it, and I actually see some slight value in this unused implementation where you can specify custom prices for future romhacks. I did figure out a way to make the split between items and regular items more implicit:
```c
#define FRONTIER_MART_ITEMS_START_ID 0
#define FRONTIER_MART_TMS_START_ID 26

BOOL ScrCmd_GetExchangeServiceCornerItemAndCost(ScriptContext *ctx)
{
    ...
    static const u16 prizeList[][2] = {
        [FRONTIER_MART_ITEMS_START_ID] = { ITEM_PROTEIN, 1 },
        { ITEM_CALCIUM, 1 },
        ...
        [FRONTIER_MART_TMS_START_ID] = { ITEM_TM06, 32 },
        { ITEM_TM73, 32 },
        ...
        { ITEM_TM26, 80 }
    };

    if (martID == 1) {
        startID = FRONTIER_MART_TMS_START_ID;
    } else {
        startID = FRONTIER_MART_ITEMS_START_ID;
    }
```

lhearachel: Honestly wasn't certain if this was well-defined behavior; it appears that it is:

> A designator causes the following initializer to initialize of the array element described by the designator. Initialization then continues forward in order, beginning with the next element after the one described by the designator. (since C99)

```c
int n[5] = {[4]=5,[0]=1,2,3,4}; // holds 1,2,3,4,5
```

I think this does reduce the potential for confusing mistakes. Does the compiler emit a warning if you implement this and then add a new entry to the non-TMs vendor?

Viperio19: There is no warning, and it even still matches. It seems to work like this:
```c
int n[5] = {[0]=1,2,3,6,[3]=4,5}; // holds 1,2,3,4,5
```

lhearachel: Yikes. That's frustrating.

lhearachel: Let's go with what you proposed and make a comment next to the first TM entry that remarks to update the constant when adding items above that line.

Viperio19: done!

lhearachel (approved):

PR #1126, opened 2026-05-01
Document distortion world map object rotator

Kuruyia:
This continues documentation of the distortion world overlay by documenting the code that's responsible for handling billboard rotation. This is used when map objects are standing on rotated floating platforms, or when certain camera angles are used.

When initializing, the distortion world system allocates a fixed amount of rotators which can be bound to billboards that need to support rotation. One rotator is always bound to the player map object. One example of a non-player map object that uses this system is B2F Cynthia, who is standing on a rotated floating platform. Whether a rotator is bound to a map object is controlled by the `rotated` field in the distortion world map object event structure.

The actual rotation is done by a material animation loaded from the field effect NARC and added to the map object render object. It seems that this animation has 360 frames, where the `n`th frame rotates the billboard by `n` degrees. The current frame for this animation is continuously set in the render callback of the bound billboard.


src/overlay009/ov9_02249960.c
```diff
@@ -2248,286 +2254,272 @@ static BOOL DoCameraTransition(DistWorldSystem *system, const DistWorldCameraAng
...
 
-    v0->unk_08 = FieldEffectManager_AllocAndReadNARCWholeMember(param0->fieldSystem->fieldEffMan, 197, 1);
-    v0->unk_0C = NNS_G3dGetAnmByIdx(v0->unk_08, 0);
+    HeapExp_FndInitAllocator(&mapObjRotatorMan->allocator, HEAP_ID_FIELD1, 4);
+    mapObjRotatorMan->animData = FieldEffectManager_AllocAndReadNARCWholeMember(system->fieldSystem->fieldEffMan, 197, TRUE);
+    mapObjRotatorMan->anim = NNS_G3dGetAnmByIdx(mapObjRotatorMan->animData, 0);
 }
 
-static void ov9_0224A228(UnkStruct_ov9_0224A228 *param0, UnkStruct_ov9_0224A294 *param1, Billboard *param2)
+static void DistWorldMapObjectRotatorManager_PrepareRotator(DistWorldMapObjectRotatorManager *mapObjRotatorMan, DistWorldMapObjectRotator *mapObjRotator, Billboard *billboard)
```
lhearachel, line 2271: **question**: `Prepare` is interesting wording; what made you choose this over `Init`?

Kuruyia: A few reasons:
- `DistWorldMapObjectRotator_Init` already exists, so I wanted to find another name to avoid potential confusion
- This function is the more complex one of the two; it initializes some `DistWorldMapObjectRotator` members but also "prepares" the billboard to be rotated
- It being called may be deferred to the rotator manager ticking, whereas `DistWorldMapObjectRotator_Init` is always called when binding a billboard/map object to a rotator

lhearachel: Thanks for the explanation!

lhearachel (approved):

PR #1109, opened 2026-04-24
Document some Battle Hall Frontier Scrcmds

VicSevenT:
This is a hodgepodge of Battle Hall frontier script documentation. The main points are that it identifies the file dedicated to Battle Hall-related frscrcmds, identifies the main `BattleHall` struct and creates `BH_FUNC` constants to help with future Battle Hall documentation.

PR #1127, opened 2026-05-01
Document some facility-specific frscrcmds and structs

VicSevenT:
Identifies `frscrcmd_battle_<facility>` files and some functions within, and the main structs for managing the Battle Factory, Arcade, and Castle.

Essentially the same work as done in https://github.com/pret/pokeplatinum/pull/1109 applied to all the other facilities.

Check first commit for more specific diffs without file renames getting in the way.


asm/macros/frscrcmd.inc
```diff
@@ -566,17 +566,17 @@
...
     .byte \arg2
-    .short \arg3
+    .short \destVar
     .endm
 
-    .macro FrontierScrCmd_68 arg0
-    .short FRSCRCMD_68
-    .short \arg0
+    .macro BattleFactory_CheckWonBattle destVar
+    .short FRSCRCMD_BATTLEFACTORY_CHECKWONBATTLE
```
lhearachel, line 578: **thought**: Does the `BATTLE` prefix provide much here?

include/constants/battle_arcade_functions.h
lhearachel, line 1: **thought**: Would it make sense to merge all of these into `battle_frontier_functions`? Is there a reason to separate them?

src/overlay104/frscrcmd_battle_castle.c
```diff
@@ -0,0 +1,580 @@
...
+
+    SetupBattleCastleAppArgs(args, battleCastle);
+    sub_0209B988(ctx->scriptMan->unk_00, &sBattleCastleNullTemplate, args, 0, StoreBattleCastleAppResult);
+
+    return TRUE;
+}
+
+BOOL FrontierScrCmd_BattleCastle_CleanupBattle(FrontierScriptContext *ctx)
+{
+
```
lhearachel: **polish**: Remove this empty newline.

src/overlay104/frscrcmd_battle_castle.c
```diff
@@ -0,0 +1,580 @@
...
+        battleCastle->unk_13 = 1;
+        break;
+    case BC_FUNC_UNK_44:
+        if (battleCastle->challengeType == 3) {
+            if (arg1 == 0) {
+                battleCastle->unk_22 = BattleFrontierStats_GetStat(SaveData_GetBattleFrontier(v11->saveData), BattleFrontierStats_GetCastleLatestCPIndex(battleCastle->challengeType), BattleFrontierStats_GetHostFriendIdx(BattleFrontierStats_GetCastleLatestCPIndex(battleCastle->challengeType)));
+                BattleFrontierStats_SetStat(SaveData_GetBattleFrontier(v11->saveData), BattleFrontierStats_GetCastleLatestCPIndex(battleCastle->challengeType), BattleFrontierStats_GetHostFriendIdx(BattleFrontierStats_GetCastleLatestCPIndex(battleCastle->challengeType)), battleCastle->unk_20);
+            } else {
+                BattleFrontierStats_SetStat(SaveData_GetBattleFrontier(v11->saveData), BattleFrontierStats_GetCastleLatestCPIndex(battleCastle->challengeType), BattleFrontierStats_GetHostFriendIdx(BattleFrontierStats_GetCastleLatestCPIndex(battleCastle->challengeType)), battleCastle->unk_22);
+            }
```
lhearachel, line 417: **note**: Just making a note that these lines are very long and difficult to read. Probably worth splitting up if/when you document them properly.

src/overlay104/frscrcmd_battle_factory.c
```diff
@@ -0,0 +1,473 @@
...
+        battleFactory->unk_4DC[i] = 0;
+    }
+
+    BattleFactoryAppArgs *args = Heap_Alloc(HEAP_ID_FIELD2, sizeof(BattleFactoryAppArgs));
+    MI_CpuClear8(args, sizeof(BattleFactoryAppArgs));
+
+    args->saveData = v3->saveData;
+    args->challengeType = battleFactory->challengeType;
+    args->unk_05 = battleFactory->unk_05;
+    args->isExchangeMode = 1;
```
lhearachel: **question**: Should this use a `BOOL` constant?

lhearachel (commented):
I tried to limit my feedback solely to the functions documented here rather than address them all.

include/constants/battle_arcade_functions.h
VicSevenT: I thinks it's a horse a piece. Combining them yields less files overall, but that file is massive and has 5 enums that, despite all following the same structure, are technically never used together or in the same file/context. I don't think combining them would result in any less `#includes` in the repo as no two enums are ever used in any one file anyway.

asm/macros/frscrcmd.inc
VicSevenT: Do you specifically mean just the `BATTLE` in the constant, or `Battle` overall as a prefix to the macro? For the former, that is just following the convention of the constants being verbatim `FRSCRCMD_<MACRO_NAME>`. 

For the latter, I think it's subjective. I personally prefer it because while with say the `Factory` or `Castle` removing the `Battle` could be fine, I kind of find `Hall` without `Battle` weird. I also don't think the names were getting so long that removing `Battle` was that worthwhile.

src/overlay104/frscrcmd_battle_factory.c
VicSevenT: Yes, fixed

VicSevenT:
`enumproc` has since been merged, so I've also updated the PR to use enums for all the sets of functions, including moving `generated/battle_tower_functions` -> `constants/battle_tower_functions`.

include/constants/battle_arcade_functions.h
```diff
@@ -1,58 +1,60 @@
...
-#define BA_FUNC_UNK_44                   44
-#define BA_FUNC_UNK_45                   45
-#define BA_FUNC_UNK_46                   46
-#define BA_FUNC_UNK_47                   47
-#define BA_FUNC_UNK_48                   48
-#define BA_FUNC_UNK_49                   49
-#define BA_FUNC_UNK_50                   50
-#define BA_FUNC_UNK_51                   51
-#define BA_FUNC_UNK_52                   52
+enum BattleArcadeFunction {
```
lhearachel, line 4: **praise**: Congrats on being the first actual use of `enumproc` in the repo. 😁

lhearachel (approved):

PR #1128, opened 2026-05-02
document mail

scbroede:
`PARTY_MENU_MODE_GIVE_ITEM`, `PARTY_MENU_MODE_GIVE_ITEM_DONE`, `PARTY_MENU_MODE_GIVE_MAIL`, and `PARTY_MENU_MODE_GIVE_MAIL_DONE` might be mislabeled; the "done" ones seem to refer to giving from the party menu while the others seem to refer to giving from the bag


include/item.h
lhearachel: **polish**: We should either update these comments or get rid of them.

include/item.h
```diff
@@ -280,15 +280,15 @@ u8 Item_IsMail(u16 item);
  * @param item
  * @return The mail number for the item, or ITEM_NONE if the item is not mail.
  */
-u8 Item_MailNumber(u16 item);
+u8 Item_MailType(u16 item);
```
lhearachel: **suggestion**: `Item_GetMailType`

src/applications/mail.c
```diff
@@ -0,0 +1,374 @@
...
+#include "string_gf.h"
+#include "unk_02014A84.h"
+#include "unk_0209747C.h"
+
+FS_EXTERN_OVERLAY(overlay20);
+FS_EXTERN_OVERLAY(mail_viewer);
+
+typedef struct MailApp {
+    enum HeapID heapID;
+    u8 padding[4];
```
lhearachel, line 29: **question**: Does this match if this becomes a `u32` instead?

src/applications/mail_viewer.c
```diff
@@ -0,0 +1,952 @@
...
+    Heap_Free(mailViewerApp->bgConfig);
+}
+
+static void MailViewer_DrawMail(MailViewerApp *mailViewerApp)
+{
+    NNSG2dCharacterData *charData;
+    NNSG2dPaletteData *plttData;
+    int tilesNarcIndex = grass_mail_tiles_NCGR + mailViewerApp->args->mailType;
+    int screenDataNarcIndex = grass_mail_NSCR + mailViewerApp->args->mailType;
+    int plttNarcIndex = grass_mail_tiles_NCLR + mailViewerApp->args->mailType;
```
lhearachel, line 706: **thought**: Should we make it explicit to a user that they have to edit the order-file if they add new mail-types?

src/applications/mail.c
scbroede: it matches as anything that takes up 4 bytes. it could be a `u32`, it could be an `int`, it could be any kind of pointer, etc.

it's just there for alignment and making sure the struct is the right size, so i prefer to leave these as u8 arrays with the exact amount of bytes they need to take up

src/applications/mail_viewer.c
scbroede: probably a good idea. do you think it would be better to put that warning here or in the meson.build for the mail narc? i'm leaning toward the latter since users would have to add the pngs there anyway afaik

lhearachel: In the `meson.build`, IMO. It's entirely localized to the asset compilation; the code is just a consequence.

lhearachel (approved):

PR #1130, opened 2026-05-03
tools: Integrate dataproc as the JSON interface for msgenc

lhearachel:
Self-explanatory.

This leaves `datagen_events.cpp` as the sole user of `rapidjson` as a dependency.

### DIFF
--- a/tools/msgenc/Json.cpp
+++ b/tools/msgenc/Json.cpp
@@ -12,12 +12,7 @@
 #include <string_view>
 #include <vector>
 
-#include "rapidjson/document.h"
-#include "rapidjson/error/en.h"
-#include "rapidjson/error/error.h"
-#include "rapidjson/prettywriter.h"
-#include "rapidjson/rapidjson.h"
-#include "rapidjson/stringbuffer.h"
+#include "dataproc.h"
 
 #include "MessagesConverter.h"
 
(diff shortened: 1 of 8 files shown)


tools/msgenc/Json.cpp
```diff
@@ -97,103 +92,63 @@ void Json::WriteHeader(const string &_filename) {
...
-        if (membGarbage != entry.MemberEnd()) {
-            if (!membGarbage->value.IsInt()) {
-                sprintf(errbuf, "Bank.messages[%d].garbage: expected an integer.", i);
-                throw runtime_error(errbuf);
+        datanode_t messages = dp_get(&this->doc, ".messages");
+        std::size_t numMessages = dp_arrlen(messages);
+        for (std::size_t i = 0; i < numMessages; i++) {
+            string message;
+            datanode_t elem = dp_arrelem(messages, i);
+            if (dp_hasmemb(elem, "garbage")) {
```
Kuruyia: **suggestion**: If `garbage` and `en_US` are mutually exclusive, then we might want to print a warning/error if both are specified.

lhearachel: I think we should just invert the priority; `garbage` is a matching-artifact, not something we should expect users to add.

Kuruyia: Sounds good to me

lhearachel: Done and also applied to instances in `frontierproc` and `trainerproc`.

Kuruyia (approved):
LGTM!

PR #1136, opened 2026-05-04
Document distortion world map info file

Kuruyia:
This continues documentation of the distortion world overlay by documenting the map info file structure.

This file is the very first file of the main distortion world archive (`tw_arc.narc`), and it contains an array that holds information about each distortion world map.

The following information can be found in each array item:
- the map header ID to which this item applies to
- the index of the main data file for this map in the main distortion
  world archive, minus one
- the offsets used by the land data manager when rendering the map

This also renames the `DistWorldFile` struct to `DistWorldMapFile` to better show its relation to the newly-introduced `DistWorldMapInfoFile` struct.


lhearachel (approved):
Beautiful work, no comments here. 😁

PR #1138, opened 2026-05-05
Document distortion world floor loading

Kuruyia:
This continues documentation of the distortion world overlay by documenting the code related to loading maps.

In the distortion world, up to two floors are loaded at any given time. One of them is called the active floor, which is where the player is currently located. The other one is the inactive floor, and is the floor that is rendered below. The active floor is managed by the regular field system's land data manager. The inactive floor is managed by the distortion world system's land data manager.

Floors in the distortion world have two main constraints: there is a one-to-one relationship between floors and map headers, and its map matrix cannot be more than 2x2.

The player has two ways of travelling between floors: the main one is by using an elevator platform, and the other one is when cascading on the waterfall between B4F and B5F. Floors are loaded asynchronously.

The logic when travelling to the previous floor and to the next floor is a bit different. Here's the gist of it.

When travelling to the previous floor:
- unload the currently inactive floor
- switch the currently active floor to be the new inactive floor
- load the new active floor

When travelling to the next floor:
- unload the currently active floor
- switch the currently inactive floor to be the new active floor
- load the new inactive floor

### DIFF
--- a/src/overlay009/ov9_02249960.c
+++ b/src/overlay009/ov9_02249960.c
@@ -7,9 +7,10 @@
 #include "constants/field/dynamic_map_features.h"
 #include "constants/field/map.h"
 #include "constants/graphics.h"
+#include "constants/heap.h"
 #include "constants/map_object.h"
+#include "constants/quadrant.h"
 #include "constants/scrcmd.h"
-#include "constants/species.h"
 #include "constants/types.h"
 #include "generated/map_headers.h"
 #include "generated/movement_actions.h"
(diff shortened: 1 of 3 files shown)


src/overlay009/ov9_02249960.c
```diff
@@ -4727,45 +4758,45 @@ static void ov9_0224C8E8(DistWorldSystem *param0)
...
     }
 
     {
         int v10 = Player_GetXPos(fieldSystem->playerAvatar);
         int v11 = Player_GetZPos(fieldSystem->playerAvatar);
 
-        LandDataManager_DistortionWorldInitialLoad(v2->unk_18, v10, v11);
-        v2->unk_08 = 1;
+        LandDataManager_DistortionWorldInitialLoad(v2->landDataMan, v10, v11);
+        v2->ready = 1;
```
lhearachel: **polish**: Update these references to use `BOOL` constants.

src/overlay009/ov9_02249960.c
```diff
@@ -4727,45 +4758,45 @@ static void ov9_0224C8E8(DistWorldSystem *param0)
...
     }
 
-    v2->unk_04 = 1;
+    v2->valid = 1;
```
lhearachel: **polish**: Ditto.

src/overlay009/ov9_02249960.c
```diff
@@ -4982,261 +5011,252 @@ static void Dummy0224CC4C(DistWorldSystem *system)
...
 {
-    UnkStruct_ov9_0224CBD8 *v0 = &param0->unk_1CD0;
-
-    if (v0->unk_00 != 0) {
-        return 1;
-    }
-
-    return 0;
+    DistWorldFloorLoader *floorLoader = &system->floorLoader;
+    return floorLoader->loadDir != FLOOR_LOAD_NONE;
```
lhearachel: **question**: I presume this doesn't match as the following?


```suggestion
    return system->floorLoader.loadDir != FLOOR_LOAD_NONE;
```

Kuruyia: This totally matches, I just didn't notice it ^^

lhearachel (approved):

PR #1139, opened 2026-05-05
Document union room script

Viperio19:
This PR documents the union room script and parts of the relevant code in `unk_0205B33C.c`. There are still a lot of unnamed script commands throughout the script, but that felt out of scope for this PR.


src/unk_0205B33C.c
```diff
@@ -745,138 +748,138 @@ void sub_0205BAAC(int param0)
     CommSys_SendData(101, &v0, 1);
 }
 
-static const int Unk_020ED560[2][2] = {
-    { 0x2C, 0x68 },
-    { 0x2D, 0x69 }
+static const int sMessagesShowingTrainerCase[2][2] = {
```
lhearachel: **polish**: Remove the explicit bounds on the first layer of these arrays:


```suggestion
static const int sMessagesShowingTrainerCase[][2] = {
```

src/unk_0205B33C.c
```diff
@@ -907,198 +907,189 @@ static int sub_0205BAE8(UnkStruct_0205B43C *param0, int param1)
...
 
     switch (v4->unk_1C) {
     case 4:
     case 11:
-        return Unk_020ED548[v0];
+        return sMessagesTrainerAppearsBusy[gender];
         break;
     case 5:
-        return Unk_020ED560[LCRNG_Next() % 2][v0];
+        return sMessagesShowingTrainerCase[LCRNG_Next() % 2][gender];
```
lhearachel: **polish**: Use `NELEMS(sMessageShowingTrainerCase)` so that the RNG naturally grows if modified.


```suggestion
        return sMessagesShowingTrainerCase[LCRNG_Next() % NELEMS(sMessagesShowingTrainerCase)][gender];
```

This also applies to the other instances.

lhearachel (commented):
Just some comments in the documented C code. Looks good to me, otherwise! 😁

src/unk_0205B33C.c
Viperio19: For some reason this doesn't match and I have no clue why

lhearachel: Try casting it to `(int)`

Viperio19: that worked, just pushed a commit for the two suggestions!

lhearachel (approved):

PR #1146, opened 2026-05-09
initial easy chat documentation

scbroede:
leaving this here before it gets any bigger


include/unk_02014A84.h
```diff
@@ -12,17 +12,17 @@ void Sentence_InitWithType(Sentence *sentence, u32 type);
 void sub_02014AB4(Sentence *sentence);
 void sub_02014AC4(Sentence *sentence, int param1);
 String *Sentence_AsString(const Sentence *sentence, enum HeapID heapID);
-String *sub_02014BA0(const Sentence *sentence, u32 param1);
+String *Sentence_GetTemplateString(const Sentence *sentence, u32 param1);
```
lhearachel, line 15: **thought**: Out of scope for this PR, but we should probably rename `Sentence` at some point.

src/easy_chat_args.c
```diff
@@ -0,0 +1,177 @@
...
+    args->unk_05 = FALSE;
+    args->isUnmodified = TRUE;
+    args->wasUpdated = FALSE;
+    args->frame = Options_Frame(SaveData_GetOptions(saveData));
+
+    if (type == EASY_CHAT_TYPE_SENTENCE) {
+        Sentence_InitWithType(&args->sentence, 3);
+    } else {
+        for (int v1 = 0; v1 < MAX_EASY_CHAT_WORDS; v1++) {
+            args->words[v1] = 0xffff;
```
lhearachel: **polish**: Prefer `-1` if it matches, otherwise `0xFFFF` (uppercase).

scbroede: i left these as-is for now because i was going to add a WORD_NONE constant in my next pr

scbroede: actually never mind, just pushed the change here

lhearachel (approved):

PR #1147, opened 2026-05-10
tools: Implement itemproc

lhearachel:
Commits for this PR are split into logical increments to hopefully make them easier to view on GitHub.

This is an implementation of `dataproc` that makes no use of `metang`. This should hopefully make my vision for the end-state of the build more clear. In essence, the `enum` defining the constants used to access associated data files and text bank entries is the single-source-of-truth defining both the ordering and the some part of the name of expected input files. The processing tool is responsible for loading that `enum` to a mapping, translating them into the expected input files, and declaring dependencies on the ones that it finds.


tools/dataproc/src/common.c
```diff
@@ -83,6 +83,56 @@ static FILE* open_depfile(const char *depfile_path) {
...
+        result.size--;
+    }
+
+    u16 count = 0;
+    if (params.filter) {
+        for (size_t i = 0; i < result.size; i++) {
+            count += params.filter(result.members[i].name);
+        }
+    }
+    else count = (u16)result.size & 0xFFFF;
```
scbroede: nitpick: the `& 0xFFFF` is redundant when already casting to u16

res/items/data/hm05.json
```diff
@@ -0,0 +1,31 @@
...
+  "plural": "HM05s",
+  "article": "an",
+  "description": [
+    "Obstacles are moved, reducing the\n",
+    "foe’s evasion stat. It can also be\n",
+    "used to clear deep fog."
+  ],
+  "icon": {
+    "sprite": "hm_NCGR",
+    "palette": "tm_water_NCLR"
```
scbroede, line 12: do we want a comment here for the palette mistake? (even if not, we'll want to update the bugs and glitches doc to reflect that this was moved here)

scbroede (commented):
just a couple of nitpicks. i'd also like to see that "unused" ncgr/nclr get documented better and the battle use funcs get a proper enum, but i'm happy to leave those for later

res/items/data/hm05.json
lhearachel: It's JSON, we cannot have comments.

But good shout on the docs update.

scbroede (approved):

PR #1150, opened 2026-05-11
Document script_battles.s and trainer_encounter.c

Viperio19:
This PR documents `script_battles.s` and the relevant parts for that in `trainer_see.c`. The main part is `SysTask_ApproachingTrainer` which uses a state machine for a trainer approaching the player, although this state machine uses a list of functions (`sApproachingTrainerTask`) instead of one function with a `switch (state)` statement.


src/map_object_move.c
```diff
@@ -1015,17 +1015,17 @@ int Direction_GetOpposite(int dir)
     return sOppositeDirections[dir];
 }
 
-int sub_02064488(int param0, int param1, int param2, int param3)
+int Direction_GetBetweenCoords(int xSrc, int zSrc, int xDst, int zDst)
```
lhearachel: **suggestion**: I think the primitives here suggest that we should call this `GetDirectionBetweenPoints`.

src/trainer_encounter.c
```diff
@@ -38,20 +38,20 @@ typedef struct ApproachingTrainerTemplate {
     MapObject *trainerMapObj;
 } ApproachingTrainerTemplate;
 
-typedef struct {
-    int unk_00;
-    int unk_04;
+typedef struct ApproachingTrainerData {
+    int state;
+    int done;
```
lhearachel: **polish**: This seems like a `BOOL`.

src/trainer_encounter.c
```diff
@@ -386,313 +386,332 @@ BOOL FieldSystem_IsTrainerDefeated(FieldSystem *fieldSystem, MapObject *mapObj)
...
-static void sub_0206806C(SysTask *task, void *param1)
+static void SysTask_ApproachingTrainer(SysTask *task, void *inData)
 {
-    UnkStruct_020EF6D0 *v0 = param1;
+    ApproachingTrainerData *data = inData;
 
-    while (Unk_020EF6D0[v0->unk_00](v0) == 1) {
+    while (sApproachingTrainerTask[data->state](data) == 1) {
         (void)0;
     }
```
lhearachel: **suggestion**:


```suggestion
    while (sApproachingTrainerTask[data->state](data) == 1); // wait
```

src/map_object_move.c
Viperio19: I thought that public functions should (always?) have a prefix, but I do agree `GetDirectionBetweenPoints` is better.

lhearachel: They don't necessarily need to, if the parameters are all primitives like this one.

src/trainer_encounter.c
Viperio19: `clang-format` wants to change it to this, is that still preferred?
```diff
-    while (sApproachingTrainerTask[data->state](data) == 1); // wait
+    while (sApproachingTrainerTask[data->state](data) == 1)
+        ; // wait
```

lhearachel: Yes.

Viperio19: Done!

lhearachel (approved):

PR #1154, opened 2026-05-13
Document common scripts

Viperio19:
This PR documents `scripts_common.s`, naming unnamed common scripts and shortening some names in previously documented ones. This also includes using the object event ids for pokecenter nurses, which required making the object event ids for pokecenters globally unique.


Viperio19:
As you can see I didn't manage to name all the script entries. Most of those don't seem to get referenced anywhere (not by their hex value, decimal value, or by `(COMMON_SCRIPTS, script id - offset)`, so I could call those `CommonScript_UnusedScript[ID]`, but I'm not 100% certain that there isn't some other way they could potentially be referenced. My question is here is if I can just name those `CommonScript_UnusedScript[ID]`.
There is also one script entry that gets used but I had no clue how to call it:
`ScriptEntry _139D @ 0x7F1`
This is for the school kid in the battle tower who asks you stuff about how you feel. I think this is used to set the phrases (intro, win and defeat text) for in the battle tower wifi challenge, I just don't know what would be a discriptive name for that.

include/data/scripts/scrcmd.h
```diff
@@ -303,7 +303,7 @@ ScriptCommand(SCRCMD_STARTCATCHINGTUTORIAL,                                ScrCm
 ScriptCommand(SCRCMD_SLATHERHONEYTREE,                                     ScrCmd_SlatherHoneyTree)
 ScriptCommand(SCRCMD_GETHONEYTREESTATUS,                                   ScrCmd_GetHoneyTreeStatus)
 ScriptCommand(SCRCMD_STARTHONEYTREEBATTLE,                                 ScrCmd_StartHoneyTreeBattle)
-ScriptCommand(SCRCMD_12A,                                                  ScrCmd_12A)
+ScriptCommand(SCRCMD_HONEYTREESTOPSHAKING,                                 ScrCmd_HoneyTreeStopShaking)
```
lhearachel: **polish**: This should be `StopHoneyTreeShaking` to keep the verb in front.

include/data/scripts/scrcmd.h
```diff
@@ -827,7 +827,7 @@ ScriptCommand(SCRCMD_332,                                                  ScrCm
 ScriptCommand(SCRCMD_SETPLAYERVOLUME,                                      ScrCmd_SetPlayerVolume)
 ScriptCommand(SCRCMD_ADDTOGAMERECORD,                                      ScrCmd_AddToGameRecord)
 ScriptCommand(SCRCMD_ADDTOGAMERECORDBIGVALUE,                              ScrCmd_AddToGameRecordBigValue)
-ScriptCommand(SCRCMD_336,                                                  ScrCmd_336)
+ScriptCommand(SCRCMD_ISHALLOFFAMECORRUPTED,                                ScrCmd_IsHallOfFameCorrupted)
```
lhearachel: **polish**: `CheckIsHallOfFameCorrupted` for consistency.

lhearachel (commented):
> As you can see I didn't manage to name all the script entries. Most of those don't seem to get referenced anywhere (not by their hex value, decimal value, or by (`COMMON_SCRIPTS`, script id - offset), so I could call those `CommonScript_UnusedScript[ID]`, but I'm not 100% certain that there isn't some other way they could potentially be referenced. My question is here is if I can just name those `CommonScript_UnusedScript[ID]`.

I would keep them named as `CommonScript_[ID]`. Don't flag things as Unused unless we know they are not used.

lhearachel (approved):

PR #1156, opened 2026-05-15
Document Sunyshore Gym persistent features

VicSevenT:
Documents the positioning, rotation, and collision of the gears in Sunyshore Gym.

The collision works by defining 2x1 regions (5x1 for the vertical gears) on each side of the gear, so 4 regions per gear. Based on how many "arms" the gear has (dead end, L, or T) and the current orientation of the gears, certain regions will be active and keep the player from walking off the paths.

Some gear models have an "alt" variant where the paths are slightly rotated wrt to the gear teeth so that the gears mesh properly.


src/overlay008/ov8_02249960.c
```diff
@@ -117,9 +117,17 @@
...
+#define SUNYSHORE_PROP_OFFSET FX32_CONST(8)
+
+#define SUNYSHORE_NUM_ROTATION_STATES 4
+
+#define SUNYSHORE_ROOM_1_NUM_GEARS 3
+#define SUNYSHORE_ROOM_2_NUM_GEARS 6
+#define SUNYSHORE_ROOM_3_NUM_GEARS 13
+
+#define SUNYSHORE_90_DEGREES_ROTATION  0x4000
+#define SUNYSHORE_180_DEGREES_ROTATION 0x8000
```
scbroede: these also match as 
`FX_DEG_TO_IDX(FX32_CONST(90))`
and
`FX_DEG_TO_IDX(FX32_CONST(180))`

do you think that would be better? (or ideally we would use one of the two macros already defined for this)

src/overlay008/ov8_02249960.c
```diff
@@ -1606,579 +1608,604 @@ BOOL CanalaveGym_DynamicMapFeaturesCheckCollision(FieldSystem *fieldSystem, cons
...
-    FieldSystem *fieldSystem = FieldTask_GetFieldSystem(param0);
-    UnkStruct_ov8_0224997C *v2 = FieldTask_GetEnv(param0);
-    UnkStruct_ov8_0224ABD4 *v3 = (UnkStruct_ov8_0224ABD4 *)fieldSystem->unk_04->dynamicMapFeaturesData;
+    FieldSystem *fieldSystem = FieldTask_GetFieldSystem(task);
+    int *state = FieldTask_GetEnv(task);
+    SunyshoreGymSystem *gymSystem = fieldSystem->unk_04->dynamicMapFeaturesData;
 
-    switch (v2->unk_00) {
+    switch (*state) {
     case 0: {
```
scbroede, line 2154: remove scope block

src/persisted_map_features_init.c
```diff
@@ -70,37 +69,37 @@ void PersistedMapFeatures_InitForCanalaveGym(FieldSystem *fieldSystem)
...
+        entranceZ = 21;
         break;
     case 2:
-        data->unk_00 = 0;
-        v2 = 25;
+        data->rotationState = 0;
+        entranceZ = 25;
         break;
     default:
         GF_ASSERT(0);
```
scbroede:
```suggestion
        GF_ASSERT(FALSE);
```

src/overlay008/ov8_02249960.c
```diff
@@ -1606,579 +1608,604 @@ BOOL CanalaveGym_DynamicMapFeaturesCheckCollision(FieldSystem *fieldSystem, cons
...
 
-UnkStruct_ov8_0224E740 Unk_ov8_0224E740[3] = {
+static struct {
+    int numRegions;
+    const u8 *regionsList[SUNYSHORE_NUM_ROTATION_STATES];
+} sSunyshoreCollisionLists[SUNYSHORE_GYM_NUM_ROOMS] = {
     {
-        0x6,
+        6,
         {
```
scbroede:
```suggestion
        .numRegions = 6,
        .regionsList = {
```
(same for the other array entries)

src/overlay008/ov8_02249960.c
```diff
@@ -1606,579 +1608,604 @@ BOOL CanalaveGym_DynamicMapFeaturesCheckCollision(FieldSystem *fieldSystem, cons
...
 
-    v1 = 0;
-    GF_ASSERT(param0->unk_07_0 <= 3);
+    fx16 newAngle = 0;
+    GF_ASSERT(gearSetup->initialRotation <= SUNYSHORE_NUM_ROTATION_STATES - 1);
 
-    v1 += (0x4000 * param0->unk_07_0);
-    GF_ASSERT(param1 <= 3);
+    newAngle += SUNYSHORE_90_DEGREES_ROTATION * gearSetup->initialRotation;
+    GF_ASSERT(rotationState <= 3);
```
scbroede:
```suggestion
    GF_ASSERT(rotationState <= SUNYSHORE_NUM_ROTATION_STATES - 1);
```

src/overlay008/ov8_02249960.c
VicSevenT: Yes, I do think that is better. I thought I looked for some sort of translation macro but I guess I didn't look hard enough.

I'm going to move the shorthand macro `F32_DEG_TO_IDX` from `field_camera.c` to `math_util.h` so I can use it here. It may need some renaming and further refactoring to use elsewhere in the repo but I'm leaving that as out of scope.

src/overlay008/ov8_02249960.c
VicSevenT: Removing this scope block gives me the clangd warning `Label followed by a declaration is a C23 extension`. Up to this point my convention has been to keep the scope to remove that warning, even though it doesn't actually affect the project build (not sure if it will if we ever try to build with a standard gcc or clang). 

If removing the scope is still preferred, I will look into a way to globally suppress the warning.

lhearachel: Definitely do not remove scope blocks when they are used to permit a declaration after a label.

scbroede (approved):

PR #1158, opened 2026-05-16
decompile bn_mod_exp_mont

scbroede:
i hope the person who decided to throw away the perfectly good openssl implementation and replace it with this stubs their toe or something

### DIFF
--- a/lib/crypto/src/bignum.c
+++ b/lib/crypto/src/bignum.c
@@ -1358,7 +1358,7 @@ u32 BN_mod_inverse_word(u32 word)
     return ret;
 }
 
-int BN_gen_exp_bits(BIGNUM *param0, u8 **param1, int unused, BN_CTX *param3)
+int BN_gen_exp_bits(const BIGNUM *param0, u8 **param1, int unused, BN_CTX *param3)
 {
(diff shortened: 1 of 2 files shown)


lib/crypto/src/bignum.c
```diff
@@ -2012,562 +2012,193 @@ BOOL BN_mod_exp_recp(BIGNUM *r, const BIGNUM *a, const BIGNUM *p, const BIGNUM *
...
+            break;
+        }
+    }
+
+    for (int i = nTop; i < nTop * 2; i++) {
+        v11[i] = 0;
+    }
+    bn_from_montgomery_words(rr->d, v11, v14, nTop, v12);
+
+here:
```
lhearachel, line 2185: **question**: Is there a better name for this label?

lhearachel (commented):
Generally looks good, just one question.

lib/crypto/src/bignum.c
scbroede: almost certainly, but coming up with that name would require a level of understanding of what this function is doing that i don't currently have

lhearachel (approved):

PR #1157, opened 2026-05-15
First pass at comm_manager.c

adrienntindall:
190 functions documented to the best of my ability for `comm_manager.c` as part of my ongoing effort to properly document communications related code for a project.

PR #1161, opened 2026-05-17
Field comm manager

adrienntindall:
Follows #1157 

Document 55 functions in field_comm_manager.c and finish documenting the FieldCommManager struct.
New constants: `comm_sync.h` (comm sync states)


src/field_comm_manager.c
```diff
@@ -236,102 +300,132 @@ static void Task_StartBattleServer(void)
...
     }
 
     CommInfo_SendPlayerInfo();
-    FieldCommMan_SetTask(sub_02059980, 0);
+    FieldCommManager_SetTask(FieldCommTask_WaitBattleClient, 0);
 }
 
-static void sub_02059980(void)
+/**
+ * @brief Inidicator task that the Field Communication Manager initialized as a client is waiting for battles. Does nothing when called.
```
scbroede:
```suggestion
 * @brief Indicator task that the Field Communication Manager initialized as a client is waiting for battles. Does nothing when called.
```

src/field_comm_manager.c
```diff
@@ -236,102 +300,132 @@ static void Task_StartBattleServer(void)
...
+static void FieldCommTask_ReturnToBattleRoom(void)
 {
-    void *v0;
+    void *commPlayerData;
 
-    if (CommTiming_IsSyncState(98)) {
-        v0 = Heap_Alloc(HEAP_ID_COMMUNICATION, CommPlayer_Size());
-        CommPlayerMan_Init(v0, sFieldCommMan->fieldSystem, 0);
+    if (CommTiming_IsSyncState(SYNC_CHANGE_TO_BATTLE_ROOM)) {
+        commPlayerData = Heap_Alloc(HEAP_ID_COMMUNICATION, CommPlayer_Size());
```
scbroede:
```suggestion
        void *commPlayerData = Heap_Alloc(HEAP_ID_COMMUNICATION, CommPlayer_Size());
```
does this match as the above? (same goes for `FieldCommTask_EnterBattleRoom` below)

src/field_comm_manager.c
```diff
@@ -416,43 +519,58 @@ static void sub_02059B74(void)
...
 }
 
-static void sub_02059C2C(BOOL param0, const Party *party)
+/**
+ * @brief Callback for sub_0205AB10. Effectively acts like a task, as it changes the task when called.
+ *
+ * @param param0
+ * @param party
+ */
+static void FieldCommManager_UnknownCallback(BOOL param0, const Party *party)
```
scbroede, line 541: imo it's better to leave it at `sub_` than to give it a temporary name like this

src/field_comm_manager.c
```diff
@@ -464,257 +582,340 @@ static void sub_02059C8C(void)
...
+/**
+ * @brief Task that sets the preview party and starts a Vs battle, then deletes the Field Communication Manager
+ */
+static void FieldCommTask_StartVsLinkBattle(void)
 {
-    BOOL v0 = TRUE;
-    int v1;
-    u8 v2[6];
+    int battleType;
+    u8 party[6];
```
scbroede:
```suggestion
    u8 party[MAX_PARTY_SIZE];
```

also would it be better to call this partyOrder to match the naming in `Encounter_NewVsLinkWithRecording`?

src/field_comm_manager.c
```diff
@@ -464,257 +582,340 @@ static void sub_02059C8C(void)
...
         return;
     }
 
-    FieldCommMan_SetTask(sub_02059964, 10);
+    FieldCommManager_SetTask(FieldCommTask_SendPlayerInfoClient, 10);
 }
 
-void sub_02059EAC(int param0, int unused1, void *unused2, void *unused3)
+/**
+ * @brief Sets the flag that indicates the player indicated by the netId had their trainer case successfuly copied
```
scbroede:
```suggestion
 * @brief Sets the flag that indicates the player indicated by the netId had their trainer case successfully copied
```

src/field_comm_manager.c
```diff
@@ -464,257 +582,340 @@ static void sub_02059C8C(void)
...
+    FieldCommManager_SetTask(FieldCommManager_Delete, 0);
 }
 
-SecretBase *FieldCommMan_GetCurrentOccupiedSecretBase(SaveData *saveData)
+/**
+ * @brief Gets the current secret base specified by UndergroundMan_GetCurrentOccupiedSecretBase(saveData)
+ *
+ * @param saveData
+ *
+ * @return The current SecretBase* if underground, NULL if not underground
```
scbroede:
```suggestion
 * @return The current occupied SecretBase* if underground, NULL if not underground and in a secret base
```

src/field_comm_manager.c
```diff
@@ -33,47 +35,60 @@
...
+FieldCommunicationManager *FieldCommManager_Get(void)
 {
     return sFieldCommMan;
 }
 
-static void sub_02059658(void)
+/**
+ * @brief Deadstripped function. Does nothing when called.
+ */
+static void FieldCommManager_deadstripped_02059658(void)
```
scbroede:
```suggestion
static void FieldCommManager_Dummy(void)
```

src/field_comm_manager.c
adrienntindall: The bottleneck with this function is ``sub_0205AB10`` since it's a callback for that function, the name is going to be replaced with ``FieldCommManager_[whatever sub_0205AB10 does]Callback`` eventually and I think it's valuable (for me reading the code at least) to have it labeled as a callback as opposed to confusing it for a task (since it effectively acts like one, but isn't a part of the task manager proper) in the meantime.

src/field_comm_manager.c
adrienntindall: spelling is hard =(

scbroede (approved):

PR #1162, opened 2026-05-17
Implementation of `SNELEMS`

Tai-disc:
This PR aims to implement the macro `SNELEMS` in `include/pch/global_pch.h` and replace each instance of `(int)NELEMS` with `SNELEMS`. That way, we don't need to cast `NELEMS` to `int` anymore.


lhearachel (approved):
Thank you for doing this! ❤️

PR #1166, opened 2026-05-18
Document Eterna Gym persistent features

VicSevenT:
A lot of the Eterna code was already documented. The main additions here are the code for updating the clock graphics as the the clock and the camera manager which moves the FOV to show the clock changing and the fountains going down.

### DIFF
--- a/src/overlay008/ov8_02249960.c
+++ b/src/overlay008/ov8_02249960.c
@@ -66,17 +66,25 @@
 #include "res/field/props/models/prop_models.naix"
 #include "res/text/bank/eterna_city_gym.h"
 
-#define ETERNA_CLOCK_DIAMETER     13
-#define ETERNA_CLOCK_LEFT_BOUND   5
-#define ETERNA_CLOCK_TOP_BOUND    7
-#define ETERNA_CLOCK_RIGHT_BOUND  (ETERNA_CLOCK_LEFT_BOUND + ETERNA_CLOCK_DIAMETER - 1)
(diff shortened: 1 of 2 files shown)


src/overlay008/ov8_02249960.c
```diff
@@ -177,58 +185,58 @@ typedef struct SunyshoreGymCollisionRegion {
...
+} EternaGymClockHand;
 
-typedef struct {
+typedef struct EternaGymSystem {
     fx32 fxHour;
     fx32 fxMinute;
-    int unk_08;
-    UnkStruct_ov8_0224AF58 unk_0C[2];
+    int rotateHourHand;
+    EternaGymClockHand clockHands[2];
```
lhearachel, line 210: **thought**: Is it more sensible / does it result in cleaner code to separate this into `hourHand` and `minuteHand`?

VicSevenT: The only problem I have is that at line 2467 there is a pointer created, `EternaGymClockHand *clockHands = gymsystem->clockHands`, that is then iterated over. It is still possible to get matching code if `hourHand` and `minuteHand` are broken up here, but I do think the resulting code is a little less intuitive. The individual hands are only indexed explicitly once each (around line 2426), so I don't think the gain there is worth the change to the iteration segment.

lhearachel: Understandable! We can leave it as-is.

lhearachel (approved):

PR #1169, opened 2026-05-19
Document hidden items, poffin, and record chatot cry scripts

Viperio19:
This PR documents `scripts_hidden_items.s`, `scripts_poffin_common.s` and `scripts_record_chatot_cry.s`.


res/field/scripts/scripts_hidden_items.s
```diff
@@ -2,413 +2,413 @@
...
+    ScriptEntry HiddenItems_Item
+    ScriptEntry HiddenItems_Item
+    ScriptEntry HiddenItems_Item
+    ScriptEntry HiddenItems_Item
+    ScriptEntry HiddenItems_Item
+    ScriptEntry HiddenItems_Item
+    ScriptEntry HiddenItems_Item
+    ScriptEntry HiddenItems_Item
+    ScriptEntry HiddenItems_Item
     ScriptEntryEnd
```
Viperio19: Similar to the battle script entries, I think it would be good to make these generated. My suggestion would be to at some point make it so this is generated from `gHiddenItems`.

lhearachel: This can be added to `itemproc` starting from using `countof` on `gHiddenItems` after including the appropriate header, if you feel up to it. 🙂

lhearachel (approved):

lhearachel:
Approving as-is, unless you want to tackle the `itemproc` bit to generate those script entries.

tools/dataproc/src/itemproc.c
```diff
@@ -376,6 +382,21 @@ static void add_idmap(datafile_t *df, size_t i, size_t data_id) {
             i, data_id, icon, palette, gba_id);
 }
 
+static void emit_hidden_scripts(void) {
```
lhearachel, line 385: **praise**: Dope, thank you! ❤️

PR #1176, opened 2026-05-21
Implement moveproc, restructure move-related assets, and begin documenting the dataproc toolchain

lhearachel:
I apologize in advance for the size of this PR.

## Description

This PR implements the `moveproc` compiler for handling move-related asset files. It processes the `data.json` file associated with each in-game move and produces:

1. The in-game data table,
2. Text banks for in-game text (names, descriptions, in-battle messages), and
3. The expected ordering of the animation and execution script archives.

The changes to `.json` and `.s` files can largely be ignored. The structure for the move-related assets is now, summarized:

```
res/moves
├── absorb
│   ├── anim.s
│   ├── data.json
│   └── script.s
├── acid
│   ├── anim.s
│   ├── data.json
│   └── script.s
├── acid_armor
│   ├── anim.s
│   ├── data.json
│   └── script.s
...
```

One such subdirectory exists for each move. Additionally, the individual subdirectories for the bogus move entries (e.g., `0468`, `0469`, etc.) have been removed. Data for those bogus entries is now embedded into the `moveproc` tool, and scripts for them have been ported to `res/moves/.shared`.

> [!NOTE]
> This structure is a bit of a preview of my envisioned changes for field maps, which will look quite similar!

The build definitions are a bit scattered. Move execution scripts (`script.s`) are located in each `res/moves` subdirectory, but they cannot be built inside of `res/moves` due to a dependency on the text archives (which would create a circular dependency between `res/text` and `res/moves`). Thus, we just declare the existence of these files in `res/moves` and continue building them in `res/battle/scripts`.

## Additional Changes

Some changes are also made here to unify names for certain move-related constants.

## Caveats

- I considered breaking `res/battle/scripts` down to `res/battle_scripts`, but opted against it for now to keep scope from bloating any further.


lhearachel:
Some additional thoughts:

1. I have some dummy-scripts that I used to modify all of the files which may be of interest to end-users. Maybe those should be included in the repository?
2. Do we want to keep the flag-implementation as is for moves? Or do we want to change it to a similar structure as item parameters where we have an object of known fields?

docs/datafiles/moves.md
```diff
@@ -0,0 +1,214 @@
...
+        "effect": "CONTEST_EFFECT_BASIC",
+        "type": "CONTEST_TYPE_BEAUTY"
+    }
+}
+```
+
+## `anim.s` — Animation Script
+
+This script defines the commands executed to play the move's animation when used
+in a battle or contest. Commands are defined by `asm/macros/btlanimcmd.inc`.
```
scbroede: do we want to point users to `btlanimfunc.inc` here too? it is included in `btlanimcmd.inc` so it wouldn't be too hard to find, but stuff like `Func_FadeBattlerSprite` and `Func_Shake` in the below example is defined there instead

docs/datafiles/overview.md
```diff
@@ -0,0 +1,111 @@
...
+| `itemproc`     | `res/items/data/*.json` + `include/constants/items.h`      | Item metadata, hidden-item script definitions, item-to-icon mappings, TM-to-move mappings. |
+| `moveproc`     | `res/moves/<name>/data.json` + `include/constants/moves.h` | Move metadata, in-game battle text, and animation scripts.                                 |
+| `npctradeproc` | `res/npc_trades/*.json` + `include/constants/npc_trades.h` | Pokémon offered by NPCs for an in-game trade.                                              |
+| `resdatproc`   | Sprite template JSON + NARC ID maps                        | Serialized sprite templates to be loaded from the filesystem.                              |
+| `speciesproc`  | `res/pokemon/<species>/data.json`                          | Pokémon base stats, evolutions, learnsets, and sprite positions.                           |
+| `trainerproc`  | `res/trainers/data/*.json`                                 | Trainer metadata, including their parties and any in-battle messages.                      |
```
scbroede: nitpick: it's not just in-battle messages; all trainer-related messages are generated here, including the ones when talking to them in the overworld (except the ones where a trainer can only be battled at morning/night and you're talking to them outside of that period, those are in the map text banks)

scbroede (commented):
the move animation entries in `bugs_and_glitches.md` need to be updated with the new paths

docs/datafiles/moves.md
lhearachel: Yeah, that makes sense to me.

scbroede (approved):

PR #1182, opened 2026-05-29
Document Veilstone Gym persistent features and finish up Overlay 8

VicSevenT:
The Veilstone Gym features include the punching bags and tire stacks that exist around the gym. The overlay 8 codes deals largely with determining the behavior of a punching bag when you come in contact with one, and there is a dedicated ov5 file that handles rendering the objects.

This is the last part of overlay 8 to be documented, so as part of this I also renamed the file+overlay. I think this file should ultimately go inside a field/overworld subfolder, so until that structure exists I left the file in `overlay008/` to indicate that it still should be moved eventually.


include/overlay008/gym_features.h
```diff
@@ -63,31 +67,36 @@ typedef struct SunyshoreGymPersistedFeatures {
...
 void VeilstoneGym_DynamicMapFeaturesFree(FieldSystem *fieldSystem);
 BOOL VeilstoneGym_DynamicMapFeaturesCheckCollision(FieldSystem *fieldSystem, const int tileX, const int tileZ, const fx32 height, BOOL *isColliding);
-BOOL ov8_0224BF4C(FieldSystem *fieldSystem);
+BOOL VeilstoneGym_HitPunchingBag(FieldSystem *fieldSystem);
+
+void PastoriaGym_DynamicMapFeaturesInit(FieldSystem *fieldSystem);
+BOOL PastoriaGym_DynamicMapFeaturesCheckCollision(FieldSystem *fieldSystem, const int tileX, const int tileZ, const fx32 height, BOOL *isColliding);
+void PastoriaGym_PressButton(FieldSystem *fieldSystem);
+
 void HearthomeGym_DynamicMapFeaturesInit(FieldSystem *fieldSystem);
```
lhearachel, line 85: **question**: Do we want to list these between Eterna and Veilstone to preserve Platinum's gym-ordering?

src/overlay005/veilstone_gym_object_renderer.c
```diff
@@ -0,0 +1,430 @@
...
+
+        Heap_Free(renderer->animations);
+    }
+}
+
+static void TireStackAnim_Update(OverworldAnimManager *animManager, void *context)
+{
+    VeilstoneGymTireStackRenderer *renderer = context;
+
+    switch (renderer->updateAnimations) {
```
lhearachel, line 273: **question**: Does this match as an `if (renderer->updateAnimations == TRUE)` with no `else` block?

src/overlay005/veilstone_gym_object_renderer.c
```diff
@@ -0,0 +1,430 @@
...
+    Simple3D_DrawRenderObjWithPos(&renderer->renderObj, &pos);
+}
+
+static const OverworldAnimManagerFuncs sTireStackAnimFuncs = {
+    sizeof(VeilstoneGymTireStackRenderer),
+    TireStackAnim_Init,
+    TireStackAnim_Free,
+    TireStackAnim_Update,
+    TireStackAnim_Render
+};
```
lhearachel, line 308: **polish**: Add designated initializers here.

src/overlay008/gym_features.c
```diff
@@ -86,6 +87,32 @@
...
+#define VEILSTONE_NUM_TIRE_STACKS    11
+#define VEILSTONE_NUM_PUNCHING_BAGS  9
+
+#define VEILSTONE_OBJ_IS_TIRE_STACK(obj) (MapObject_GetDataAt(obj, 0) == FALSE)
+
+#define VEILSTONE_TILE_FLAG_BAG_CANT_MOVE_THAT_WAY (1 << 0)
+#define VEILSTONE_TILE_FLAG_TIRE_STACK_PRESENT     (1 << 1)
+#define VEILSTONE_TILE_FLAG_BAG_PAUSE_POINT        (1 << 2)
+
+#define VEILSTONE_ANIM_STATE_INIT                  0
```
lhearachel: **question**: Can / should this be an `enum`? Maybe not necessarily a named one. It might also be valuable to add a `VEILSTONE_ANIM_STATE_MAX` for the `11`s scattered through the implementation.

src/overlay005/veilstone_gym_object_renderer.c
VicSevenT: It does not

include/overlay008/gym_features.h
VicSevenT: I organized as much as I could using the original DP ordering. IMO even though its different in plat the original order is still how I think of it, plus the badge enum and trainer case still use the original order.

lhearachel (approved):

PR #1185, opened 2026-06-01
Document FieldBattleDTO_New[BattleFacility] functions

Viperio19:
This PR documents the functions that create a `FieldBattleDTO` for a battle frontier battle, for example `FieldBattleDTO_NewBattleTower` in `FrontierScrCmd_BattleTower_StartBattle`.

PR #1189, opened 2026-06-02
Document Battle Hall struct Init and save data

VicSevenT:
Documents most of the init and free functions for the `BattleHall` struct, as well as the `BattleHallSave` struct that stores all the information necessary to resume a Battle Hall run when the player saves mid-challenge.

Also managed to identify the remaining unknown `BattleFrontierStatsIndex`s and some other miscellaneous battle hall-related functions.


src/overlay104/ov104_02234DB4.c
```diff
@@ -15,124 +13,111 @@
...
 
-        for (v5 = 0; v5 < v4; v5++) {
-            v10->unk_260[v5] = (u8)sub_02030198(v0, 3, v5, 0, NULL);
+        for (i = 0; i < partySize; i++) {
+            battleHall->partySlots[i] = BattleHallSave_GetMember(currentChallenge, 3, i, 0, NULL);
         }
 
-        for (v5 = 0; v5 < (10 * 2); v5++) {
-            v10->unk_18[v5] = (u16)sub_02030198(v0, 2, v5, 0, NULL);
+        for (i = 0; i < (10 * 2); i++) {
```
lhearachel: **question**: This `10 * 2` is a bit suspicious to me. Might inform the `[20]` uses in the struct-defs?

src/battle_hall_save.c
lhearachel, line 1: **question**: Should this be `save_battle_hall` to keep it with the other `save`-related modules?

VicSevenT: We currently have `save_player`, `save_catchrecords`, and `save_info_window`, but also `daycare_save`, `trainer_case_save_data`, and `wifi_history_save_data`, so I'm not sure a standard has been decided?

src/overlay104/ov104_02234DB4.c
VicSevenT: Rebasing after https://github.com/pret/pokeplatinum/pull/1185 revealed that `unk_18` is `trainerIDs`, so I think the 10 is the battles per round * 2 for when it's two trainers per battle in a multi battle. Adding a constant for that.

There is another `10 * 2` right below this for `unk_268` that I am going to leave untouched for now until `unk_268` is identified.

src/battle_hall_save.c
lhearachel: Hm, I think this is worth bike-shedding at some point, but probably not in this PR.

src/battle_hall_save.c
```diff
@@ -0,0 +1,147 @@
...
+    case 2:
+        return save->trainerIDs[index];
+    case 3:
+        return save->partySlots[index];
+    case 4:
+        return save->unk_2C[index];
+    default:
+        GF_ASSERT(FALSE);
+        break;
+    }
```
lhearachel, line 72: **suggestion**: Make constants for these field accesses.

lhearachel (approved):
Just one more suggestion, the rest looks good to me. 😁

PR #1191, opened 2026-06-03
Document catch records: unk_0202F180

zackhudg:
Rename `unk_0202F180` to `catchrecords_save`


src/save_catchrecords.c
```diff
@@ -1,7 +1,6 @@
-#include "unk_0202F180.h"
+#include "catchrecords_save.h"
 
 #include <nitro.h>
-#include <string.h>
```
zackhudg, line 4: assuming removing unused headers is correct?

include/save_catchrecords.h
lhearachel, line 1: **thought**: Could we name this `save_catchrecords.[h,c]` instead?

src/save_catchrecords.c
lhearachel: Yes! Honestly, the `nitro.h` include could also be removed here.

lhearachel (commented):
Thanks for doing this! Just one little nitpick-thought about the name.

include/save_catchrecords.h
zackhudg: sure thing - just curious on the thought process for future purposes, thanks. Saw some were prefixed and some suffixed so was not sure which to do.

lhearachel: In this case, I think it's valuable to keep the file closer to other `save`-related files, hence the prefix.

lhearachel (approved):

PR #1193, opened 2026-06-05
document sentences and words

scbroede:


include/constants/string_padding_mode.h
lhearachel, line 1: **question**: Should this live with `constants/string.h`? I think ideally I'd like to compress constants-headers as much as possible.

lhearachel (commented):
I'm not a huge fan of this naming (it feels extremely generic), but I don't have a strong alternative suggestion given how many distinct domains make use of it.

Just one question below.

include/constants/string_padding_mode.h
scbroede: makes sense to me. moved

lhearachel (approved):

PR #1194, opened 2026-06-05
Document battle tower wifi mode dummy dtos

Viperio19:
This PR documents the dummy arrays of `FrontierPokemonDataDTO` and `FrontierTrainerDataDTO` for the battle tower wifi mode. Possibly it's too fast for now to already name them `sDummyPokemonDTOs` and `sDummyTrainerDTOs` but I wanted to at least make the suggestion, and I mainly just wanted to decorate the structs.


include/struct_defs/struct_0202D63C.h
```diff
@@ -20,10 +22,10 @@ typedef struct UnkStruct_0202D63C_t {
...
         u8 unk_C8_val2;
     };
-    u8 unk_C9;
-    u16 unk_CA[4];
-    u16 unk_D2[4];
-    u16 unk_DA[4];
+    u8 trainerType;
+    u16 introMsg[4];
+    u16 winMsg[4];
+    u16 loseMsg[4];
```
lhearachel, line 28: **question**: Should these have type `charcode_t`?

lhearachel (commented):
Just one question; this looks fine otherwise. 😁

include/struct_defs/struct_0202D63C.h
Viperio19: From a quick glance, they seem to just be message or sentence ids, so I don't think so

lhearachel (approved):

PR #1195, opened 2026-06-06
nitrogfx: add support for extended palettes

scbroede:
upstream pr: https://github.com/red031000/nitrogfx/pull/55

### DIFF
--- a/tools/nitrogfx/gfx.c
+++ b/tools/nitrogfx/gfx.c
@@ -1340,6 +1340,7 @@ void ReadNtrPalette(char *path, struct Palette *palette, int bitdepth, int palIn
 {
     int fileSize;
     bool inverted = false;
+    palette->extendedLength = 0;
     unsigned char *data = ReadWholeFile(path, &fileSize);
 
     if (memcmp(data, "RLCN", 4) != 0 && memcmp(data, "RPCN", 4) != 0) //NCLR / NCPR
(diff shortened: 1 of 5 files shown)


tools/nitrogfx/gfx.c
VicSevenT, line 1: Is generating a PNG that converts back to NCGR+NCLR supposed to be supported? When I try to create a PNG without `-convertTo8Bpp`, I am able to reverse it to get a NCGR, but generating a NCLR fails with `Unrecognized option "-extended"`.

Image

Using `-convertTo8Bpp`, I can no longer recover a matching NCGR from the PNG, and the NCLR generation fails again with an unrecognized option (although the Suggested option changes from `-extended 1` to `-extended 16`.

Image

VicSevenT, line 1: When generating PAL files and then trying to convert back to NCLR, I'm getting an error about the number of colors.

Image

VicSevenT (commented):
I tried this out using a graphics and palette file from `frontier_bg.narc` and ran into some issues/questions.

[files.zip](https://github.com/user-attachments/files/28814916/files.zip)

tools/nitrogfx/gfx.c
scbroede: should be fixed now

scbroede: generating an extended nclr from a png isn't supported. i think in order to do that with these palettes the image would have to be 12bpp

VicSevenT: Yup working for me now

scbroede:
converting to draft until we work out how we want to store the palette

scbroede:
new usage:
`./nitrogfx.exe test_0.pal test.NCLR -extended 1 -paddingtype 1`

where extended is the amount of non-padding palettes and paddingtype is 0 for padding with 0s and 1 for padding with that weird-ass pattern. if we ever don't want padding (which i'm not sure is a thing for these, but could be possible), the `-nopad` flag will handle that

VicSevenT:
Looks good to me. I've been able to wire this up with meson and start unpacking the extended palettes in `frontier_bg.narc` and it's all working well. Thanks again for doing this!

lhearachel (commented):
Matching `nitrogfx` PR: https://github.com/red031000/nitrogfx/pull/58

lhearachel (approved):
Upstream PR has been merged

PR #1201, opened 2026-06-14
Standardize dummy/unused script blocks/entries

Viperio19:
This PR attempts to standardize the names for unused script blocks/entries. The introduced standard is:
- `[Map]_Dummy[X]` for dummy script entries (an entry containing just `End` or also a dummy message) where X is the script id 
- `[Map]_[Description]_Unused` for unused script blocks/entries with a clear purpose (e.g. `CelesticTownNorthwestHouse_Vendor_Unused`)
- `[Map]_Unused[X]` for unused script blocks that are more just left over garbage where X is just an enumerator for the amount of unused script blocks
There are some cases where I didn't exactly follow this just because some script blocks/entries don't fully fall into just one of these categories, but I hope this at least sets a standard that is mostly easy to follow.

### DIFF
--- a/res/field/scripts/scripts_amity_square.s
+++ b/res/field/scripts/scripts_amity_square.s
@@ -55,7 +55,7 @@
     ScriptEntry AmitySquare_Warp25
     ScriptEntry AmitySquare_Warp26
     ScriptEntry AmitySquare_Warp27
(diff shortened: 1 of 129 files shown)


lhearachel (approved):
A quick skim looks good to me

PR #1206, opened 2026-06-16
Polish scripts pt 2 (F-O)

### LINKED ISSUE #1188: FLAG_OBTAINED_LUNAR_WING Is a Misnomer
This flag, which is set after interacting with Cresselia for the first time on Fullmoon Island, erroneously implies that it is set when you pick up the Lunar Wing. This is not true as the Lunar Wing is handled through a typical ground item script that is separate from the script for interacting with Cresselia, and thus does not reflect the player obtaining the Lunar Wing.

A better name would be `FLAG_MET_CRESSELIA`, akin to `FLAG_MET_BEBE`. This says nothing about Cresselia's roaming status, nor about the Lunar Wing.

Viperio19:
Continuation of alphabetically going through scripts to improve documentation. Closes #1188


lhearachel (approved):
A quick skim looks good to me

PR #1210, opened 2026-06-18
Document Frontier script startup process

VicSevenT:
Includes partial documentation of `ScrCmd_LaunchBattleFrontierScene` and its internals, which is the main entrypoint into the frontier scripts. Also documents the struct that helps pass global structs into the frontier context and documents more of the scenes data.


include/struct_decls/battle_frontier_decl.h
lhearachel, line 1: Just checking that this forward declaration is required

VicSevenT: Yeah, `BattleFrontier` has some circular references with `FrontierGraphics` and `FrontierScriptManager`. Eventually it may make sense to move all the frontier-related structs into a single .h file, but for now this `decl` is needed.

lhearachel (approved):

PR #1215, opened 2026-06-21
document wallpaper passwords

scbroede:
fuck this narc tbh

so i put the narc in its own folder because it doesn't really fit into any of the existing ones, but i don't really see anything else going into the new folder in the future either. maybe this is a case where we want to create a more general `data` folder or something that future narcs might fit in to?


res/words/meson.build
```diff
@@ -0,0 +1,64 @@
...
+    '-ipa', 'file',
+    '-interworking',
+    '-inline', 'on,noauto',
+    '-char', 'signed',
+    '-nosyspath',
+    '-stdinc',
+    '-sym', 'on',
+]
+
+word_bank = static_library('word_bank',
```
lhearachel: **question**: Is there a reason to compile this in this manner instead of with `arm-none-eabi-gcc` like we do for the trainer AI script?

We could also split the table in the source-file into its own `.c` file, then include that in the source file in the ROM.

tools/nitroarc/src/common.h
lhearachel, line 1: **issue**: Make sure to file a PR to the upstream repository (or I can port the changes myself and associate them with you via other means)

tools/nitroarc/src/main.c
```diff
@@ -114,6 +115,7 @@ static const option_cfg_t s_options[] = {
 
     [exclude]      = { "exclude",      NULL, 'E', 0, true  },
     [exclude_from] = { "exclude-from", NULL, 'X', 0, true  },
+    [file_from]    = { "file-from",    NULL, 't', 0, true  },
```
lhearachel: **issue**: This cannot use `'t'` (`list` supersedes it above); I think `file-from` is a bit of a misnomer, anyways, maybe that could be `single-file`? And then this could be `'s'`.

res/words/meson.build
```diff
@@ -0,0 +1,64 @@
+word_bank_srcs = files(
+    meson.global_source_root() / 'src/password_word_bank.c'
+)
+
+word_bank_c_args = [
```
lhearachel: **question**: Why not just the standard args?

res/words/meson.build
```diff
@@ -0,0 +1,64 @@
...
+
+word_bank = static_library('word_bank',
+    sources: word_bank_srcs,
+    c_args: word_bank_c_args,
+    include_directories: [
+        public_includes,
+        toplevel_includes,
+    ],
+    dependencies: [nitrosdk_dep],
+    pic: false
```
lhearachel: **note**: This should be unnecessary for a `static_library`.

lhearachel (changes requested):

tools/nitroarc/src/common.h
```diff
@@ -60,7 +60,7 @@ struct options {
     const char *directory;    // -C, --directory=DIR
     const char *file;         // -f, --file=ARCHIVE
     const char *format;       // -F, --format=FORMAT
-    const char *file_from;    // -t  --file=FILE
+    const char *single_file;    // -t  --file=FILE
```
lhearachel: **nitpick**: Align the comment with the others

tools/nitroarc/src/main.c
```diff
@@ -166,7 +166,7 @@ int parse_opts(const char **argv, options_t *opts) {
         case directory:    target_optarg(opts->directory);
         case file:         target_optarg(opts->file);
         case format:       target_optarg(opts->format);
-        case file_from:    target_optarg(opts->file_from);
+        case single_file:    target_optarg(opts->single_file);
```
lhearachel: **nitpick**: Align the statement with the others

lhearachel (changes requested):

tools/nitroarc/src/common.h
scbroede: done

lhearachel (approved):

PR #1217, opened 2026-06-21
Documenting src/unk_0202D05C.c and renaming it to wifi_battle_tower_save.c

Land0fChocolate:


include/overlay096/struct_ov96_0223B450_sub2.h
lhearachel, line 1: **issue**: Delete files rather than just comment out their content.

include/struct_decls/struct_0202D060_decl.h
lhearachel, line 1: **issue**: Move structure definitions into headers that contains the functions for mutating them, then delete these declaration-only files.

include/wifi_battle_tower_save.h
lhearachel, line 1: **issue**: Rename the header and source files, per the PR description.

lhearachel (changes requested):

include/overlay096/struct_wifi_player_profile.h
lhearachel, line 1: **question**: Why is this struct still separate? It clearly does not belong to overlay 96, because it is used in the save structure.

Land0fChocolate: It's separate because I don't know where it goes yet. It probably doesn't fit with the Wifi Battle Tower stuff. It might fit in with some code that is currently undocumented. I left it as it is until I have more information.

Land0fChocolate: `WifiPlayerProfile_Size` in `src/wifi_battle_tower_save.c` also probably doesn't belong there. Again, I've left it as it is until I can find a better home for it.

src/wifi_battle_tower_save.c
```diff
@@ -0,0 +1,545 @@
...
+        save->unk_03 += param1;
+    }
+
+    if (save->unk_04 + param2 < 65535) {
+        save->unk_04 += param2;
+    }
+
+    if (save->unk_06 + param3 < 65535) {
+        save->unk_06 += param3;
+    }
```
lhearachel: **suggestion**: Try these to see if they match, just to encode the type information into the comparison:

```suggestion
    if (save->unk_03 + param1 < (u8)(-1)) {
        save->unk_03 += param1;
    }

    if (save->unk_04 + param2 < (u16)(-1)) {
        save->unk_04 += param2;
    }

    if (save->unk_06 + param3 < (u16)(-1)) {
        save->unk_06 += param3;
    }
```

We might want to make these special constants, if they aren't already available to our toolchain.

src/wifi_battle_tower_save.c
```diff
@@ -0,0 +1,545 @@
...
+        record->flags &= mask;
+        break;
+    case 1:
+        record->flags |= mask;
+        break;
+    case 0:
+        return (BOOL)((record->flags >> bitIdx) & 0x1);
+    }
+
+    return 0;
```
lhearachel: **polish**: `FALSE`

src/wifi_battle_tower_save.c
```diff
@@ -0,0 +1,545 @@
...
+{
+    BattleFrontierSave *frontier = SaveData_SaveTable(saveData, SAVE_TABLE_ENTRY_FRONTIER);
+    EasyChatSentence_Copy(&(frontier->unk_950.easyChatMessages.sentences[sentenceIdx]), sentence);
+}
+
+EasyChatSentence *FrontierEasyChatMessages_GetSentence(SaveData *saveData, int sentenceIdx)
+{
+    BattleFrontierSave *frontier = SaveData_SaveTable(saveData, SAVE_TABLE_ENTRY_FRONTIER);
+
+    return &(frontier->unk_950.easyChatMessages.sentences[sentenceIdx]);
```
lhearachel: **polish**: No need for these `parens` around the argument to a address-of operator (`&`).

src/wifi_battle_tower_save.c
```diff
@@ -0,0 +1,545 @@
...
+}
+
+void WifiBattleTowerDownloadData_MarkOpponent(WifiBattleTowerDownloadData *downloadData, u8 rank, u8 opponentIdx, RTCDate *date)
+{
+    u8 byteIdx, bitIdx;
+    u8 mask = 1;
+    u16 flatIdx;
+
+    if ((opponentIdx == 0) || (opponentIdx > 200)) {
```
lhearachel: **polish**: Remove unnecessary parens on such comparisons

src/wifi_battle_tower_save.c
```diff
@@ -0,0 +1,545 @@
...
+    downloadData->lastDownloadDate = Date_Encode(date);
+}
+
+void WifiBattleTowerDownloadData_Reset(WifiBattleTowerDownloadData *downloadData)
+{
+    MI_CpuClear8(downloadData->downloadedOpponents, 250);
+    MI_CpuClear8(&downloadData->lastDownloadDate, sizeof(u32));
+}
+
+static BOOL Date_IsAfter(RTCDate *dateA, RTCDate *dateB)
```
lhearachel, line 409: **polish**: Use `TRUE` and `FALSE` as return values in functions that return `BOOL`

lhearachel (commented):
Just a few polish bits left after getting everything else squared away, nice work!

lhearachel (approved):

PR #1219, opened 2026-06-22
Improve names regular vars and local, regular, and daily flags

Viperio19:
This PR standardizes the names of vars and flags (using `FLAG_RECEIVED_` for being given an item instead of `FLAG_OBTAINED_` for picking it up, using full location names where relevant), improves the names of some other flags and vars, and marks the map local and daily flags as such.


generated/vars_flags.txt
```diff
@@ -1,68 +1,70 @@
...
-FLAG_HIDE_OBSTACLE_30
-FLAG_UNK_0x003E
-FLAG_UNK_0x003F
-FLAG_UNK_0x0040
+FLAG_UNUSED_0x0000
+MAP_LOCAL_FLAGS_START
+FLAG_MAP_LOCAL_0x0001 = MAP_LOCAL_FLAGS_START
+FLAG_MAP_LOCAL_0x0002
+FLAG_MAP_LOCAL_0x0003
+FLAG_MAP_LOCAL_0x0004
```
Viperio19: Do we want to rename these to for example `FLAG_MAP_LOCAL_1` to reflect the map local vars (`VAR_MAP_LOCAL_0`)? I personally don't love that there is `VAR_MAP_LOCAL_A` and `VAR_MAP_LOCAL_10`, but maybe there is some middle ground that work nicely here such as just shortening to `FLAG_MAP_LOCAL_0x1` and then also `VAR_MAP_LOCAL_0xA` and `VAR_MAP_LOCAL_0x10`

lhearachel: I think that that is sound, yeah. I personally don't have an issue with `VAR_MAP_LOCAL_A`, but I can get how that might be confusing to an end-user.

generated/vars_flags.txt
```diff
@@ -1,68 +1,70 @@
...
-FLAG_HIDE_OBSTACLE_27
-FLAG_HIDE_OBSTACLE_28
-FLAG_HIDE_OBSTACLE_29
-FLAG_HIDE_OBSTACLE_30
-FLAG_UNK_0x003E
-FLAG_UNK_0x003F
-FLAG_UNK_0x0040
+FLAG_UNUSED_0x0000
+MAP_LOCAL_FLAGS_START
+FLAG_MAP_LOCAL_0x1 = MAP_LOCAL_FLAGS_START
```
lhearachel: **suggestion**: Use a leading zero for consistency in width: `0x01`, `0x02`, etc.

res/field/scripts/scripts_oreburgh_mine_b1f.s
```diff
@@ -75,7 +75,7 @@ OreburghMineB1F_RemoveHiker:
     End
 
 OreburghMineB1F_YouHaveAnOveradundance:
```
lhearachel: **typo**: `Overabundance`

generated/vars_flags.txt
Viperio19: done!

lhearachel (approved):

PR #1179, opened 2026-05-26
comm_command

adrienntindall:
Builds on #1161 

This PR is long but ultimately has 3 goals:

1) Fully document the contents of `comm_command.c`. This file should be reviewed thoroughly, and is thankfully short.

2) Give general names and corresponding constants to each other `comm_command_` file, named as such when they contain a `CommCmdTable` list of functions. These are not intended to be documented further and only the tables themselves need to be looked at when reviewing, since they will contain all already documented functions that had their names changed.  

3) Use the `COMM_CMD_` constants in appropriate locations in order to maintain functionality when adding new communication functions to the base table. I'm pretty sure I found all instances, but if I missed any I would like that to mentioned as a part of this review.

PR #1220, opened 2026-06-23
Document most of the Battle Hall logic

VicSevenT:
This covers most of the battle hall-specific frontier script commands, including all the logic for picking the next opponent. The main chunk of logic left remaining for the Battle Hall is the communication code, which I am leaving until after #1179 is merged. 

`battle_hall_helpers.c` contains some massive lists used to help generate the next opponent. One of these is a listing of frontier trainer classes, and I was able to generate this in `frontierproc`. The other two are related to the pokemon in `pl_bsdpm.narc`, which is still not unpacked. I hope to be able to unpack this and generate those lists to in the near future.


asm/macros/frscrcmd.inc
```diff
@@ -766,9 +766,9 @@
     .short \partySlot2
     .endm
 
-    .macro FrontierScrCmd_8C arg0
-    .short FRSCRCMD_8C
-    .short \arg0
+    .macro BattleHall_SetupNextOpponent unused
```
lhearachel: **question**: Can we provide a default value for `unused`?

asm/macros/frscrcmd.inc
```diff
@@ -800,21 +800,21 @@
     .short \destVar
     .endm
 
-    .macro FrontierScrCmd_93 arg0, arg1, arg2
-    .short FRSCRCMD_93
-    .short \arg0
-    .short \arg1
-    .short \arg2
+    .macro BattleHall_SendCommMessage command, arg, returnVar
```
lhearachel, line 803: **suggestion**: I would like to annotate the arguments to script-commands with `:req` to denote that they are required. Otherwise, it's a pretty big foot-gun for an end-user. That can be a separate PR, but we should definitely do it.

I have a basic `awk`-script for this, if that interests you:

```awk
#!/usr/bin/env -vS awk -f

! /\.macro/ { print $0 }

/\.macro/ {
    for (i = 3; i <= NF; i++) {
        if ($i ~ /.*,/) {
            sub(/,/, ":req,", $i)
        } else {
            $i = $i ":req"
        }
    }

    print "    " $0
}
```

Stick this in a file, make it executable, and run it against the input file:

```bash
./annotate.awk asm/macros/frscrcmd.inc > asm/macros/frscrcmd.inc.new
```

Then you can check the output and replace the old file.

src/overlay104/battle_hall.c
```diff
@@ -276,115 +264,107 @@ void BattleHall_Save(BattleHall *battleHall, u8 saveType)
...
 
-u16 ov104_02235578(BattleHall *param0)
+u16 BattleHall_GetEarnedBP(BattleHall *battleHall)
 {
-    u8 v0;
-    static const u8 v1[18 + 1] = { 0, 1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 6, 6, 8, 8, 10, 10, 12 };
-    static const u8 v2[18 + 1] = { 0, 6, 6, 6, 8, 8, 8, 10, 10, 10, 12, 12, 14, 15, 17, 17, 20, 20, 23 };
+    u8 bp;
+    static const u8 sBPPerRoundSolo[18 + 1] = { 0, 1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 6, 6, 8, 8, 10, 10, 12 };
+    static const u8 sBPPerRoundWiFi[18 + 1] = { 0, 6, 6, 6, 8, 8, 8, 10, 10, 10, 12, 12, 14, 15, 17, 17, 20, 20, 23 };
```
lhearachel: **polish**: Is there something magic to these `18` values? Is that just the maximum number of rounds? Can we encode that?

src/overlay104/battle_hall.c
```diff
@@ -452,64 +429,61 @@ void BattleHall_CalcPlayerLevelSqrt(BattleHall *battleHall)
...
-    v6 = (FX_FX32_TO_F32(param0->playerLevelSqrt) * 5.0);
+    baseLevel = FX32_CONST(playersLevel) - battleHall->playerLevelSqrt * 3;
+    float rankFactor = FX_FX32_TO_F32(battleHall->playerLevelSqrt) * 5.0;
 
-    if ((v2 / v6) < 1.0) {
-        v6 = 1.0;
-        v6 = (float)(v5 + 1 - 1);
+    if (playersLevel / rankFactor < 1.0) {
+        rankFactor = 1.0;
+        rankFactor = typeRank + 1 - 1;
```
lhearachel, line 448: **question**: Does this match without the `+ 1 - 1`?

src/overlay104/battle_hall_helpers.c
```diff
@@ -0,0 +1,1631 @@
...
+                }
+            } else {
+                if (selectedType == sBattleHallPotentialOpponentTypes[opponentIndex][0] || selectedType == sBattleHallPotentialOpponentTypes[opponentIndex][1]) {
+                    if (playersSpecies != sBattleHallPotentialOpponents[opponentIndex]) {
+                        opponentIndices[currentBattleIndex + opponentsFound] = opponentIndex + 1;
+
+                        opponentsFound++;
+                    }
+                }
+            }
```
lhearachel: **suggestion**: Unwind a level of nesting here by making the `else` an `else if`

src/overlay104/battle_hall_helpers.c
```diff
@@ -0,0 +1,1631 @@
...
+    case FRONTIER_CHALLENGE_DOUBLE:
+        return 2;
+    }
+
+    return 1;
+}
+
+u8 BattleHall_GetOpponentPartySize(u8 challengeType)
+{
+    switch (challengeType) {
```
lhearachel: **question**: Does this function not match with an `if else`? Or as a ternary expression?

src/overlay104/battle_hall.c
VicSevenT: No

src/overlay104/battle_hall_helpers.c
VicSevenT: Matches with a conditional

src/overlay104/battle_hall.c
VicSevenT: Yeah, I guess it would be the number of "unique rounds". Round 18 is the first round after the Gold battle, so after that rewards and logic don't change. I'll create a constant for that.

asm/macros/frscrcmd.inc
VicSevenT: Makes sense, but as you say, I think this should be in a separate PR.

lhearachel (approved):

PR #1222, opened 2026-06-27
Documenting stuff in overlay096

Land0fChocolate:
~~TODO: rename files and organise them. Maybe make a wifi subdirectory.~~


include/overlay096/struct_world_exchange_trainer_error.h
```diff
@@ -1,6 +1,9 @@
 #ifndef POKEPLATINUM_STRUCT_WORLD_EXCHANGE_TRAINER_ERROR_H
 #define POKEPLATINUM_STRUCT_WORLD_EXCHANGE_TRAINER_ERROR_H
 
+// validationError and systemError values come from Nintendo's closed API;
+// the exact semantics are not documented.
```
lhearachel: **question**: Do we need to keep this? "The exact semantics" is a bit confusing to me.

include/wifi_battle_tower/application.h
```diff
@@ -0,0 +1,29 @@
...
+int WifiBattleTower_AppInit(ApplicationManager *appMan, int *state);
+int WifiBattleTower_AppMain(ApplicationManager *appMan, int *state);
+int WifiBattleTower_AppExit(ApplicationManager *appMan, int *unused);
+void WifiBattleTower_BuildAffineSpriteTemplate(AffineSpriteListTemplate *template, WifiBattleTowerAppState *appState, SpriteResourcesHeader *spriteResourceHeader, int vramType);
+Menu *WifiBattleTower_CreateYesNoMenu(BgConfig *bgConfig, int tilemapTop, int baseTile);
+void WifiBattleTower_SetState(WifiBattleTowerAppState *appState, int state, int nextState);
+int WifiBattleTower_GetSignalStrength(void);
+void WifiBattleTower_SetExitMode(WifiBattleTowerAppState *appState, int exitMode, int unused);
+
+enum WifiBattleTowerLoopState {
```
lhearachel, line 11: **polish**: If this is meant to be used by `SetState` etc. above, then move the `enum` definition above the function declarations and update the parameter types accordingly.

include/wifi_battle_tower/application.h
```diff
@@ -0,0 +1,29 @@
...
+#include "wifi_battle_tower/app_state.h"
+
+#include "bg_window.h"
+#include "menu.h"
+#include "overlay_manager.h"
+#include "sprite.h"
+
+int WifiBattleTower_AppInit(ApplicationManager *appMan, int *state);
+int WifiBattleTower_AppMain(ApplicationManager *appMan, int *state);
+int WifiBattleTower_AppExit(ApplicationManager *appMan, int *unused);
```
lhearachel: **polish**: `unused` should be `state`, per the typical application contract.

lhearachel: **suggestion**: Mirror the declarations for other applications:


```suggestion
BOOL WifiBattleTower_Init(ApplicationManager *appMan, int *state);
BOOL WifiBattleTower_Main(ApplicationManager *appMan, int *state);
BOOL WifiBattleTower_Exit(ApplicationManager *appMan, int *state);
```

include/wifi_battle_tower/screen.h
```diff
@@ -0,0 +1,28 @@
...
+int WifiBattleTower_ScreenInit(WifiBattleTowerAppState *appState, int unused);
+int WifiBattleTower_ScreenMain(WifiBattleTowerAppState *appState, int unused);
+int WifiBattleTower_ScreenExit(WifiBattleTowerAppState *appState, int unused);
+void Window_DrawTextAligned(Window *window, String *string, int xOffset, int yOffset, int alignment, TextColor textColor);
+void Window_DrawSystemTextAligned(Window *window, String *string, int xOffset, int yOffset, int alignment, TextColor textColor);
+void WifiBattleTower_InitNetworkIcon(WifiBattleTowerAppState *appState);
+
+#define POPUP_TILEMAP_TOP        13
+#define WINDOW_BORDER_PLTT_SLOT  11
+#define WINDOW_BORDER_TILE_START 31 // (1 + (18 + 12))
```
lhearachel: **suggestion**: I doubt that this needs to be included in the header itself, but this should be computed using constants from `render_window.h`. Specifically, `(18 + 12)` should be `SCROLLING_MESSAGE_BOX_TILE_COUNT`.

include/wifi_battle_tower/screen.h
```diff
@@ -0,0 +1,28 @@
...
+
+int WifiBattleTower_ScreenInit(WifiBattleTowerAppState *appState, int unused);
+int WifiBattleTower_ScreenMain(WifiBattleTowerAppState *appState, int unused);
+int WifiBattleTower_ScreenExit(WifiBattleTowerAppState *appState, int unused);
+void Window_DrawTextAligned(Window *window, String *string, int xOffset, int yOffset, int alignment, TextColor textColor);
+void Window_DrawSystemTextAligned(Window *window, String *string, int xOffset, int yOffset, int alignment, TextColor textColor);
+void WifiBattleTower_InitNetworkIcon(WifiBattleTowerAppState *appState);
+
+#define POPUP_TILEMAP_TOP        13
+#define WINDOW_BORDER_PLTT_SLOT  11
```
lhearachel: **polish**: Ditto to shuffle these into the source file if they are not needed elsewhere or in multiple source-files.

include/wifi_battle_tower/save.h
lhearachel, line 1: **thought**: I guess we should talk about this now; do we want this to be a standalone subfolder, or should we only shuffle the application code away (for overlay 96)?

src/overlay005/field_menu.c
```diff
@@ -159,7 +159,7 @@ void FieldMenuManager_ShowSingleColumnMenu(FieldMenuManager *menuManager)
         menuManager->anchorY -= menuManager->optionsCount * GLYPH_ROW_TILES_HEIGHT;
     }
 
-    Window_Add(menuManager->fieldSystem->bgConfig, &menuManager->menuWindow, BG_LAYER_MAIN_3, menuManager->anchorX, menuManager->anchorY, menuWidth, menuManager->optionsCount * GLYPH_ROW_TILES_HEIGHT, PLTT_13, ((1 + (10 * 4)) + (10 * 2)));
+    Window_Add(menuManager->fieldSystem->bgConfig, &menuManager->menuWindow, BG_LAYER_MAIN_3, menuManager->anchorX, menuManager->anchorY, menuWidth, menuManager->optionsCount * GLYPH_ROW_TILES_HEIGHT, PLTT_13, (1 + (10 * 4)) + (10 * 2));
```
lhearachel, line 163: **thought**: This looks like some base-tile math to me, though I don't know which for sure.

src/overlay005/field_menu.c
```diff
@@ -159,7 +159,7 @@ void FieldMenuManager_ShowSingleColumnMenu(FieldMenuManager *menuManager)
...
     LoadStandardWindowGraphics(menuManager->fieldSystem->bgConfig, BG_LAYER_MAIN_3, 1024 - (18 + 12) - 9, PLTT_11, STANDARD_WINDOW_SYSTEM, HEAP_ID_FIELD1);
```
lhearachel: **suggestion**: `1024 - (18 + 12) - 9` is `BASE_TILE_STANDARD_WINDOW_FRAME` from `constants/field_base_tiles.h`.

src/wifi_battle_tower/application.c
```diff
@@ -0,0 +1,389 @@
...
+
+    SpriteTransfer_RequestChar(appState->spriteResource[0][0]);
+    SpriteTransfer_RequestPlttWholeRange(appState->spriteResource[0][1]);
+    NARC_dtor(narc);
+}
+
+static const u16 sSelectionArrowPositions[][2] = {
+    { 0xE0, 0x6F },
+    { 0xE0, 0x81 }
+};
```
lhearachel: **question**: Does this match with `CoordinatesU16`?


```suggestion
static const CoordinatesU16 sSelectionArrowPositions[] = {
    { .x = 0xE0, .y = 0x6F },
    { .x = 0xE0, .y = 0x81 },
};
```

src/wifi_battle_tower/application.c
```diff
@@ -0,0 +1,389 @@
...
+    template->heapID = HEAP_ID_68;
+}
+
+static void WifiBattleTower_CreateSelectionArrows(WifiBattleTowerAppState *appState)
+{
+    SpriteResourcesHeader_Init(&appState->arrowSpriteHeader, 0, 0, 0, 0, 0xffffffff, 0xffffffff, 0, 0, appState->resourceCollection[0], appState->resourceCollection[1], appState->resourceCollection[2], appState->resourceCollection[3], NULL, NULL);
+    AffineSpriteListTemplate template;
+    WifiBattleTower_BuildAffineSpriteTemplate(&template, appState, &appState->arrowSpriteHeader, NNS_G2D_VRAM_TYPE_2DMAIN);
+
+    for (int i = 0; i < 2; i++) {
```
lhearachel: **polish**: Use `NELEMS` / `SNELEMS` for the upper-bound on this loop.

src/wifi_battle_tower/http.c
```diff
@@ -0,0 +1,429 @@
...
+
+#include "http/http.h"
+#include "overlay096/struct_wifi_player_profile.h"
+#include "overlay096/struct_world_exchange_trainer.h"
+#include "overlay096/struct_world_exchange_trainer_error.h"
+#include "wifi_battle_tower/screen.h"
+
+#include "enums.h"
+
+typedef enum {
```
lhearachel: **style**: We generally don't `typedef` enums.

src/wifi_battle_tower/http.c
```diff
@@ -0,0 +1,429 @@
...
+    HTTP_STATE_AWAITING_UPLOAD,
+    HTTP_STATE_ERROR_UPLOAD,
+    HTTP_STATE_AWAITING_INFO,
+    HTTP_STATE_ERROR_INFO,
+    HTTP_STATE_AWAITING_SET_PROFILE,
+    HTTP_STATE_ERROR_SET_PROFILE,
+    HTTP_STATE_COMPLETE
+} BattleTowerHttpRequestState;
+
+typedef struct {
```
lhearachel: **style**: Tag the struct with the same name as its `typedef`.

src/wifi_battle_tower/screen.c
```diff
@@ -0,0 +1,1666 @@
...
+static int WifiBattleTower_State_DisconnectingPrompt(WifiBattleTowerAppState *appState);
+static int WifiBattleTower_State_CleanupInet(WifiBattleTowerAppState *appState);
+static int WifiBattleTower_State_GoodbyePrompt(WifiBattleTowerAppState *appState);
+static int Window_CalcAlignedTextX(Window *window, String *string, int xOffset, int alignment, u32 unused, int font);
+static void WifiBattleTower_ShowWaitDial(WifiBattleTowerAppState *appState);
+static void WifiBattleTower_HideWaitDial(WifiBattleTowerAppState *appState);
+
+// clang-format off
+static int (*sStateFunctions[])(WifiBattleTowerAppState *) = {
+    WifiBattleTower_State_InitialPrompt,           // 0
```
lhearachel: **note**: If we need to keep state numbers aligned, then we should have an `enum` and use designated-initializers here.

src/wifi_battle_tower/screen.c
```diff
@@ -0,0 +1,1666 @@
...
+}
+
+static void WifiBattleTower_LoadGraphics(WifiBattleTowerAppState *appState)
+{
+    BgConfig *bgConfig = appState->bgConfig;
+    NARC *narc = NARC_ctor(NARC_INDEX_DATA__WIFIP2PMATCH, HEAP_ID_68);
+
+    Graphics_LoadPaletteFromOpenNARC(narc, 3, 0, 0, 0, HEAP_ID_68);
+    Graphics_LoadPaletteFromOpenNARC(narc, 3, 4, 0, 0, HEAP_ID_68);
+    Font_LoadScreenIndicatorsPalette(0, 13 * 0x20, HEAP_ID_68);
```
lhearachel: **polish**: `13 * 0x20` is `PLTT_OFFSET(13)`.

lhearachel (changes requested):

include/wifi_battle_tower/save.h
Land0fChocolate: I moved `save.c` and `save.h` (which were previously named `wifi_battle_tower_save`) into the wifi_battle_tower folders simply because that was where the other wifi battle tower code was and it seemed neat and tidy. Though it might be unconventional to mix main binary files with overlay files. I don't feel too strongly about one way or the other.

lhearachel: I think we might want to instead shuffle all of the save-related data structures and functions into a separate folder? I'd like to hear other maintainers' thoughts here.

Land0fChocolate: I think this PR has been hanging around for quite a while and I would like to finish things up.

So options for organizing the files that manage save data like `battle_frontier_save.c`, `battle_hall_win_records.c`, etc:

1. Keep things as they are. Which means those files stay in the `src` root.
2. We make a subdirectory like `src/save` and put those files in there.
3. They get placed with other code of their group. So `battle_frontier_save.c` would go into a subdirectory that would be like `src/battle_frontier` and share its home with other battle frontier files.

After thinking on it over the past week, I'm leaning towards the second option.

lhearachel: I think it's fine to keep things as they are for now and discuss more with other maintainers as to the best option. Lots of folks have been busy, so there hasn't been much opportunity to talk it out. Let's revert the changes to the save code for now and address them in a future PR.

lhearachel:
There are still unresolved comments that have not been addressed; please make those changes, too.

include/wifi_battle_tower/application.h
Land0fChocolate: Due to a call to some of those functions by the application manager at `\src\unk_0203D1B8.c:1341` in `sub_0203E1AC()`, the type must be an *int. 

`WifiBattleTower_SetState` however uses the `WifiBattleTowerState` enum and can be changed.

lhearachel (approved):

PR #1231, opened 2026-07-03
Document Battle Anim Sprite Funcs

Fexty12573:
Documents all of the "Sprite Funcs" as part of the Battle Animation Subsystem.

When battle anim scripts create a sprite with `AddSpriteWithFunc` they can specify the ID of a function (`SPRITE_FUNC_*`) that will get called on it, allowing complex animations to be performed on them.

If you want to review in order, there is a table (`sBattleAnimSpriteFuncs`) with all of them in `src/battle_anim/script_func_tables.c`. Tho I imagine just going through files would be easier.

Relevant Files are:
* `script_funcs_[0-4].h/c`
* `script_funcs_status.h/c`

The rest is mostly supplementary or constants.


include/constants/battle/battle_anim.h
```diff
@@ -449,5 +449,39 @@
 #define EMITTER_PRIORITY_MODE_BACK  2 //< Send to back
 #define EMITTER_PRIORITY_MODE_UB    3 //< Invokes undefined behavior 😀
 
+#define SPRITE_FUNC_EXAMPLE            0
```
lhearachel, line 452: **note**: You can probably make this a proper `enum` now that `enumproc` is a thing.

src/battle_anim/script_funcs_status.c
```diff
@@ -0,0 +1,562 @@
...
+            ctx->state++;
+            AlphaFadeContext_Init(
+                &ctx->alphaFade,
+                FREEZE_SPRITE_END_ALPHA,
+                FREEZE_SPRITE_START_ALPHA,
+                FREEZE_SPRITE_DEFAULT_ALPHA - FREEZE_SPRITE_END_ALPHA,
+                FREEZE_SPRITE_DEFAULT_ALPHA - FREEZE_SPRITE_START_ALPHA,
+                FREEZE_SPRITE_ALPHA_FADEOUT_FRAMES);
+        }
+    } break;
```
lhearachel: **style**: Super nitpicky, but I think we prefer sticking the `break` inside the scope-block.

src/battle_anim/script_funcs_3.c
```diff
@@ -3552,59 +3629,49 @@ void BattleAnimScriptFunc_RevolveEmitter(BattleAnimSystem *system)
...
+    5,
+    5,
+    5,
+    4,
+    4,
+    4,
+    3,
+    3,
+    3,
+    FOLLOW_ME_SPRITE_OFFSET_SWITCH_PRIO,
```
lhearachel, line 3688: **note**: This looks like a stray find-replace?

lhearachel (commented):
I'm mostly skimming these changes and making notes on things that jump out at me

include/constants/battle/battle_anim.h
Fexty12573: True, but I wonder if there is any benefit to doing so, these funcs are only invoked from scripts and I don't see a situation where you'd wanna call them directly from C as they're pretty specialized. But I don't mind either way so do lmk

src/battle_anim/script_funcs_3.c
Fexty12573: Actually those are intentional to signal that these values have a special meaning in that function and isn't just "offset 0". But I realize I forgot to actually use them in the function below

include/constants/battle/battle_anim.h
lhearachel: Probably not, feel free to leave them as-is.

lhearachel (approved):

PR #1232, opened 2026-07-05
Apply some graphics macros, enums, and constants

CharlesFolz4:
Replacing some magic numbers with macros, enums, and constants that have already been defined, but were previously not extensively applied.


lhearachel (commented):
I actually think the `PLTT_` constants are pretty low-value. I would rather we purge them from the repository entirely. What do you think?

CharlesFolz4:
I agree that the PLTT_ constants from the Palette enum is fairly low value, but if it exists at all, it should be applied consistently.  Removing that would involve changing the fields `textPltt` and `framePltt` in the struct `MainMenuWindow` from `enum Palette` to an int, as well as the relevant params from the functions `MainMenuUtil_InitWindow()` and `LoadMoveSelectPltt()`.   
I'm not opposed to making that change, but I do want to confirm that that is a desirable change first.

lhearachel:
> I'm not opposed to making that change, but I do want to confirm that that is a desirable change first.

I think that's a fine change, yes. Go for it. :grin:

lhearachel (approved):

PR #1234, opened 2026-07-13
document field map change flags

zebben:
Documents unk_02070428 from #325. The issue mentioned "Field system flags" but I went with "field_map_change_flags" based on the surrounding code

### DIFF
--- a/include/spawn_locations.h
+++ b/include/spawn_locations.h
@@ -10,10 +10,10 @@
 int FieldOverworldState_GetDefaultWarpID(void);
 void Location_InitFly(int flyDestination, Location *location);
 void Location_InitBlackOut(int blackOutDestination, Location *location);
-int GetMapBlackOutWarpId(int mapId);
-int GetMapFlyWarpId(int mapId);
-int GetSpawnIdByMapAndCoords(int mapID, int param1, int param2);
-void TryUnlockFlyLocationByMap(FieldSystem *fieldSystem, int param1);
-BOOL CheckFlyLocationUnlocked(FieldSystem *fieldSystem, enum MapHeaderID mapHeaderID);
+int GetMapBlackOutWarpId(enum MapHeaderID mapHeaderID);
+int GetMapFlyWarpId(enum MapHeaderID mapHeaderID);
+int GetSpawnIdByMapAndCoords(enum MapHeaderID mapHeaderID, int param1, int param2);
+void TryUnlockFlyLocationByMap(FieldSystem *fieldSystem, enum MapHeaderID mapHeaderID);
+BOOL CheckFlyLocationUnlocked(FieldSystem *fieldSystem, int flyLocation);
 
 #endif // POKEPLATINUM_SPAWN_LOCATIONS_H
--- a/include/spawn_locations.h
+++ b/include/spawn_locations.h
@@ -3,6 +3,8 @@
 
 #include "field/field_system_decl.h"
 
+#include "generated/map_headers.h"
+
 #include "location.h"
 
 int FieldOverworldState_GetDefaultWarpID(void);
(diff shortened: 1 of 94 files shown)


include/spawn_locations.h
```diff
@@ -12,6 +12,6 @@ int GetMapBlackOutWarpId(int mapId);
 int GetMapFlyWarpId(int mapId);
 int GetSpawnIdByMapAndCoords(int mapID, int param1, int param2);
 void TryUnlockFlyLocationByMap(FieldSystem *fieldSystem, int param1);
-BOOL CheckFlyLocationUnlocked(FieldSystem *fieldSystem, int param1);
+BOOL CheckFlyLocationUnlocked(FieldSystem *fieldSystem, int mapId);
```
lhearachel: **question**: Should `mapId` be `mapHeaderID`? If so, does this match using the corresponding `enum` for its type?

src/field_map_change_flags.c
```diff
@@ -44,22 +44,22 @@ void FieldSystem_InitFlagsOnMapChange(FieldSystem *fieldSystem)
     fieldSystem->wildBattleMetadata.encounterAttempts = 0;
 
     if (SystemFlag_CheckSafariGameActive(SaveData_GetVarsFlags(fieldSystem->saveData)) == FALSE) {
-        SpecialEncounter *v0;
+        SpecialEncounter *speEnc;
```
lhearachel: **style**: Prefer to use inline declarations, if matching permits.

```c
    SpecialEncounter *specialEncounter = SaveData_GetSpecialEncounters(fieldSystem->saveData);
    RoamingPokemon_UpdatePlayerRecentRoutes(specialEncounter, fieldSystem->location->mapId);
    RoamingPokemon_MoveAllLocations(specialEncounter);
```

This is a general comment, so it applies throughout your changes.

src/field_map_change_flags.c
```diff
@@ -73,26 +73,26 @@ void FieldSystem_InitFlagsWarp(FieldSystem *fieldSystem)
...
 
     {
-        PlayerData *v2 = FieldOverworldState_GetPlayerData(SaveData_GetFieldOverworldState(fieldSystem->saveData));
+        PlayerData *playerData = FieldOverworldState_GetPlayerData(SaveData_GetFieldOverworldState(fieldSystem->saveData));
 
-        if ((v2->playerState == PLAYER_AVATAR_CYCLING) && (MapHeader_IsBikeAllowed(fieldSystem->location->mapId) == 0)) {
-            v2->playerState = PLAYER_AVATAR_WALKING;
-        } else if (v2->playerState == PLAYER_AVATAR_SURFING) {
-            v2->playerState = PLAYER_AVATAR_WALKING;
+        if ((playerData->playerState == PLAYER_AVATAR_CYCLING) && (MapHeader_IsBikeAllowed(fieldSystem->location->mapId) == 0)) {
```
lhearachel: **polish**: Remove unnecessary parens around either side of the conditional and replace `0` with `FALSE` for conditional expressions.

src/field_map_change_flags.c
```diff
@@ -126,24 +126,24 @@ void FieldSystem_ClearPartnerTrainer(FieldSystem *fieldSystem)
...
+void FieldSystem_RandomizeRoamingPokemonLocations(FieldSystem *fieldSystem)
 {
     RoamingPokemon_RandomizeAllLocations(SaveData_GetSpecialEncounters(fieldSystem->saveData));
 }
 
-static BOOL sub_020705DC(FieldSystem *fieldSystem)
+static BOOL CreateFlyLocationJournalEvent(FieldSystem *fieldSystem)
 {
-    int v0 = GetMapFlyWarpId(fieldSystem->location->mapId);
+    int flyLocation = GetMapFlyWarpId(fieldSystem->location->mapId);
```
lhearachel: **question**: Is `flyLocation` also a map header?

lhearachel (commented):
Thanks for the contribution! This looks very nice, and I agree with the chosen name.

include/spawn_locations.h
zebben: Want me to refactor `mapId` on `Location` as well?

lhearachel: Yes, if they look like header IDs.

src/field_map_change_flags.c
zebben: undid this change. `flyLocation` appears to be an index of `sSpawnLocations`

lhearachel (approved):

PR #1241, opened 2026-07-24
Document Battle Factory save data

xXAbieGamingXx:
unk_0202FF4C.c from #325


src/battle_factory_save.c
```diff
@@ -0,0 +1,130 @@
...
+    case FACTORY_SAVE_STREAK_FLAGS:
+        if (*value_dupe >= 1) {
+            flags->streakActiveFlags |= (1 << challengeType);
+        } else {
+            flags->streakActiveFlags &= (0xff ^ (1 << challengeType));
+        }
+        break;
+    }
+
+    return;
```
lhearachel: **question**: Does this match as:


```suggestion
    if (field == FACTORY_SAVE_STREAK_FLAGS) {
        if (*value_dupe >= 1) {
            flags->streakActiveFlags |= (1 << challengeType);
        } else {
            flags->streakActiveFlags &= (0xff ^ (1 << challengeType));
        }
    }
```

lhearachel (commented):
@VicSevenT would you like to also issue a review, since you've been looking at frontier stuff lately?

src/battle_factory_save.c
xXAbieGamingXx: Yes, that code matches

src/battle_factory_save.c
```diff
@@ -0,0 +1,126 @@
...
+            flags->streakActiveFlags &= (0xff ^ (1 << challengeType));
+        }
+    }
+}
+
+u32 BattleFactoryStreakFlags_GetFlag(BattleFactoryStreakFlags *flags, u8 field, u8 challengeType, void *unused)
+{
+    switch (field) {
+    case FACTORY_SAVE_STREAK_FLAGS:
+        return (u32)((flags->streakActiveFlags >> challengeType) & 0x1);
```
VicSevenT: Remove the explicit cast and extra parens if possible
```suggestion
        return (flags->streakActiveFlags >> challengeType) & 0x1;
```

src/battle_factory_save.c
```diff
@@ -0,0 +1,126 @@
...
+        save->opponentPersonality[index] = *(u32 *)value;
+        break;
+    }
+}
+
+u32 BattleFactorySave_GetMember(BattleFactorySave *save, u8 field, u8 index, void *unused)
+{
+    switch (field) {
+    case FACTORY_SAVE_CHALLENGE_TYPE:
+        return (u32)save->challengeType;
```
VicSevenT: All the `(u32)` casts in this function I think can be removed.

src/battle_factory_save.c
```diff
@@ -0,0 +1,126 @@
...
+{
+    return save->didSave;
+}
+
+void BattleFactorySave_RecordSave(BattleFactorySave *save, BOOL didSave)
+{
+    save->didSave = didSave;
+}
+
+// set value
```
VicSevenT: I don't think there's any value to keeping this comment with the function now named.

include/battle_factory_save.h
```diff
@@ -0,0 +1,50 @@
...
+typedef struct BattleFactoryStreakFlags {
+    u8 streakActiveFlags;
+    u8 unused[3];
+} BattleFactoryStreakFlags;
+
+typedef struct BattleFactorySave {
+    u8 isOpenLevel : 1;
+    u8 challengeType : 3;
+    u8 didSave : 1;
+    u8 unused : 3;
```
VicSevenT: This field definition can be removed entirely. The compiler will align the next field to the next byte automatically.

VicSevenT (commented):
A couple of small points of polish, but otherwise looks good to me.

VicSevenT (approved):

PR #1245, opened 2026-07-26
Document Battle Frontier Communication

xXAbieGamingXx:


src/overlay104/frontier_communication.c
```diff
@@ -0,0 +1,1304 @@
...
+    battleHall_dupe->msgsReceived++;
+
+    if (CommSys_CurNetId() == dummy1) {
+        return;
+    }
+}
+
+BOOL FrontierCommunication_Unreachable2(BattleHall *battleHall)
+{
+    int unused, dummy1;
```
lhearachel: **question**: Does `unused` get optimized out by the compiler?

src/overlay104/frontier_communication.c
```diff
@@ -0,0 +1,1304 @@
...
+void FrontierCommunication_Unreachable1(int dummy1, int unused1, void *dummy2, void *battleHall)
+{
+    BattleHall *battleHall_dupe = battleHall;
+    const u16 *unused2 = dummy2;
+
+    battleHall_dupe->msgsReceived++;
+
+    if (CommSys_CurNetId() == dummy1) {
+        return;
+    }
```
lhearachel, line 35: **question**: Does this match without the conditional (but still with the function call)?

src/overlay104/frontier_communication.c
```diff
@@ -0,0 +1,1304 @@
...
+
+BOOL ov104_0222EE14(BattleHall *battleHall, u8 param1)
+{
+    u16 *data;
+    int result, dataSize = 44;
+    data = battleHall->commBuffer;
+
+    *data = param1;
+
+    if (CommSys_SendData(37, data, dataSize) == 1) {
```
lhearachel, line 173: **suggestion**: Use `NELEMS(battleHall->commBuffer)` (or `SNELEMS`) in place of `dataSize` if it matches. This also applies for cases where `commHugeBuffer` is used instead.

src/overlay104/frontier_communication.c
```diff
@@ -0,0 +1,1304 @@
...
+    dummy = 0;
+    battleArcade_dupe->msgsReceived++;
+
+    if (CommSys_CurNetId() == netID) {
+        return;
+    }
+
+    if (CommSys_CurNetId() == 0) {
+        return;
+    }
```
lhearachel: **suggestion**: Merge these conditionals, if it matches.

src/overlay104/frontier_communication.c
```diff
@@ -0,0 +1,1304 @@
...
+            battleCastle->unk_A1B = battleCastle->unk_A18;
+        } else {
+            if ((battleCastle->unk_A1B - 6) == 4) {
+                if (battleCastle->unk_A18 != 4) {
+                    battleCastle->unk_A1B = battleCastle->unk_A18;
+                }
+            }
+        }
+    } else {
+        if (battleCastle->unk_A1B == 4) {
```
lhearachel: **suggestion**: Replace such code patterns with `else if (...)`.

src/overlay104/frontier_communication.c
xXAbieGamingXx: That didn't match

src/overlay104/frontier_communication.c
xXAbieGamingXx: Yes, modified all instances

lhearachel (approved):

PR #1247, opened 2026-07-30
Transition to compiled arm7 binary

ljtpetersen:
As title says.

Should this wait for a new nitrorom release, so that the .wrap file remains a tag instead of a commit hash?


lhearachel (approved):
I'm signing off on the first commit here; I'd like for a second contributor to handle the sign-off for my portion (and squash this into a single commit).

The new code is just a copy-paste from the upstream repo. The only thing that should require review is the `README.md` and the changes to the build-scripts.

scbroede (approved):

PR #1250, opened 2026-08-01
Replace SDATTool with nitrosfx v0.1

h2o-DS:
Replaces `SDATTool` with `nitrosfx` v0.1

Unpacks `swar` files into `wav`.
Replaces `generated/sdat.h` with `res/sound/pl_sound_data.naix` generated from the sdat file.
Moves files related to pokemon cries into their respective species folders. Adding species will cause the added cries to be added to the sdat's private folder, but will require editing `res/sound/pl_sound_data.json` in order to be added to the SDAT; this will be addressed in a future PR.

Files in `res/sound/WAVARC` are IMA-ADPCM encoded due to some of them not currently being able to round-tripped in PCM16.


docs/datafiles/moves.md
```diff
@@ -96,7 +96,7 @@ L_0:
     SetExtraParams 0, 2, 2, 0, 0, 0
     Func_FadeBattlerSprite BATTLE_ANIM_DEFENDER, 0, 1, BATTLE_COLOR_RED, 10, 0
     Func_Shake 1, 0, 1, 2, BATTLE_ANIM_BATTLER_SPRITE_DEFENDER
-    PlayLoopedSoundEffectR SEQ_SE_DP_W052, 3, 4
+    PlayLoopedSoundEffectR SEQ_SE_DP_W052_sseq, 3, 4
```
lhearachel, line 99: **question**: Could we strip the extension before generating the NAIX? That would greatly reduce the scope of changes. As-is, GitHub is choking on this PR quite a bit due to the number of files that have been touched.

h2o-DS: I would prefer the extensions, but that can be reverted if required

PR #1255, opened 2026-08-08
Replace SDATTool with nitrosfx v0.1

h2o-DS:
Replaces SDATTool with nitrosfx v0.1

Unpacks swar files into wav.
Replaces generated/sdat.h with res/sound/pl_sound_data.naix generated from the sdat file.
Moves files related to pokemon cries into their respective species folders. Adding species will cause the added cries to be added to the sdat's private folder, but will require editing res/sound/pl_sound_data.json in order to be added to the SDAT; this will be addressed in a future PR.

Files in res/sound/WAVARC are IMA-ADPCM encoded due to some of them not currently being able to round-tripped in PCM16.

Replacing #1250 to break into commits small enough to review. Commits can be looked at in any order.


tools/nitrosfx/cJSON.c
lhearachel, line 1: **thought**: Is it possible to make an abstract JSON interface so we can link this against `dataproc` rather than `cJSON`?

lhearachel: We discussed this offline, this isn't a concern right now.

h2o-DS:
Resolved merge conflicts. See most recent commit for several script dependency changes.

lhearachel (approved):

PR #1260, opened 2026-08-13
Document `object_event_gfx_data.c`

Gudf:
This file contains the mappings between object event graphics IDs and the files they use, the functions used to initialize/update/etc. them, as well as some other graphics-related properties (reflections, shadow, tracks, animations...).


include/overlay005/object_event_gfx_data.h
```diff
@@ -0,0 +1,121 @@
...
+
+enum TrackType {
+    TRACK_TYPE_NONE = 0,
+    TRACK_TYPE_FOOTSTEPS,
+    TRACK_TYPE_LINE,
+};
+
+typedef struct ObjectEventGfxRendererEntry {
+    enum ObjectEventGfx graphicsID;
+    const UnkStruct_ov5_021FB0F0 *renderer;
```
lhearachel, line 77: **question**: Is it out-of-scope to name this type?

src/map_object_move.c
```diff
@@ -46,28 +46,28 @@ static void sub_0206363C(MapObject *mapObj);
...
-static void sub_02063C60(MapObject *mapObj, u8 currTileBehavior, u8 prevTileBehavior, const UnkStruct_ov5_021ECD10 *param3);
-static void sub_02063C94(MapObject *mapObj, u8 currTileBehavior, u8 prevTileBehavior, const UnkStruct_ov5_021ECD10 *param3);
-static void sub_02063CC8(MapObject *mapObj, u8 currTileBehavior, u8 prevTileBehavior, const UnkStruct_ov5_021ECD10 *param3);
-static void sub_02063CFC(MapObject *mapObj, u8 currTileBehavior, u8 prevTileBehavior, const UnkStruct_ov5_021ECD10 *param3);
-static void sub_02063D30(MapObject *mapObj, u8 currTileBehavior, u8 prevTileBehavior, const UnkStruct_ov5_021ECD10 *param3);
-static void sub_02063DA8(MapObject *mapObj, u8 currTileBehavior, u8 prevTileBehavior, const UnkStruct_ov5_021ECD10 *param3);
-static void MapObject_DetermineElevatedBridgeStatus(MapObject *mapObj, u8 currTileBehavior, u8 prevTileBehavior, const UnkStruct_ov5_021ECD10 *unused);
-static void MapObject_EmptyFunction(MapObject *mapObj, u8 currTileBehavior, u8 prevTileBehavior, const UnkStruct_ov5_021ECD10 *param3);
+static void MapObject_SinkIntoTerrain(MapObject *mapObj, u8 currTileBehavior, u8 prevTileBehavior, const ObjectEventGfxRenderDetailsEntry *param3);
```
lhearachel: **suggestion**: Let's proactively rename `param3` for these functions.

src/map_object_move.c
```diff
@@ -931,7 +931,7 @@ int MapObject_RecalculateObjectHeight(MapObject *mapObj)
         pos.y = updatedPos.y;
         MapObject_SetPos(mapObj, &pos);
         MapObject_SetYPrev(mapObj, MapObject_GetY(mapObj));
-        MapObject_SetY(mapObj, (((pos.y) >> 3) / FX32_ONE));
+        MapObject_SetY(mapObj, ((pos.y) >> 3) / FX32_ONE );
```
lhearachel: **polish**: Remove the extraneous space (does `clang-format` assert its presence...?)

src/overlay005/object_event_gfx_data.c
```diff
@@ -1017,140 +955,140 @@ const UnkStruct_ov5_021ED2D0 Unk_ov5_021FB5BC[] = {
...
+    { .startFrame = 0, .endFrame = 0, .animType = BILLBOARD_ANIM_TYPE_TABLE_END }
 };
 
-static const BillboardAnim Unk_ov5_021FB284[] = {
-    { 0, 2, BILLBOARD_ANIM_TYPE_LOOP },
-    { 3, 5, BILLBOARD_ANIM_TYPE_LOOP },
-    { 6, 9, BILLBOARD_ANIM_TYPE_LOOP },
-    { 10, 11, BILLBOARD_ANIM_TYPE_LOOP },
-    { 0, 0, BILLBOARD_ANIM_TYPE_TABLE_END }
+static const BillboardAnim sSurfBillboardAnims[] = {
```
lhearachel, line 987: **question**: Should this and others like it demarcate that they refer to the player?

src/map_object_move.c
Gudf: Ah, I didn't intend on the parens getting removed in the first place, my IDE used a more recent version of clang-format (not sure why?) that removed those automatically and inserted this extraneous space in several places for some reason. I reverted the others but I guess I missed this one lol. I'll ask if anyone is working on this file and revert this one as well if so, otherwise I might as well go and remove all of the extra parens I suppose.
Ah, actually I was thinking of a different file, I suppose I simply didn't notice in this file as well. Still, nobody seems to be working on it so i'm removing the space.

src/overlay005/object_event_gfx_data.c
Gudf: Eh, they're not inherently player-specific and it makes the identifier longer, I'd rather leave them as is.

include/overlay005/object_event_gfx_data.h
Gudf: I'm not sure, maybe not? But I felt like it'd make more sense to do it when documenting the file(s) that actually use the struct, though I guess we could consider we have enough information right now.

Gudf: Ultimately decided it's out of scope, I'll probably work on the relevant TU next.

lhearachel (approved):

PR #1262, opened 2026-08-14
Unpack graphic/ending.narc

anastarawneh:


res/graphics/end_credits/meson.build
```diff
@@ -0,0 +1,204 @@
...
+ending_narc = custom_target('ending.narc',
+    output: [
+        'ending.narc',
+        'ending.naix',
+    ],
+    input: [
+        copy_gen.process(ending_ncgr),
+        copy_gen.process(ending_nclr),
+        copy_gen.process(ending_ncer),
+        copy_gen.process(ending_nanr),
```
lhearachel: **issue**: This creates two targets for each file: one to create the Nitro file, and another to copy it into place.

Instead, use the generators provided at the top level `res/meson.build`, which removes the copy for each file.

include/cutscenes/end_credits/defs.h
```diff
@@ -54,17 +54,17 @@ enum EndCreditsSkipState {
...
+    END_CREDITS_MEMORY_5,
+    END_CREDITS_MEMORY_6,
+    END_CREDITS_MEMORY_7,
+
+    END_CREDITS_MEMORY_8,
+    END_CREDITS_MEMORY_9,
+    END_CREDITS_MEMORY_10,
 
     END_CREDITS_MEMORY_COUNT
 };
```
lhearachel, line 70: **question**: Why change to less descriptive names?

lhearachel (changes requested):

include/cutscenes/end_credits/defs.h
anastarawneh: The memories, contrary to what I had assumed for some reason, are not the same for Lucas/Dawn.

res/graphics/end_credits/meson.build
anastarawneh: I changed most of them, except the extended palettes, which already use the `custom_target` function to remove the extra `_0` in other parts of the repo.

lhearachel (approved):

PR #1264, opened 2026-08-17
feat: renamed BattleUseFunc and add enum

K4ddate:
the JSON field `battleUseFunc` for items is only used in `src/battle/battle_sub_menus/battle_bag.c` : `TryUseItem()`

https://github.com/pret/pokeplatinum/blob/6b2e9a5bf94c714ef8ef7cda5be06c0ccee01c70/src/battle_sub_menus/battle_bag.c#L412

The code only checks if its value is `3` which corresponds to the items *Poke Doll* and *Fluffy Tail*.
- All _Poke ball_ have the value `1`, and is never used.
- All _Healing_ items have the `2`, and is never used.
- All other items have the value `0`. and is never used.

The current naming is misleading, as the value doesn't point to a function, so I renamed it `battleUseCategory`. This is obviously open to any suggestion that better fits the project naming.

All JSON item files have been edited and this new enum has been added : 

```c
#define BATTLE_USE_CATEGORY_NONE      0
#define BATTLE_USE_CATEGORY_POKE_BALL 1
#define BATTLE_USE_CATEGORY_HEALING   2
#define BATTLE_USE_CATEGORY_ESCAPING  3
```

`make check` passes on my side, but I don't know if you have more tests I should apply.

*Made with luv :3*


tools/dataproc/src/itemproc.c
```diff
@@ -187,7 +188,7 @@ static void proc_item(datafile_t *df, const char *member) {
         .fieldPocket      = towidth(enum_u8(".fieldPocket", POCKET), 4),
         .battlePocket     = towidth(enum_u8(".battlePocket", BATTLE_POCKET_MASK), 5),
         .fieldUseFunc     = enum_u8(".fieldUseFunc", ITEM_USE_FUNC),
-        .battleUseFunc    = u8(".battleUseFunc"),
+        .battleUseCategory     = enum_u8(".battleUseCategory", BATTLE_USE_CATEGORY),
```
lhearachel, line 191: **style**: For this section of the code, could you keep the `=`s aligned for the other fields? e.g., this line should become:

```suggestion
        .battleUseCategory = enum_u8(".battleUseCategory", BATTLE_USE_CATEGORY),
```

And all other `=`s for this struct declaration should be vertically-aligned with that one.

lhearachel (commented):
Just one nitpick; could you also update `docs/datafiles/items.md` as needed?

K4ddate:
Hello, thanks for your review.

I will indeed update the documentation. Regarding the C linting, I feel like it should be done via workflows? I will lint it correctly for the PR, but it could be interesting to implement some workflows for the tools too.

adrienntindall:
> Hello, thanks for your review.
> 
> I will indeed update the documentation. Regarding the C linting, I feel like it should be done via workflows? I will lint it correctly for the PR, but it could be interesting to implement some workflows for the tools too.

You can run ``make format`` to fix any formatting issues

lhearachel:
> > Hello, thanks for your review.
> > I will indeed update the documentation. Regarding the C linting, I feel like it should be done via workflows? I will lint it correctly for the PR, but it could be interesting to implement some workflows for the tools too.
> 
> You can run `make format` to fix any formatting issues

This is not true for the `tools` folder. We don't apply the typical formatting rules to those because the tools are so disparate in their practices.

K4ddate:
Hi there,

Documentation updated and proc_item() fields linted.

lhearachel (approved):

PR #1267, opened 2026-08-20
Use common TEXT_COLOR constants throughout repo

VicSevenT:
I found some macros for common `TEXT_COLOR`s in `pokemon_summary_screen/main.h`, and I figured it'd make sense to elevate them and use them repo-wide.


lhearachel (commented):
Just one overarching thought: is it worth noting somehow that these refer to palette-indices, and so they may not necessarily be that exact color? I believe that these are the *standard* colors, but the text-printers should always be using the same palette as the message window, which might change.

VicSevenT:
Ah..., that actually completely slipped my mind. And now as I look into it more, the original constants from the summary screen *are not* accurate for the entire code base. It'd be nice to have some sort of constants for these long-term, but it needs to be looked at further. I'm going to close this out for the time-being.

PR #1268, opened 2026-08-21
Document the WFC Battle Facility Selector

VicSevenT:
This is a small script/scene that launches when the Player selects "Battle Frontier" in the Wifi Club. From here, the player and their partner select the facility to challenge and the pokemon they will use, and then it launches the proper facility script.


scbroede:
build failure

VicSevenT:
Fixed

lhearachel (approved):

PR #1281, opened 2026-09-05
more rankings documentation

zebben:


include/rankings.h
```diff
@@ -9,7 +9,42 @@
...
+    RANKING_STAT_POKEMON_CAUGHT = 7,
+    RANKING_STAT_POKEMON_HATCHED = 8,
+    RANKING_STAT_POKEMON_FISHED = 9,
+
+    RANKING_STAT_CONTEST_WINS = 10,
+    RANKING_STAT_CONTEST_ENTRIES = 11,
+    RANKING_STAT_RIBBONS_EARNED = 12,
+
+    RANKING_STAT_MAX = 13
+};
```
lhearachel, line 34: **issue**: Don't explicitly assign values to the `enum` members unless necessary to skip over values. Just let the members auto-increment where possible. The exception to this is the first member, which we stylistically set to `0` to be explicit.

lhearachel (dismissed):

lhearachel (changes requested):
Imagine I hit this button instead of Approve above

lhearachel (approved):

Issue #1282, opened 2026-09-06
[Vanilla bug] Expert AI does not use Thunder scoring branch


SunakazeKun:
In ``Expert_Main``, there are two jumps based on the ``BATTLE_EFFECT_SKIP_CHARGE_TURN_IN_SUN``, effectively preventing the second branch to ``Expert_UnusedSolarbeam``. Based on the functionality of that unreached code, this appears to be intended for scoring the move Thunder, not Solar Beam. Further, the effect that follows ``BATTLE_EFFECT_SKIP_CHARGE_TURN_IN_SUN`` (151) is actually ``BATTLE_EFFECT_THUNDER`` (152), indicating it really is a typo on GameFreak's end.

The line in question: https://github.com/pret/pokeplatinum/blob/bca37652996330898fdd2408281ea419b8c995c7/src/battle/trainer_ai/script.s#L1717

Since a lot of other bugs are detailed in the trainer AI code, I suppose that adding a comment and renaming the ``UnusedSolarbeam`` branches accordingly does the trick.

PR #1283, opened 2026-09-06
Noted Expert AI Thunder scoring bug

SunakazeKun:
Solves #1282

### DIFF
--- a/src/battle/trainer_ai/script.s
+++ b/src/battle/trainer_ai/script.s
@@ -1715,8 +1715,8 @@ Expert_Main:
     IfCurrentMoveEffectEqualTo BATTLE_EFFECT_CHARGE_TURN_DEF_UP, Expert_ChargeTurnNoInvuln
     IfCurrentMoveEffectEqualTo BATTLE_EFFECT_SKIP_CHARGE_TURN_IN_SUN, Expert_ChargeTurnNoInvuln
 
-    // Bug: Thunder is not properly scored. This is supposed to check for BATTLE_EFFECT_THUNDER.
-    IfCurrentMoveEffectEqualTo BATTLE_EFFECT_SKIP_CHARGE_TURN_IN_SUN, Expert_UnusedThunder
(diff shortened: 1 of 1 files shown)


src/battle/trainer_ai/script.s
```diff
@@ -1714,7 +1714,10 @@ Expert_Main:
     IfCurrentMoveEffectEqualTo BATTLE_EFFECT_MIRROR_COAT, Expert_MirrorCoat
     IfCurrentMoveEffectEqualTo BATTLE_EFFECT_CHARGE_TURN_DEF_UP, Expert_ChargeTurnNoInvuln
     IfCurrentMoveEffectEqualTo BATTLE_EFFECT_SKIP_CHARGE_TURN_IN_SUN, Expert_ChargeTurnNoInvuln
-    IfCurrentMoveEffectEqualTo BATTLE_EFFECT_SKIP_CHARGE_TURN_IN_SUN, Expert_UnusedSolarbeam
+
+    // Bug: Thunder is not properly scored. This is supposed to check for BATTLE_EFFECT_THUNDER.
+    IfCurrentMoveEffectEqualTo BATTLE_EFFECT_SKIP_CHARGE_TURN_IN_SUN, Expert_UnusedThunder
```
lhearachel: **suggestion**: We should probably adjust the label (and its sub-labels), since it's definitely used.

Additionally, fully-capitalize `BUG`.

lhearachel (commented):
Thanks for the contribution! Could you also add an entry in `docs/bugs_and_glitches.md` that provides a patch for the bug?

src/battle/trainer_ai/script.s
SunakazeKun: Both aspects are covered by the newest commit.

SunakazeKun:
> Thanks for the contribution! Could you also add an entry in `docs/bugs_and_glitches.md` that provides a patch for the bug?

I see that this document currently contains none of the trainer AI bugs noted in the trainer AI script. Hence, I didn't add anything to it.

lhearachel (approved):

PR #1289, opened 2026-09-14
Document Battle Frontier script change commands

VicSevenT:


asm/macros/frscrcmd.inc
```diff
@@ -24,19 +44,19 @@
     .short FRSCRCMD_END
     .endm
 
-    .macro FrontierScrCmd_02
-    .short FRSCRCMD_02
+    .macro ExitFrontier
```
scbroede, line 47: assuming i understand what this does correctly, would it make more sense to call it `ExitFacility`?

VicSevenT: Hmm, I'm not sure. I used `ExitFrontier` to be consistent with how I've been using `BattleFrontier`/`Frontier` in the code to refer to the frontier script system, as this command triggers a complete exit from that system.

scbroede: if it's consistent with usage elsewhere it's fine imo

scbroede (approved):

